PR #1, opened 2025-04-07
chore: add prettier

nirinchev:
Figured that since we'll be actively collaborating, it's best to align on the code style early to avoid getting merge conflicts due to different formatting.

For vscode, it's easiest to install the prettier extension and enable the "Format on save" config option.


blva (approved):
LGTM

PR #7, opened 2025-04-08
ci: add eslint

fmenezes:
- added eslint (via `npm run lint`)
- added github action to check eslint
- changed github action `prettier` to actually check if code needs reformatting
- added `eslint-config-prettier` to make sure eslint does not touch formatting without `prettier`

### DIFF
--- a/.github/workflows/code_health.yaml
+++ b/.github/workflows/code_health.yaml
@@ -37,7 +37,6 @@ jobs:
                     exit 1
                   fi
     prettier:
-        name: Prettier Check
         runs-on: ubuntu-latest
         steps:
             - uses: GitHubSecurityLab/actions-permissions/monitor@v1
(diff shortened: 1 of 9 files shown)


.github/workflows/code_health.yaml
```diff
@@ -37,22 +37,47 @@ jobs:
...
-              id: prettier-run
-              uses: rutajdash/prettier-cli-action@d42c4325a3b344f3bd4be482bc34de521998d557
+            - uses: actions/setup-node@v4
               with:
-                  config_path: ./.prettierrc.yml
-            - name: Prettier Output
-              if: ${{ failure() }}
-              shell: bash
+                  node-version-file: package.json
+            - name: install dependencies
```
gagik:
```suggestion
            - name: Install dependencies
```

.github/workflows/code_health.yaml
```diff
@@ -37,22 +37,47 @@ jobs:
...
+              run: |
+                  npm ci
+            - run: |
+                  npm run reformat
+            - name: Check for uncommitted files
```
gagik: do we need this kind of check? haven't seen this before

.github/workflows/code_health.yaml
```diff
@@ -37,22 +37,47 @@ jobs:
...
+                  config: ${{ vars.PERMISSIONS_CONFIG }}
+            - name: Checkout Repository
+              uses: actions/checkout@v4
+            - uses: actions/setup-node@v4
+              with:
+                  node-version-file: package.json
+            - name: install dependencies
+              run: |
+                  npm ci
+            - run: |
```
gagik, line 72: could we combine this with the pretttier? I guess this allows us to run in parallel but dependency install will take the most time regardless

.github/workflows/code_health.yaml
```diff
@@ -37,22 +37,47 @@ jobs:
...
+        runs-on: ubuntu-latest
+        steps:
+            - uses: GitHubSecurityLab/actions-permissions/monitor@v1
+              with:
+                  config: ${{ vars.PERMISSIONS_CONFIG }}
+            - name: Checkout Repository
+              uses: actions/checkout@v4
+            - uses: actions/setup-node@v4
+              with:
+                  node-version-file: package.json
```
gagik, line 67:
```suggestion
                  node-version-file: package.json
                  cache: "npm"
```

eslint.config.js
```diff
@@ -0,0 +1,12 @@
+import { defineConfig } from "eslint/config";
+import js from "@eslint/js";
+import globals from "globals";
+import tseslint from "typescript-eslint";
+import eslintConfigPrettier from "eslint-config-prettier/flat";
+
+export default defineConfig([
```
gagik, line 7: could we piggy back by extending https://www.npmjs.com/package/@mongodb-js/eslint-config-devtools

also related for introducing `.prettierrc`
https://github.com/mongodb-js/mongosh/blob/cf4f9eed0dcef6dbbdd434c90c36d7575782ea4d/packages/errors/.prettierrc.json#L1

.github/workflows/code_health.yaml
fmenezes: this is to make sure `npm run reformat` did not change anything, without it simply `prettier -c .` will allow warnings

gagik: if the idea is to check for prettier, we can use https://prettier.io/docs/cli#--check and look for the exit code

gagik: hmm, I'm not 100% on this but I think if `check` does return 1, the action should fail

eslint.config.js
fmenezes: we are using latest eslint 9.x and `@mongodb-js/eslint-config-devtools` requires 7.x, how about a followup PR?

.github/workflows/code_health.yaml
fmenezes: perfect, changed

.github/workflows/code_health.yaml
fmenezes: I haven't merged them yet, it feels easier to address when you know what is wrong

gagik: okay they're pretty quick actually; I'm fine leaving them as is for the time being 👍

gagik (approved):
Looks good; would be good to follow-up later with aligning eslint and [.prettier.json](https://github.com/mongodb-js/mongosh/blob/cf4f9eed0dcef6dbbdd434c90c36d7575782ea4d/packages/errors/.prettierrc.json#L1) with devtools things in some form

PR #8, opened 2025-04-08
Refactor the tool registration

nirinchev:
Refactor the way tools are defined and registered:
- Creates a base class for tools that encapsulates the execution logic along with the metadata
- Move the Atlas tools implementations to their own classes inside tools/atlas
- Register all tools in tools/atlas/index.ts and tools/mongodb/index.ts

Additionally, it adds a few QOL improvements:
- Enable sourcemaps
- Add a launch profile for debugging the server with vscode


package.json
```diff
@@ -6,20 +6,21 @@
...
     "build": "npm run build:clean && npm run build:compile && npm run build:addshebang && npm run build:chmod",
-    "inspect": "npm run build && npx @modelcontextprotocol/inspector -- dist/index.js",
+    "inspect": "npm run build && mcp-inspector -- dist/index.js",
     "prettier": "prettier",
     "check": "npm run check:lint && npm run check:format",
-    "check:lint": "npx eslint .",
-    "check:format": "npx prettier -c .",
-    "reformat": "npx prettier -- --write ."
+    "check:lint": "eslint .",
+    "check:format": "npm run prettier -c .",
```
gagik:
```suggestion
    "check:format": "prettier -c .",
```
I think this should work

src/tools/atlas/index.ts
```diff
@@ -0,0 +1,20 @@
+import { McpServer } from "@modelcontextprotocol/sdk/server/mcp.js";
```
blva: super nit: can we call this toolIndex or tools or any other name? Atlas has index api commands and that was my first though when I saw this line and you probably will add index support

package.json
fmenezes: or simply rebase on main

fmenezes (approved):
LGTM, left one question

src/tools/tool.ts
```diff
@@ -0,0 +1,62 @@
+import { McpServer } from "@modelcontextprotocol/sdk/server/mcp.js";
+import { z, ZodRawShape, ZodTypeAny } from "zod";
+import { log } from "../logger.js";
+import { CallToolResult } from "@modelcontextprotocol/sdk/types.js";
+import { State } from "../state.js";
+
+export abstract class ToolBase<Args extends ZodRawShape> {
+    protected state: State = undefined!; // We should never use this before it's set
```
gagik: we could avoid this if we use a private constructor i.e. 
```ts
private constructor(state: State);
```
and then use
```ts
static register(server: McpServer, state: State): ToolBase {
	const base = new ToolBase(state);
        /* .... */
       return base;
}
```

src/server.ts
```diff
@@ -1,22 +1,18 @@
 import { McpServer } from "@modelcontextprotocol/sdk/server/mcp.js";
-import { CallToolResult } from "@modelcontextprotocol/sdk/types.js";
-import { Transport } from "@modelcontextprotocol/sdk/shared/transport.js";
-import { z, ZodOptional, ZodString } from "zod";
-import { ApiClient, AtlasCluster } from "./client.js";
+import { ApiClient } from "./client.js";
```
fmenezes, line 2: should we perhaps move client and all API stuff to `AtlasToolBase`? I think only atlas will use it

blva: telemetry will use it I think, which won't be atlas

src/tools/atlas/index.ts
nirinchev: Hm... happy to rename it to something else, but the index pattern is fairly well established in js projects, so when we add index support, we'll probably need to call the file something else anyway 😅

src/server.ts
nirinchev: Hm... not super familiar with how we're going to be doing telemetry, so will leave it as is for now, but we can definitely move it over if it turns out it's not needed for telemetry.

src/tools/atlas/index.ts
blva: np! no strong opinion and less knowledgeable of js xD

blva (approved):

fmenezes (approved):

fmenezes:
@nirinchev make sure to delete dist folder on your branch before merging

fmenezes (approved):

PR #13, opened 2025-04-08
feat: add unit test framework

blva:
- adds jest as a testing framework


gagik:
Don't have strong feelings about this but we might have a better time using https://vitest.dev/ as it has built-in TypeScript and ESM support

blva:
> Don't have strong feelings about this but we might have a better time using https://vitest.dev/ as it has built-in TypeScript and ESM support

no strong opinions on my side as well, is it something we use in other js repos within MDB? if so I'm happy to shift!

gagik:
@blva It's newer so doesn't seem to be used by other projects much; I originally was gonna suggest mocha as that's what we use in other repos but actually jest seems simpler to setup for our purpose right now. Thinking more, maybe let's just stick to jest since the basic configuration looks already done anyhow

blva:
> @blva It's newer so doesn't seem to be used by other projects much; I originally was gonna suggest mocha as that's what we use in other repos but actually jest seems simpler to setup for our purpose right now. Thinking more, maybe let's just stick to jest since the basic configuration looks already done anyhow

cool! I'll continue with this then, I'd rather stick with technologies we've been working with for now, so jest it is and we can revisit!

jest.config.js
```diff
@@ -0,0 +1,207 @@
+/**
+ * For a detailed explanation regarding each configuration property, visit:
```
blva: auto-generated via `npm init jest@latest`, we can clean up the comments

tests/unit/index.test.ts
```diff
@@ -0,0 +1,27 @@
...
+                    return new Promise((resolve) => {
+                        resolve(transport);
+                    });
+                }),
+            };
+        }),
+    };
+});
+
+describe("Server initialization", () => {
```
blva, line 22: adding a smalll test for now so that we can add tests in next PRs and I'll slowly backfill tests

.github/workflows/code_health.yaml
```diff
@@ -19,3 +19,19 @@ jobs:
...
+            - uses: GitHubSecurityLab/actions-permissions/monitor@v1
+            - uses: actions/checkout@v4
+            - uses: actions/setup-node@v4
+              with:
+                  node-version-file: package.json
+                  cache: "npm"
+            - name: Install dependencies
+              run: npm ci
+            - name: Build project
+              run: npm run build
```
fmenezes: do we need build?

blva: nope

tsconfig.jest.json
```diff
@@ -0,0 +1,10 @@
+{
+  "extends": "./tsconfig.json",
+  "compilerOptions": {
+    "module": "commonjs",
+    "target": "esnext",
+    "isolatedModules": true,
+    "allowSyntheticDefaultImports": true
+  },
+  "include": ["src/**/*.ts", "__tests__/**/*.ts"]
```
fmenezes: our tests are under `/tests` here we are importing `__tests__`, to me we should stick to `__tests__` and add the readme parts to contributing.md

tsconfig.jest.json
```diff
@@ -0,0 +1,10 @@
+{
+  "extends": "./tsconfig.json",
+  "compilerOptions": {
+    "module": "commonjs",
```
fmenezes: this is actually contradicting main tsconfig.json, do we need it?

blva: we need the module, but I can update to `esnext` as it seems to work

blva: sorry actually, we can't use `nodenext` here, it doesn't work well w/ isolated modules, but `esnext` works

tsconfig.jest.json
blva: yep, makes sensed, I'll move the tests to `__tests__`

gagik: can we do `unit-tests/**/*.ts` pattern instead? no real reason to use `__tests__`

blva: no strong opinion, would `tests/unit/**/*.ts` work for both of you?

tsconfig.jest.json
```diff
@@ -0,0 +1,10 @@
+{
+  "extends": "./tsconfig.json",
+  "compilerOptions": {
+    "module": "esnext",
+    "target": "esnext",
```
gagik, line 5:
```suggestion
    "module": "NodeNext",
    "esModuleInterop": true,
```
I think this should help but yeah jest works weirdly with an esm setup (which is what we're using)

blva: let me try that one

tsconfig.jest.json
gagik: sounds good to me

tsconfig.jest.json
blva: doesn't work, tried nodenext a lot but no success, we end up with import errors and I tried a bunch of different mapping/trasnformations but still couldn't make it work

gagik:
@blva https://jestjs.io/docs/ecmascript-modules Seems like ESM support is experimental in Jest. 
With us using `"type": "module"` (which we should because ESM is what's recommended now, CommonJS is more of a legacy thing). We can either
a) continue trying with Jest, I think some configuration is possible with ESM but it does seem like library mocking is not going to be great regardless with our setup.
b) maybe use Mocha + Chai + Sinon as they do support ESM and would also be more consistent with rest of DevTools. If you need help setting this up let me know, I could create a PR copying some of the mongosh configuration.

gagik:
@blva I opened #50 to try to see how Mocha goes, it actually seems simpler to set up because of the ESM stuff, wasn't too familiar with how jest was working with it. Let me know what you think

blva:
@gagik works for me, can you fix the tests in the PR so we can merge to start adding mroe tests

gagik:
@blva Actually upon further investigation, I think this is actually the best experience after all, mocha isn't much different 😅 Sorry for the bunch of back and forth, was just trying to make sure we're not regressing the TS setup but this should work best.

If we can rename `__tests__/` => `tests/`, LGTM

gagik (approved):

PR #14, opened 2025-04-08
feat: add data access tools

nirinchev:
Adds the following data-access tools:
- connect
- list-databases
- list-collections
- collection-schema
- collection-indexes
- collection-storage-size
- db-stats
- find
- aggregate
- count
- create-index
- insert-one
- insert-many
- delete-one
- delete-many
- drop-collection
- drop-database
- update-one
- update-many
- rename-collection

Follow-ups:
- update-one/many doesn't support aggregation update

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,6 +14,7 @@
   "publishConfig": {
     "access": "public"
   },
+  "type": "module",
   "scripts": {
     "prepare": "npm run build",
     "build:clean": "rm -rf dist",
@@ -40,11 +41,16 @@
     "globals": "^16.0.0",
     "prettier": "^3.5.3",
     "typescript": "^5.8.2",
-    "typescript-eslint": "^8.29.1",
-    "zod": "^3.24.2"
+    "typescript-eslint": "^8.29.1"
   },
   "dependencies": {
-    "@types/express": "^5.0.1"
+    "@mongodb-js/devtools-connect": "^3.7.2",
+    "@mongosh/service-provider-node-driver": "^3.6.0",
+    "@types/express": "^5.0.1",
+    "bson": "^6.10.3",
+    "mongodb": "^6.15.0",
+    "mongodb-schema": "^12.6.2",
+    "zod": "^3.24.2"
   },
   "engines": {
     "node": ">=23.0.0"
--- a/src/config.ts
+++ b/src/config.ts
@@ -5,7 +5,7 @@ import { fileURLToPath } from "url";
 const __filename = fileURLToPath(import.meta.url);
 const __dirname = path.dirname(__filename);
 
-const packageMetadata = fs.readFileSync(path.resolve(path.join(__dirname, "..", "package.json")), "utf8");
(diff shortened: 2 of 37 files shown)


package.json
```diff
@@ -14,6 +14,7 @@
   "publishConfig": {
     "access": "public"
   },
+  "type": "module",
```
nirinchev, line 17: Had to add this since I was seeing complaints that node was first trying to parse it as cjs, but was seeing an esm. I don't have a strong opinion on whether we should go with one or the other.

src/config.ts
```diff
@@ -1,7 +1,11 @@
 import path from "path";
 import fs from "fs";
+import { fileURLToPath } from "url";
 
-const packageMetadata = fs.readFileSync(path.resolve("./package.json"), "utf8");
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = path.dirname(__filename);
+
+const packageMetadata = fs.readFileSync(path.join(__dirname, "..", "package.json"), "utf8");
```
nirinchev, line 8: This may have regressed at some point, but `path.resolve("./package.json")` is looking at `dist/config.js` rather than in the parent directory. For some reason, Claude is not happy about relative paths, so `"../package.json"` didn't work either, so I needed the `__dirname` shenanigans above, since esm doesn't have `__dirname` built in.

src/tools/tool.ts
```diff
@@ -1,22 +1,24 @@
...
 import { State } from "../state.js";
 
+export type ToolArgs<Args extends ZodRawShape> = z.objectOutputType<Args, z.ZodNever>;
+
 export abstract class ToolBase<Args extends ZodRawShape> {
     protected abstract name: string;
 
     protected abstract description: string;
 
     protected abstract argsShape: Args;
```
```suggestion
export abstract class ToolBase {
    protected abstract name: string;

    protected abstract description: string;

    protected abstract argsShape: ZodRawShape;
    
    protected abstract execute(args: ToolArgs<typeof this.argsShape>): Promise<CallToolResult>;
```
I went on a rabbithole trying to get rid of some redundancy here. Unfortunately I don't think the `Args` generic is doing anything for us.

I would instead get rid of it and use `typeof this.argsShape`. This is still not really going to be a strict type as is. For example in `find.ts` in both cases you can define execute to use `ToolArgs<ZodRawShape>` and it'd still pass fine so unfortunately TypeScript isn't that strict/smart anyhow.

Same for `MongoDBToolBase`.

Then we can simply define `argsShape` inside the class as is without replicating it

gagik (changes requested):

src/tools/tool.ts
nirinchev: Yeah, I couldn't get typescript to behave exactly how I wanted it to, but figured it's better this way than rewriting the entire server in a proper language 🥲 Anyway, agree with your assessment and have removed the args generic.

gagik: yeah I was also hoping there's a better way but couldn't come up with an alternative either 🥲

PR #15, opened 2025-04-08
chore: move atlas projects and clusters into resources

fmenezes:


fmenezes:
after testing resources I realized Visual Studio Code and Claude did not pick up information out of the box, only when I explicitly added the context manually, closing for now we might revisit in future.

PR #29, opened 2025-04-09
feat: add atlas-create-free-cluster atlas-inspect-cluster tools

fmenezes:
- add `atlas-create-free-cluster` tool
- add `atlas-inspect-cluster` tool
- rename `auth` tool to `atlas-auth` tool
- rename `listClusters` tool to `atlas-list-clusters` tool
- rename `listProjects` tool to `atlas-list-projects` tool
- add `npm run generate` to auto generate required API types from openapi spec


package.json
```diff
@@ -27,21 +27,32 @@
     "check": "npm run check:lint && npm run check:format",
     "check:lint": "eslint .",
     "check:format": "prettier -c .",
-    "reformat": "prettier --write ."
+    "reformat": "prettier --write .",
+    "generate:download": "curl -Lo ./scripts/spec.json https://github.com/mongodb/openapi/raw/refs/heads/main/openapi/v2/openapi-2025-03-12.json",
+    "generate:filter": "tsx ./scripts/filter.ts > ./scripts/filteredSpec.json < ./scripts/spec.json",
+    "generate:bundle": "redocly bundle --ext json --remove-unused-components ./scripts/filteredSpec.json --output ./scripts/bundledSpec.json",
+    "generate:openapi": "openapi-typescript ./scripts/bundledSpec.json --root-types-no-schema-prefix --root-types --output ./src/common/atlas/openapi.d.ts",
+    "generate:clear": "rm -rf ./scripts/bundledSpec.json ./scripts/filteredSpec.json ./scripts/spec.json",
```
nirinchev: Super nit, but these seem to be very dependent on the order of execution, so not sure if it's that useful to have them as separate scripts - it doesn't seem like they'd be really useful on their own.

src/common/atlas/client.ts
```diff
@@ -120,10 +102,13 @@ export class ApiClient {
                 ...options?.headers,
             },
         };
+
+        console.error(`Calling Atlas API: ${url.toString()}`);
+        console.error(`with: ${JSON.stringify(opt)}`);
```
nirinchev: [nit] This is inconsequential until we have https://github.com/mongodb-js/mongodb-mcp-server/issues/26, but probably a good idea to already start using `logger.ts/log` instead of console logging. Also, do we want to keep this code - might be risky depending on what's in this object, but definitely looks like it could leak some sensitive headers.

nirinchev (approved):
Looks good, just a few minor notes.

package.json
fmenezes: the only reason they are separated is that it was challenging to implement in one go, we could move it to bash but it also felt a bit weird to do so

src/common/atlas/client.ts
fmenezes: Left by mistake, I was debugging, removed

package.json
```diff
@@ -24,24 +24,35 @@
     "build": "npm run build:clean && npm run build:compile && npm run build:addshebang && npm run build:chmod",
     "inspect": "npm run build && mcp-inspector -- dist/index.js",
     "prettier": "prettier",
-    "check": "npm run check:lint && npm run check:format",
+    "check": "npm run build && npm run check:lint && npm run check:format",
```
fmenezes, line 27: @nirinchev I'm adding build here, I've seen some weird bug where the source was not compiling but the styles were passing 🤷‍♂️

package.json
```diff
@@ -24,24 +24,35 @@
...
     "check:lint": "eslint .",
     "check:format": "prettier -c .",
-    "reformat": "prettier --write ."
+    "reformat": "prettier --write .",
+    "generate:download": "curl -Lo ./scripts/spec.json https://github.com/mongodb/openapi/raw/refs/heads/main/openapi/v2/openapi-2025-03-12.json",
+    "generate:filter": "tsx ./scripts/filter.ts > ./scripts/filteredSpec.json < ./scripts/spec.json",
+    "generate:bundle": "redocly bundle --ext json --remove-unused-components ./scripts/filteredSpec.json --output ./scripts/bundledSpec.json",
+    "generate:openapi": "openapi-typescript ./scripts/bundledSpec.json --root-types-no-schema-prefix --root-types --output ./src/common/atlas/openapi.d.ts",
```
blva: nit: can create an issue if it requires more changes - can we have per-path or per operationID files?

package.json
nirinchev: Yeah, sounds good. I was thinking of adding a build step on CI anyway, but that's also fine. Our project builds fairly quickly still, so I'm not really worried about the extra couple of seconds it adds.

blva (dismissed):

package.json
blva: nit: can be a follow-up PR- we should update our readme with the right commands to run

package.json
fmenezes: replaced with a bash

package.json
fmenezes: right now `npm run check` should check everything

package.json
fmenezes: I don't think the tool supports it

blva (approved):

nirinchev (approved):

PR #32, opened 2025-04-09
feat: add atlas access list tools

### LINKED ISSUE #30: [Atlas] Implement tool to add IP/CIDR access to Clusters/Projects
Clusters are not accessible unless we whitelist IPs/CIDR blocks

fmenezes:
fixes #30 

- added `atlas-inspect-access-list` tool
- added `atlas-create-access-list` tool

### DIFF
--- a/src/tools/atlas/createAccessList.ts
+++ b/src/tools/atlas/createAccessList.ts
@@ -2,7 +2,6 @@ import { z } from "zod";
 import { CallToolResult } from "@modelcontextprotocol/sdk/types.js";
 import { AtlasToolBase } from "./atlasTool.js";
 import { ToolArgs } from "../tool.js";
-import { NetworkPermissionEntry } from "../../common/atlas/openapi.js";
 
 const DEFAULT_COMMENT = "Added by Atlas MCP";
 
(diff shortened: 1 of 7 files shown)


src/tools/atlas/createAccessList.ts
```diff
@@ -0,0 +1,58 @@
...
+    protected async execute({
+        projectId,
+        ipAddresses,
+        cidrBlocks,
+        comment,
+    }: ToolArgs<typeof this.argsShape>): Promise<CallToolResult> {
+        await this.ensureAuthenticated();
+
+        if (!ipAddresses?.length && !cidrBlocks?.length) {
+            throw new Error("Either ipAddresses or cidrBlocks must be provided.");
```
nirinchev, line 30: Should we default to the device's public IP address? Alternatively, we can lookup the current IP address and respond with a prompt whether to add it - something like:

```ts
return {
  content: [{
    type: "text",
    text: "Either an IP address or a CIDR block must be provided. The current device's public IP address is: a.b.c.d - do you want to use that?"
  }]
}
```

nirinchev (approved):

src/tools/atlas/createAccessList.ts
fmenezes: thanks for pointing out, I'll handle current ips on a follow up PR, I'll check if makes sense to have a resource for this one

fmenezes: added https://github.com/mongodb-js/mongodb-mcp-server/issues/36

Issue #21, opened 2025-04-09
[MongoDB] Add config options for defaults we don't want to set for every operation


nirinchev:
We should allow users to configure read preference, write concern, etc. without exposing them for every operation.

fmenezes:
most MCP servers are configured via env vars our config.ts already reads env vars, I think we should assume that is the safest bet

PR #35, opened 2025-04-09
feat: add logging

nirinchev:


fmenezes:
have you considered https://modelcontextprotocol.io/specification/2025-03-26/server/utilities/logging ?

nirinchev:
Yes, I wanted to integrate that as well, but got pretty late last night, so decided to leave it for today 😅 I still think it'd be valuable to both send the logs to the client, but also store them on disk.

fmenezes:
If you want to add it to disk we run the risk of adding to too many logs needing to implement log rotation, where to you want to send the logs to? Should we truncate?

nirinchev:
@fmenezes I'm using https://github.com/mongodb-js/devtools-shared/tree/main/packages/mongodb-log-writer for writing the logs on disk. It's the same package we use in mongosh and Compass and it implements log cleanup based on retention policy and storage limits. Right now, I've hardcoded some default values, but once we get to https://github.com/mongodb-js/mongodb-mcp-server/issues/21, we'd be able to expose them as config options for users to tweak.

src/config.ts
```diff
@@ -15,6 +16,19 @@ export const config = {
...
+function getLocalDataPath() {
+    if (process.platform === "win32") {
+        const appData = process.env.APPDATA;
+        const localAppData = process.env.LOCALAPPDATA ?? process.env.APPDATA;
+        if (localAppData && appData) {
+            return path.join(localAppData, "mongodb", "mongodb-mcp");
+        }
+    }
+
+    return path.join(os.homedir(), ".mongodb", "mongodb-mcp");
```
nirinchev, line 33: This is the same parent folder as the one we use for mongosh/compass.

src/logger.ts
```diff
@@ -1,4 +1,119 @@
...
+    const logDir = path.join(config.localDataPath, ".app-logs");
+    await fs.mkdir(logDir, { recursive: true });
+
+    const manager = new MongoLogManager({
+        directory: path.join(config.localDataPath, ".app-logs"),
+        retentionDays: 30,
+        onwarn: console.warn,
+        onerror: console.error,
+        gzip: false,
+        retentionGB: 1,
```
nirinchev, line 112: Defaulting to 30 day retention or 1 gb of storage. We'll want to expose these as config options at a later point.

PR #40, opened 2025-04-10
chore: update readme tools

blva:


fmenezes:
@nirinchev I don't know if it makes sense to prefix all tools with mongodb-<something> and altas ones as `mongodb-atlas-<something>` because if we have many other mcp servers these ones can be hard to find/use

PR #43, opened 2025-04-10
chore: refactor package.json import

fmenezes:


src/types/mongodb-redact.d.ts
```diff
@@ -0,0 +1,4 @@
+declare module "mongodb-redact" {
```
fmenezes, line 1: for some reason `mongodb-redact` types are not compatible

src/config.ts
```diff
@@ -1,13 +1,6 @@
...
-import fs from "fs";
-import { fileURLToPath } from "url";
 import os from "os";
-
-const __filename = fileURLToPath(import.meta.url);
-const __dirname = path.dirname(__filename);
-
-const packageMetadata = fs.readFileSync(path.join(__dirname, "..", "package.json"), "utf8");
-const packageJson = JSON.parse(packageMetadata);
+import packageJson from "../package.json" with { type: "json" };
```
nirinchev, line 3: Does this work on windows?

fmenezes: I haven't tested anything on windows yet, perhaps we should start with integration tests and run them on linux, windows and macos

fmenezes: added https://github.com/mongodb-js/mongodb-mcp-server/issues/44

nirinchev (approved):
Looks good - we can look into adding some smoke tests later today or tomorrow.

src/types/mongodb-redact.d.ts
gagik: try adding `"allowSyntheticDefaultImports": "true"` to tsconfig.json and running `npm i` again

gagik (changes requested):
Would be good to add `allowSyntheticDefaultImports`

src/types/mongodb-redact.d.ts
gagik: you should be able to delete this file now

fmenezes:
> Nice 🚀 but would be good to add `allowSyntheticDefaultImports` so we avoid the manual types

@gagik I've enabled it but in the case of `mongodb-redact` it does not solve the issue

gagik (approved):
Oh okay, I suppose they misconfigured something there 🤷

Issue #20, opened 2025-04-09
Investigate secure storage options


nirinchev:
We need to persist the Atlas credentials in an encrypted secure storage location.

PR #45, opened 2025-04-10
feat: add support for supplying config options

### LINKED ISSUE #21: [MongoDB] Add config options for defaults we don't want to set for every operation

nirinchev:
Adds the following mechanisms for supplying config (in order of precedence):
- cli args
- environment variables (field names are converted to `MDB_MCP_UPPER_SNAKE_CASE`
- `mongodb-mcp.conf` placed in the `/etc` folder on Unix or `%LOCAL_APP_DATA%\mongodb\mongodb-mcp`.

Fixes #21

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -13,17 +13,20 @@ interface UserConfig extends Record<string, string> {
     apiBaseUrl: string;
     clientId: string;
     stateFile: string;
-    projectId: string;
 }
 
 const defaults: UserConfig = {
     apiBaseUrl: "https://cloud.mongodb.com/",
     clientId: "0oabtxactgS3gHIR0297",
     stateFile: path.join(localDataPath, "state.json"),
-    projectId: "",
 };
 
-const mergedUserConfig = Object.assign({}, defaults, getFileConfig(), getEnvConfig(), getCliConfig());
+const mergedUserConfig = {
+    ...defaults,
+    ...getFileConfig(),
+    ...getEnvConfig(),
+    ...getCliConfig(),
+};
 
 const __filename = fileURLToPath(import.meta.url);
 const __dirname = path.dirname(__filename);
@@ -68,12 +71,12 @@ function getEnvConfig(): Partial<UserConfig> {
     };
 
     const result: Partial<UserConfig> = {};
-    Object.keys(defaults).forEach((key) => {
+    for (const key of Object.keys(defaults)) {
         const envVarName = `MDB_MCP_${camelCaseToSNAKE_UPPER_CASE(key)}`;
         if (process.env[envVarName]) {
-            result[key as keyof UserConfig] = process.env[envVarName];
+            result[key] = process.env[envVarName];
         }
-    });
+    }
 
     return result;
 }
--- a/src/config.ts
+++ b/src/config.ts
@@ -4,7 +4,7 @@ import argv from "yargs-parser";
 
 import packageJson from "../package.json" with { type: "json" };
 import fs from "fs";
-const localDataPath = getLocalDataPath();
+const { localDataPath, configPath } = getLocalDataPath();
 
 // If we decide to support non-string config options, we'll need to extend the mechanism for parsing
 // env variables.
@@ -37,22 +37,28 @@ const config = {
 
 export default config;
 
-function getLocalDataPath(): string {
-    let result: string | undefined;
+function getLocalDataPath(): { localDataPath: string; configPath: string } {
+    let localDataPath: string | undefined;
+    let configPath: string | undefined;
 
     if (process.platform === "win32") {
         const appData = process.env.APPDATA;
         const localAppData = process.env.LOCALAPPDATA ?? process.env.APPDATA;
         if (localAppData && appData) {
-            result = path.join(localAppData, "mongodb", "mongodb-mcp");
+            localDataPath = path.join(localAppData, "mongodb", "mongodb-mcp");
+            configPath = path.join(localAppData, "mongodb", "mongodb-mcp.conf");
         }
     }
 
-    result ??= path.join(os.homedir(), ".mongodb", "mongodb-mcp");
+    localDataPath ??= path.join(os.homedir(), ".mongodb", "mongodb-mcp");
+    configPath ??= "/etc/mongodb-mcp.conf";
 
-    fs.mkdirSync(result, { recursive: true });
+    fs.mkdirSync(localDataPath, { recursive: true });
 
-    return result;
+    return {
+        localDataPath,
+        configPath,
+    };
 }
 
 // Gets the config supplied by the user as environment variables. The variable names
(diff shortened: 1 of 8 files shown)


src/config.ts
```diff
@@ -3,25 +3,45 @@ import fs from "fs";
...
+    clientId: string;
+    stateFile: string;
+    projectId: string;
+}
+
+const cliConfig = argv(process.argv.slice(2)) as unknown as Partial<UserConfig>;
+
+const defaults: UserConfig = {
+    apiBaseUrl: "https://cloud.mongodb.com/",
+    clientId: "0oabtxactgS3gHIR0297",
```
gagik, line 19: where is this clientId coming from?

src/config.ts
```diff
@@ -32,3 +52,50 @@ function getLocalDataPath() {
...
+    for (const key of Object.keys(defaults)) {
+        for (const partialConfig of partialConfigs) {
+            if (partialConfig[key]) {
+                mergedConfig[key] = partialConfig[key];
+            }
+        }
+    }
+
+    return mergedConfig;
+}
```
```suggestion
function mergeConfigs(defaults: UserConfig, ...partialConfigs: Array<Partial<UserConfig>>): UserConfig {
  return Object.assign({}, defaults, ...partialConfigs);
}
```

I believe this should do the same trick; we could get rid of the function also. 
The functional difference can be that if partial config has more fields than the default those would also get assigned but considering they're meant to be partial, not a case we need to worry about necessarily, unless we're worried about some fields or hidden options being overriden

src/config.ts
nirinchev: It was there already - @fmenezes?

src/tools/mongodb/connect.ts
```diff
@@ -71,5 +72,7 @@ export class ConnectTool extends MongoDBToolBase {
         });
 
         this.mongodbState.serviceProvider = provider;
+        this.state.connectionString = connectionString;
+        await saveState(this.state);
```
gagik, line 76: do I understand correct that we'd save the conectionString with its credentials in a plaintext file this way?

nirinchev: Yes - this should be handled in https://github.com/mongodb-js/mongodb-mcp-server/issues/20.

gagik (approved):

src/tools/mongodb/connect.ts
fmenezes: question: why do we need `mongodbState` and also `state` ? does `mongodbState` get's persisted?

src/config.ts
```diff
@@ -3,32 +3,95 @@ import fs from "fs";
...
+        return str.replace(/([a-z])([A-Z])/g, "$1_$2").toUpperCase();
+    };
+
+    const result: Partial<UserConfig> = {};
+    Object.keys(defaults).forEach((key) => {
+        const envVarName = `MDB_MCP_${camelCaseToSNAKE_UPPER_CASE(key)}`;
+        if (process.env[envVarName]) {
+            result[key as keyof UserConfig] = process.env[envVarName];
+        }
+    });
```
```suggestion
    for (key of Object.keys(defaults)) {
        const envVarName = `MDB_MCP_${camelCaseToSNAKE_UPPER_CASE(key)}`;
        if (process.env[envVarName]) {
            result[key as keyof UserConfig] = process.env[envVarName];
        }
    }
```
nit: more of personal preference but better to use loops generally

src/config.ts
fmenezes: that is correct, it is needed for oauth device flow

gagik (approved):
LGTM

src/config.ts
```diff
@@ -3,32 +3,95 @@ import fs from "fs";
...
+    clientId: string;
+    stateFile: string;
+    projectId: string;
+}
+
+const defaults: UserConfig = {
+    apiBaseUrl: "https://cloud.mongodb.com/",
+    clientId: "0oabtxactgS3gHIR0297",
+    stateFile: path.join(localDataPath, "state.json"),
+    projectId: "",
```
fmenezes: I think we have removed this `projectId` configuration

src/tools/mongodb/connect.ts
nirinchev: No - mongodb state is a set of globals that will not get stored on disk. I was debating between adding them to state and then special-casing which fields get persisted but at the end opted to create a separate object.

src/config.ts
```diff
@@ -3,32 +3,95 @@ import fs from "fs";
...
+};
+
+const mergedUserConfig = Object.assign({}, defaults, getFileConfig(), getEnvConfig(), getCliConfig());
```
fmenezes: nit: something like this would also work?
```suggestion
const mergedUserConfig = {
  ...defaults,
  ...getFileConfig(),
  ...getEnvConfig(),
  ...getCliConfig()
};
```

fmenezes:
@nirinchev are we merging state and config this way? in cli our approach has been one file locations are as follows

- windows `%APPDATA%/atlascli/config.toml`
- macos `$HOME/Library/Application Support/atlascli/config.toml`
- linux `$XDG_CONFIG_HOME/atlascli/config.toml` (in absence of `$XDG_CONFIG_HOME` we use `$HOME/.config`)

which seems to be the most common across cli tools I've seen

src/tools/mongodb/connect.ts
fmenezes: so far config is read only from env vars / defaults and state is something we expect to store (as in preferences).

I think we are merging these two concepts together in this PR, which I'm ok with.

In theory `mongodbState` is `mongodbConfig` then.

nirinchev:
Hm... good point - I thought the app directory could be convenient place to store it, but it's probably not that intuitive. Happy to change it to either the Atlas CLI or the mongosh locations.

fmenezes:
I think we should do 

- windows `%APPDATA%/mongodb-mcp-server/config.json`
- macos `$HOME/Library/Application Support/mongodb-mcp-server/config.json`
- linux `$XDG_CONFIG_HOME/mongodb-mcp-server/config.json` (in absence of $XDG_CONFIG_HOME we use `$HOME/.config`)

and replace `state.json` with `credentials.json`.

src/tools/mongodb/connect.ts
nirinchev: Technically, it's not even a config - it's just a place to stash certain things we may need for later (such as the MongoDB client after `connect` is invoked). This is technically session state as opposed to the current state, which is persistent across sessions. I'll probably clean it up in a separate PR.

fmenezes (approved):
All of my comments are nits and I'm good to resolve, in fact we would be ok with just env vars and arguments for configuration

nirinchev:
Okay, so for now we'll look for config in:

- `/etc/mongodb-mcp.conf` on macOS/Linux
- `%APPDATA%\mongodb\mongodb-mcp\mongodb-mcp.conf` on Windows

This almost matches mongosh (except for the fallbacks on macOS). Since I wanted to keep it simple for now and keep the config getting loaded synchronously, I'll keep it like that. We can (and should) definitely refactor it to load the config asynchronously at some point, in which case we can add the extra fallback locations.

I'll look into the state situation at a follow-up PR.

Issue #33, opened 2025-04-09
[Atlas] Investigate openapi-fetch


fmenezes:
Lets check if https://openapi-ts.dev/openapi-fetch/ might simplify api client

PR #47, opened 2025-04-10
refactor: use openapi-fetch

fmenezes:
fixes https://github.com/mongodb-js/mongodb-mcp-server/issues/33

### DIFF
--- a/src/common/atlas/apiClient.ts
+++ b/src/common/atlas/apiClient.ts
@@ -36,10 +36,7 @@ export class ApiClientError extends Error {
         message ||= `error calling Atlas API`;
         try {
             const text = await response.text();
-            return new ApiClientError(
-                `${message}: [${response.status} ${response.statusText}] ${text}`,
-                response
-            );
+            return new ApiClientError(`${message}: [${response.status} ${response.statusText}] ${text}`, response);
         } catch {
             return new ApiClientError(`${message}: ${response.status} ${response.statusText}`, response);
         }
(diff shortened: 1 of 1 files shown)


src/common/atlas/apiClient.ts
```diff
@@ -48,32 +39,40 @@ export interface ApiClientOptions {
...
+            if (await apiClient.validateToken()) {
+                request.headers.set("Authorization", `Bearer ${apiClient.token?.access_token}`);
+                return request;
+            }
+        },
+    });
+    private errorMiddleware = (): Middleware => ({
+        async onResponse({ response }) {
+            if (!response.ok) {
+                throw new ApiClientError(`Error calling Atlas API: ${await response.text()}`, response);
```
gagik: are we guaranteed the response text? is there a chance this could error as well?

gagik (approved):
Looks good to me, I think I'll try to introduce some more linting rules for more consistency with i.e. strings and arrow functions later.

src/common/atlas/apiClient.ts
fmenezes: addressed

src/common/atlas/apiClient.ts
```diff
@@ -48,32 +39,51 @@ export interface ApiClientOptions {
...
+                return request;
+            }
+        },
+    });
+    private errorMiddleware = (): Middleware => ({
+        async onResponse({ response }) {
+            if (!response.ok) {
+                try {
+                    const text = await response.text();
+                    throw new ApiClientError(
```
gagik: this does mean that what we'll get is 
`Error calling Atlas API: Error calling Atlas API:... `

fmenezes: More like `Error calling Atlas API: [400 Bad Request] Invalid json or some other text from backend`

gagik: I think passing the response object is sufficient, we can just omit the response text

fmenezes: What do you mean? response is a property and does not display error message directly. You prefer to hide error message?

gagik: I'd say so. If we want to include properties of the response inside the message (including in format `[${response.status} ${response.statusText}]`) then let's do it inside the `ApiClientError` class's `super` call to message so it's standardized and enforced. And in that sense avoid anything async.

PR #48, opened 2025-04-10
feat: refactor state, store it securely

nirinchev:
Uses @napi-rs/keyring to store the credentials from the state in the OS secure storage location.

Additionally, reactors `State` to have a `persistent` field which gets stored and `session` field where we can add global state that should not persist across launches. For now, this is only the `NodeDriverServiceProvider`, which was previously in a now-deleted `MongoDBToolState`.

The node library we're using is a wrapper around https://github.com/open-source-cooperative/keyring-rs, which is a rust-based implementation wrapping the native credential stores for most desktop operating systems.

Fixes https://github.com/mongodb-js/mongodb-mcp-server/issues/20

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,6 +11,7 @@
       "dependencies": {
         "@mongodb-js/devtools-connect": "^3.7.2",
         "@mongosh/service-provider-node-driver": "^3.6.0",
+        "@napi-rs/keyring": "^1.1.6",
         "@types/express": "^5.0.1",
         "bson": "^6.10.3",
         "mongodb": "^6.15.0",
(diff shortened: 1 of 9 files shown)


fmenezes:
LGTM, personally I would separate `session` and `credentials` and remove `state` from the picture

nirinchev:
@gagik @fmenezes I used your feedback as inspiration and tried to do something in the middle 😅 I'm keeping the credentials and the rest of the state combined, but they're now encapsulated in a class responsible for saving/loading of data. I decided to keep them together until we decide what we want to do with the auth flows and whether we want to persist data at all, but should be fairly trivial to rip out if necessary.

blva (approved):
nice, LGTM 🚀

Issue #36, opened 2025-04-10
[Atlas] Add currentIp to access lists


fmenezes:
Surface current ip for create access list tool

PR #52, opened 2025-04-10
feat: add currentIp option atlas-create-access-list tool

fmenezes:
fixes https://github.com/mongodb-js/mongodb-mcp-server/issues/36

### DIFF
--- a/src/common/atlas/apiClient.ts
+++ b/src/common/atlas/apiClient.ts
@@ -252,6 +252,30 @@ export class ApiClient {
         }
     }
 
+    async getIpInfo() {
+        await this.validateToken();
+
+        const endpoint = "api/private/ipinfo";
+        const url = new URL(endpoint, config.apiBaseUrl);
+        const response = await fetch(url, {
+            method: "GET",
+            headers: {
+                Accept: "application/json",
+                Authorization: `Bearer ${this.token?.access_token}`,
+                "User-Agent": config.userAgent,
+            },
+        });
+
+        if (!response.ok) {
+            throw await ApiClientError.fromResponse(response);
+        }
+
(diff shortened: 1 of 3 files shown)


src/common/atlas/apiClient.ts
```diff
@@ -252,6 +252,30 @@ export class ApiClient {
         }
     }
 
+    async getIpInfo() {
+        await this.validateToken();
```
nirinchev: Should we check the return of this call?

src/common/atlas/apiClient.ts
```diff
@@ -252,6 +252,30 @@ export class ApiClient {
...
+
+        const endpoint = "api/private/ipinfo";
+        const url = new URL(endpoint, config.apiBaseUrl);
+        const response = await fetch(url, {
+            method: "GET",
+            headers: {
+                Accept: "application/json",
+                Authorization: `Bearer ${this.token?.access_token}`,
```
nirinchev: [nit] Assuming this.token is never null if `this.validateToken` returned successfully.

```suggestion
                Authorization: `Bearer ${this.token!.access_token}`,
```

src/tools/atlas/createAccessList.ts
```diff
@@ -36,6 +38,16 @@ export class CreateAccessListTool extends AtlasToolBase {
             comment: comment || DEFAULT_COMMENT,
         }));
 
+        if (currentIpAddress) {
+            const currentIp = await this.apiClient!.getIpInfo();
```
nirinchev:
```suggestion
            const currentIp = await this.apiClient.getIpInfo();
```

nirinchev (approved):

PR #55, opened 2025-04-10
refactor: rename state to session, combine tool registration, and clearer dependencies

gagik:
Just makes it clearer what goes where. I think session was a lot easier for me to reason about then state.
This did end up a bit all over the place because of the different changes in the codebase but hopefully is still reviewable.


nirinchev (approved):

src/index.ts
```diff
@@ -1,17 +1,33 @@
...
-export async function runServer() {
-    const server = new Server();
+try {
+    const state = new State();
+    await state.loadCredentials();
 
-    const transport = new StdioServerTransport();
-    await server.connect(transport);
-}
+    const apiClient = ApiClient.fromState(state);
```
nirinchev: When doing https://github.com/mongodb-js/mongodb-mcp-server/pull/58, I realized the server wrapper was useful for testing. Moving all of this setup to a top-level script will make it difficult to mock/test.

nirinchev (changes requested):

src/index.ts
gagik: re-introduced the class, would be happy to rebase this again after merging #58

src/tools/atlas/atlasTool.ts
```diff
@@ -1,8 +1,11 @@
 import { ToolBase } from "../tool.js";
-import { State } from "../../state.js";
+import { ApiClient } from "../../common/atlas/apiClient.js";
+import { Session } from "../../session.js";
 
 export abstract class AtlasToolBase extends ToolBase {
-    constructor(state: State) {
-        super(state);
+    private apiClient?: ApiClient;
```
fmenezes: do we need this line?
```suggestion
```

src/tools/atlas/tools.ts
```diff
@@ -10,19 +7,13 @@ import { InspectAccessListTool } from "./inspectAccessList.js";
...
-        new InspectAccessListTool(state),
-        new ListDBUsersTool(state),
-        new CreateDBUserTool(state),
-    ];
-
-    for (const tool of tools) {
-        tool.register(server);
-    }
-}
+export const AtlasTools = [
```
fmenezes, line 10: 💯

src/tools/atlas/atlasTool.ts
gagik: good catch, that was leftover from bunch of moving around

fmenezes (approved):

nirinchev (approved):

PR #56, opened 2025-04-10
Add config option for connection string

nirinchev:
Adds the following connect options:
- `readPreference`
- `readConcern`
- `writeConcern.w`
- `timeoutMS`

I've tried to strike a balance between giving users control and not going overboard wtih config options that may never get used.

Additionally, refactored the env variable parsing where instead of using the defaults to look up env variables, we now construct an object from all env variables that start with `MDB_MCP_`.


src/config.ts
```diff
@@ -66,21 +79,51 @@ function getLocalDataPath(): { localDataPath: string; configPath: string } {
...
+            obj[currentField] = {};
         }
+
+        setValue(obj[currentField] as Record<string, unknown>, path, value);
+    }
+
+    const result: Record<string, unknown> = {};
+    for (const [key, value] of Object.entries(process.env).filter(
+        ([key, value]) => value !== undefined && key.startsWith("MDB_MCP_")
+    )) {
```
```suggestion
    const mcpEnvironmentValues = Object.entries(process.env).filter(
        ([key, value]) => value !== undefined && key.startsWith("MDB_MCP_")
    ) as [string, string][];
    for (const [key, value] of mcpEnvironmentValues) {
```

gagik: very nit: yeah technically this is as bad as the type assertion but seems preferable it as the type cast happens at the point where we get the data rather than when we try to consume it

src/config.ts
```diff
@@ -66,21 +79,51 @@ function getLocalDataPath(): { localDataPath: string; configPath: string } {
...
+        const fieldPath = key
+            .replace("MDB_MCP_", "")
+            .split(".")
+            .map((part) => SNAKE_CASE_toCamelCase(part));
+
+        setValue(result, fieldPath, value!);
```
```suggestion
        setValue(result, fieldPath, value);
```

src/config.ts
```diff
@@ -4,21 +4,34 @@ import argv from "yargs-parser";
 
 import packageJson from "../package.json" with { type: "json" };
 import fs from "fs";
+import { ReadConcernLevel, ReadPreferenceMode, W } from "mongodb";
 const { localDataPath, configPath } = getLocalDataPath();
 
 // If we decide to support non-string config options, we'll need to extend the mechanism for parsing
 // env variables.
-interface UserConfig extends Record<string, string | undefined> {
+interface UserConfig extends Record<string, unknown> {
```
```suggestion
interface UserConfig {
```
nit: is there a benefit from extending Record?

src/tools/mongodb/mongodbTool.ts
```diff
@@ -19,8 +21,12 @@ export abstract class MongoDBToolBase extends ToolBase {
 
     protected abstract operationType: DbOperationType;
 
-    protected ensureConnected(): NodeDriverServiceProvider {
+    protected async ensureConnected(): Promise<NodeDriverServiceProvider> {
         const provider = this.state.serviceProvider;
+        if (!provider && config.connectionString) {
+            await connectToMongoDB(config.connectionString, this.state);
```
gagik: since `connectToMongoDB` already indirectly lives inside `MongoDBToolBase`, it could also be part of its class and then be inherited by the connect tool

gagik (approved):

Issue #39, opened 2025-04-10
[Atlas] Use service accounts for auth


fmenezes:

wtrocki:
I assume that would be TypeScriptAPI access? 
We can follow our guide: https://www.mongodb.com/developer/products/mongodb/mongodb-atlas-authentication-using-service-accounts-oauth/

PR #59, opened 2025-04-11
feat: change Atlas API auth to service accounts

fmenezes:
fixes https://github.com/mongodb-js/mongodb-mcp-server/issues/39


README.md
```diff
@@ -146,10 +149,91 @@ Paste the mcp server configuration into the file
...
+3. Configuration file
+4. Default values
+
+### Configuration Options
+
+| Option             | Description                                                                 |
+| ------------------ | --------------------------------------------------------------------------- |
+| `apiClientId`      | Atlas API client ID for authentication                                      |
+| `apiClientSecret`  | Atlas API client secret for authentication                                  |
+| `stateFile`        | Path to store application state (default ~/.mongodb/mongodb-mcp/state.json) |
```
nirinchev: We can drop this now - I forgot to remove it as part of https://github.com/mongodb-js/mongodb-mcp-server/pull/48 😅

src/tools/atlas/createAccessList.ts
```diff
@@ -39,7 +39,7 @@ export class CreateAccessListTool extends AtlasToolBase {
         }));
 
         if (currentIpAddress) {
-            const currentIp = await this.apiClient.getIpInfo();
+            const currentIp = await this.apiClient!.getIpInfo();
```
gagik, line 42: I think the TypeScript setup in VSCode might be misconfigured, there shouldn't be a need for these assertions here (otherwise I'd have expected for it to be caught by now during build).

Can you try `Command + Shift + P` > `Select TypeScript Version` > `Use workspace's version`

README.md
```diff
@@ -146,10 +149,91 @@ Paste the mcp server configuration into the file
...
+| `connectionString` | MongoDB connection string for direct database connections                   |
```
nirinchev: We should probably indicate this is optional and if not provided, the `connect` tool should be used.

src/state.ts
```diff
@@ -1,25 +1,15 @@
...
 import logger from "./logger.js";
 import { mongoLogId } from "mongodb-log-writer";
 
 interface Credentials {
-    auth: {
-        status: "not_auth" | "requested" | "issued";
-        code?: OauthDeviceCode;
-        token?: OAuthToken;
-    };
     connectionString?: string;
```
nirinchev, line 7: Per slack discussion, let's drop the connection string storage and rely on users either providing it every time or supplying it as env variable. I know this will turn the credentials field to an empty object, but we could keep it around for persisting sensitive data after we go through security review. It will also allow the security team to review the mechanisms we intend to use for secure storage.

src/tools/atlas/createAccessList.ts
fmenezes: check https://github.com/mongodb-js/mongodb-mcp-server/pull/59/files#diff-6936a378894754df6d0ac7af88a816f2728450ee9e9c4170388944fa75b8d072R8 we need them now

blva (commented):
LGTM

src/tools/atlas/atlasTool.ts
```diff
@@ -1,17 +1,20 @@
...
 export abstract class AtlasToolBase extends ToolBase {
     constructor(
         state: State,
-        protected apiClient: ApiClient
+        protected apiClient?: ApiClient
     ) {
         super(state);
     }
 
     protected async ensureAuthenticated(): Promise<void> {
```
nirinchev:
```suggestion
    protected ensureAuthenticated(): void {
```

nirinchev (approved):

src/state.ts
fmenezes: will do that on a follow up PR

src/tools/atlas/atlasTool.ts
gagik:
```suggestion
    protected ensureAuthenticated(): asserts this is { apiClient: ApiClient } {
```
should let you get rid of all `!` if you remove the await-s

PR #60, opened 2025-04-11
chore: remove configFile and state.json

fmenezes:
simplifying for now, we should be able to add them later


src/state.ts
```diff
@@ -1,34 +1,8 @@
...
-                this.credentials = JSON.parse(data);
-            }
-
-            return true;
-        } catch (err: unknown) {
-            logger.error(mongoLogId(1_000_007), "state", `Failed to load state: ${err}`);
-            return false;
-        }
-    }
+    connectionString?: string;
```
nirinchev: We should remove this - since it's no longer persisted, it is redundant.

src/state.ts
```diff
@@ -1,34 +1,8 @@
 import { NodeDriverServiceProvider } from "@mongosh/service-provider-node-driver";
-import { AsyncEntry } from "@napi-rs/keyring";
```
nirinchev, line 2: We should remove this package dependency if we're not going to use it.

src/config.ts
```diff
@@ -46,33 +44,28 @@ const config = {
...
-    configPath ??= "/etc/mongodb-mcp.conf";
 
-    fs.mkdirSync(localDataPath, { recursive: true });
+    const logPath = path.join(localDataPath, ".app-logs");
 
-    return {
-        localDataPath,
-        configPath,
-    };
+    fs.mkdirSync(logPath, { recursive: true });
```
gagik: unrelated to the refactor but we should handle this call failing as we might not have permissions etc. and this is outside our try/catch

gagik (approved):

nirinchev:
Looks like this is now conflicting with the integration tests. You can remove the mocks from `helpers.ts` since those methods no longer exist.

gagik (commented):
Could it be better for the apiClient to have a state rather than for a state to have an apiClient? I could explore this in #55

fmenezes:
> Could it be better for the apiClient to have a state rather than for a state to have an apiClient?

same principle for the connection provider of MongoDB tools, seems like this state is a session rather than state

PR #62, opened 2025-04-11
chore: add type-powered eslint rules

gagik:
We haven't been utilizing typescript-eslint's feature to the fullest so adds some extra configuration to enable them and benefit from stricter types


src/tools/tool.ts
```diff
@@ -28,22 +28,15 @@ export abstract class ToolBase {
...
-                logger.error(mongoLogId(1_000_000), "tool", `Error executing ${this.name}: ${error}`);
+                return await this.execute(...args);
+            } catch (error: unknown) {
+                logger.error(mongoLogId(1_000_000), "tool", `Error executing ${this.name}: ${error as string}`);
 
                 return await this.handleError(error);
             }
         };
 
-        if (this.argsShape) {
```
gagik, line 39: are we allowing not specifying argsShape? we should update the type if so

package.json
```diff
@@ -19,9 +19,8 @@
     "prepare": "npm run build",
     "build:clean": "rm -rf dist",
     "build:compile": "tsc",
-    "build:addshebang": "echo '#!/usr/bin/env node' > dist/index2.js && cat dist/index.js >> dist/index2.js && mv dist/index2.js dist/index.js",
     "build:chmod": "chmod +x dist/index.js",
-    "build": "npm run build:clean && npm run build:compile && npm run build:addshebang && npm run build:chmod",
+    "build": "npm run build:clean && npm run build:compile && npm run build:chmod",
```
fmenezes, line 23: if you're removing shebang you should also remove chmod, that was meant to help before npm package is sorted

src/tools/tool.ts
nirinchev: Oh, that's probably a leftover - at some point I was allowing it, but now we're just passing an empty object.

package.json
gagik: shebang is still included in https://github.com/mongodb-js/mongodb-mcp-server/pull/62/files#diff-a2a171449d862fe29692ce031981047d7ab755ae7f84c707aef80701b3ea0c80

Issue #54, opened 2025-04-10
Implement tool disabling


nirinchev:
We should allow users to disable tools by name and/or limit the server to non-destructive operations only.

PR #63, opened 2025-04-11
add a mechanism to disable tools from the config

nirinchev:
Allows disabling of tools by category, operation type, or name. Example config:

```json
env: {
  "MDB_MCP_DISABLED_TOOLS": "atlas, write, list-indexes"
}
```

or as command line args:

```shell
--disableTools atlas write list-indexes
```

Fixes https://github.com/mongodb-js/mongodb-mcp-server/issues/54


src/tools/atlas/createAccessList.ts
```diff
@@ -1,13 +1,14 @@
...
 import { AtlasToolBase } from "./atlasTool.js";
-import { ToolArgs } from "../tool.js";
+import { ToolArgs, OperationType } from "../tool.js";
 
 const DEFAULT_COMMENT = "Added by Atlas MCP";
 
 export class CreateAccessListTool extends AtlasToolBase {
     protected name = "atlas-create-access-list";
     protected description = "Allow Ip/CIDR ranges to access your MongoDB Atlas clusters.";
+    protected operationType: OperationType = "create";
```
nirinchev, line 11: @fmenezes @blva I've tried to map the atlas tools to operation types, but feel free to suggest more appropriate ones (or we can introduce new types if that'd make sense).

src/config.ts
```diff
@@ -74,6 +76,12 @@ function getEnvConfig(): Partial<UserConfig> {
                 return;
             }
 
+            // Try to parse an array of values
+            if (value.indexOf(",") !== -1) {
+                obj[currentField] = value.split(",").map((v) => v.trim());
```
gagik, line 84: if this supports an array anyways we could just always do the split without the `indexOf` check

src/tools/tool.ts
```diff
@@ -46,6 +59,35 @@ export abstract class ToolBase {
         }
     }
 
+    // Checks if a tool is allowed to run based on the config
+    private verifyAllowed(): CallToolResult | undefined {
```
gagik: we could make the category a static field and avoid registering the tool itself? then claude will actually avoid it

gagik: https://github.com/mongodb-js/mongodb-mcp-server/pull/55/files#diff-8a8ae07582c9d433ec8c2e5c4310ff8901e604f4965c5b90a49117ad46c47595R39-R42 would be useful for this

nirinchev: Yeah, I was going back and forth on whether to not register the tools or just prevent using them at runtime - seems like there's precedent for both, though I don't have strong opinion which one we should go with.

gagik: I think the big pro about not registering is that Claude/the Agent won't discover it which I think is a much better experience

gagik: Otherwise we're creating errors for Claude/the Agent to run into and spend credits on

nirinchev: Fair enough, I'll change it

gagik: it may be easier to filter with #55

fmenezes:
have we seen this pattern before? are there other MCP servers that allow disabling tools?

nirinchev:
Yes, there's a handful of database-focused servers I've seen implement some form of protection against destructive operations. E.g. [mcp-server-mysql](https://github.com/benborla/mcp-server-mysql) has env variables such as `ALLOW_INSERT_OPERATION`, `ALLOW_DELETE_OPERATION`, etc. to control what tools are allowed. Alternatively, [mongodb-lens](https://github.com/furey/mongodb-lens#tutorial-5-working-with-confirmation-protection) has a 2-step confirmation, where it responds with a code and then verifies that the user has input it to go through with the operation. [supabase-mcp-server](https://github.com/alexander-zuev/supabase-mcp-server?tab=readme-ov-file#database-query-tools) similarly implements a 3-tier safety system.

I think this is ultimately important to instil confidence in users, especially with some agents being fairly opaque with what they're doing. I know most allow you to disable tools one-by-one, but with server updates, you run the risk of letting a destructive tool sneak in and wreak havoc on your data.

fmenezes:
@nirinchev I was thinking here, what if all tools declared a list of tags instead of one category? this way we could have something like disableTools `write` and keep everything readonly but also we could disable `atlas,destroy`, allowing you to do writes without destroys but no access to atlas.

nirinchev:
Right now we're kind of there, but with the caveat that all tools have exactly 2 "tags" - the category and the operation type. Therefore if a user specifies `atlas, write, delete`, they're disabling all tools in the atlas category, as well as all tools that have operation type `write` or `delete`. We can definitely make it even more flexible/generic by allowing arbitrary number of tags, but I wonder if that wouldn't overcomplicate things.

fmenezes:
> Rigth now we're kind of there, but with the caveat that all tools have exactly 2 "tags" - the category and the operation type. Therefore if a user specifies `atlas, write, delete`, they're disabling all tools in the atlas category, as well as all tools that have operation type `write` or `delete`. We can definitely make it even more flexible/generic by allowing arbitrary number of tags, but I wonder if that wouldn't overcomplicate things.

That works for me, just make sure to have a category for the main tools, could be 'tools', 'driver' or 'server'. At least we would be consistent.

nirinchev:
What would be in the `tools` category? Right now I've used the same category names as the folder structure (that is atlas and mongodb) - happy to find better names, though `driver` is probably not what I'd go with.

gagik:
Maybe `clusters-read` / `clusters-write` + `database-read` `database-write`? We can also have `read`/`write` which combines the two.

fmenezes (approved):
it works for me, just make sure to leave documentation on README.md otherwise it will be difficult for people to find out how to use these values

PR #64, opened 2025-04-11
chore: auto generate apiClient

fmenezes:


src/common/atlas/apiClientError.ts
```diff
@@ -0,0 +1,19 @@
...
+    response?: Response;
+
+    constructor(message: string, response: Response | undefined = undefined) {
+        super(message);
+        this.name = "ApiClientError";
+        this.response = response;
+    }
+
+    static async fromResponse(response: Response, message?: string): Promise<ApiClientError> {
+        message ||= `error calling Atlas API`;
```
gagik:
```suggestion
    static async fromResponse(response: Response, message: string = 'error calling Atlas API'): Promise<ApiClientError> {
```

scripts/apply.ts
```diff
@@ -0,0 +1,96 @@
+import fs from "fs";
```
gagik:
```suggestion
import fs from "fs/promises";
```
try this to avoid promisify 🙂

src/logger.ts
```diff
@@ -98,20 +99,10 @@ class ProxyingLogger extends LoggerBase {
...
-        fs.mkdir(path, options, (err, resultPath) => {
-            if (err) {
-                reject(err);
-            } else {
-                resolve(resultPath);
-            }
-        });
-    });
-}
+const mkdirAsync = promisify(fs.mkdir);
```
gagik: same here with `fs/promises`

gagik (approved):
Nice 🚀 A bit scary to mix auto-generated and and regular code in one place but it does give a nice DevEx so seems worth it

fmenezes:
> Nice 🚀 A bit scary to mix auto-generated and and regular code in one place but it does give a nice DevEx so seems worth it

I started with a template, but unfortunately the devex was much worse

PR #67, opened 2025-04-11
fix: version

fmenezes:


package.json
```diff
@@ -58,6 +55,9 @@
     "yaml": "^2.7.1"
   },
   "dependencies": {
+    "@types/simple-oauth2": "^5.0.7",
+    "@types/yargs-parser": "^21.0.3",
```
nirinchev: @types don't need to be in dependencies since we're transpiling to pure js.

nirinchev (approved):

PR #72, opened 2025-04-11
chore(release): use app token to create the release PR

nirinchev:


.github/workflows/publish.yaml
```diff
@@ -15,8 +15,6 @@ jobs:
     steps:
       - uses: GitHubSecurityLab/actions-permissions/monitor@v1
       - uses: actions/checkout@v4
-        with:
-          fetch-depth: 0
```
fmenezes: Right now we need it because we check against past tags

nirinchev: Oh yeah, good point. Will revert.

.github/workflows/prepare_release.yaml
```diff
@@ -0,0 +1,43 @@
...
+          cache: "npm"
+      - name: Bump version
+        id: bump-version
+        run: |
+          echo "NEW_VERSION=$(npm version ${{ inputs.version }} --no-git-tag-version)" >> $GITHUB_OUTPUT
+      - name: Create release PR
+        uses: peter-evans/create-pull-request@271a8d0340265f705b14b6d32b9829c1cb33d45e # 7.0.8
+        with:
+          title: "Release v${{ steps.bump-version.outputs.NEW_VERSION }}"
+          token: ${{ steps.app-token.outputs.token }}
```
fmenezes, line 37: Ideally we should set author and committer to bots details

gagik (approved):

fmenezes (approved):

Issue #73, opened 2025-04-14
Suggest configured connection string in case of connection failure


nirinchev:
Sometimes models will provide dumb connection strings (e.g. verbatim `process.env.MONGO_URI`) that will inevitably fail to connect. If the user has configured a connection string already, we should respond with a suggestion to use that value instead.

PR #75, opened 2025-04-14
feat: suggest the config connection string on connect failure

nirinchev:
If the user has a configured connection string and the connect fails (e.g. due to the model infering an incorrect connection string from the code), suggest using the config connection string instead.

Fixes https://github.com/mongodb-js/mongodb-mcp-server/issues/73

### DIFF
--- a/src/tools/mongodb/metadata/connect.ts
+++ b/src/tools/mongodb/metadata/connect.ts
@@ -4,6 +4,7 @@ import { DbOperationType, MongoDBToolBase } from "../mongodbTool.js";
 import { ToolArgs } from "../../tool.js";
 import { ErrorCodes, MongoDBError } from "../../../errors.js";
 import config from "../../../config.js";
+import { MongoError as DriverError } from "mongodb";
 
 export class ConnectTool extends MongoDBToolBase {
     protected name = "connect";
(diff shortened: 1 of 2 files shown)


gagik (approved):
Nice

src/tools/mongodb/metadata/connect.ts
```diff
@@ -57,10 +58,32 @@ export class ConnectTool extends MongoDBToolBase {
             throw new MongoDBError(ErrorCodes.InvalidParams, "Invalid connection options");
         }
 
-        await this.connectToMongoDB(connectionString);
+        try {
+            await this.connectToMongoDB(connectionString);
+            return {
+                content: [{ type: "text", text: `Successfully connected to ${connectionString}.` }],
+            };
+        } catch (error) {
```
fmenezes, line 66: nit: alternatively you can override `handleError` here

nirinchev: Unfortunately, it's a bit more complicated since I wanted to have the extra check comparing the connection string we tried connecting to to the connection string in the config. `handleError` right now doesn't supply the original args used to invoke the tool and even if it did, in the case of connecting by cluster name, we wouldn't know what the resolved connection string is.

PR #76, opened 2025-04-16
feat: add createCollection tool

nirinchev:
chore: add integration tests for listCollections

### DIFF
--- a/tests/integration/tools/mongodb/create/createCollection.test.ts
+++ b/tests/integration/tools/mongodb/create/createCollection.test.ts
@@ -1,0 +1,129 @@
+import {
+    connect,
+    jestTestCluster,
+    jestTestMCPClient,
+    getResponseContent,
+    validateParameters,
+} from "../../../helpers.js";
+import { toIncludeSameMembers } from "jest-extended";
+import { McpError } from "@modelcontextprotocol/sdk/types.js";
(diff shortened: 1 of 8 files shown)


gagik (approved):

tests/integration/tools/mongodb/create/createCollection.test.ts
```diff
@@ -0,0 +1,129 @@
...
+    jestTestMCPClient,
+    getResponseContent,
+    validateParameters,
+} from "../../../helpers.js";
+import { toIncludeSameMembers } from "jest-extended";
+import { McpError } from "@modelcontextprotocol/sdk/types.js";
+import { ObjectId } from "bson";
+
+describe("createCollection tool", () => {
+    const client = jestTestMCPClient();
```
gagik, line 13: not blocking: this isn't from this PR but naming feels weird in retrospect as jest is an implementation detail. maybe `mock` or just `getTestMCPClient`?

nirinchev: Well, it does also call the before/after hooks, so it kind of depends on jest (or a similar testing framework that has the same hooks with similar signatures). Also, I don't think it'd be accurate to call it a mock client as it's more of an integration setup. Happy to iterate on the naming though if we feel this is not the right one.

PR #79, opened 2025-04-16
chore: refactor integration test setup

nirinchev:
Noticed that we're always running all the setup functions, so it made sense to just combine them into one.


gagik (approved):
🚀

PR #82, opened 2025-04-16
add create project tool

blva:


src/tools/atlas/createProject.ts
```diff
@@ -0,0 +1,32 @@
...
+        organizationId: z.string().describe("Organization ID for the new project"),
+    };
+
+    protected async execute({ projectName, organizationId }: ToolArgs<typeof this.argsShape>): Promise<CallToolResult> {
+        this.session.ensureAuthenticated();
+
+        const input = {
+            name: projectName,
+            orgId: organizationId,
+        } as Group;
```
nirinchev, line 56: This is suspicious. Not something we need to address in this PR, but the fact we have to type assert like this means that the type definitions for the API client are imprecise. If a partial Group is acceptable as the body for a `createProject` call, this should be expressed in the type system, rather than be something we need to read in the docs or - worse - trial and error to understand.

src/tools/atlas/createProject.ts
```diff
@@ -0,0 +1,32 @@
...
+import { ToolArgs, OperationType } from "../tool.js";
+import { Group } from "../../common/atlas/openapi.js";
+
+export class CreateProjectTool extends AtlasToolBase {
+    protected name = "atlas-create-project";
+    protected description = "Create a MongoDB Atlas project";
+    protected operationType: OperationType = "create";
+    protected argsShape = {
+        projectName: z.string().describe("Name for the new project"),
+        organizationId: z.string().describe("Organization ID for the new project"),
```
nirinchev: When testing, have you noticed models be consistently good about discovering the correct org id or is this something users will be prompted to fill out?

tests/integration/helpers.ts
```diff
@@ -196,3 +204,48 @@ export function validateParameters(tool: ToolInfo, parameters: ParameterInfo[]):
...
+    const apiClient = new ApiClient({
+        baseUrl: "http://localhost:3000",
+        userAgent: "AtlasMCP-Test",
+        credentials: {
+            clientId: "test-client-id",
+            clientSecret: "test-client-secret",
+        },
+    });
+
+    jest.spyOn(apiClient, "createProject").mockResolvedValue(jestTestAtlasData.project);
```
nirinchev: I have mixed feelings about this as originally, I was intending for the `integration` tests to be actual integration tests where we run operations against the real things and validate using other tools that what the MCP server does is what the user intended. At the same time, I know from experience that setting up integration tests against Atlas is fairly involved, so can understand the appeal of mocks here.

I wouldn't block the PR on this as I don't think we should be spending the time right now, but we should definitely flag it as a future improvement to remove these mocks and run the tests against cloud-qa or a similar environment and use the Atlas CLI or another tool to validate that the things that should be happening are indeed happening.

tests/integration/tools/atlas/createProject.test.ts
```diff
@@ -0,0 +1,70 @@
...
+                    const mcpError = error as McpError;
+                    expect(mcpError.code).toEqual(-32602);
+                    expect(mcpError.message).toContain("Invalid arguments for tool atlas-create-project");
+                }
+            });
+        }
+    });
+
+    describe("with valid arguments", () => {
+        it("creates a new project", async () => {
```
nirinchev: Here we should probably assert that the expected apiClient methods are called with the supplied arguments.

tests/integration/tools/atlas/createProject.test.ts
```diff
@@ -0,0 +1,70 @@
...
+                    expect(error).toBeInstanceOf(McpError);
+                    const mcpError = error as McpError;
+                    expect(mcpError.code).toEqual(-32602);
+                    expect(mcpError.message).toContain("Invalid arguments for tool atlas-create-project");
+                }
+            });
+        }
+    });
+
+    describe("with valid arguments", () => {
```
nirinchev: As an illustration of the limitations of the mocked client, here we can't encode what the expected behavior is for corner cases such as trying to create a project in a non-existent organization or an org the user only has read access to or creating a project with a duplicate name. Again, not a blocker, but something to consider once we get the ability to run tests against Atlas.

nirinchev (commented):
I'm sorry about the merge conflicts - I really should have merged the integration tests refactor earlier, but got distracted. Hope it's not too much hassle to resolve the conflicts.

tests/integration/helpers.ts
blva: Makes sense, perharps the right approach for now would be to create this under a unit folder since we're mocking it and later set it up (I think we should have that for sure!)

src/tools/atlas/createProject.ts
blva: fair point, it does show up in other creation tools and might be a limitation of us using auto-generated objects from openapi. the definitions are not imprecise, I think they're too precise and we end up needing to add fields that are not required for the api call, including readonly, example:

```shell

Type '{ name: string; orgId: string; }' is missing the following properties from type '{ readonly clusterCount: number; readonly created: string; readonly id?: string | undefined; readonly links?: { href?: string | undefined; rel?: string | undefined; }[] | undefined; name: string; orgId: string; regionUsageRestrictions: "COMMERCIAL_FEDRAMP_REGIONS_ONLY" | "GOV_REGIONS_ONLY"; tags?: { ...; }[] | undef...': clusterCount, created, regionUsageRestrictions, withDefaultAlertsSettingsts(2739)

```

tests/integration/helpers.ts
blva: for now I'll delete this test completely until we find how we'll configure atlas tests, there is a PR open from @fmenezes

src/tools/atlas/createProject.ts
nirinchev: This seems like it's the definition of a Group response rather than the input expected for `createProject`.

src/tools/atlas/createProject.ts
blva: no, this one is a bit more tricky since it involves project creation it can be expected that uses didn't do any operations to their org. but your comment made me realize that we might have room to generate the project name at least, we don't need to prompt for that

src/tools/atlas/createProject.ts
blva: Yes, the other option is to fetch only the write values using `FetchOptions`, which is probably what you expect. The annoying part is that we need to configure even the optional fields for the request and it's not how other create requests are implemented in the code base, but I'll change to that

nirinchev: To be clear, I don't feel strongly this should happen in this PR - we could definitely do it as a follow up (or even after we go ahead with the MVP release). While mildly annoying that we don't have the precise type definitions for the API client, having tests will help ensure we're still sending the correct payloads - it's more of a continued maintenance/extensibility issue.

src/tools/atlas/createProject.ts
nirinchev: I think the org id would be quite annoying for people to figure out - is there no way to obtain the org id from the token or some other API? If I'm chatting with an LLM and I need to go open the browser to figure out what my org id is, I might as well go ahead and create my project then and there.

src/tools/atlas/createProject.ts
fmenezes: @nirinchev this has been a difficulty of our endpoints in general we don't do separate annotations for request and response bodies across the API, which, as you realized here, tends to become imprecise. Agreed not to touch it for now.

src/tools/atlas/createProject.ts
fmenezes: I think we will need an `atlas-list-orgs` tool otherwise this is unusable

tests/integration/helpers.ts
fmenezes: I'm working on atlas tests, here is a snek peak https://github.com/mongodb-js/mongodb-mcp-server/pull/83

fmenezes: I'm planning to run tests aginst atlas itself, here they will run daily as opposed to every PR

src/tools/atlas/createProject.ts
fmenezes: I wouldn't create a tool to create orgs given service accounts are scoped per orgs

tests/integration/helpers.ts
blva: yep! deleted for now

src/tools/atlas/createProject.ts
blva: ok then I'll keep as is, for the record the other option I was going to test:

```typescript

  const createProjectOpts: FetchOptions<operations["createProject"]> = {
            body: {
                name: projectName,
                clusterCount: 1,
                regionUsageRestrictions: "COMMERCIAL_FEDRAMP_REGIONS_ONLY",
                withDefaultAlertsSettings: true,
                orgId: organizationId,
                created: new Date().toISOString(),
            },
        };
```

We can investigate updating the tools later. Resolving for now!

src/tools/atlas/createProject.ts
blva: ok, I'll add list orgs and try

blva: added it, however I'm wondering if we should ask the user first if we can create it with the found org id, but i added using assumption

src/tools/atlas/createProject.ts
```diff
@@ -0,0 +1,58 @@
+import { z } from "zod";
+import { CallToolResult } from "@modelcontextprotocol/sdk/types.js";
+import { AtlasToolBase } from "./atlasTool.js";
+import { ToolArgs, OperationType } from "../tool.js";
+import { Group } from "../../common/atlas/openapi.js";
+import logger from "../../logger.js";
```
gagik: unused variable (seems to be caught by lint)

src/tools/atlas/createProject.ts
```diff
@@ -0,0 +1,58 @@
...
+                const organizations = await this.session.apiClient.listOrganizations();
+                if (!organizations?.results?.length) {
+                    return {
+                        content: [{ type: "text", text: "No organizations were found in your MongoDB Atlas account. Please create an organization first." }],
+                        isError: true,
+                    };
+                }
+                organizationId = organizations.results[0].id;
+                assumedOrg = true;
+            } catch (error) {
```
gagik:
```suggestion
            } catch {
```

gagik (approved):
Looks good 🚀  just the unused vars

blva:
I'll merge with the coverage decreased as the tests for the atlas tool are being added in a separate PR

PR #83, opened 2025-04-16
chore: add atlas tests

fmenezes:


tests/integration/tools/atlas/accessLists.test.ts
```diff
@@ -0,0 +1,107 @@
...
+            it("should create an access list", async () => {
+                const projectId = getProjectId();
+
+                const response = (await integration.mcpClient().callTool({
+                    name: "atlas-create-access-list",
+                    arguments: {
+                        projectId,
+                        ipAddresses: ["8.8.8.8"],
+                        cidrBlocks: ["9.9.9.9/24"],
+                        currentIpAddress: true,
```
blva: nit: i think we generate random ips in AtlasCLI, could we do this here as well?

blva (approved):
🚢

PR #84, opened 2025-04-16
chore: add integration tests for create-index

nirinchev:


src/tools/mongodb/metadata/connect.ts
```diff
@@ -37,25 +36,21 @@ export class ConnectTool extends MongoDBToolBase {
...
-                };
-            }
+        if (
+            connectionStringOrClusterName.startsWith("mongodb://") ||
+            connectionStringOrClusterName.startsWith("mongodb+srv://")
+        ) {
+            connectionString = connectionStringOrClusterName;
         } else {
-            throw new MongoDBError(ErrorCodes.InvalidParams, "Invalid connection options");
+            // TODO:
```
fmenezes: ?

src/tools/mongodb/mongodbTool.ts
```diff
@@ -19,9 +19,10 @@ export abstract class MongoDBToolBase extends ToolBase {
     protected category: ToolCategory = "mongodb";
 
     protected async ensureConnected(): Promise<NodeDriverServiceProvider> {
-        const provider = this.session.serviceProvider;
+        let provider = this.session.serviceProvider;
         if (!provider && config.connectionString) {
             await this.connectToMongoDB(config.connectionString);
+            provider = this.session.serviceProvider;
```
fmenezes: It seems you're assigning the same value, would you double check if this is needed?

nirinchev: `connectToMongoDB` - if successful - will set the session provider.

src/tools/mongodb/metadata/connect.ts
nirinchev: This is unchanged code, but it's the hook where we'd want to call any atlas API that would allow us to generate a connection string based on a cluster name (if possible).

gagik: should we add some text to that todo then?

nirinchev: The text we return from the next statement is pretty much what we would write in the TODO comment, but happy to update it and add a link to the issue.

src/tools/mongodb/mongodbTool.ts
gagik: is there a scenario where `NodeDriverServiceProvider.connect` returns undefined? seems like it'd throw instead right (and not be caught by `connectToMongoDB`)?

src/tools/mongodb/mongodbTool.ts
```diff
@@ -19,9 +19,10 @@ export abstract class MongoDBToolBase extends ToolBase {
     protected category: ToolCategory = "mongodb";
 
     protected async ensureConnected(): Promise<NodeDriverServiceProvider> {
-        const provider = this.session.serviceProvider;
+        let provider = this.session.serviceProvider;
```
gagik: maybe we should directly use `this.session.serviceProvider` at all times to avoid confusion

src/tools/mongodb/mongodbTool.ts
nirinchev: Based on the types, if the promise resolves, we should get a valid `NodeServiceProvider` and not `undefined`. If the `connect` attempt fails, this will throw an error which will eventually be bubbled up to the error handler of whatever tool the user was calling.

gagik: do I understand correctly then that the second `if (!this.session.serviceProvider)` would never occur? We could make the `connectToMongoDB` return the provider and set it to the local provider variable (I guess going back to that..) to make that type logic work

gagik (approved):

src/tools/mongodb/mongodbTool.ts
nirinchev: I changed it to always use `this.session.serviceProvider` and removed the local `provider` var.

PR #85, opened 2025-04-17
chore: add integration tests for insert-many

nirinchev:
Also, removes insertOne as it seems the model is good enough to figure out the usage of insert-many even with a single doc.


src/tools/mongodb/create/insertMany.ts
```diff
@@ -27,7 +27,7 @@ export class InsertManyTool extends MongoDBToolBase {
         return {
             content: [
                 {
-                    text: `Inserted \`${result.insertedCount}\` documents into collection \`${collection}\``,
+                    text: `Inserted \`${result.insertedCount}\` document(s) into collection "${collection}"`,
```
gagik, line 30: is there a reason for the change of the \`?

nirinchev: Mostly consistency - it doesn't really matter as in my experience, the model will not return the result as-is, but rather reformat it. But we've been using quotes in other places, so figured I should standardize around it.

gagik (approved):

tests/integration/tools/mongodb/create/insertMany.test.ts
```diff
@@ -0,0 +1,141 @@
...
+
+    describe("with invalid arguments", () => {
+        const args = [
+            {},
+            { collection: "bar", database: 123, documents: [] },
+            { collection: [], database: "test", documents: [] },
+            { collection: "bar", database: "test", documents: "my-document" },
+            { collection: "bar", database: "test", documents: { name: "Peter" } },
+        ];
+        for (const arg of args) {
```
gagik, line 39:
```suggestion
        for (const arg of [
            {},
            { collection: "bar", database: 123, documents: [] },
            { collection: [], database: "test", documents: [] },
            { collection: "bar", database: "test", documents: "my-document" },
            { collection: "bar", database: "test", documents: { name: "Peter" } },
        ]) {
```

PR #87, opened 2025-04-17
feat: core telemetry functionality

blva:
- adds telemetry framework
- adds auth and unauth telemetry emission
- adds caching if events can't be published
- adds support to disable telemetry 

next:
- add more events
- add unit tests 
- check if we can add int tests


src/config.ts
```diff
@@ -39,9 +40,21 @@ const mergedUserConfig = {
...
+    device_id: "id", // TODO: use @mongodb-js/machine-id
+    platform: process.platform,
+    arch: process.arch,
+    os_type: process.platform,
+    os_version: process.version,
+}
+
 const config = {
     ...mergedUserConfig,
+    ...machineMetadata,
```
gagik: is there an advantage of having machineMetadata inside the config?

src/index.ts
```diff
@@ -23,8 +24,16 @@ try {
     const transport = new StdioServerTransport();
 
     await server.connect(transport);
-} catch (error: unknown) {
-    logger.emergency(mongoLogId(1_000_004), "server", `Fatal error running server: ${error as string}`);
+}
 
+// Start the server
+try {
+    await main();
```
gagik: purely personal preference but I don't see much benefit in this main function as it stands as leads to more back and forth referencing. One advantage of main though is the ability to export it for external use but I'm not sure this should be encouraged anyhow

src/session.ts
```diff
@@ -1,19 +1,48 @@
...
+import { mongoLogId } from "mongodb-log-writer";
+import { Implementation } from "@modelcontextprotocol/sdk/types.js";
 
 export class Session {
+    sessionId?: string;
     serviceProvider?: NodeDriverServiceProvider;
-    apiClient?: ApiClient;
+    apiClient: ApiClient;
+    agentClientName?: string;
+    agentClientVersion?: string;
```
gagik:
```suggestion
    agentRunner?: {
    	name: string;
    	version: string;
    };
```
it can also be agentClient, it just becomes confusing with who's the client of the agent, the user or the software

blva:
note: this is a draft, will do some cleanup and change tests, looking for feedback on the code architecture

src/session.ts
```diff
@@ -1,19 +1,48 @@
...
 
 export class Session {
+    sessionId?: string;
     serviceProvider?: NodeDriverServiceProvider;
-    apiClient?: ApiClient;
+    apiClient: ApiClient;
+    agentClientName?: string;
+    agentClientVersion?: string; 
+
+    constructor() {
```
gagik: we should probably pass the config into the constructor to make this dependency more explicit

src/session.ts
```diff
@@ -1,19 +1,46 @@
 import { NodeDriverServiceProvider } from "@mongosh/service-provider-node-driver";
 import { ApiClient } from "./common/atlas/apiClient.js";
 import config from "./config.js";
+import { Implementation } from "@modelcontextprotocol/sdk/types.js";
 
 export class Session {
+    sessionId?: string;
     serviceProvider?: NodeDriverServiceProvider;
-    apiClient?: ApiClient;
+    apiClient: ApiClient;
```
gagik, line 14: nice, I like that the apiClient is always defined now 🥳

src/session.ts
```diff
@@ -1,19 +1,46 @@
...
+    agentClientName?: string;
+    agentClientVersion?: string;
```
gagik:
```suggestion
    agentRunner?: {
    	name: string;
   	version: string;
   }
```
(can also be agentClient, that makes sense; I just find the agent's client to possibly be the user?)

src/session.ts
```diff
@@ -1,19 +1,46 @@
...
+    }
+
+    setAgentClientData(agentClient: Implementation | undefined) {
+        this.agentClientName = agentClient?.name;
+        this.agentClientVersion = agentClient?.version;
+    }
 
     ensureAuthenticated(): asserts this is { apiClient: ApiClient } {
-        if (!this.apiClient) {
+        if (!this.apiClient || !this.apiClient.hasCredentials()) {
```
gagik:
```suggestion
        if (!this.apiClient.hasCredentials()) {
```

src/session.ts
```diff
@@ -1,19 +1,46 @@
...
             if (!config.apiClientId || !config.apiClientSecret) {
                 throw new Error(
                     "Not authenticated make sure to configure MCP server with MDB_MCP_API_CLIENT_ID and MDB_MCP_API_CLIENT_SECRET environment variables."
                 );
             }
 
+            // Initialize or reinitialize API client with credentials
             this.apiClient = new ApiClient({
```
gagik: in what scenario would it end up reinitialized with credentials? does the global config object get updated in the middle of a session? we should probably keep anything that is dynamic in a session as part of session then

src/index.ts
blva: I think this was me tweaking with getting the client config, I think it's not necessary so let me try to revert

src/telemetry/telemetry.ts
```diff
@@ -0,0 +1,203 @@
...
+                error: error instanceof Error ? error : new Error(String(error)),
+            };
+        }
+    }
+
+    /**
+     * Reads cached events from disk
+     * Returns empty array if no cache exists or on read error
+     */
+    private async readCache(): Promise<BaseEvent[]> {
```
gagik: not sure we do disk cache for mongosh or Compass so it might be an overkill for the MCP; also makes the presence of telemetry more visible on the filesystem, we could just rely on in-memory and do our best to flush them on a graceful exit.

src/telemetry/telemetry.ts
```diff
@@ -0,0 +1,203 @@
...
+    os_type: string;
+    os_version?: string;
+    session_id?: string;
+};
+
+export class Telemetry {
+    private readonly commonProperties: CommonProperties;
+
+    constructor(private readonly session: Session) {
+        // Ensure all required properties are present
```
gagik: I'm having trouble understanding this comment so unsure if this is the intention but Object.freeze doesn't do any assertions to ensure the required properties are present, it prevents them from being modified

src/telemetry/telemetry.ts
```diff
@@ -0,0 +1,203 @@
...
+
+    /**
+     * Emits a tool event with timing and error information
+     * @param command - The command being executed
+     * @param category - Category of the command
+     * @param startTime - Start time in milliseconds
+     * @param result - Whether the command succeeded or failed
+     * @param error - Optional error if the command failed
+     */
+    public async emitToolEvent(
```
gagik: this seems a bit redundant with `createToolEvent`

src/telemetry/telemetry.ts
```diff
@@ -0,0 +1,203 @@
...
+        }
+
+        const event = this.createToolEvent(command, category, startTime, result, error);
+        await this.emit([event]);
+    }
+
+    /**
+     * Creates a tool event with common properties and timing information
+     */
+    private createToolEvent(
```
gagik: More subjective but to keep this service flexible, I'd move this inside `tool.ts` and keep the telemetry service with the core functions + type definitions

gagik (commented):
Great start 🚀

src/config.ts
blva: my line of thought is that we might need it every time we send an event because we can't support identify events, so i wanted to make it easy to fetch, could also live in telemetry, no strong opinion! what would you prefer?

src/session.ts
blva: I think if it has no credentials and they're set via env variables the client needs to be re-initialized, I can try to update the logic to check for that

src/telemetry/telemetry.ts
blva: we do it for atlascli, but I agree it might not be a strong requirement for now. cc @fmenezes wdyt?

src/config.ts
gagik: I see we could keep it separate from config and just use it separately; just so we have a separation between what is actually configurable and what is set in stone metadata

blva: great point! thank you, will do

src/telemetry/telemetry.ts
blva: yep! removed createToolEvent

src/telemetry/telemetry.ts
blva: fair enough as we'll have other sorts of events, let me try

src/session.ts
blva: updated

src/config.ts
blva: updated!

src/telemetry/telemetry.ts
blva: yep, this comment is misleading, i'll update it

src/common/atlas/apiClient.ts
```diff
@@ -62,10 +66,10 @@ export class ApiClient {
     constructor(options?: ApiClientOptions) {
         this.options = {
             ...options,
-            baseUrl: options?.baseUrl || "https://cloud.mongodb.com/",
+            baseUrl: options?.baseUrl || "https://cloud-dev.mongodb.com/",
```
fmenezes:
```suggestion
            baseUrl: options?.baseUrl || "https://cloud.mongodb.com/",
```

src/config.ts
```diff
@@ -42,6 +43,8 @@ const mergedUserConfig = {
 const config = {
     ...mergedUserConfig,
     version: packageJson.version,
+    mcpServerName: "MdbMcpServer",
```
fmenezes: will this ever change?

src/config.ts
```diff
@@ -11,6 +11,7 @@ interface UserConfig {
     apiBaseUrl?: string;
     apiClientId?: string;
     apiClientSecret?: string;
+    telemetry?: "enabled" | "disabled";
```
fmenezes, line 13: we tend to support DO_NOT_TRACK as per https://consoledonottrack.com/, we should follow that if possible

src/telemetry/telemetry.ts
fmenezes: cli is short lived process MCP is long lived process, makes sense to keep in memory.

src/config.ts
blva: nice, added

src/common/atlas/apiClient.ts
blva: sorry, was testing it out! will fix

src/config.ts
blva: not AFAIK

src/telemetry/telemetry.ts
blva: np! I'll remove then for now, we can always add it back

src/telemetry/constants.ts
```diff
@@ -0,0 +1,16 @@
...
+import config from "../config.js";
+import { getMachineIdSync } from "native-machine-id";
+
+/**
+ * Machine-specific metadata formatted for telemetry
+ */
+export const MACHINE_METADATA = Object.freeze({
+    device_id: getMachineIdSync(),
+    mcp_server_version: pkg.version,
+    mcp_server_name: config.mcpServerName,
```
fmenezes: this seems constant, do we need this? maybe we add useragent instead?

src/config.ts
```diff
@@ -42,6 +43,8 @@ const mergedUserConfig = {
 const config = {
     ...mergedUserConfig,
     version: packageJson.version,
+    mcpServerName: "MdbMcpServer",
+    isTelemetryEnabled: true,
```
fmenezes: are you using this? is it needed?

src/session.ts
fmenezes: I don't think we need this whole section anymore, you already added a logic to initialize apiClient with/without credentials, we just need to `if (!this.apiClient.hasCredentials()) { throw ... }` IMO

src/config.ts
fmenezes: It would be nice if constants lived outside config, maybe a `constants.ts` file?

something like
```
export const SERVER_NAME = "MdbMcpServer";
export const SERVER_VERSION = packageJson.version;
```

blva: ok, I added but I thought we could keep it in config for now

src/session.ts
blva: I'll wait for you to merge your PR then i can test how it will look like

src/config.ts
blva: not isTelemetryEnabled, thanks! removing

src/telemetry/constants.ts
blva: we need it for telemetry right? we could give user agent, but i figure that having the data split is better.. no strong opinion as we hope not to change server_name, but we might change after public preview

src/telemetry/constants.ts
```diff
@@ -0,0 +1,16 @@
+import pkg from "../../package.json" with { type: "json" };
+import config from "../config.js";
+import { getMachineIdSync } from "native-machine-id";
+
+/**
+ * Machine-specific metadata formatted for telemetry
+ */
+export const MACHINE_METADATA = Object.freeze({
```
gagik: I have just been learning more about Object.freeze but I would use `as const` (appended in the end) instead of `Object.freeze` as we don't really need runtime enforcement, type-level enforcement is fine

blva: np! removed

src/types/native-machine-id.d.ts
```diff
@@ -0,0 +1,34 @@
+/**
+ * Type definitions for native-machine-id
+ * Provides functionality to retrieve the machine ID of the current device.
+ */
+
+declare module "native-machine-id" {
```
gagik, line 6: I find it odd that type definitions were not included in the import, will take a look at this as they work on mongosh

src/telemetry/constants.ts
```diff
@@ -5,12 +5,12 @@ import { getMachineIdSync } from "native-machine-id";
...
+export const MACHINE_METADATA = {
     device_id: getMachineIdSync(),
     mcp_server_version: pkg.version,
     mcp_server_name: config.mcpServerName,
     platform: process.platform,
     arch: process.arch,
     os_type: process.platform,
     os_version: process.version,
-});
+};
```
gagik: this is a very confusing thing in TypeScript but adding `as const` here will make it stricter (combined with `const` in the start.

src/types/native-machine-id.d.ts
gagik: will probably be addressed by https://github.com/mongodb-js/devtools-shared/pull/529/commits/4111cff5418f794b059be259dd1a82f7522ed30f

src/telemetry/constants.ts
blva: thanks! updated

src/types/native-machine-id.d.ts
blva: nice! can update afterwards

src/common/atlas/apiClient.ts
gagik: I think this is back, we might want to consider an ENV variable here for our own testing

src/session.ts
```diff
@@ -1,25 +1,73 @@
...
-import config from "./config.js";
+import defaultConfig from "./config.js";
+import { Implementation } from "@modelcontextprotocol/sdk/types.js";
+
+// Define the type for configuration used by Session
+interface SessionConfig {
+    apiBaseUrl?: string;
+    apiClientId?: string;
+    apiClientSecret?: string;
+    [key: string]: unknown;
```
gagik: is there a reason to add this index signature?

src/common/atlas/apiClient.ts
```diff
@@ -116,6 +129,32 @@ export class ApiClient {
...
+    async sendEvents(events: BaseEvent[]): Promise<void> {
+        let endpoint = "api/private/unauth/telemetry/events";
+        const headers: Record<string, string> = {
+            Accept: "application/json",
+            "Content-Type": "application/json",
+            "User-Agent": this.options.userAgent,
+        };
+
+        const accessToken = await this.getAccessToken();
+        if (accessToken) {
```
gagik, line 143: what do we do if there's no access token? will the request fail?

src/common/atlas/apiClient.ts
```diff
@@ -116,6 +129,32 @@ export class ApiClient {
...
+
+        const url = new URL(endpoint, this.options.baseUrl);
+        const response = await fetch(url, {
+            method: "POST",
+            headers,
+            body: JSON.stringify(events),
+        });
+
+        if (!response.ok) {
+            throw await ApiClientError.fromResponse(response);
```
gagik, line 156: do we want to throw in case of failure with telemetry? we may want to just log this

src/config.ts
```diff
@@ -41,7 +45,8 @@ const mergedUserConfig = {
 
 const config = {
     ...mergedUserConfig,
-    version: packageJson.version,
+    version: SERVER_VERSION,
+    mcpServerName: SERVER_NAME,
```
gagik: I'd move this to `MACHINE_METADATA` and maybe get rid of individual global constants

src/config.ts
```diff
@@ -41,7 +45,8 @@ const mergedUserConfig = {
...
 };
 
 export default config;
```
```suggestion
export const config;
```

gagik: It'd be good to avoid default exports. This might lead to a lot of import changes so we could do it separately

src/session.ts
```diff
@@ -1,25 +1,73 @@
 import { NodeDriverServiceProvider } from "@mongosh/service-provider-node-driver";
 import { ApiClient } from "./common/atlas/apiClient.js";
-import config from "./config.js";
+import defaultConfig from "./config.js";
```
gagik:
```suggestion
import { config as defaultConfig } from "./config.js";
```
more subjective and I realize this was already the case with config but we should try to avoid default exports and particularly renaming default exports; makes it hard to refactor/debug later

src/session.ts
```diff
@@ -1,25 +1,73 @@
...
 export class Session {
+    sessionId?: string;
     serviceProvider?: NodeDriverServiceProvider;
-    apiClient?: ApiClient;
+    apiClient: ApiClient;
+    agentRunner?: {
+        name: string;
+        version: string;
+    };
+    private credentials?: { clientId: string; clientSecret: string };
```
gagik: do we use credentials outside of the apiClient? we could just keep it as part of client and not copy in session if not

src/telemetry/eventCache.ts
```diff
@@ -0,0 +1,46 @@
...
+    private static instance: EventCache;
+    private events: BaseEvent[] = [];
+
+    private constructor() {}
+
+    /**
+     * Gets the singleton instance of EventCache
+     * @returns The EventCache instance
+     */
+    public static getInstance(): EventCache {
```
gagik, line 29: nice!

src/telemetry/telemetry.ts
```diff
@@ -0,0 +1,181 @@
...
+    os_version?: string;
+    session_id?: string;
+};
+
+export class Telemetry {
+    private readonly commonProperties: CommonProperties;
+
+    constructor(private readonly session: Session) {
+        // Create an immutable object with all telemetry properties
+        this.commonProperties = Object.freeze({
```
gagik: I think `readonly` is enough for our purposes as well, no need for freeze

src/telemetry/telemetry.ts
```diff
@@ -0,0 +1,181 @@
...
+
+    /**
+     * Reads cached events from memory
+     * Returns empty array if no cache exists
+     */
+    private readCache(): BaseEvent[] {
+        try {
+            return EventCache.getInstance().getEvents();
+        } catch (error) {
+            logger.warning(
```
gagik: nit: I think these wrapper functions are not needed and the try/catch aren't really ever going to handle much considering these are array operations, I'd use the `EventCache` directly

related idea: an instance of `EventCache` could just be passed to the constructor of Telemetry and live inside it for better dependency injection

src/common/atlas/apiClient.ts
fmenezes: This makes sense to me, we use authenticated or non-authenticated requests, I think given this is a client, I would have two methods and call the appropriate one on caller, but up to @blva

src/telemetry/eventCache.ts
```diff
@@ -0,0 +1,46 @@
...
+     */
+    public getEvents(): BaseEvent[] {
+        return [...this.events];
+    }
+
+    /**
+     * Sets the cached events, replacing any existing events
+     * @param events - The events to cache
+     */
+    public setEvents(events: BaseEvent[]): void {
```
gagik: is there a chance we'd want to add events multiple times rather than set them in batches at a time?

src/telemetry/types.ts
```diff
@@ -0,0 +1,51 @@
+/**
+ * Result type constants for telemetry events
+ */
+export const TELEMETRY_RESULT = {
+    SUCCESS: "success" as const,
+    FAILURE: "failure" as const,
+};
+
+export type TelemetryResult = (typeof TELEMETRY_RESULT)[keyof typeof TELEMETRY_RESULT];
```
gagik:
```suggestion
export type TelemetryResult = "success" | "failure";
```
I don't think we benefit much from this enum-like constant, I'd just do a string-based type definition and use string directly; enums are weird in JS from my experience and we get all benefits from auto-completion for union string types anyways

src/common/atlas/apiClient.ts
blva: if there's no access token it then calls the unauth endpoint, since the access token logic lives in the client (to fetch and check if it's valid) I left this for the client to handle instead of exposing two methods

src/session.ts
blva: removing

src/telemetry/types.ts
blva: updated!

src/session.ts
blva: removed w/ the merge

src/config.ts
blva: since this is already quite big, I'll defer to change this in another PR as its not in the scope of the PR

src/config.ts
blva: We use it in the client though, so wanted to keep it here where is common

src/common/atlas/apiClient.ts
blva: we throw it threat the exception later at the telemetry layer, it's not the api client responsibility to know that it should swallow the exception IMO

src/session.ts
blva: this has been removed w/ the merge

src/telemetry/eventCache.ts
blva: not afaik specially since the endpoint always expects a list of events anyways

src/session.ts
blva: updated w/ the merge, this was broadly removed

src/telemetry/telemetry.ts
blva: done

src/config.ts
fmenezes: added in https://github.com/mongodb-js/mongodb-mcp-server/pull/101

src/telemetry/types.ts
```diff
@@ -0,0 +1,46 @@
...
+        session_id?: string;
+    } & Event["properties"];
+}
+
+/**
+ * Interface for tool events
+ */
+export interface ToolEvent extends BaseEvent {
+    properties: {
+        command: string;
```
fmenezes, line 39: right now we use this as the toolname, we also have resources, I think this should contain `"tool" | "resource" | "prompt"` and another one maybe called `tool_name`

blva: yes, we should have a category property where we can set "tool" but the idea is to keep command for all based on the tracking plan

src/config.ts
gagik: it's a bit confusing for me as it is inside a configuration, which to me means it is configurable

src/telemetry/eventCache.ts
gagik: I was just wondering if this could benefit from being addictive rather than replacing all every time. Could be a set if we're worried about duplication. But works for me either way 👍

blva: oh got it! that has been updated, we are always appending them now and i am using an lru cache logic

src/tools/tool.ts
```diff
@@ -22,28 +24,72 @@ export abstract class ToolBase {
 
     protected abstract argsShape: ZodRawShape;
 
+    private readonly telemetry: Telemetry;
+
     protected abstract execute(...args: Parameters<ToolCallback<typeof this.argsShape>>): Promise<CallToolResult>;
 
-    protected constructor(protected session: Session) {}
+    protected constructor(protected session: Session) {
+        this.telemetry = new Telemetry(session);
```
gagik: maybe we could define one telemetry object for all tools and pass that into the constructor? so we don't have bunch of duplicated telemetry objects

gagik (approved):
Awesome work  🚀  one last note about Telemetry being initialized in every tool that may be worth addressing

src/tools/tool.ts
blva: good point! updated

src/telemetry/types.ts
fmenezes: But how can you tell which is which?

src/tools/tool.ts
```diff
@@ -26,27 +28,63 @@ export abstract class ToolBase {
...
                 );
 
-                return await this.execute(...args);
+                const result = await this.execute(...args);
+                await this.emitToolEvent(startTime, result);
+                return result;
             } catch (error: unknown) {
                 logger.error(mongoLogId(1_000_000), "tool", `Error executing ${this.name}: ${error as string}`);
 
+                const toolResult: CallToolResult = {
```
fmenezes: we should do this after `await this.handleError`

blva: can i add in the next PR? will need some more tweak to handle the calculating the duration time

src/telemetry/types.ts
blva: category: "tool"
command: "connect" | "list-projects" ...

src/tools/tool.ts
fmenezes: I think you just need to do

```
const toolResult: CallToolResult = await this.handleError(...);
await this.emitToolEvent(...);
return toolResult;
```

fmenezes (approved):

src/tools/tool.ts
blva: my bad, added

Issue #80, opened 2025-04-16
Connect tool defaults to altas cluster even if we have connection string configured


nirinchev:
We should test out a few strategies - for now I'm thinking we could:
1. Expose the connection string as a resource and rely on the model picking it up
2. Add different tools for connecting with atlas cluster name or connection string
3. Tweak the connect argument schema to see if the model will try to connect with no arguments

PR #88, opened 2025-04-22
fix: tweak the connect tool to prefer preconfigured connection string

nirinchev:
Fixes https://github.com/mongodb-js/mongodb-mcp-server/issues/80

### DIFF
--- a/tests/integration/tools/mongodb/metadata/connect.test.ts
+++ b/tests/integration/tools/mongodb/metadata/connect.test.ts
@@ -13,9 +13,10 @@ describe("Connect tool", () => {
 
         validateParameters(connectTool, [
             {
-                name: "connectionStringOrClusterName",
-                description: "MongoDB connection string (in the mongodb:// or mongodb+srv:// format) or cluster name",
-                type: "string",
+                name: "options",
+                description:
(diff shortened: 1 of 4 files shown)


tests/integration/tools/mongodb/metadata/connect.test.ts
```diff
@@ -13,9 +13,10 @@ describe("Connect tool", () => {
 
         validateParameters(connectTool, [
             {
-                name: "connectionStringOrClusterName",
-                description: "MongoDB connection string (in the mongodb:// or mongodb+srv:// format) or cluster name",
-                type: "string",
+                name: "options",
+                description:
+                    "Options for connecting to MongoDB. If not provided, the connection string from the config://connection-string resource will be used. If the user hasn't specified Atlas cluster name or a connection string explicitly and the `config://connection-string` resource is present, always invoke this with no arguments.",
```
gagik, line 18: out of curiosity, does this work well in practice? I'm afraid this may be specific to models being used though don't have a better solution for this either.

gagik (approved):

tests/integration/tools/mongodb/metadata/connect.test.ts
nirinchev: When I tested it, it seemed to work well with Claude, Cursor, and Windsurf, though would be nice if others can verify it works as expected in their setups.

PR #91, opened 2025-04-22
chore: add tests for metadata actions

nirinchev:
This PR adds integration tests and fixes issues discovered for the following metadata actions:
- `collection-schema`. Here we're additionally changing the response from table to json as the table was getting quite unwieldy with nested fields.
- `collection-storage-size`

This is a bit a hairy as at some point I got annoyed by some of the more repetitive tests, so extracted helpers for them which resulted in more changes than I planned for. In most files the changes are purely moving functionality to helpers, while the newly created files contain the actual new code.

### DIFF
--- a/src/tools/mongodb/metadata/collectionStorageSize.ts
+++ b/src/tools/mongodb/metadata/collectionStorageSize.ts
@@ -4,7 +4,7 @@ import { ToolArgs, OperationType } from "../../tool.js";
 
 export class CollectionStorageSizeTool extends MongoDBToolBase {
     protected name = "collection-storage-size";
-    protected description = "Gets the size of the collection in MB";
+    protected description = "Gets the size of the collection";
     protected argsShape = DbOperationArgs;
 
     protected operationType: OperationType = "metadata";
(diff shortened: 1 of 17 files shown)


src/tools/mongodb/metadata/collectionStorageSize.ts
```diff
@@ -14,17 +14,55 @@ export class CollectionStorageSizeTool extends MongoDBToolBase {
...
+        return super.handleError(error, args);
+    }
+
+    private static getStats(value: number): { value: number; units: string } {
+        const kb = 1024;
+        const mb = kb * 1024;
+        const gb = mb * 1024;
+
+        if (value > gb) {
+            return { value: value / gb, units: "GB" };
```
gagik, line 57: non-blocking: we're losing precision to JS floats here aren't we? considering we're giving this information back to a robot, is unit conversion valuable?

tests/integration/helpers.ts
```diff
@@ -223,12 +224,84 @@ export const dbOperationParameters: ParameterInfo[] = [
     { name: "collection", type: "string", description: "Collection name", required: true },
 ];
 
-export function validateParameters(tool: ToolInfo, parameters: ParameterInfo[]): void {
+export const dbOperationInvalidArgTests = [{}, { database: 123 }, { foo: "bar", database: "test" }, { database: [] }];
+
+export async function validateToolMetadata(
```
gagik: other validate functions call `it` statements themselves but this one is contents of an it statement; would be good to rename if there's an intentional difference or make it consistent with others

tests/integration/helpers.ts
```diff
@@ -223,12 +224,84 @@ export const dbOperationParameters: ParameterInfo[] = [
...
+    mcpClient: Client,
```
gagik: related: this takes an mcpClient while others take the integration test

tests/integration/helpers.ts
```diff
@@ -223,12 +224,84 @@ export const dbOperationParameters: ParameterInfo[] = [
...
+        const response = await integration.mcpClient().callTool({
+            name,
+            arguments: validation().args,
+        });
+        const content = getResponseContent(response.content);
+        expect(content).toContain("You need to connect to a MongoDB instance before you can access its data.");
+    });
+}
+
+export function validateThrowsForInvalidArguments(
```
gagik, line 293: we could make this also include the describe statement

gagik (approved):
Good to go after considering the comments

src/tools/mongodb/metadata/collectionStorageSize.ts
nirinchev: If that robot was doing actual maths rather than autocompleting, I'd be more comfortable not doing the conversion ourselves 😄 When I was testing it a while back I got very mixed results with vscode just giving me the raw bytes in the response, while Claude trying to do a conversion to human-understandable units. With the explicit conversion, it seems that all models are returning the number as we provide it, which I prefer.

gagik: sounds good. great to hear our robot overlords aren't that great at math yet 🙏

PR #98, opened 2025-04-23
fix: close mongo client between runs

nirinchev:
Looks like we lost the `mongoClient?.close()` cleanup in https://github.com/mongodb-js/mongodb-mcp-server/pull/83, which caused jest to complain with `A worker process has failed to exit gracefully and has been force exited`.


blva (approved):

gagik (approved):

nirinchev:
Looks like we have some atlas tests failures due to authorization issues. Those should be unrelated to the changes in the PR, so will go ahead and merge.

PR #101, opened 2025-04-23
refactor: move configuration to be passed

fmenezes:


tests/integration/helpers.ts
```diff
@@ -315,14 +307,3 @@ export function validateThrowsForInvalidArguments(
         }
     });
 }
-
-export function describeAtlas(name: number | string | Function | jest.FunctionLike, fn: jest.EmptyFunction) {
```
fmenezes, line 319: drive-by: duplicated from `atlasHelpers.ts`

src/version.ts
```diff
@@ -0,0 +1,3 @@
+import packageJson from "../package.json" with { type: "json" };
+
+export default packageJson.version;
```
gagik:
```suggestion
export const version = packageJson.version;
```

gagik: if we're using this in only one place we could do:
```ts
import { version } from "../package.json" with { type: "json" };
```
and not have this file

tests/integration/helpers.ts
```diff
@@ -29,7 +27,7 @@ export interface IntegrationTest {
     randomDbName: () => string;
 }
 
-export function setupIntegrationTest(): IntegrationTest {
+export function setupIntegrationTest(cfg: UserConfig = config): IntegrationTest {
```
gagik: nit
```suggestion
export function setupIntegrationTest(userConfig: UserConfig = config): IntegrationTest {
```

src/server.ts
```diff
@@ -5,15 +5,22 @@ import { AtlasTools } from "./tools/atlas/tools.js";
...
 import logger, { initializeLogger } from "./logger.js";
 import { mongoLogId } from "mongodb-log-writer";
-import config from "./config.js";
+import { UserConfig } from "./config.js";
+import version from "./version.js";
 
 export class Server {
     public readonly session: Session;
-    private readonly mcpServer: McpServer;
+    private readonly mcpServer: McpServer = new McpServer({
```
gagik: it'd be better for the server to be passed in the MCP server + session as it was before, from dependency injection perspective. We might i.e. have other constructs that use the MCP server or the session, so we don't want it to be a derivative of the server (though one can also argue this class is like the main function).

src/session.ts
```diff
@@ -1,22 +1,27 @@
...
+    apiClientId?: string;
+    apiClientSecret?: string;
+}
 
 export class Session {
     serviceProvider?: NodeDriverServiceProvider;
     apiClient: ApiClient;
 
-    constructor() {
+    constructor(options?: SessionOptions) {
```
gagik:
```suggestion
    constructor(options: SessionOptions) {
```

gagik (approved):
Looks good, minor comments.
I think it'd be better to pass dependencies into the parent class rather than define them internally though

tests/integration/helpers.ts
```diff
@@ -58,13 +56,7 @@ export function setupIntegrationTest(): IntegrationTest {
             }
         );
 
-        mcpServer = new Server({
```
gagik, line 61: I think it's beneficial for example for us to create a test server here as opposed to expanding the config

src/session.ts
gagik: (or setting a default)

src/version.ts
fmenezes: It's read in two places right now, renamed to `packageInfo.ts` added another string in there, we can remove it later

PR #102, opened 2025-04-23
chore: add tests for read operations

nirinchev:


src/tools/mongodb/read/aggregate.ts
```diff
@@ -2,10 +2,10 @@ import { z } from "zod";
 import { CallToolResult } from "@modelcontextprotocol/sdk/types.js";
 import { DbOperationArgs, MongoDBToolBase } from "../mongodbTool.js";
 import { ToolArgs, OperationType } from "../../tool.js";
+import { EJSON } from "bson";
 
 export const AggregateArgs = {
     pipeline: z.array(z.object({}).passthrough()).describe("An array of aggregation stages to execute"),
-    limit: z.number().optional().default(10).describe("The maximum number of documents to return"),
```
gagik, line 8: I'm a bit worried about this, we should run the MCP against a very very large collection and ask for all documents: https://github.com/mongodb-js/mongodb-mcp-server/issues/108

nirinchev: That wasn't previously used, so figured it's better to remove it rather than lie to the model. Agree that we should investigate mechanisms to mitigate that risk post initial release.

PR #103, opened 2025-04-23
refactor: split test helpers

fmenezes:
followup from https://github.com/mongodb-js/mongodb-mcp-server/pull/101


tests/integration/tools/mongodb/mongodbHelpers.ts
```diff
@@ -0,0 +1,113 @@
...
+import { UserConfig, config } from "../../../../src/config.js";
+
+interface MongoDBIntegrationTest {
+    mongoClient: () => MongoClient;
+    connectionString: () => string;
+    connectMcpClient: () => Promise<void>;
+    randomDbName: () => string;
+}
+
+export function describeMongoDB(
```
gagik: nit: maybe something like `describeWithMongoDB`?

tests/integration/helpers.ts
```diff
@@ -23,16 +23,9 @@ type ToolInfo = Awaited<ReturnType<Client["listTools"]>>["tools"][number];
 export interface IntegrationTest {
     mcpClient: () => Client;
     mcpServer: () => Server;
-    mongoClient: () => MongoClient;
-    connectionString: () => string;
-    connectMcpClient: () => Promise<void>;
-    randomDbName: () => string;
 }
 
 export function setupIntegrationTest(userConfig: UserConfig = config): IntegrationTest {
```
gagik, line 28: what if we were to mix MongoDB + Atlas flows? would that be its own describe statement?

tests/integration/tools/mongodb/mongodbHelpers.ts
```diff
@@ -0,0 +1,113 @@
...
+    name: number | string | Function | jest.FunctionLike,
+    fn: (integration: IntegrationTest & MongoDBIntegrationTest) => void
+): void {
+    describe("mongodb", () => {
```
gagik, line 19: do we benefit from having this parent MongoDB describe? we could also set the variables inside the child

gagik (approved):
Nice, I like the describe-based structure!

tests/integration/helpers.ts
fmenezes: I think we would need to do something like `describeAtlasMongoDB(...)`

tests/integration/tools/mongodb/mongodbHelpers.ts
fmenezes: the reason why I added the main one was because need separation of beforeAll/AfterAll from main setup and MongoDB one other ordering is not correct

PR #104, opened 2025-04-23
ci: split atlas and mongodb tests

fmenezes:
splitting atlas and mongodb tests to run them in isolation


.github/workflows/code_health.yaml
```diff
@@ -55,10 +74,42 @@ jobs:
...
+        run: npm test -- --testPathIgnorePatterns "tests/integration/tools/mongodb" --testPathIgnorePatterns "tests/integration/[^/]+\.ts"
+      - name: Upload test results
+        uses: actions/upload-artifact@v4
+        with:
+          name: atlas-test-results
+          path: coverage/lcov.info
+  coverage:
+    runs-on: ubuntu-latest
+    needs: [run-tests, run-atlas-tests]
+    if: always()
```
gagik, line 87: what does this do?

.github/workflows/code_health.yaml
```diff
@@ -55,10 +74,42 @@ jobs:
...
+          name: test-results
+          path: coverage/mongodb
+      - name: Download atlas test results
+        uses: actions/download-artifact@v4
+        with:
+          name: atlas-test-results
+          path: coverage/atlas
+      - name: Install lcov-result-merger
+        run: |
+          npm install -g lcov-result-merger
```
gagik: we could maybe do `npx -y lcov-result-merger` in one go below

gagik (approved):

.github/workflows/code_health.yaml
fmenezes: runs coverage even if one of the tests fails (say mongodb tests pass and altas fails, it will still compute coverage)

PR #105, opened 2025-04-23
chore: add remaining mongodb integration tests

nirinchev:
Add final set of tests for updateMany, renameCollection, and explain.

### DIFF
--- a/src/tools/mongodb/metadata/explain.ts
+++ b/src/tools/mongodb/metadata/explain.ts
@@ -47,14 +47,31 @@ export class ExplainTool extends MongoDBToolBase {
         const method = methods[0];
 
         if (!method) {
-            throw new Error("No method provided");
+            return {
+                content: [
+                    {
+                        text: "No method provided. Expected one of the following: `aggregate`, `find`, or `count`",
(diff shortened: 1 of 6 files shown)


src/tools/mongodb/metadata/explain.ts
```diff
@@ -47,14 +47,31 @@ export class ExplainTool extends MongoDBToolBase {
         const method = methods[0];
 
         if (!method) {
-            throw new Error("No method provided");
+            return {
+                content: [
+                    {
+                        text: "No method provided. Expected one of the following: `aggregate`, `find`, or `count`",
+                        type: "text",
```
gagik: does `isError: true,` not fit here? is it too destructive?

gagik (approved):
nice, might need updates after https://github.com/mongodb-js/mongodb-mcp-server/pull/103

src/tools/mongodb/metadata/explain.ts
fmenezes: also you can simply `throw new Error("No method ...")` and handleError will kick in

PR #109, opened 2025-04-24
chore: add telemetry unit tests

blva:


tests/unit/telemetry.test.ts
```diff
@@ -0,0 +1,200 @@
...
+        }
+
+        if (appendEventsCalledWith) {
+            expect(appendEvents[0]?.[0]).toEqual(appendEventsCalledWith);
+        }
+    }
+
+    beforeEach(() => {
+        // Reset mocks before each test
+        jest.clearAllMocks();
```
gagik, line 71: I'm not too familiar with jest but since the original file mock happens in the global scope and doesn't get re-run; doesn't that imply this should clear that as well? 
I assume this works I'm just trying to understand if jest doesn't clear file mocks here or something

gagik (approved):
Nice; just a question I'm wondering about

tests/unit/telemetry.test.ts
blva: yep thanks for the question, the global scope ones guarantee that we won't use the actual implementation while the clear will simply clean up any leftover state between runs

PR #111, opened 2025-04-24
refactor: split loggers

fmenezes:


src/logger.ts
```diff
@@ -47,22 +52,39 @@ class ConsoleLogger extends LoggerBase {
...
-        private server: McpServer
-    ) {
+class DiskLogger extends LoggerBase {
+    private logWriter?: MongoLogWriter;
+
+    constructor(private logPath: string) {
         super();
     }
 
+    async initialize(): Promise<void> {
```
nirinchev: Rather than having an async instance method for initializing the logger, wouldn't it be cleaner to turn it into an async factori that calls a private constructor? That way we won't have ambiguous optionality for the `logWriter`.

Something like:

```ts
class DiskLogger extends LoggerBase {
  private constructor(private logWriter: MongoLogWriter) {
    super();
  }

  public static async create(logPath: string) {
    ...
    const logWriter = await manager.createLogWriter();
    return new DiskLogger(logWriter);
  }
}
```

nirinchev (approved):
Looks good - I have a minor suggestion but it's not blocking.

src/logger.ts
```diff
@@ -86,31 +108,58 @@ class Logger extends LoggerBase {
...
+            logger.log(level, id, context, message);
+        }
+    }
+}
+
+const logger = new CompositeLogger();
+export default logger;
+
+export async function initializeLogger(server: McpServer, logPath: string): Promise<void> {
+    logger.setLoggers(new McpLogger(server), new DiskLogger(logPath));
```
gagik: combined with Nikola's comment, I'd move the logger from global variable into this scope and return it; that way the logger variable will only exist in a valid, initalized state similar to https://github.com/mongodb-js/mongosh/blob/0ad691f4878dc52a8782ecaf66fc46928b2613f4/packages/logging/src/logging-and-telemetry.ts#L59-L67

fmenezes: I wanted to do this originally, but it became a bigger change, leaving for another PR

PR #117, opened 2025-04-24
chore: disable telemetry

blva:


src/config.ts
```diff
@@ -31,6 +31,7 @@ const defaults: UserConfig = {
         timeoutMS: 30_000,
     },
     disabledTools: [],
+    telemetry: "disabled",
```
fmenezes, line 34: why are we doing this?

fmenezes: I think I understand, you want to enable this once backend is fully working and tested

blva: yep, just so we're not blocked from releasing

Issue #95, opened 2025-04-23
Remove connect tool


GaurabAryal:
Should we consider removing 'connect' tool? LLM often decides to call `connect` as the first step whenever it needs to call a MongoDB tool as it always assumes the first step before calling the tool is to connect to a cluster (even though the connection has already been established earlier in the "session").

nirinchev:
I haven't experienced the LLM calling connect multiple times - can you share the prompts you used that caused connect to be called again?

Issue #92, opened 2025-04-23
Connection string is not retained when LLM tries to reconnect


GaurabAryal:
If I pass connection string in the launch command or define it in an env var, when Cursor calls the `Connect` tool the second time (with some amount of delay between the first successful connection attempt), sometimes MCP doesn't appear to know what the connection string is.

nirinchev:
We recently resolved https://github.com/mongodb-js/mongodb-mcp-server/issues/80, which appears to be somewhat related here. Do you think you can test with latest main and confirm you're still seeing this problem?

GaurabAryal:
@nirinchev – Pulled main and tested again - it does appear to be better for sure!

Though I'm wondering why the LLM cannot fetch this from the launch command on the first attempt to connect?

![Image](https://github.com/user-attachments/assets/72999ba6-fb71-4d7a-9582-168d690bdb16)
![Image](https://github.com/user-attachments/assets/1121270b-e3cf-4fe1-808d-d1b340b40089)

nirinchev:
This is not the correct syntax for providing a connection string. Right now we don't take in positional arguments, so you'll need to pass in the arg name like `--connectionString mongodb://localhost:58321`

gagik:
By the way the `npx -y /..../dist/index.js` seems incorrect and not sure how that works out but you're probably looking for this
in Cursor:
```json
 "MongoDB": {
      "command": "< your full path >/dist/index.js --connectionString mongodb://localhost:58321",
    }
```
and in VSCode:
```json
 "MongoDB": {
      "type": "stdio",
      "command": "< your full path >/dist/index.js",
      "args": ["--connectionString", "mongodb://localhost:58321"]
    }
```

nirinchev:
I think the args in vscode should be separate, no? Like `"args": [ "--connectionString", "mongodb://localhost:58321" ]`

PR #118, opened 2025-04-24
feat: update the connect tool based on connectivity status

nirinchev:
This turned out to be a lot more involved than anticipated and revealed a handful of upstream bugs, which we've worked around.

There are two main changes in this PR:
1. we've removed the resource exposing the preconfigured connection string in favor of a `config` resource, listing a redacted subset of the UserConfig options. Right now it lists the following:
  - telemetry - enabled/disabled
  - logPath - full log path
  - connectionString - set/not set + a hint to the model on what tool to use
  - connectOptions - the read/write concern etc.
2. Completely reworked the `connect` tool. Now it updates its metadata based on whether we have a connection string configured and whether it has run before. If we have connected already or have a connection string, it changes its name to `switch-connection` and updates its metadata to more clearly hint at the model in which situations it should be invoked.

Based on my testing with Claude, Cursor, and Windsurf, this works pretty solidly, but would appreciate if others could give it a go.

### TODO
- [x] Check how that behaves when the connection string in the config is incorrect.

Fixes https://github.com/mongodb-js/mongodb-mcp-server/issues/92
Fixes https://github.com/mongodb-js/mongodb-mcp-server/issues/95


src/server.ts
```diff
@@ -21,7 +21,7 @@ export class Server {
     public readonly session: Session;
     private readonly mcpServer: McpServer;
     private readonly telemetry: Telemetry;
-    private readonly userConfig: UserConfig;
+    public readonly userConfig: UserConfig;
```
nirinchev, line 25: Still not super happy with where we are at wrt configs, but this was a stopgap - there were a bunch of places in the tests where we were updating fields in the global config expecting this to be reflected in the server config, which only worked by accident.

src/server.ts
```diff
@@ -86,6 +87,32 @@ export class Server {
...
+                    "Server configuration, supplied by the user either as environment variables or as startup arguments",
+            },
+            (uri) => {
+                const result = {
+                    telemetry: this.userConfig.telemetry,
+                    logPath: this.userConfig.logPath,
+                    connectionString: this.userConfig.connectionString
+                        ? "set; no explicit connect needed, use switch-connection tool to connect to a different connection if necessary"
+                        : "not set; before using any mongodb tool, you need to call the connect tool with a connection string",
+                    connectOptions: this.userConfig.connectOptions,
```
nirinchev, line 151: These are the config options we're exposing as a resource

tests/integration/helpers.ts
```diff
@@ -22,13 +22,14 @@ export interface IntegrationTest {
     mcpServer: () => Server;
 }
 
-export function setupIntegrationTest(userConfig: UserConfig = config): IntegrationTest {
+export function setupIntegrationTest(userConfigGetter: () => UserConfig = () => config): IntegrationTest {
```
nirinchev: This was necessary to allow the user config to be constructed as part of the test initialization rather than statically.

tests/integration/helpers.ts
```diff
@@ -67,10 +69,19 @@ export function setupIntegrationTest(userConfig: UserConfig = config): Integrati
...
     beforeEach(async () => {
-        config.telemetry = "disabled";
+        if (mcpServer) {
+            mcpServer.userConfig.telemetry = "disabled";
+        }
         randomDbName = new ObjectId().toString();
     });
 
+    afterEach(async () => {
+        if (mcpServer) {
```
nirinchev, line 74: Moved this from the mongodb helpers because the session cleanup seemed like something common.

tests/integration/helpers.ts
gagik:
```suggestion
export function setupIntegrationTest(getUserConfig: () => UserConfig = () => config): IntegrationTest {
```
nit

tests/integration/helpers.ts
```diff
@@ -22,13 +22,14 @@ export interface IntegrationTest {
...
 
-export function setupIntegrationTest(userConfig: UserConfig = config): IntegrationTest {
+export function setupIntegrationTest(userConfigGetter: () => UserConfig = () => config): IntegrationTest {
     let mcpClient: Client | undefined;
     let mcpServer: Server | undefined;
 
     let randomDbName: string;
 
     beforeAll(async () => {
+        const userConfig = userConfigGetter();
```
gagik:
```suggestion
        const userConfig = getUserConfig();
```

gagik (approved):
haven't tested it out myself but changes look good

nirinchev:
I'm going to go ahead and merge this, even though we're going to rework it once more.

PR #119, opened 2025-04-25
chore: add type-safe ESLint

### LINKED ISSUE #113: Enable eslint on tests and scripts

gagik:
Fixes #113 

Largely did the minimum viable actions needed to get the linting to pass.

Also added `expectDefined` so we don't need to keep asserting our types manually.


tests/integration/helpers.ts
```diff
@@ -203,8 +203,9 @@ export function validateThrowsForInvalidArguments(
             it(`throws a schema error for: ${JSON.stringify(arg)}`, async () => {
                 try {
                     await integration.mcpClient().callTool({ name, arguments: arg });
-                    expect.fail("Expected an error to be thrown");
+                    throw new Error("Expected an error to be thrown");
```
gagik, line 202: this was not working correctly so we had the false impression that extra fields are problematic with our zod setup; they're not so I removed places where we were using that as an invalid arg scenario

tsconfig.json
```diff
@@ -8,7 +8,7 @@
     "strict": true,
     "strictNullChecks": true,
     "esModuleInterop": true,
-    "types": ["node"],
+    "types": ["node", "jest"],
```
gagik, line 11: This just seems to be the best way to get VSCode to stop complaining about the types once in a while but I think can work without too as long as others are represented

tests/integration/helpers.ts
```diff
@@ -26,6 +25,8 @@ export function setupIntegrationTest(userConfig: UserConfig = config): Integrati
     let mcpClient: Client | undefined;
     let mcpServer: Server | undefined;
 
+    // This gets used in the scope of tests.
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
nirinchev: This is really unused. Must be some leftover from the split, but the randomDbName we're using is setup in `setupMongoDBIntegrationTest`.

tests/integration/helpers.ts
```diff
@@ -203,8 +205,9 @@ export function validateThrowsForInvalidArguments(
             it(`throws a schema error for: ${JSON.stringify(arg)}`, async () => {
                 try {
                     await integration.mcpClient().callTool({ name, arguments: arg });
-                    expect.fail("Expected an error to be thrown");
+                    throw new Error("Expected an error to be thrown");
                 } catch (error) {
+                    expect((error as Error).message).not.toEqual("Expected an error to be thrown");
```
nirinchev, line 204: Isn't the assertion just below sufficient? I don't imagine `Error` is an instance of `McpError`.

tests/integration/server.test.ts
```diff
@@ -47,13 +47,13 @@ describe("Server integration test", () => {
                 });
             });
 
-            it("should return capabilities", async () => {
+            it("should return capabilities", () => {
                 const capabilities = integration.mcpClient().getServerCapabilities();
-                expect(capabilities).toBeDefined();
+                expectDefined(capabilities);
                 expect(capabilities?.completions).toBeUndefined();
                 expect(capabilities?.experimental).toBeUndefined();
```
nirinchev: I guess the `?`-s are no longer necessary here?

tests/integration/helpers.ts
gagik: it is but then the error message is more confusing (it just says expected `McpError` got `Error` which originally made me think it was erroring differently instead of not erroring at all). I want to differentiate an external error from the intentional "did not end up erroring" error

PR #123, opened 2025-04-25
chore: bump version

gagik:
This bumps the version so we can unblock doing a release and test out the CI publishing.


gagik:
Will use the action instead

PR #129, opened 2025-04-25
chore: add config flags

blva:
- add a boolean value if the user has atlas api keys and if the user has connection string set

### DIFF
--- a/src/telemetry/types.ts
+++ b/src/telemetry/types.ts
@@ -61,7 +61,7 @@ export type CommonStaticProperties = {
     arch: string;
     os_type: string;
     os_version?: string;
-}
+};
 
 /**
  * Common properties for all events that might change.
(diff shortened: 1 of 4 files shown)


src/telemetry/types.ts
```diff
@@ -3,6 +3,7 @@
  */
 export type TelemetryResult = "success" | "failure";
 export type ServerCommand = "start" | "stop";
+export type TelemetryConfigSet = "true" | "false";
```
fmenezes: nit: let's call this a boolean
```suggestion
export type TelemetryBoolSet = "true" | "false";
```

fmenezes (approved):

src/telemetry/types.ts
blva: updated!

PR #130, opened 2025-04-25
feat: add readOnly flag

blva:
- adds `readOnlyMode` flag

### DIFF
--- a/README.md
+++ b/README.md
@@ -150,6 +150,7 @@ The MongoDB MCP Server can be configured using multiple methods, with the follow
 | `connectionString` | MongoDB connection string for direct database connections (optional users may choose to inform it on every tool call) |
 | `logPath`          | Folder to store logs                                                                                                  |
 | `disabledTools`    | An array of tool names, operation types, and/or categories of tools that will be disabled.                            |
+| `readOnlyMode`     | When set to true, only allows read and metadata operation types, disabling create/update/delete operations            |
 
 #### `logPath`
 
@@ -180,6 +181,19 @@ Operation types:
 - `delete` - Tools that delete resources, such as delete document, drop collection, etc.
 - `read` - Tools that read resources, such as find, aggregate, list clusters, etc.
 - `metadata` - Tools that read metadata, such as list databases, list collections, collection schema, etc.
+
+#### Read-Only Mode
+
+The `readOnlyMode` configuration option allows you to restrict the MCP server to only use tools with "read" and "metadata" operation types. When enabled, all tools that have "create", "update", "delete", or "cluster" operation types will not be registered with the server.
+
+This is useful for scenarios where you want to provide access to MongoDB data for analysis without allowing any modifications to the data or infrastructure.
+
+You can enable read-only mode using:
+
+- **Environment variable**: `export MDB_MCP_READ_ONLY_MODE=true`
+- **Command-line argument**: `--readOnlyMode=true`
+
+When read-only mode is active, you'll see a message in the server logs indicating which tools were prevented from registering due to this restriction.
 
 ### Atlas API Access
 
--- a/README.md
+++ b/README.md
@@ -191,7 +191,7 @@ This is useful for scenarios where you want to provide access to MongoDB data fo
 You can enable read-only mode using:
 
 - **Environment variable**: `export MDB_MCP_READ_ONLY=true`
-- **Command-line argument**: `--readOnly=true`
+- **Command-line argument**: `--readOnly`
 
 When read-only mode is active, you'll see a message in the server logs indicating which tools were prevented from registering due to this restriction.
 
@@ -235,6 +235,7 @@ export MDB_MCP_API_CLIENT_SECRET="your-atlas-client-secret"
 export MDB_MCP_CONNECTION_STRING="mongodb+srv://username:password@cluster.mongodb.net/myDatabase"
 
 export MDB_MCP_LOG_PATH="/path/to/logs"
+
 ```
 
 #### Command-Line Arguments
--- a/README.md
+++ b/README.md
@@ -184,7 +184,7 @@ Operation types:
 
 #### Read-Only Mode
 
-The `readOnly` configuration option allows you to restrict the MCP server to only use tools with "read" and "metadata" operation types. When enabled, all tools that have "create", "update", "delete", or "cluster" operation types will not be registered with the server.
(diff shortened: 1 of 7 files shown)


src/logger.ts
```diff
@@ -106,6 +106,11 @@ class McpLogger extends LoggerBase {
     }
 
     log(level: LogLevel, _: MongoLogId, context: string, message: string): void {
+        // Only log if the server is connected
+        if (this.server?.isConnected() === false) {
```
blva: fly-by - when server is not connect, this can crash, so I've added this check.

fmenezes:
```suggestion
        if (!this.server?.isConnected()) {
```

README.md
```diff
@@ -181,6 +182,19 @@ Operation types:
...
+#### Read-Only Mode
+
+The `readOnlyMode` configuration option allows you to restrict the MCP server to only use tools with "read" and "metadata" operation types. When enabled, all tools that have "create", "update", "delete", or "cluster" operation types will not be registered with the server.
+
+This is useful for scenarios where you want to provide access to MongoDB data for analysis without allowing any modifications to the data or infrastructure.
+
+You can enable read-only mode using:
+
+- **Environment variable**: `export MDB_MCP_READ_ONLY_MODE=true`
+- **Command-line argument**: `--readOnlyMode=true`
```
fmenezes: Could we do book cli arg style? simply `--readOnly`

blva: yep! updating

README.md
```diff
@@ -181,6 +182,19 @@ Operation types:
...
+#### Read-Only Mode
+
+The `readOnly` configuration option allows you to restrict the MCP server to only use tools with "read" and "metadata" operation types. When enabled, all tools that have "create", "update", "delete", or "cluster" operation types will not be registered with the server.
+
+This is useful for scenarios where you want to provide access to MongoDB data for analysis without allowing any modifications to the data or infrastructure.
+
+You can enable read-only mode using:
+
+- **Environment variable**: `export MDB_MCP_READ_ONLY=true`
+- **Command-line argument**: `--readOnly=true`
```
nirinchev:
```suggestion
- **Command-line argument**: `--readOnly`
```

README.md
```diff
@@ -181,6 +182,19 @@ Operation types:
 - `read` - Tools that read resources, such as find, aggregate, list clusters, etc.
 - `metadata` - Tools that read metadata, such as list databases, list collections, collection schema, etc.
 
+#### Read-Only Mode
+
+The `readOnly` configuration option allows you to restrict the MCP server to only use tools with "read" and "metadata" operation types. When enabled, all tools that have "create", "update", "delete", or "cluster" operation types will not be registered with the server.
```
nirinchev: This is not a problem with this PR, but just noticing the `cluster` operation type doesn't make a lot of sense and in fact it's not used. If you're up for it, do you mind removing it as an option and also updating the readme here.

src/server.ts
```diff
@@ -36,6 +36,16 @@ export class Server {
 
     async connect(transport: Transport) {
         this.mcpServer.server.registerCapabilities({ logging: {} });
+
+        // Log read-only mode status if enabled
+        if (this.userConfig.readOnly) {
+            logger.info(
+                mongoLogId(1_000_005),
```
nirinchev: Can we move this to a constant under `logger/LogId` and update the value to be consistent with the other constants there.

src/server.ts
```diff
@@ -116,6 +126,7 @@ export class Server {
 
         if (command === "start") {
             event.properties.startup_time_ms = commandDuration;
+            event.properties.read_only_mode = this.userConfig.readOnly || false;
```
nirinchev, line 119: Unrelated to this PR, but do we want to track the `disabledTools` option from the user config as well? Might be something to consider as we're fleshing out the telemetry aspects.

src/tools/tool.ts
```diff
@@ -86,6 +86,17 @@ export abstract class ToolBase {
     // Checks if a tool is allowed to run based on the config
     protected verifyAllowed(): boolean {
         let errorClarification: string | undefined;
+
+        // Check read-only mode first
+        if (this.config.readOnly && !["read", "metadata"].includes(this.operationType)) {
+            logger.debug(
+                mongoLogId(1_000_010),
```
nirinchev: (after merging/rebasing on main)

```suggestion
                LogId.toolDisabled,
```

nirinchev (approved):
Looks good. I would have probably preferred to create a synthetic `write` operation type which expands to `creade/update/delete` and keep only `disabledTools`, but not a hill I'm willing to die on 😅 Your approach is more in line with other servers and probably easier to reason about, so happy to go that route as well.

README.md
blva: 100%, removed!

src/server.ts
blva: I actually decided to remove since we're not doing anything other than logging here, but will do from now on

src/server.ts
blva: yea! good idea

src/tools/tool.ts
```diff
@@ -109,7 +109,11 @@ export abstract class ToolBase {
     // Checks if a tool is allowed to run based on the config
     protected verifyAllowed(): boolean {
         let errorClarification: string | undefined;
-        if (this.config.disabledTools.includes(this.category)) {
+
+        // Check read-only mode first
+        if (this.config.readOnly && !["read", "metadata"].includes(this.operationType)) {
```
blva, line 114: keeping it like this so we don't alter the configured disallowed tools. i also think adding the virtual "write" would be an extra layer of complexity for users that id' like to avoid. 

the other thing we could do is remove this check and set the disallowed operation types during initalization, I can do that if you prefer, but I don't have a strong opinion!

blva: (cc @nirinchev )

nirinchev: I think this is totally fine!

One thing I noticed is that the SDK recently added the ability to enable/disable tools - https://github.com/modelcontextprotocol/typescript-sdk/pull/247. We should probably look into it post-public-preview and see if it makes sense to use that mechanism instead. Might also be interesting to see if there'd be a way to allow users to dynamically update the server config without restarting it.

nirinchev (approved):

src/tools/tool.ts
blva: nice, yep, that'd be awesome, I think in VSCode it gives you a checklist so I think it's possible depending on the tool already

PR #132, opened 2025-04-25
Add dependabot config

nirinchev:


.github/dependabot.yml
```diff
@@ -0,0 +1,6 @@
+version: 2
+updates:
```
fmenezes, line 2: how about github actions?

blva (approved):

PR #142, opened 2025-04-28
chore: disable the connect tool

nirinchev:


tests/integration/server.test.ts
```diff
@@ -1,14 +1,9 @@
...
 
 describe("Server integration test", () => {
-    describe("without atlas", () => {
-        const integration = setupIntegrationTest(() => ({
-            ...config,
-            apiClientId: undefined,
-            apiClientSecret: undefined,
-        }));
-
+    describeWithMongoDB("without atlas", (integration) => {
```
fmenezes: It was the way it was to be resilient against providing env vars on CI/terminal

fmenezes: nvm I see you've changed describeAtlas

nirinchev: You mean the undefining of the apiClientId/Secret - I can bring it back, the important thing here was to provide the connection string as otherwise the server was failing due to neither atlas credentials, nor connection string being configured.

src/common/atlas/apiClient.ts
```diff
@@ -63,12 +63,11 @@ export class ApiClient {
         },
     };
 
-    constructor(options?: ApiClientOptions) {
+    constructor(options: ApiClientOptions) {
         this.options = {
             ...options,
-            baseUrl: options?.baseUrl || "https://cloud.mongodb.com/",
```
fmenezes, line 69: the code was structured like that originally we moved `https://cloud.mongodb.com/` to be part of apiClient to keep everything about the API in one place, any reason to move it back?

src/tools/mongodb/mongodbTool.ts
```diff
@@ -4,12 +4,31 @@ import { NodeDriverServiceProvider } from "@mongosh/service-provider-node-driver
...
 import logger, { LogId } from "../../logger.js";
+import { UserConfig } from "../../config.js";
+import { Session } from "../../session.js";
 
 export const DbOperationArgs = {
     database: z.string().describe("Database name"),
     collection: z.string().describe("Collection name"),
 };
 
+export async function connectToMongoDB(connectionString: string, config: UserConfig, session: Session): Promise<void> {
```
fmenezes: this is slighly confusing to have `connectToMongoDB` pure function plus `connectToMongoDB` as member of `MongoDBToolBase` class on on the same file.

Have you considered moving this to session given it currently holds instance of `NodeDriverServiceProvider` ?

src/server.ts
```diff
@@ -33,7 +34,7 @@ export class Server {
         this.userConfig = userConfig;
     }
 
-    async connect(transport: Transport) {
+    async connect(transport: Transport): Promise<void> {
```
gagik, line 36: I think this should stay

gagik (approved):

src/tools/mongodb/mongodbTool.ts
gagik: in similar realm, a solution could be to have a wrapper class for `NodeDriverServiceProvider` (with an unconnected state) live inside `session` similar to `apiClient`. And that client would have the connect function.

My thinking is session is the entry to i.e. `apiClient` is the "plugin" for Atlas stuff, `databaseClient` (or something like that) is the "plugin" for connections, each with their own modular states.

tests/integration/tools/mongodb/mongodbHelpers.ts
```diff
@@ -14,24 +14,30 @@ interface MongoDBIntegrationTest {
 export function describeWithMongoDB(
     name: string,
     fn: (integration: IntegrationTest & MongoDBIntegrationTest & { connectMcpClient: () => Promise<void> }) => void,
-    getUserConfig: (mdbIntegration: MongoDBIntegrationTest) => UserConfig = () => config
+    getUserConfig: (mdbIntegration: MongoDBIntegrationTest) => UserConfig = () => config,
+    describeFn = describe
 ) {
-    describe(name, () => {
+    describeFn(name, () => {
```
gagik, line 20: nit: no strong preference but if this is just for skipping, could it be better to make this skippable instead?

otherwise, can we make these part of an options object? otherwise named optional parameters can go out of hand easily

src/common/atlas/apiClient.ts
nirinchev: I think it makes more sense to have the defaults for the config options together rather than disconnected like that. That way we can see that the `baseUrl` option is actually always set and the exact value it is set to.

tests/integration/tools/mongodb/mongodbHelpers.ts
nirinchev: This is temporary - I intend to remove it once we bring back the connect tool.

src/server.ts
nirinchev: Not sure what you mean - what should stay?

gagik: nevermind, I think I just saw it flipped where the type definition was being removed 😅

src/tools/mongodb/mongodbTool.ts
nirinchev: Not opposed to this, but I'd suggest doing it separately - I'd prefer that this PR is the bare minimum that disables the connect tool, rather than a broader restructuring of the codebase.

fmenezes (approved):

PR #143, opened 2025-04-28
fix: use record instead of a passthrough object

### LINKED ISSUE #139: Can't use MCP server in VSCode
When a user tries to use the MCP server enabled in VSCode with all tools enabled, they get:

```shell
Failed to validate tool 9f1_find: TypeError: Cannot use 'in' operator to search for 'type' in true
```

gagik:
Fixes #139.
Needs additional testing with other editors to ensure a record of unknowns is similarly respected. Tested with Claude.

### DIFF
--- a/src/tools/mongodb/create/insertMany.ts
+++ b/src/tools/mongodb/create/insertMany.ts
@@ -9,7 +9,7 @@ export class InsertManyTool extends MongoDBToolBase {
     protected argsShape = {
         ...DbOperationArgs,
         documents: z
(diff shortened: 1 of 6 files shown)


fmenezes (approved):
💯

nirinchev:
Something looks wrong with the checks - not sure why GHA hasn't run the workflow.

blva (approved):

nirinchev (approved):

PR #146, opened 2025-04-28
ci: set PR auto merge version bump PR

fmenezes:


.github/workflows/prepare_release.yaml
```diff
@@ -16,12 +16,12 @@ jobs:
   create-pr:
     runs-on: ubuntu-latest
     steps:
+      - uses: GitHubSecurityLab/actions-permissions/monitor@v1
```
fmenezes, line 19: drive-by: should always be first

blva (approved):

PR #159, opened 2025-04-29
chore: add orgId and projectid

blva:
- infers the values from args
- if using a mongodb tool, checks if it's connected to an atlas cluster

### DIFF
--- a/src/logger.ts
+++ b/src/logger.ts
@@ -17,6 +17,7 @@ export const LogId = {
     telemetryEmitFailure: mongoLogId(1_002_002),
     telemetryEmitStart: mongoLogId(1_002_003),
     telemetryEmitSuccess: mongoLogId(1_002_004),
+    telmetryMetadataError: mongoLogId(1_002_005),
 
     toolExecute: mongoLogId(1_003_001),
     toolExecuteFailure: mongoLogId(1_003_002),
--- a/src/logger.ts
+++ b/src/logger.ts
@@ -17,7 +17,7 @@ export const LogId = {
     telemetryEmitFailure: mongoLogId(1_002_002),
     telemetryEmitStart: mongoLogId(1_002_003),
     telemetryEmitSuccess: mongoLogId(1_002_004),
-    telmetryMetadataError: mongoLogId(1_002_005),
+    telemetryMetadataError: mongoLogId(1_002_005),
 
     toolExecute: mongoLogId(1_003_001),
     toolExecuteFailure: mongoLogId(1_003_002),
--- a/src/session.ts
+++ b/src/session.ts
@@ -26,7 +26,6 @@ export class Session extends EventEmitter<{
         username: string;
         projectId: string;
         clusterName: string;
-        organizationId?: string;
         expiryDate: Date;
     };
 
--- a/src/session.ts
+++ b/src/session.ts
@@ -26,6 +26,7 @@ export class Session extends EventEmitter<{
         username: string;
         projectId: string;
         clusterName: string;
+        organizationId?: string;
         expiryDate: Date;
     };
 
--- a/src/telemetry/telemetry.ts
+++ b/src/telemetry/telemetry.ts
@@ -1,6 +1,6 @@
 import { Session } from "../session.js";
 import { BaseEvent, CommonProperties } from "./types.js";
-import { config } from "../config.js";
+import { UserConfig } from "../config.js";
 import logger, { LogId } from "../logger.js";
 import { ApiClient } from "../common/atlas/apiClient.js";
 import { MACHINE_METADATA } from "./constants.js";
@@ -16,6 +16,7 @@ export class Telemetry {
 
     constructor(
         private readonly session: Session,
+        private readonly userConfig: UserConfig,
         private readonly eventCache: EventCache = EventCache.getInstance()
     ) {
         this.commonProperties = {
@@ -51,7 +52,7 @@ export class Telemetry {
             mcp_client_name: this.session.agentRunner?.name,
             session_id: this.session.sessionId,
             config_atlas_auth: this.session.apiClient.hasCredentials() ? "true" : "false",
-            config_connection_string: config.connectionString ? "true" : "false",
+            config_connection_string: this.userConfig.connectionString ? "true" : "false",
         };
     }
 
@@ -64,7 +65,7 @@ export class Telemetry {
      */
     public isTelemetryEnabled(): boolean {
         // Check if telemetry is explicitly disabled in config
-        if (config.telemetry === "disabled") {
+        if (this.userConfig.telemetry === "disabled") {
             return false;
         }
 
--- a/src/telemetry/telemetry.ts
+++ b/src/telemetry/telemetry.ts
@@ -11,23 +11,6 @@ type EventResult = {
     error?: Error;
 };
 
-/**
- * Checks if telemetry is currently enabled
- * This is a method rather than a constant to capture runtime config changes
- *
- * Follows the Console Do Not Track standard (https://consoledonottrack.com/)
- * by respecting the DO_NOT_TRACK environment variable
- */
-export function isTelemetryEnabled(): boolean {
-    // Check if telemetry is explicitly disabled in config
-    if (config.telemetry === "disabled") {
-        return false;
-    }
-
-    const doNotTrack = "DO_NOT_TRACK" in process.env;
-    return !doNotTrack;
-}
-
 export class Telemetry {
     private readonly commonProperties: CommonProperties;
 
@@ -46,7 +29,7 @@ export class Telemetry {
      */
     public async emitEvents(events: BaseEvent[]): Promise<void> {
         try {
-            if (!isTelemetryEnabled()) {
+            if (!this.isTelemetryEnabled()) {
                 logger.info(LogId.telemetryEmitFailure, "telemetry", `Telemetry is disabled.`);
                 return;
             }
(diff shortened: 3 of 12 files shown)


src/tools/mongodb/mongodbTool.ts
```diff
@@ -73,4 +74,18 @@ export abstract class MongoDBToolBase extends ToolBase {
     protected connectToMongoDB(connectionString: string): Promise<void> {
         return this.session.connectToMongoDB(connectionString, this.config.connectOptions);
     }
+
+    protected resolveToolMetadata(
```
fmenezes:
```suggestion
    protected resolveTelemetryMetadata(
```

src/tools/tool.ts
```diff
@@ -149,4 +131,89 @@ export abstract class ToolBase {
...
+     *
+     * Resolves the tool metadata from the arguments passed to the tool
+     *
+     * @param args - The arguments passed to the tool
+     * @returns The tool metadata
+     */
+    protected resolveToolMetadata(...args: Parameters<ToolCallback<typeof this.argsShape>>): ToolMetadata {
+        const toolMetadata: ToolMetadata = {};
+        try {
+            if (!args[0] || typeof args[0] !== "object") {
```
fmenezes:
```suggestion
            if (!args.length) {
```

blva: thanks, updateed

src/tools/mongodb/mongodbTool.ts
blva: updated

src/session.ts
```diff
@@ -26,6 +26,7 @@ export class Session extends EventEmitter<{
         username: string;
         projectId: string;
         clusterName: string;
+        organizationId?: string;
```
fmenezes: do you need this?

blva: nope removing

src/telemetry/telemetry.ts
```diff
@@ -11,6 +11,31 @@ type EventResult = {
...
+ * Follows the Console Do Not Track standard (https://consoledonottrack.com/)
+ * by respecting the DO_NOT_TRACK environment variable
+ */
+export function isTelemetryEnabled(): boolean {
+    // Check if telemetry is explicitly disabled in config
+    if (config.telemetry === "disabled") {
+        return false;
+    }
+
+    const doNotTrack = process.env.DO_NOT_TRACK;
```
fmenezes:
```suggestion
    const doNotTrack = "DO_NOT_TRACK" in process.env;
```

we normally check if env var is set and don't care about the value.

src/tools/tool.ts
```diff
@@ -3,14 +3,18 @@ import type { McpServer, RegisteredTool, ToolCallback } from "@modelcontextproto
...
-import { Telemetry } from "../telemetry/telemetry.js";
+import { Telemetry, isTelemetryEnabled } from "../telemetry/telemetry.js";
 import { type ToolEvent } from "../telemetry/types.js";
 import { UserConfig } from "../config.js";
 
 export type ToolArgs<Args extends ZodRawShape> = z.objectOutputType<Args, ZodNever>;
 
 export type OperationType = "metadata" | "read" | "create" | "delete" | "update";
 export type ToolCategory = "mongodb" | "atlas";
+export type ToolMetadata = {
```
fmenezes:
```suggestion
export type TelemetryToolMetadata = {
```

src/telemetry/telemetry.ts
blva: thx, updated

src/tools/tool.ts
```diff
@@ -149,4 +131,89 @@ export abstract class ToolBase {
...
+            // Extract projectId using type guard
+            if ("projectId" in data && typeof data.projectId === "string" && data.projectId.trim() !== "") {
+                toolMetadata.projectId = data.projectId;
+            }
+
+            // Extract orgId using type guard
+            if ("orgId" in data && typeof data.orgId === "string" && data.orgId.trim() !== "") {
+                toolMetadata.orgId = data.orgId;
+            }
+        } catch (error) {
```
fmenezes: [q] when might catch happen?

src/tools/tool.ts
blva: updated!

src/tools/tool.ts
blva: hopefully never I just don't want telemetry to cause any issues but can remove

src/tools/tool.ts
```diff
@@ -149,4 +131,89 @@ export abstract class ToolBase {
...
+     * @param startTime - Start time in milliseconds
+     * @param result - Whether the command succeeded or failed
+     * @param args - The arguments passed to the tool
+     */
+    private async emitToolEvent(
+        startTime: number,
+        result: CallToolResult,
+        ...args: Parameters<ToolCallback<typeof this.argsShape>>
+    ): Promise<void> {
+        if (!isTelemetryEnabled()) {
```
gagik: is this not redundant by the later `emitEvents` handling? we can also use `telemetry` object here and avoid moving it to a global function

blva: Yes, a bit, while I wan tto ensure telemetry is not emitted at the emit level, I also want to avoid processing metadata when we're not going to emit it, so I thought we could benefit from also letting toolEmission skip it right away

src/tools/tool.ts
fmenezes: I see no method that might have side effects, which line might throw? perhaps `argsShape.safeParse`?

perhaps the try catch is better suited for `emitToolEvent`, no?

blva: can be! moving

blva: it's just that we could still emit if this goes wrong, but no strong opinions

src/tools/tool.ts
gagik: sounds good; we have access to `telemetry` object so we can still do `telemetry.isTelemetryEnabled()` though

src/tools/tool.ts
```diff
@@ -149,4 +131,89 @@ export abstract class ToolBase {
...
+            return;
+        }
+        const duration = Date.now() - startTime;
+        const metadata = this.resolveTelemetryMetadata(...args);
+        const event: ToolEvent = {
+            timestamp: new Date().toISOString(),
+            source: "mdbmcp",
+            properties: {
+                ...this.telemetry.getCommonProperties(),
```
```suggestion
```
no longer necessary

src/tools/tool.ts
gagik: we could also telemetry take a function which is meant to return the events to emit

src/tools/tool.ts
blva: thx, updated

src/tools/tool.ts
blva: sure let me move to telemetry.isTelemetryEnabled()

blva: updated

src/logger.ts
```diff
@@ -17,6 +17,7 @@ export const LogId = {
     telemetryEmitFailure: mongoLogId(1_002_002),
     telemetryEmitStart: mongoLogId(1_002_003),
     telemetryEmitSuccess: mongoLogId(1_002_004),
+    telmetryMetadataError: mongoLogId(1_002_005),
```
gagik:
```suggestion
    telemetryMetadataError: mongoLogId(1_002_005),
```

tests/integration/tools/mongodb/mongodbHelpers.ts
```diff
@@ -22,10 +22,12 @@ export function describeWithMongoDB(
         const integration = setupIntegrationTest(() => ({
             ...getUserConfig(mdbIntegration),
             connectionString: mdbIntegration.connectionString(),
+            telemetry: "disabled", // Explicitly disable telemetry
         }));
 
         beforeEach(() => {
             integration.mcpServer().userConfig.connectionString = mdbIntegration.connectionString();
+            integration.mcpServer().userConfig.telemetry = "disabled"; // Ensure telemetry stays disabled
```
fmenezes: [q] do we need to do this for atlas as well?

fmenezes: if we are changing this value on the fly, it would be better for the test that changes it to change it back.

src/tools/tool.ts
```diff
@@ -149,4 +131,85 @@ export abstract class ToolBase {
...
     }
+
+    /**
+     *
+     * Resolves the tool metadata from the arguments passed to the tool
+     *
+     * @param args - The arguments passed to the tool
+     * @returns The tool metadata
+     */
+    protected resolveTelemetryMetadata(
```
fmenezes: shouldn't we move this code into atlasTool.ts?

blva: yes, since the args will never be provided to mdb tools

tests/integration/tools/mongodb/mongodbHelpers.ts
blva: I've been testing this, need to change it slightly, will udpate

src/tools/tool.ts
blva: moved

tests/integration/tools/mongodb/mongodbHelpers.ts
blva: updated, and atlas is not emitting, only mdbtools were missing

tests/integration/helpers.ts
```diff
@@ -23,8 +23,12 @@ export interface IntegrationTest {
 export function setupIntegrationTest(getUserConfig: () => UserConfig): IntegrationTest {
     let mcpClient: Client | undefined;
     let mcpServer: Server | undefined;
+    let oldDoNotTrackValue: string | undefined;
 
     beforeAll(async () => {
+        // GET DO_NOT_TRACK value
+        oldDoNotTrackValue = process.env.DO_NOT_TRACK;
+        process.env.DO_NOT_TRACK = "1";
         const userConfig = getUserConfig();
```
fmenezes: Instead of keeping track of DO_NOT_TRACK, why not?
```suggestion
        const userConfig = {
           ...getUserConfig(),
           telemetry: "disabled"
       };
```

blva: actually, let me try to sort our default test config everywhere

blva: I think I had this somewhere too, but I can add it if https://github.com/mongodb-js/mongodb-mcp-server/pull/159/commits/3eee38c5e4617feba2d05a70060fc77e51e6d743 doesn't work

blva: this won't work too cause telemetry imports config so it's not injected, only env variables would work or we can refactor

blva: I'll refactor a bit

fmenezes (approved):

PR #161, opened 2025-04-29
 chore: defer machine ID resolution 

### LINKED ISSUE #122: Defer machine ID resolution
We're using `getMachineIdSync` in the global scope which blocks the entire thread at startup regardless of us using telemetry of not. `native-machine-id` is very quick (developed by a great team 😉) but we'd likely want to refactor our telemetry setup to instead asynchronously wait for the device ID to be resolved and cache events until then. 

Similar to https://github.com/mongodb-js/mongosh/pull/2411/files#diff-d717df5f752b4b9552410572829eab7bda35a181aeff5d57eab9f07a657f25d9

gagik:
Fixes #122
This defers the device ID resolution and add a timeout for resolution. Buffers events in the meantime.


src/telemetry/telemetry.ts
```diff
@@ -5,22 +5,76 @@ import logger, { LogId } from "../logger.js";
...
-        private readonly eventCache: EventCache = EventCache.getInstance()
-    ) {
-        this.commonProperties = {
-            ...MACHINE_METADATA,
-        };
+        private readonly commonProperties: CommonProperties,
+        private readonly eventCache: EventCache
+    ) {}
+
+    static create(
```
gagik, line 36: while this is not asyncronous. I'm using `create(...)` to make it clear there are associated side effects

.github/workflows/code_health.yaml
```diff
@@ -74,13 +74,7 @@ jobs:
           node-version-file: package.json
           cache: "npm"
       - name: Install dependencies & build
-        run: npm ci
-      - name: Remove dev dependencies
-        run: |
-          rm -rf node_modules
-          npm pkg set scripts.prepare="exit 0"
-          npm install --omit=dev
-      - run: npx -y @modelcontextprotocol/inspector --cli --method tools/list -- node dist/index.js --connectionString "mongodb://localhost"
```
gagik, line 83: this always fails as even if inspector works correctly, the CI will correctly fail because of invalid connection; removing for now

gagik:
This has revealed a very major problem with our Build vs Jest setup targetting entirely different module resolutions... Unfortunately I can't think of a good solution... `node-machine-id` is the only CommonJS-only dependency we have and it completely breaks our ability to test and build

tests/integration/tools/mongodb/mongodbHelpers.ts
```diff
@@ -76,8 +76,6 @@ export function setupMongoDBIntegrationTest(): MongoDBIntegrationTest {
         let dbsDir = path.join(tmpDir, "mongodb-runner", "dbs");
         for (let i = 0; i < 10; i++) {
             try {
-                // TODO: Fix this type once mongodb-runner is updated.
-                // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-unsafe-call
```
gagik, line 80: drive-by

src/deferred-promise.ts
```diff
@@ -0,0 +1,60 @@
+type DeferredPromiseOptions<T> = {
+    timeout?: number;
+    onTimeout?: (resolve: (value: T) => void, reject: (reason: Error) => void) => void;
+};
+
+/** Creates a promise and exposes its resolve and reject methods, with an optional timeout. */
+export class DeferredPromise<T> extends Promise<T> {
+    resolve!: (value: T) => void;
+    reject!: (reason: unknown) => void;
```
nirinchev: Do we need the assertions here?

```suggestion
    resolve: (value: T) => void;
    reject: (reason: unknown) => void;
```

src/deferred-promise.ts
```diff
@@ -0,0 +1,60 @@
...
+        });
+
+        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+        this.resolve = resolveFn!;
+        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+        this.reject = rejectFn!;
+
+        if (timeout !== undefined) {
+            this.timeoutId = setTimeout(() => {
+                onTimeout?.(this.resolve, this.reject);
```
nirinchev: Does it make sense to provide `timeout` without `onTimeout`? As far as I can understand the code, we'd be setting a timeout that does nothing - am I missing something here?

src/deferred-promise.ts
```diff
@@ -0,0 +1,60 @@
...
+            }, timeout);
+        }
+
+        if (executor) {
```
nirinchev: `executor` is required in the ctor - in what case are we getting falsy value here?

src/telemetry/telemetry.ts
```diff
@@ -5,23 +5,101 @@ import logger, { LogId } from "../logger.js";
...
+    }
+
+    public async close(): Promise<void> {
+        this.deviceIdPromise?.resolve("unknown");
+        this.isBufferingEvents = false;
+        await this.emitEvents(this.eventCache.getEvents());
+    }
+
+    /**
+     * @returns A hashed, unique identifier for the running device or `undefined` if not known.
```
nirinchev:
```suggestion
     * @returns A hashed, unique identifier for the running device or `unknown` if not known.
```

src/telemetry/telemetry.ts
```diff
@@ -48,6 +126,7 @@ export class Telemetry {
     public getCommonProperties(): CommonProperties {
         return {
             ...this.commonProperties,
+            device_id: this.commonProperties.device_id,
```
nirinchev: What's the point of this?

src/deferred-promise.ts
gagik: yeah, this is more of a late intialization assertion. typescript isn't smart enough to understand they're being set in the constructor

src/telemetry/telemetry.ts
gagik: leftover from some weird issues I was running into 😅 will remove

src/deferred-promise.ts
nirinchev: Not sure why though - when testing locally, I don't see it complaining when I remove the assertions - in what situations are you seeing errors?

nirinchev (approved):

src/deferred-promise.ts
gagik: ah, the structure seems to have changed back to a different setup, this is what I was thinking about: [`25d9059` (#161)](https://github.com/mongodb-js/mongodb-mcp-server/pull/161/commits/25d905907f738591e7fc1d8d22708ca513c9829a) (cleaner to do this imo)

gagik: okay I'm realizing now the assertions were leftover from an older implementation that had a different structure; removed them now. as we assert them later on instead.

PR #166, opened 2025-04-29
fix: use proper ESM with jest

gagik:
We did not have a full ESM configuration for jest and therefore had a discrepancy between our build vs test. This discrepancy leads to issues with CommonJS modules.


fmenezes (approved):
🚀

nirinchev (approved):

PR #176, opened 2025-04-30
fix: improve api error messages

fmenezes:


src/common/atlas/apiClientError.ts
```diff
@@ -1,21 +1,51 @@
...
+            text = body.reason || "unknown error";
+            if (body.detail && body.detail.length > 0) {
+                text = `${text}; ${body.detail}`;
+            }
         } catch {
-            return new ApiClientError(`${message}: ${response.status} ${response.statusText}`, response);
+            try {
+                text = await response.text();
+            } catch {
+                text = "";
```
blva: should we set to something like "couldn't fetch error" or "unknown error" too

fmenezes: ok

blva (approved):

src/common/atlas/apiClientError.ts
```diff
@@ -1,21 +1,48 @@
-export class ApiClientError extends Error {
-    response?: Response;
+import { ApiError } from "./openapi.js";
 
-    constructor(message: string, response: Response | undefined = undefined) {
+export class ApiClientError extends Error {
+    private constructor(
+        message: string,
+        public readonly response?: Response,
+        public readonly body?: ApiError
```
gagik: is there a benefit to us storing the body?

gagik:
Looks good but I agree we should hold off from merging/releasing for the time being like with https://github.com/mongodb-js/mongodb-mcp-server/pull/161 as it's not super critical yet.

src/common/atlas/apiClientError.ts
fmenezes: Getting the ApiError model gives us insight into backend error codes, some tools might require an if/else depending on the backend error code.

fmenezes:
@gagik I think this change fixes type errors from apiClient

gagik (approved):
Nice

PR #178, opened 2025-04-30
fix: bad auth error

fmenezes:
This essentially keeps trying to connect because after user is created it might take some time for it to be available.

note: also making deletion of user async on disconnect to reduce change of timeout

### DIFF
--- a/src/tools/atlas/metadata/connectCluster.ts
+++ b/src/tools/atlas/metadata/connectCluster.ts
@@ -101,12 +101,17 @@ export class ConnectClusterTool extends AtlasToolBase {
         cn.searchParams.set("authSource", "admin");
         const connectionString = cn.toString();
 
+        let lastError: Error | undefined = undefined;
+
         for (let i = 0; i < 20; i++) {
             try {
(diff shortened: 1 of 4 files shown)


src/tools/atlas/metadata/connectCluster.ts
```diff
@@ -100,7 +105,31 @@ export class ConnectClusterTool extends AtlasToolBase {
         cn.searchParams.set("authSource", "admin");
         const connectionString = cn.toString();
 
-        await this.session.connectToMongoDB(connectionString, this.config.connectOptions);
+        let lastError: Error | undefined = undefined;
+
+        for (let i = 0; i < 20; i++) {
+            try {
```
blva, line 111: nit: can we add tests for this in a follow-up?

blva (approved):

src/tools/atlas/metadata/connectCluster.ts
fmenezes: in the integration tests we can't force a user to take longer to be created, we could try with unit tests, but none of the tools are covered in unit at the moment

PR #186, opened 2025-05-02
Incorrect link of VSCode's MCP usage

up1:
- Change link of MCP usage


gagik:
Good catch, thank you for the contribution 🙂 Seems like we need to adjust the CI a little to make it work for outside contributions but should be good to go after that.

gagik (approved):

fmenezes (approved):

PR #187, opened 2025-05-02
fix: db user test error

fmenezes:


gagik (approved):
It's weird that CI didn't catch it earlier; not sure how that happened 🫨

PR #194, opened 2025-05-02
fix: fork checks

fmenezes:
- Renames lint to checks
- Moves dependency checks to the main checks workflow
- Enables the checks workflow to run on forks

### DIFF
--- a/.github/workflows/code_health_fork.yaml
+++ b/.github/workflows/code_health_fork.yaml
@@ -42,8 +42,6 @@ jobs:
     permissions:
       pull-requests: write
       contents: write
-    needs:
-      - run-tests
     steps:
       - name: Enable auto-merge for Dependabot PRs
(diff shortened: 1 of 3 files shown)


.github/workflows/code_health_fork.yaml
```diff
@@ -42,8 +42,6 @@ jobs:
     permissions:
       pull-requests: write
       contents: write
-    needs:
```
fmenezes, line 45: drive-by: no need, PRs will wait for tests

fmenezes:
cc @nirinchev

.github/workflows/code_health_fork.yaml
gagik: I don't see any harm in keeping it though. If we decide to remove these checks for whatever reason we don't want the dependabot to pop up and actually do this

PR #196, opened 2025-05-05
chore: switch to `@mongodb-js/device-id`

gagik:
This package is meant to minimize code replication and to make sure the approach we're taking with hashing is consistent and timeouts are automatically handled.

See https://github.com/mongodb-js/devtools-shared/pull/532


jest.config.cjs
```diff
@@ -1,5 +1,5 @@
 /** @type {import('ts-jest').JestConfigWithTsJest} **/
-export default {
```
gagik, line 2: I ran into https://github.com/jestjs/jest/issues/15312, possibly because of a TypeScript update, so just going to keep it as a `.cjs` file for the time being

src/telemetry/telemetry.ts
```diff
@@ -57,50 +56,35 @@ export class Telemetry {
...
         }
-        this.deviceIdPromise = DeferredPromise.fromPromise(this.getDeviceId(), {
-            timeout: DEVICE_ID_TIMEOUT,
-            onTimeout: (resolve) => {
-                resolve("unknown");
-                logger.debug(LogId.telemetryDeviceIdTimeout, "telemetry", "Device ID retrieval timed out");
+        this.deviceIdPromise = getDeviceId({
+            getMachineId: () => this.getRawMachineId(),
+            onError: (reason, error) => {
+                switch (reason) {
```
gagik, line 62: honestly seeing this used like this does make me think we should instead throw typed error objects and check `instanceof` as reason feels a bit redundant but this is fine

nirinchev: Yeah, I was debating between the two for a while - ended up going with a `reason` because that was more explicit about the possible values as opposed to `Error` inheritors, where you'd need to figure out what the options are.

PR #199, opened 2025-05-05
Update connection string app name if not present

nirinchev:


src/session.ts
```diff
@@ -97,7 +100,16 @@ export class Session extends EventEmitter<{
...
-    async connectToMongoDB(connectionString: string, connectOptions: ConnectOptions): Promise<void> {
+    async connectToMongoDB(
+        connectionString: string,
+        connectOptions: ConnectOptions,
+        telemetry: Telemetry
+    ): Promise<void> {
+        connectionString = setAppNameParamIfMissing({
+            connectionString,
+            defaultAppName: `${packageInfo.mcpServerName} ${packageInfo.version}`,
+            telemetryAnonymousId: await telemetry.deviceIdPromise,
```
gagik: this does mean we're delaying connecting to a cluster for the sake of telemetry 😢 could we not append anonymous ID yet, especially since we use the segment anonymous ID in all other places at the moment so this won't match it anyhow

gagik (approved):

Issue #211, opened 2025-05-06
[Bug]: ObjectId query syntax in fails as a filter when querying in MCP tools with "unknown operator: $oid" error


dovstern:
### App

- [x] Cursor
- [ ] Windsurf
- [ ] VSCode
- [ ] VSCode Insiders
- [ ] Claude Desktop
- [ ] Other

### Affected Models (if applicable)

- [ ] Claude 3.5 Sonnet
- [x] Claude 3.7 Sonnet
- [ ] GPT-4a
- [ ] o4-mini
- [ ] Other

### Bug Description

Title:
ObjectId query syntax fails in MCP tools with "unknown operator: $oid" error
Description:
Problem
When attempting to query documents using ObjectId references in MongoDB MCP tools, queries consistently fail with the error "unknown operator: $oid". This occurs when using standard MongoDB syntax for querying by ObjectId.
Steps to reproduce
Attempt to query a collection using the MCP MongoDB tools
Use a filter containing an ObjectId reference with standard syntax:
Apply to mcp.json
}
Observe the error: "Error running find: unknown operator: $oid"
Failed queries
The following query patterns all produce the same error:
{"_id": {"$oid": "..."}}
{"userId": {"$oid": ""..."}}
{"_id": {"$eq": {"$oid": ""..."}}}
Using the string format without $oid operator does not retrieve documents:
{"_id": ""..."}
Expected behavior
The MCP tools should properly translate the ObjectId query syntax to the appropriate MongoDB query, allowing users to query documents by their ObjectId fields using standard MongoDB syntax.
Environment
MCP tools in Cursor/AI assistance environment
Querying a MongoDB database with collections containing ObjectId fields
Impact
This issue significantly impacts the ability to query documents by their IDs, which is a fundamental operation when working with MongoDB.

nirinchev:
Thanks for reporting this - I was able to reproduce and hope to have a fix in the next release.

PR #218, opened 2025-05-07
fix: use ejson parsing for stdio messages

nirinchev:
This works around the issue where we receive bson types serialized as ejson, but not converted into the corresponding types.

Fixes https://github.com/mongodb-js/mongodb-mcp-server/issues/211


blva (approved):

gagik (approved):
I think for our situation it's an acceptable enough workaround.

dovstern:
Thank you!

PR #222, opened 2025-05-08
fix: validate creds

blva:


src/common/atlas/apiClient.ts
```diff
@@ -124,12 +128,41 @@ export class ApiClient {
...
+                // For 401, fall through to unauthenticated endpoint
+                delete headers["Authorization"];
+            } catch (error) {
+                // If the error is not a 401, rethrow it
+                if (!(error instanceof ApiClientError) || error.response.status !== 401) {
+                    throw error;
+                }
+
+                // For 401 errors, fall through to unauthenticated endpoint
+                delete headers["Authorization"];
```
nirinchev, line 159: Can we ever get here? Doesn't hurt to have it just in case, but also wondering if I'm not seeing something.

nirinchev (approved):

src/common/atlas/apiClient.ts
blva: Yes, if someone sets connection string + invalid api keys, we let them start the server

gagik: ~if fetch errors, we'd get here right?~ actually not sure

blva: we'd get here if we get a 401, if any other error, we do nothing

nirinchev: Yeah, my point was that we need to get a 401 wrapped in a `ApiClientError`, which `fetch` shouldn't be doing, should it? The only place we can throw `ApiClientError` is on line 146, but that explicitly avoids throwing for 401s.

blva: got it! i think you're right, let me try to fix it

PR #224, opened 2025-05-08
fix: improve uncaught exception for getAccessToken

blva:
- improves send events to make it clearer when we send auth/unauth
- adds one more test

### DIFF
--- a/src/common/atlas/apiClient.ts
+++ b/src/common/atlas/apiClient.ts
@@ -119,13 +119,12 @@ export class ApiClient {
     }
 
     public async sendEvents(events: TelemetryEvent<CommonProperties>[]): Promise<void> {
-        if (!this.hasCredentials()) {
+        if (!this.options.credentials) {
             await this.sendUnauthEvents(events);
             return;
         }
 
         try {
-            await this.validateAccessToken();
             await this.sendAuthEvents(events);
         } catch (error) {
             if (error instanceof ApiClientError) {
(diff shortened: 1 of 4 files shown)


src/common/atlas/apiClient.ts
```diff
@@ -89,9 +89,8 @@ export class ApiClient {
         return !!(this.oauth2Client && this.accessToken);
     }
 
-    public async hasValidAccessToken(): Promise<boolean> {
-        const accessToken = await this.getAccessToken();
-        return accessToken !== undefined;
+    public async validateAccessToken(): Promise<void> {
+        await this.getAccessToken();
```
gagik, line 93: not sure if I got it right but this can end up just not finding the access token right? so it'd effectively not validate it always?

blva: it validates if it exists otherwise it will just return undefined, which is fine

src/common/atlas/apiClient.ts
```diff
@@ -119,48 +118,57 @@ export class ApiClient {
...
-                delete headers["Authorization"];
-            } catch (error) {
-                // If the error is not a 401, rethrow it
-                if (!(error instanceof ApiClientError) || error.response.status !== 401) {
-                    throw error;
-                }
+    private async sendAuthEvents(events: TelemetryEvent<CommonProperties>[]): Promise<void> {
+        const accessToken = await this.getAccessToken();
+        if (!accessToken) {
+            throw new Error("No access token available");
```
wtrocki, line 146: [Not Actionable] Known use case so best to not throw error and resurect to log.

wtrocki (approved):
LGTM. 
It is nice to see how you have wrapped Authentication mechanism.

Issue #216, opened 2025-05-07
Add support to list all alerts in one Atlas project


blva:
- Start introducing some monitoring tools for Atlas
- https://www.mongodb.com/docs/atlas/reference/api-resources-spec/v2/#tag/Alerts/operation/listAlerts is the API endpoint

blva:
closed with https://github.com/mongodb-js/mongodb-mcp-server/pull/230, thanks @wtrocki !

PR #230, opened 2025-05-12
feat: Alerts Listing

wtrocki:
# Add MongoDB Atlas Alerts List Tool

This PR adds a new tool to list alerts from MongoDB Atlas projects. The tool provides a tabular view of alerts with their status, creation time, and other relevant details.

## Changes

- Added `ListAlertsTool` class to handle listing alerts from Atlas projects
- Added `listAlerts` method to the Atlas API client
- Added OpenAPI type definitions for the alerts endpoint
- Added integration tests for the new tool
- Added the tool to the list of available Atlas tools

## Testing

The changes include integration tests that verify:
- Tool metadata and schema validation
- Table formatting of alert results
- Proper error handling for projects with no alerts

## Example Usage

```typescript
// List alerts for a project
await mcpClient.callTool({
    name: "atlas-list-alerts",
    arguments: { projectId: "your-project-id" }
});
```

The tool returns alerts in a table format with the following columns:
- Alert ID
- Status (OPEN/CLOSED/CANCELLED)
- Created (timestamp)
- Updated (timestamp)
- Type (event type name)
- AlertAck Comment - if present

## API Reference

The tool uses the MongoDB Atlas API endpoint:
`https://cloud.mongodb.com/api/atlas/v2/groups/{groupId}/alerts`

Docs:
https://www.mongodb.com/docs/atlas/reference/api-resources-spec/v2/#tag/Access-Tracking/operation/listAccessLogsByHostname

## Related Issues

https://github.com/mongodb-js/mongodb-mcp-server/issues/216

## Checklist

- [x] Added new tool implementation
- [x] Added API client method
- [x] Added integration tests
- [x] Updated tool registry
- [x] Verified error handling
- [x] Followed existing code patterns

## Tests

Test Suites: 6 skipped, 26 passed, 26 of 32 total
Tests:       26 skipped, 284 passed, 310 total
Snapshots:   0 total
Time:        12.155 s
Ran all test suites.


src/tools/atlas/read/listAlerts.ts
```diff
@@ -0,0 +1,44 @@
...
+                    groupId: projectId,
+                },
+            },
+        });
+
+        if (!data?.results?.length) {
+            throw new Error("No alerts found in your MongoDB Atlas project.");
+        }
+
+        // Format alerts as a table
```
nirinchev, line 27: [nit] I know that this is how we've returned results in Atlas, but in my experience, the models rarely render the response 1:1, so the extra effort we're going through to get a markdown table is not really useful for users. This was not something we wanted to spend too much time on, but now that we're post-public-preview, we can explore if alternative response shapes perform similarly and if so - perhaps, choose a more natural way to represent things (e.g. json).

src/tools/atlas/read/listAlerts.ts
```diff
@@ -0,0 +1,44 @@
...
+        const data = await this.session.apiClient.listAlerts({
+            params: {
+                path: {
+                    groupId: projectId,
+                },
+            },
+        });
+
+        if (!data?.results?.length) {
+            throw new Error("No alerts found in your MongoDB Atlas project.");
```
nirinchev: This doesn't seem like a real error though - if a user legitimately may end up in a situation with no Atlas alerts, we should return a graceful response informing them rather than an error (the model will sometimes treat errors as an indication that something is wrong and attempt to correct itself).

nirinchev (commented):
Some initial thoughts - obviously get some input from @blva or @fmenezes as our resident Atlas experts, but nothing seems obviously wrong.

src/tools/atlas/read/listAlerts.ts
blva: I think this is a broad issue in our Atlas tools, we follow this pattern and probably should change everywhere. Agree it's not an error unless I'm missing something @fmenezes

wtrocki: What would be your preference? Empty table?

wtrocki: Changed to present useful information to user.

src/tools/atlas/read/listAlerts.ts
wtrocki: Yes. We should support output types like atlas CLI does. 
Specifically for non chatbot use cases like copilot everyone wants output to consume least tokens possible - JSON/YAML

blva (approved):
LGTM

PR #238, opened 2025-05-13
feat: docker support

fmenezes:
fixes https://github.com/mongodb-js/mongodb-mcp-server/issues/236

### DIFF
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -1,5 +1,4 @@
-# See RELEASING.md#DockerImage for more details about the steps in this workflow.
-name: Daily Release AtlasCLI Docker Image
+name: Daily Release Docker Image
 on:
   schedule:
     - cron: "0 1 * * *" # Every day at 1:00 AM
@@ -46,5 +45,5 @@ jobs:
     #     uses: imjohnbo/issue-bot@572eed14422c4d6ca37e870f97e7da209422f5bd
     #     with:
     #       labels: failed-release
-    #       title: Release Failure for Atlas CLI Docker Image ${{ env.TAG_PREFIX }}${{ steps.set-properties.outputs.VERSION }}-${{ steps.set-properties.outputs.DATE }}
+    #       title: Release Failure for Docker Image ${{ env.TAG_PREFIX }}${{ steps.set-properties.outputs.VERSION }}-${{ steps.set-properties.outputs.DATE }}
     #       body: See https://github.com/${{ github.repository }}/actions/runs/${{ github.run_id }}
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -8,12 +8,12 @@ permissions:
   contents: read
   issues: write
 jobs:
-  build_images:
-    name: Build and publish docker image
+  properties:
+    name: Set properties
     runs-on: ubuntu-latest
-    env:
-      IMAGE_REPOSITORY: docker.io/mongodb/apix_test
-      TAG_PREFIX: mongodb-mcp-server-
+    outputs:
+      date: ${{ steps.set-properties.outputs.DATE }}
+      version: ${{ steps.set-properties.outputs.VERSION }}
     steps:
       - uses: GitHubSecurityLab/actions-permissions/monitor@v1
         with:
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -109,6 +109,75 @@ You can use environment variables in the config file or set them and run the ser
 - Connection String via environment variables in the MCP file [example](#connection-string-with-environment-variables)
 - Atlas API credentials via environment variables in the MCP file [example](#atlas-api-credentials-with-environment-variables)
 
+#### Option 5: Using Docker
+
+You can run the MongoDB MCP Server in a Docker container, which provides isolation and doesn't require a local Node.js installation.
+
+#### Run with Environment Variables
+
+You need to provide either a MongoDB connection string OR Atlas API credentials:
```
nirinchev: We don't need the connection string anymore, though we should probably preface that with a warning that if you don't, you'll be exposing the connection string to the LLM.

fmenezes: updated readme

.github/workflows/docker.yaml
```diff
@@ -0,0 +1,57 @@
...
+  issues: write
+jobs:
+  push:
+    runs-on: ubuntu-latest
+    env:
+      IMAGE_REPOSITORY: docker.io/mongodb/mongodb-mcp-server
+    steps:
+      - uses: GitHubSecurityLab/actions-permissions/monitor@v1
+        with:
+          config: ${{ vars.PERMISSIONS_CONFIG }}
```
nirinchev, line 15: What is this? Doesn't seem like it's created.

.github/workflows/docker.yaml
```diff
@@ -0,0 +1,57 @@
...
+      - name: Check out code
+        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@b5ca514318bd6ebac0fb2aedd5d36ec1b5c232a2
+      - name: Login to Docker Hub
+        uses: docker/login-action@343f7c4344506bcbf9b4de18042ae17996df046d
+        with:
+          username: "${{ secrets.DOCKERHUB_USERNAME }}"
+          password: "${{ secrets.DOCKERHUB_SECRET }}"
```
nirinchev: Not blocking, but should we rename this secret to `DOCKERHUB_PASSWORD`?

.github/workflows/docker.yaml
```diff
@@ -0,0 +1,57 @@
...
+    - cron: "0 1 * * *" # Every day at 1:00 AM
+  workflow_dispatch: # Run the action manually
+permissions:
+  contents: read
+  issues: write
+jobs:
+  push:
+    runs-on: ubuntu-latest
+    env:
+      IMAGE_REPOSITORY: docker.io/mongodb/mongodb-mcp-server
```
nirinchev: Should we move this to a repo-level variable?

.github/workflows/docker.yaml
```diff
@@ -0,0 +1,57 @@
...
+      - name: Check out code
+        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@b5ca514318bd6ebac0fb2aedd5d36ec1b5c232a2
+      - name: Login to Docker Hub
+        uses: docker/login-action@343f7c4344506bcbf9b4de18042ae17996df046d
+        with:
+          username: "${{ secrets.DOCKERHUB_USERNAME }}"
+          password: "${{ secrets.DOCKERHUB_SECRET }}"
+      - name: Set properties
```
nirinchev: [nit] This name is a bit vague, forcing readers to actually read the script being executed - would it make sense to replace with "Set date and version"?

.github/workflows/docker.yaml
```diff
@@ -0,0 +1,57 @@
...
+        uses: docker/build-push-action@14487ce63c7a62a4a324b0bfb37086795e31c6c1
+        with:
+          context: .
+          platforms: linux/amd64,linux/arm64
+          tags: ${{ env.IMAGE_REPOSITORY }}:latest, ${{ env.IMAGE_REPOSITORY }}:${{ steps.set-properties.outputs.VERSION }}, ${{ env.IMAGE_REPOSITORY }}:${{ steps.set-properties.outputs.VERSION }}-${{ steps.set-properties.outputs.DATE }}
+          file: Dockerfile
+          push: true
+          build-args: |
+            VERSION=${{ steps.set-properties.outputs.VERSION }}
+      - uses: mongodb-js/devtools-shared/actions/setup-bot-token@main
```
nirinchev, line 42: We typically use the bot in cases where the actions token wouldn't work - e.g. if we want to open a PR, but also run the associated workflows (which wouldn't happen with the GHA token). In this case, unless we want to run follow-up workflows on the newly created issue, we should be able to just use the GHA token. Alternatively, if we need the bot token, we can remove the `issues: write` permissions as those don't really affect it.

nirinchev (approved):
Just nits around the GHA workflows from me. I've turned off the auto-merge so you get a chance to look at the comments and address those that make sense to you, but no objections to merging it once we update the workflow permissions.

.github/workflows/docker.yaml
fmenezes: According to https://github.com/GitHubSecurityLab/actions-permissions/tree/v1/monitor, it is advisable to keep a variable even if it does not exist yet.

.github/workflows/docker.yaml
fmenezes: This was intentional, as we intend to run GitHub actions on GitHub Issues (Jira integration).

If an issue is created by Github Actions user, it won't run further GitHub Actions.

PR #242, opened 2025-05-13
chore: test with node 20

gagik:
This is a test PR, not meant to be merged.

We want to test on Node 20 to make sure we enforce the compatibility with old APIs.


gagik:
Closing in favor of #246.
Actually might want to follow up to set our minimum version anyways.

PR #244, opened 2025-05-13
feat: create and update Search Index

edgarw19:
This is specifically for search indexes. The create vector index definition seems sufficiently different that I think it's worth separating.

VSCode copilot at least from my testing does better at distinguishing the two and understanding the fields to set when they are separate.

Tests will be added in a separate PR.

### DIFF
--- a/src/tools/mongodb/create/createSearchIndex.ts
+++ b/src/tools/mongodb/create/createSearchIndex.ts
@@ -1,0 +1,38 @@
+import { DbOperationArgs, MongoDBToolBase, SearchIndexArgs } from "../mongodbTool.js";
+
+export class CreateSearchIndexTool extends MongoDBToolBase {
+    constructor() {
+        super(...arguments);
+        this.name = "create-search-index";
+        this.description = "Create an Atlas Search index for a collection";
+        this.argsShape = {
+            ...DbOperationArgs,
+            name: SearchIndexArgs.name,
+            type: SearchIndexArgs.type,
+            analyzer: SearchIndexArgs.analyzer,
+            mappings: SearchIndexArgs.mappings,
+        };
+        this.operationType = "create";
+    }
+    async execute({ database, collection, name, type, analyzer, mappings, }) {
+        const provider = await this.ensureConnected();
+        const indexes = await provider.createSearchIndexes(database, collection, [
+            {
+                name,
+                type,
+                definition: {
+                    analyzer,
+                    mappings,
+                },
+            },
+        ]);
+        return {
+            content: [
+                {
+                    text: `Created the index "${indexes[0]}" on collection "${collection}" in database "${database}"`,
+                    type: "text",
+                },
+            ],
+        };
+    }
+}
--- a/src/tools/mongodb/create/createSearchIndex.ts
+++ b/src/tools/mongodb/create/createSearchIndex.ts
@@ -2,13 +2,14 @@ import { CallToolResult } from "@modelcontextprotocol/sdk/types.js";
 import { DbOperationArgs, MongoDBToolBase, SearchIndexArgs } from "../mongodbTool.js";
 import { OperationType, ToolArgs } from "../../tool.js";
 
+const SEARCH_INDEX_TYPE = "search";
+
 export class CreateSearchIndexTool extends MongoDBToolBase {
     protected name = "create-search-index";
     protected description = "Create an Atlas Search index for a collection";
     protected argsShape = {
         ...DbOperationArgs,
         name: SearchIndexArgs.name,
-        type: SearchIndexArgs.type,
         analyzer: SearchIndexArgs.analyzer,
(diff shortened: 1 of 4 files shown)


src/tools/mongodb/update/updateSearchIndex.ts
```diff
@@ -0,0 +1,41 @@
...
+        database,
+        collection,
+        name,
+        analyzer,
+        mappings,
+    }: ToolArgs<typeof this.argsShape>): Promise<CallToolResult> {
+        const provider = await this.ensureConnected();
+        // @ts-expect-error: Interface expects a SearchIndexDefinition. However,
+        // passing analyzer/mappings at the root for the definition is necessary for the call to succeed.
+        await provider.updateSearchIndex(database, collection, name, {
```
edgarw19, line 27: If the index does not exist, MCP inspector shows a descriptive error message

`"Error running update-search-index: Search index sample_mflix.comments.test3 cannot be found"`

src/tools/mongodb/mongodbTool.ts
```diff
@@ -10,6 +10,61 @@ export const DbOperationArgs = {
     collection: z.string().describe("Collection name"),
 };
 
+export const SearchIndexArgs = {
```
edgarw19, line 13: This currently includes a few key arguments as a starting point

src/tools/mongodb/mongodbTool.ts
```diff
@@ -10,6 +10,61 @@ export const DbOperationArgs = {
...
+    name: z.string().describe("The name of the index"),
```
edgarw19, line 14: This currently includes a few common arguments as a starting point.

It excludes custom analyzers/synonyms/storedsource since but we can add those if needed

src/tools/mongodb/mongodbTool.ts
```diff
@@ -10,6 +10,59 @@ export const DbOperationArgs = {
...
+    analyzer: z
+        .string()
+        .optional()
+        .default("lucene.standard")
+        .describe(
+            "The analyzer to use for the index. Can be one of the built-in lucene analyzers (`lucene.standard`, `lucene.simple`, `lucene.whitespace`, `lucene.keyword`), a language-specific analyzer, such as `lucene.cjk` or `lucene.czech`, or a custom analyzer defined in the Atlas UI."
```
mihirmpatil, line 20: Might be better to link the docs page instead of saying `defined in the Atlas UI`.
https://www.mongodb.com/docs/atlas/atlas-search/analyzers/language/

src/tools/mongodb/mongodbTool.ts
```diff
@@ -10,6 +10,59 @@ export const DbOperationArgs = {
...
+        ),
+    mappings: z
```
mihirmpatil, line 22: The vector search index definition is different compared to this. See https://www.mongodb.com/docs/atlas/atlas-vector-search/vector-search-type/
I think we'll be better off with a new type for vector search index definition.

src/tools/mongodb/create/createSearchIndex.ts
```diff
@@ -0,0 +1,47 @@
...
+const SEARCH_INDEX_TYPE = "search";
+
+export class CreateSearchIndexTool extends MongoDBToolBase {
+    protected name = "create-search-index";
+    protected description = "Create an Atlas Search index for a collection";
+    protected argsShape = {
+        ...DbOperationArgs,
+        name: SearchIndexArgs.name,
+        analyzer: SearchIndexArgs.analyzer,
+        mappings: SearchIndexArgs.mappings,
```
wtrocki, line 14: [open discussion] Alternatively, consider user providing string/exposing search json schema. Complexity of arguments can work differently across the models giving users mixed experience.

Options:
1. We can use existing capacity of OpenAI and Claude understanding Search Index Syntax.
E.g We have single param that is whole search index definition (consider you have set of chats where you first ask to define index and then you want to create it)

2. Employ JSON Schema for strict validation and alignment with MongoDB syntax (like $search indexes). Tool can expose the most recent schema to not use base model knowledge (not reliable and up to date)

wtrocki: Example of what I mean:

Screenshot 2025-05-14 at 12 33 38

src/tools/mongodb/mongodbTool.ts
wtrocki: @mihirmpatil consider that those are used to infer user generic queries. Best to make relevant context clear for machines and users.

src/tools/mongodb/mongodbTool.ts
edgarw19: Yes, this PR is specific to search indexes. I've opened a vector index PR here https://github.com/mongodb-js/mongodb-mcp-server/pull/252

mihirmpatil (approved):

PR #245, opened 2025-05-13
feat: support for listing search indexes

manoj-pillay-10gen:
# Description
Adding tool support for listSearchIndexes utility
Screenshot 2025-05-13 at 3 47 27 PM

# Testing
Shell code added. Waiting for test env setup before submitting actual tests.

### DIFF
--- a/src/tools/mongodb/read/collectionSearchIndexes.ts
+++ b/src/tools/mongodb/read/collectionSearchIndexes.ts
@@ -35,8 +35,8 @@ export class CollectionSearchIndexesTool extends MongoDBToolBase {
             content: [
                 {
                     text: indexName
-                        ? `Found ${indexes.length} search indexes  in the collection "${collection}" with name "${indexName}":`
-                        : `Found ${indexes.length} search indexes  in the collection "${collection}"`,
+                        ? `Found ${indexes.length} search indexes in the collection "${collection}" with name "${indexName}":`
+                        : `Found ${indexes.length} search indexes in the collection "${collection}"`,
                     type: "text",
                 },
                 ...(indexes.map((indexDefinition) => {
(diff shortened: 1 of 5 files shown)


tests/integration/tools/mongodb/read/collectionSearchIndexes.test.ts
```diff
@@ -0,0 +1,20 @@
+import { describeWithMongoDB } from "../mongodbHelpers.js";
```
edgarw19: In the WIP PR https://github.com/mongodb-js/mongodb-mcp-server/tree/f556cc3691a62d8e98f374be4ebaa14c4a555642/tests/integration/tools/atlas-search, there's a separate directory for atlas search tests since the setup is different. 

Is it worth keeping that directory structure and moving this into an atlas-search directory?

src/tools/mongodb/read/collectionSearchIndexes.ts
```diff
@@ -0,0 +1,51 @@
...
+
+    protected operationType: OperationType = "read";
+
+    protected async execute({
+        database,
+        collection,
+        indexName,
+    }: ToolArgs<typeof this.argsShape>): Promise<CallToolResult> {
+        const provider = await this.ensureConnected();
+        const indexes = await provider.getSearchIndexes(database, collection, indexName);
```
edgarw19: nit: Would it make sense to cast the result of getSearchIndexes to a known interface with common fields (like `name`)

Ie I think in some of the tests, we may call `getSearchIndexes` as well and check if certain properties match what's expected

tests/integration/tools/mongodb/read/collectionSearchIndexes.test.ts
manoj-pillay-10gen: Altered directory structure to use proposal in #181

edgarw19 (approved):

PR #249, opened 2025-05-14
chore: add hints

blva:


src/common/atlas/apiClient.ts
```diff
@@ -34,7 +34,9 @@ export class ApiClient {
 
     private getAccessToken = async () => {
         if (this.oauth2Client && (!this.accessToken || this.accessToken.expired())) {
-            this.accessToken = await this.oauth2Client.getToken({});
+            this.accessToken = await this.oauth2Client.getToken({
+                agent: this.options.userAgent,
```
blva, line 38: flyby - user agents

src/tools/tool.ts
```diff
@@ -27,6 +27,22 @@ export abstract class ToolBase {
...
+            title: this.name,
+            description: this.description,
+        };
+
+        if (this.operationType === "read" || this.operationType === "metadata") {
+            annotations.readOnlyHint = true;
+            annotations.destructiveHint = false;
+        } else {
+            annotations.destructiveHint = this.operationType === "delete";
+        }
```
```suggestion
        switch (this.operationType) {
        	case "read":
        	case "metadata":
				annotations.readOnlyHint = true;
				annotations.destructiveHint = false;
				break;
	        case "delete": 
				annotations.destructiveHint = true;
				break;
	}
```
very nit and subjective just seems easier to understand the branching + switch helps ensure we handle all values

tests/integration/helpers.ts
```diff
@@ -206,6 +206,9 @@ export function validateToolMetadata(
         expectDefined(tool);
         expect(tool.description).toBe(description);
 
+        expectDefined(tool.annotations);
+        expect(tool.annotations.title).toBe(name);
```
gagik: can we add a small test for the readOnlyHint and stuff too?

nirinchev (approved):
Sweet!

tests/integration/helpers.ts
blva: yep! added

PR #253, opened 2025-05-14
feat: Add RunPipeline tool

vm-mishchenko:
Add a new `RunPipeline` tool that can execute aggregation pipeline without requiring an Atlas account, cluster, or collection.

The tool accepts a set of documents, an aggregation pipeline, and a search index definition, and runs them against [the Search Playground.](https://search-playground.mongodb.com/) The Search Playground internally creates an ephemeral collection and executes the pipeline in a temporary environment.

Manual testing + integration test. More tests will be added in the following prs.

**What would be the result of query?**
image

**Is the query syntax correct?**
image

**Why $search doesn't return first doc?**
image

### DIFF
--- a/src/tools/playground/runPipeline.ts
+++ b/src/tools/playground/runPipeline.ts
@@ -1,0 +1,166 @@
+import { OperationType, TelemetryToolMetadata, ToolArgs, ToolBase, ToolCategory } from "../tool.js";
+import { z } from "zod";
+import { CallToolResult } from "@modelcontextprotocol/sdk/types.js";
+import { EJSON } from "bson";
+
+const PLAYGROUND_SEARCH_URL = "https://search-playground.mongodb.com/api/tools/code-playground/search";
+
+const DEFAULT_DOCUMENTS = [
+    {
+        name: "First document",
+    },
+    {
+        name: "Second document",
+    },
+];
+
+const DEFAULT_SEARCH_INDEX_DEFINITION = {
+    mappings: {
+        dynamic: true,
+    },
+};
+
+const DEFAULT_PIPELINE = [
+    {
+        $search: {
+            index: "default",
+            text: {
+                query: "first",
+                path: {
+                    wildcard: "*",
+                },
+            },
+        },
+    },
+];
+
+const DEFAULT_SYNONYMS: Array<Record<string, unknown>> = [];
+
+export const RunPipelineOperationArgs = {
+    documents: z
+        .array(z.record(z.string(), z.unknown()))
+        .describe("Documents to run the pipeline against. 500 is maximum.")
+        .default(DEFAULT_DOCUMENTS),
+    aggregationPipeline: z
+        .array(z.record(z.string(), z.unknown()))
+        .describe("Aggregation pipeline to run on the provided documents.")
+        .default(DEFAULT_PIPELINE),
+    searchIndexDefinition: z
+        .record(z.string(), z.unknown())
+        .describe("Search index to create before running the pipeline.")
+        .optional()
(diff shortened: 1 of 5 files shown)


src/tools/playground/runPipeline.ts
```diff
@@ -0,0 +1,166 @@
...
+        },
+    },
+];
+
+const DEFAULT_SYNONYMS: Array<Record<string, unknown>> = [];
+
+export const RunPipelineOperationArgs = {
+    documents: z
+        .array(z.record(z.string(), z.unknown()))
+        .describe("Documents to run the pipeline against. 500 is maximum.")
```
edgarw19, line 43: nit: Worth adding `.max(500)` to codify this?

src/tools/playground/runPipeline.ts
```diff
@@ -0,0 +1,166 @@
...
+
+interface RunErrorResponse {
+    code: string;
+    message: string;
+}
+
+export class RunPipeline extends ToolBase {
+    protected name = "run-pipeline";
+    protected description =
+        "Run aggregation pipeline for provided documents without needing an Atlas account, cluster, or collection.";
```
edgarw19: I wonder if it's helpful to provide some description context to the LLM about when to use this tool?  Like `can be useful in cases such as x, y, z`., since the use cases seem more open ended than the other tools

src/tools/playground/runPipeline.ts
```diff
@@ -0,0 +1,166 @@
...
+export const RunPipelineOperationArgs = {
+    documents: z
+        .array(z.record(z.string(), z.unknown()))
+        .describe("Documents to run the pipeline against. 500 is maximum.")
+        .default(DEFAULT_DOCUMENTS),
+    aggregationPipeline: z
+        .array(z.record(z.string(), z.unknown()))
+        .describe("Aggregation pipeline to run on the provided documents.")
+        .default(DEFAULT_PIPELINE),
+    searchIndexDefinition: z
```
edgarw19, line 49: Are more specific types for `aggregationPipeline`/`searchIndexDefinition`/`synonyms` useful for the LLM or is it already pretty good at determining the types from the description?

For ex, the search playground looks limited to a subset of aggregation pipeline stages. Would those be helpful to include in the type?

src/tools/playground/runPipeline.ts
vm-mishchenko: nice idea, added

src/tools/playground/runPipeline.ts
vm-mishchenko: Added a small clause: `The tool can be useful for running ad-hoc pipelines for testing or debugging.` 

I agree, it's quite open ended tool so I would leave it to llm to decide when exactly it wants to use it.

src/tools/playground/runPipeline.ts
vm-mishchenko: I feel it would be hard to add more specific `zod` types here. All these entities have a complex dynamic structure unfortunately. 

I updated `Aggregation pipeline...` to `MongoDB aggregation pipeline` (same for other fields) to stress MongoDB part that hopefully nudges LLM to the right direction. 

Regarding supported stages, I’d avoid listing them here. If we hardcode them, the list will likely get out of sync over time between the Playground and MCP. I’d rather rely on the Playground’s response to flag any unsupported stages. It actually supports more than what’s in the public docs (product wants to position it as a Search only playground for now).

edgarw19 (approved):
lgtm!

mihirmpatil (approved):

PR #254, opened 2025-05-15
fix: updates count tool

blva:


src/tools/mongodb/read/count.ts
```diff
@@ -4,27 +4,38 @@ import { ToolArgs, OperationType } from "../../tool.js";
...
         .record(z.string(), z.unknown())
         .optional()
         .describe(
-            "The query filter to count documents. Matches the syntax of the filter argument of db.collection.count()"
+            "The query filter to count documents. Matches the syntax of the filter argument of db.collection.countDocuments()"
         ),
+    query: z
+        .record(z.string(), z.unknown())
+        .optional()
+        .describe("Alternative old name for filter. Will be used in db.collection.countDocuments()"),
```
gagik: is there a reason to keep providing this field?

blva: yes, since the LLM was already confusing query/filter, I figure that it can get confused the other way around and try to send it as "query" because it thinks we'll use .count() (even though we call out is countDocments, but who knows, it might have outdated data about MDB)

nirinchev: This doesn't make much sense to me. I think we should tweak the description of `filter` rather than accept two arguments that are doing the same thing.

blva: np, I'll remove and update the description

blva: we can try that for now and see if it improves, based on the issue I'm not sure the LLM will stop getting confused

gagik: I would leave that up to your testing, my (untested) intuition would be to give the LLM less options. I think it was getting confused because filter was the actual MongoDB syntax thing to do for count rather than because it found both equally valid.

blva: ok, I think I found a way to keep the changes here minimal. I updated the tool description, clarifying it uses .count(). I think the LLMs were getting confused, more specifically cursor, with the fact that count() is deprecated and assuming it was countDocuments under-the-hood.

fmenezes (approved):

nirinchev (approved):

gagik (approved):

PR #257, opened 2025-05-16
feat: add HTTP/SSE transport support and document usage in Quick Start

Rujaxx:
This pull request introduces support for running the MongoDB MCP server with HTTP or SSE transport using the new `--transportType` and `--port` command-line arguments. The default port is 5700 if not specified. The Quick Start section in the README has been updated to include Option 5, providing clear instructions and example commands for exposing the server over HTTP or SSE for network-based client integration.

- Adds support for `--transportType` (`http` or `sse`) and `--port` arguments in the server startup.
- Updates the Quick Start documentation in the README to include usage examples for HTTP/SSE transport.
- Clarifies that all other configuration options (like `--connectionString`, `--apiClientId`, etc.) are supported with these transport types.
- Helps users easily expose the MCP server for integration with clients that require a network endpoint.


nirinchev:
Thanks for this contribution. Unfortunately, we can't accept it in its current form, because the server is stateful (we already have a session that tracks connection status) and as such, doesn't lend itself to multiple clients connecting to it - e.g., you could end up executing crud updates against the wrong connection.

We do have plans to rework the design and enable stateless operation, but that's lower priority than some of the features we're working on.

fmenezes:
thank you so much for your contribution @Rujaxx we ended up implementing it internally with more tests, telemetry changes and expansion of config, feel free to take a look at https://github.com/mongodb-js/mongodb-mcp-server/pull/361

PR #282, opened 2025-06-04
Update Dockerfile

yairsabag:


fmenezes:
Thanks for your contribution! Would you mind opening an issue first to describe the problem you’re trying to solve? That helps us assess whether this is aligned with the direction of the project before reviewing the PR.

PR #289, opened 2025-06-09
chore: add deeplinks to vscode, cursor installs

nirinchev:
This is an alternative to https://github.com/mongodb-js/mongodb-mcp-server/pull/288.

The difference is that we don't try to pass on any config parameters as I don't think it's obvious users have to update the config after installation. Additionally, I've moved the badges to the top of the readme and made them consistent visual style.


blva (approved):
LGTM!

nirinchev:
Pulling windsurf out for now since we need an http link there rather than an app link to play nicely with github readmes.

PR #298, opened 2025-06-13
MCP-2: add is_container_env to telemetry

fmenezes:


src/telemetry/telemetry.ts
```diff
@@ -117,6 +163,14 @@ export class Telemetry {
         };
     }
 
+    public async getAsyncCommonProperties(): Promise<CommonProperties> {
+        return {
+            ...this.getCommonProperties(),
+            is_container_env: (await this.containerEnvPromise) ? "true" : "false",
```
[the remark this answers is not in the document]
fmenezes: not compatible TelemetryBoolSet expects strings "true" or "false"

src/telemetry/telemetry.ts
```diff
@@ -7,30 +7,69 @@ import { MACHINE_METADATA } from "./constants.js";
...
 type EventResult = {
     success: boolean;
     error?: Error;
 };
 
-export const DEVICE_ID_TIMEOUT = 3000;
+async function fileExists(filePath: string): Promise<boolean> {
+    try {
+        await fs.stat(filePath);
+        return true; // File exists
```
himanshusinghs: Is this supposed to be a strict file check? If yes then we might wanna do: `await fs.stat(filePath).isFile()`

src/telemetry/telemetry.ts
```diff
@@ -7,30 +7,69 @@ import { MACHINE_METADATA } from "./constants.js";
...
+}
+
+async function isContainerized(): Promise<boolean> {
+    for (const file of ["/.dockerenv", "/run/.containerenv", "/var/run/.containerenv"]) {
+        const exists = await fileExists(file);
+        if (exists) {
+            return true;
+        }
+    }
+    return !!process.env.container;
```
nirinchev: This is fairly minor as this is clearly not on a hot path, but would it make sense to first check for the env variable since that is cheapest? And then for the files, we can check them in parallel rather than 1 by 1.

src/telemetry/telemetry.ts
```diff
@@ -7,30 +7,69 @@ import { MACHINE_METADATA } from "./constants.js";
...
+}
 
 export class Telemetry {
-    private isBufferingEvents: boolean = true;
     /** Resolves when the device ID is retrieved or timeout occurs */
+    private bufferingEvents: number = 2;
```
nirinchev: This name no longer makes much sense. How about `pendingPropertyPromises` or something similar?

src/telemetry/telemetry.ts
```diff
@@ -117,6 +163,14 @@ export class Telemetry {
         };
     }
 
+    public async getAsyncCommonProperties(): Promise<CommonProperties> {
```
nirinchev: I don't have super strong feelings here, but I'm not a huge fan of making methods public just for the sake of tests. While I agree it's convenient to test the internals here, this and `isBufferingEvents` are implementation details, so they shouldn't be made public.

src/telemetry/telemetry.ts
fmenezes: to be honest I'm not sure if these files could be symlinks, podman/docker/etc manage these files

himanshusinghs: Not so straightforward then. Lets keep it as is.

src/telemetry/telemetry.ts
```diff
@@ -7,55 +7,97 @@ import { MACHINE_METADATA } from "./constants.js";
 import { EventCache } from "./eventCache.js";
 import nodeMachineId from "node-machine-id";
 import { getDeviceId } from "@mongodb-js/device-id";
+import fs from "fs/promises";
+
+async function fileExists(filePath: string): Promise<boolean> {
+    try {
+        await fs.stat(filePath);
```
gagik:
```suggestion
        await fs.access(filePath);
```

src/telemetry/telemetry.ts
```diff
@@ -7,55 +7,97 @@ import { MACHINE_METADATA } from "./constants.js";
...
 
-type EventResult = {
-    success: boolean;
-    error?: Error;
-};
+async function isContainerized(): Promise<boolean> {
+    if (process.env.container) {
+        return true;
+    }
+    for (const file of ["/.dockerenv", "/run/.containerenv", "/var/run/.containerenv"]) {
```
gagik: do the `/` paths work on windows?

src/telemetry/telemetry.ts
```diff
@@ -7,55 +7,93 @@ import { MACHINE_METADATA } from "./constants.js";
...
+            getContainerEnv,
+        });
 
-        void instance.start();
+        instance.start();
         return instance;
     }
 
-    private async start(): Promise<void> {
+    private start(): void {
```
gagik: it's usually always cleaner to use async function then wrap them in `finally`-s etc. I'd revert this.

src/telemetry/telemetry.ts
```diff
@@ -7,55 +7,93 @@ import { MACHINE_METADATA } from "./constants.js";
...
         if (!this.isTelemetryEnabled()) {
             return;
         }
+
         this.deviceIdPromise = getDeviceId({
```
gagik: you are probably looking for [Promise.all](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all) (or Promise.allSettled)?

src/telemetry/telemetry.ts
fmenezes: I don't need read access

src/telemetry/telemetry.ts
```diff
@@ -7,55 +7,93 @@ import { MACHINE_METADATA } from "./constants.js";
...
+    return exists.includes(true);
+}
 
 export class Telemetry {
-    private isBufferingEvents: boolean = true;
     /** Resolves when the device ID is retrieved or timeout occurs */
-    public deviceIdPromise: Promise<string> | undefined;
+    private pendingPromises: number = 2;
+    private deviceIdPromise: Promise<string> | undefined;
+    private containerEnvPromise: Promise<boolean> | undefined;
```
gagik: We can probably turn this into a single setup promise with Promise.all or something, see below:

src/telemetry/telemetry.ts
fmenezes: since these files are controlled by docker container I'm not too sure what happens when using non root / non default user over docker (meaning `docker run -u <some user>`)

gagik (changes requested):
I think we can manage the multiple promises better. Likely having 1 setup promise that combines the other 1 promises and just awaiting those instead of having a counter. Though maybe I'm missing some bit of the implementation here.

fmenezes:
> I think we can manage the multiple promises better. Likely having 1 setup promise that combines the other 1 promises and just awaiting those instead of having a counter. Though maybe I'm missing some bit of the implementation here.

I did not know either it was prior to my time, but just checked the code it should be ok just to wait on both promises, just be instant

src/telemetry/telemetry.ts
gagik: ~without read access~ without permission check access, we'd not be able to check for ~permission~ stat either.`stat` also gives you stats which we're discarding, `access` is lighter in that sense.

Node Docs example uses access to check for file existance:
https://nodejs.org/api/fs.html#fs_fs_access_path_mode_callback

src/telemetry/telemetry.ts
fmenezes: idea here was to start the promises and not wait for them to finish

gagik: what would be the benefit of that? we're already spawning an async function without waiting for it to finish (with `void` above) so this isn't blocking the main thread.

We'd want to buffer events beforehand so we can make sure telemetry stuff we send is with device_id resolved if possible

gagik:
> should be ok just to wait on both promises

yeah we can keep the async start + buffering like before and just add the second await into it with Promise.all before setting isBuffering to false

src/telemetry/telemetry.ts
```diff
@@ -106,14 +138,16 @@ export class Telemetry {
...
         return {
-            ...this.commonProperties,
+            ...MACHINE_METADATA,
             mcp_client_version: this.session.agentRunner?.version,
             mcp_client_name: this.session.agentRunner?.name,
             session_id: this.session.sessionId,
             config_atlas_auth: this.session.apiClient.hasCredentials() ? "true" : "false",
             config_connection_string: this.userConfig.connectionString ? "true" : "false",
+            is_container_env: (await this.containerEnvPromise) ? "true" : "false",
+            device_id: await this.deviceIdPromise,
```
gagik: I don't think we should be awaiting here; an explicit unawaited + buffering with an async start like it was before is easier to reason about, it just needs another to await for container inside `start`

Otherwise there's potentially hundreds of floating functions awaiting device_id or whatever else. We can assume in worst case scenario this isn't instant (and could never even resolve)

src/telemetry/telemetry.ts
```diff
@@ -73,15 +109,11 @@ export class Telemetry {
             },
             abortSignal: this.deviceIdAbortController.signal,
         });
-
-        this.commonProperties.device_id = await this.deviceIdPromise;
-
-        this.isBufferingEvents = false;
+        this.containerEnvPromise = this.getContainerEnv();
```
gagik:
```suggestion
        const [deviceId, isContainerEnv] = await Promise.all([this.deviceIdPromise, this.getContainerEnvPromise]);
        this.commonProperties.device_id = deviceId;
        this.commonProperties.is_container_env = isContainerEnv;
```

something like this

src/telemetry/telemetry.ts
fmenezes: on the device ID there is a timeout, on the container env it is just `fs/promises` so we can be sure they will resolve. These promises are private, I think the buferring logic only makes sense if we are tagging on finally or something, to me feels unneeded complication. Our current timeout is 3 seconds, no reason to be extra defensive.

src/telemetry/telemetry.ts
fmenezes: with this approach start would take up to 3s which I think it's acceptable, will change

gagik: it will have identical behavior to awaiting inside the getCommonProperties, we do not await the `start()` function anywhere so it's essentially same as declaring a combined promise, we instantly initialize the telemetry and start buffering event when emit is run.

In both cases we'd instantly be able to emit events, and in both cases if device ID takes 3 seconds, the eventual emission of events will take 3 seconds. 

But with `start` there's only 1 point where we deal with asynchronous logic, and in the other case we have a bunch of less predictable asyncronous functions listening in into 2 different promises.

src/telemetry/telemetry.ts
```diff
@@ -7,56 +7,92 @@ import { MACHINE_METADATA } from "./constants.js";
...
-        const instance = new Telemetry(session, userConfig, commonProperties, { eventCache, getRawMachineId });
+    ): Promise<Telemetry> {
+        const instance = new Telemetry(session, userConfig, {
+            eventCache,
+            getRawMachineId,
+            getContainerEnv,
+        });
 
-        void instance.start();
+        await instance.start();
```
gagik:
```suggestion
        void instance.start();
```
This was an important detail in the existing implementation: **start was not getting awaited before**. This is why the behavior is equivalent to what was trying to be accomplished before with skipping the promise. Telemetry was already not waiting for the device ID resolution.

So the create function does not need to be asynchronous and can be left as is.

src/telemetry/telemetry.ts
gagik: it's fair that we don't need to be extra defensive but either way we're still essentially going to end up buffering these events.

The difference is that with awaiting here the "buffering" would happen through a bunch of async functions in parallel in memory stack waiting on these promises as opposed to more explicit buffering where we store the functions we want to execute ourselves and execute them in order after resolving these promises in one place. 

With awaiting in common properties approach we'd, for example, not have any guarantees that these functions end up resolving in the same order. Which we don't need to care about, but it's nice to have better idea of how async execution is going to happen and have 1 point of logic where we resolve all the async issues.

With discussions with @nirinchev, we also want to create a shared telemetry service component across devtools so we would not want to deviate too much unless there are good reasons for it (this setup is modeled after mongosh and Compass telemetry).

src/telemetry/telemetry.ts
```diff
@@ -7,56 +7,92 @@ import { MACHINE_METADATA } from "./constants.js";
...
-export const DEVICE_ID_TIMEOUT = 3000;
+    return exists.includes(true);
+}
 
 export class Telemetry {
-    private isBufferingEvents: boolean = true;
-    /** Resolves when the device ID is retrieved or timeout occurs */
-    public deviceIdPromise: Promise<string> | undefined;
+    private deviceId: string | undefined;
+    private containerEnv: boolean | undefined;
```
gagik: why not have them inside `commonProperties` so it easier to see which fields in the state are used solely for common properties?

src/index.ts
```diff
@@ -20,7 +20,7 @@ try {
         version: packageInfo.version,
     });
 
-    const telemetry = Telemetry.create(session, config);
+    const telemetry = await Telemetry.create(session, config);
```
gagik:
```suggestion
    const telemetry = Telemetry.create(session, config);
```
the creation logic does not be turned async.

I'd say the main change needed is to add `this.commonProperties.is_container_env` intialization into `start` (possibly rename it to `setup` as that naming seems confusing)

src/telemetry/telemetry.ts
fmenezes: 3s is nothing to wait, I think the code looks much better to wait on promises instead of "ignoring" them, I can move to last

gagik: adding potentially 3 seconds to the MCP server startup time for no real functional benefit is absolutely non-trivial. We shouldn't be slowing down anything in our logic for the sake of telemetry. This is also not ignoring a promise, this is explicitly spawning a parallel async operation here and the `void` keyword exists for that reason.

If you want, you can separate the refactor bits of this PR into its own thing and we can have a larger discussion about those changes but in places like mongosh we absolutely do care about startup time so when consolidating telemetry we'd likely end up with the original structure eventually.

tests/unit/telemetry.test.ts
```diff
@@ -250,23 +266,17 @@ describe("Telemetry", () => {
...
-                    // Make sure the timeout doesn't happen prematurely.
-                    expect(telemetry["isBufferingEvents"]).toBe(true);
-                    expect(telemetry.getCommonProperties().device_id).toBe(undefined);
+                    const testEvent = createTestEvent();
 
-                    jest.advanceTimersByTime(DEVICE_ID_TIMEOUT);
+                    await telemetry.emitEvents([testEvent]);
 
-                    await telemetry.deviceIdPromise;
+                    await delay(5000); // Wait for timeout
```
gagik: this test will now take 5 seconds longer. if we use `jest.advanceTimersByTime` we don't have to actually wait 5 seconds.

src/telemetry/telemetry.ts
fmenezes: commonProperties is a function not a member

src/telemetry/telemetry.ts
fmenezes: I see no reason to think 3 seconds is negligible, but I'm changing the logic to send wait for 3 seconds asynchronously not blocking main flow

src/telemetry/telemetry.ts
gagik: good point, maybe then we can scope it into `machineMetadata` or `machineInfo` here? so we can mix the things we have from the global constant and things we resolve later on in initialization. 

or it can be just these 2 and then we can expand that and the constant, that works too.
We might have more fields like this in the future and we wouldn't want to keep adding fields in root level.

src/telemetry/telemetry.ts
gagik: > I'm changing the logic to send wait for 3 seconds asynchronously not blocking main flow

not sure what you mean

src/telemetry/telemetry.ts
```diff
@@ -7,114 +7,151 @@ import { MACHINE_METADATA } from "./constants.js";
...
-        return {
-            ...this.commonProperties,
-            mcp_client_version: this.session.agentRunner?.version,
-            mcp_client_name: this.session.agentRunner?.name,
-            session_id: this.session.sessionId,
-            config_atlas_auth: this.session.apiClient.hasCredentials() ? "true" : "false",
-            config_connection_string: this.userConfig.connectionString ? "true" : "false",
-        };
+    private async getCommonProperties(): Promise<CommonProperties> {
+        if (!this.cachedCommonProperties) {
```
gagik, line 107: This is still basically identical to `isBufferingEvents` as we'd be indirectly buffering by spawning async operations and awaiting them.

I can see some value in the idea that we'd only start trying to resolve the device ID etc. stuff when there's an emit emission but this will be problematic and harder to reason about:

if there's 2 `emitEvents` one after another, we'd have 2 processes at the same time trying to set to `cachedCommonProperties`.

I still think the responsibility of a 1-time resolution of machine should be inside the setup function and not `emit` or `getCommonProperties`

tests/unit/telemetry.test.ts
```diff
@@ -16,6 +16,9 @@ const MockApiClient = ApiClient as jest.MockedClass<typeof ApiClient>;
 jest.mock("../../src/telemetry/eventCache.js");
 const MockEventCache = EventCache as jest.MockedClass<typeof EventCache>;
 
+const nextTick = () => new Promise((resolve) => process.nextTick(resolve));
+const delay = (ms: number) => new Promise((resolve) => setTimeout(resolve, ms));
```
gagik:
```suggestion
```

tests/unit/telemetry.test.ts
```diff
@@ -16,6 +16,9 @@ const MockApiClient = ApiClient as jest.MockedClass<typeof ApiClient>;
 jest.mock("../../src/telemetry/eventCache.js");
 const MockEventCache = EventCache as jest.MockedClass<typeof EventCache>;
 
+const nextTick = () => new Promise((resolve) => process.nextTick(resolve));
```
gagik, line 19: I have never seen `nextTick` used in tests before, is there a difference between this and not doing anything?

src/telemetry/telemetry.ts
```diff
@@ -7,114 +7,151 @@ import { MACHINE_METADATA } from "./constants.js";
...
+            e instanceof Error &&
+            (
+                e as Error & {
+                    code: string;
+                }
+            ).code === "ENOENT"
+        ) {
+            return false; // File does not exist
+        }
+        throw e; // Re-throw unexpected errors
```
gagik, line 27: we might want to log this error for our own analytics or something but I think we shouldn't be throwing an error to the user if telemetry fails

src/telemetry/telemetry.ts
```diff
@@ -135,60 +172,62 @@ export class Telemetry {
...
-    private async emit(events: BaseEvent[]): Promise<void> {
-        if (this.isBufferingEvents) {
-            this.eventCache.appendEvents(events);
+    public async flush(events?: BaseEvent[]): Promise<void> {
+        if (!this.isTelemetryEnabled()) {
+            logger.info(LogId.telemetryEmitFailure, "telemetry", `Telemetry is disabled.`);
+            return;
+        }
+
+        if (this.flushing) {
```
gagik, line 185: isn't this identical to buffering? could we structure it as buffering?

src/telemetry/telemetry.ts
fmenezes: `getCommonProperties` happens only once and resolves during the first flush, I've seen this approach many times on backends before, we can rename `getCommonProperties` into `setup` does it make sense to you?

tests/unit/telemetry.test.ts
fmenezes: Yes, the reason why this is needed is so once we emit an event the flushing process makes a fetch to api and that does not resolve immediately.

src/telemetry/telemetry.ts
fmenezes: We are logging it over `getCommonProperies`.

src/telemetry/telemetry.ts
fmenezes: I disagree with buffering as a terminology here, if we are emptying the cache after sending events over network it means we are no longer accumulating but rather releasing resources.

The only term I've seen used to resonate with that purpose is flush, similar to flush logs after a buffer is too big for instance.

src/telemetry/telemetry.ts
gagik: > `getCommonProperties` happens only once

but `getCommonProperties` gets called by `emitEvents` every time right? 

1. First `emitEvents` call would begin the async process but it wouldn't immediately set `this.cachedCommonProperties`. 
2. For the next `emitEvents` call, `!this.cachedCommonProperties` is still true and it'd run the async functions afterwards again. 

We could check for `this.cachedCommonProperties == undefined;` and set `this.cachedCommonProperties = {};` before doing async work to prevent this.

then we'd prevent the async clash

> we can rename `getCommonProperties` to `setup`

yeah, that actually does make it much better in my view. then I do see some of the general value of starting this only when an emission happens, if that is the intended improvement.

src/telemetry/telemetry.ts
gagik: > we are no longer accumulating but rather releasing resources

Well, we are releasing already buffered resources but we're buffering all future resources. The state of the boolean to me in usage is more relevant generally to the future resources rather than the past resources, i.e. we use it largely to determine whether or not to `this.eventCache.appendEvents(events ?? []);`. But I dont have as strong feelings about this, can also be `isFlushing`

We do actually have a concept of flush along with buffering in https://github.com/mongodb-js/mongosh/blob/main/packages/logging/src/logging-and-telemetry.ts#L129

tests/unit/telemetry.test.ts
gagik: got it, could we add a test where there's 2 or more `telemetry.emitEvents` calls one after the other and check that all works as expected?

src/telemetry/telemetry.ts
```diff
@@ -7,114 +7,151 @@ import { MACHINE_METADATA } from "./constants.js";
...
-            mcp_client_name: this.session.agentRunner?.name,
-            session_id: this.session.sessionId,
-            config_atlas_auth: this.session.apiClient.hasCredentials() ? "true" : "false",
-            config_connection_string: this.userConfig.connectionString ? "true" : "false",
-        };
+    private async getCommonProperties(): Promise<CommonProperties> {
+        if (!this.cachedCommonProperties) {
+            let deviceId: string | undefined;
+            try {
+                deviceId = await getDeviceId({
```
gagik: nit, feel free to ignore: it'd still be good to do a Promise.all so we can initialize things in parallel. I know this is messier structure but would let us be much more

gagik (approved):
I don't want to drag this on longer as I am OOO for next week, thank you for all the iterations with this, so will remove the request changes as not to block this getting out.

I think we're quite aligned now, the only distinction is the preference of whether we should setup telemetry right after intialization or after the first `emitEvents` call. I still believe in the former but this I can mark as subjective and not something I have as strong feelings about. 

I only have a concern about how this would work with multiple parallel `emitEvents` calls so maybe some tests there would be good.

PR #325, opened 2025-06-26
chore: provides dedicated cli and library entrypoints - MCP-6

himanshusinghs:
## Proposed changes
Important changes in this PR are summarised here:
1. CLI entry point has been moved from `src/index.ts` to `src/cli.ts`. The same has been reflected in package json's entry as well under `bin`
2. A new entry point (`src/lib.ts`) has been added for exporting entities to be used by other projects (vscode). For now it exports only the entities necessary for a server creation as per the following [commit](https://github.com/mongodb-js/vscode/commit/7bba1397e481d80292236d3c8d8a500e55a497b1#diff-c2dbb92c81a472b4bc81088731816dd9f4676cdb1b4a8ade3fdb49c6a3b20d82).
3. MCP server is now bundled as CommonJS module, using webpack, for both cli and lib entrypoints.
    - Decision to bundle instead of emitting a CommonJS code was taken after several trial and error approaches of trying `tsc` to emit files such that they could be imported in vscode (which is using classic module resolution) without having import errors and errors in dependency resolution during runtime. So far only bundling approach worked out.
    - Types are emitted using `tsc` and also referenced in package.json for type resolution in other projects.
4. We were shipping our entire repo code as the package which was, I suppose, accidental. Added the `files` entry in package.json to fix this.

All the existing tests pass. What this PR lacks and would be great to have is some type of CLI tests for running the MCP server, possibly also doing some basic e2e. This would give more confidence in doing these bundling changes.

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
   "author": "MongoDB <info@mongodb.com>",
   "homepage": "https://github.com/mongodb-js/mongodb-mcp-server",
   "repository": {
-    "url": "https://github.com/mongodb-js/mongodb-mcp-server.git"
+    "url": "git+https://github.com/mongodb-js/mongodb-mcp-server.git"
   },
   "bin": {
     "mongodb-mcp-server": "dist/cli.cjs"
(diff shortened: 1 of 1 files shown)


himanshusinghs:
Closing this in favor of #371

PR #340, opened 2025-07-04
chore: avoid log files on disk and move them to console for containers

fmenezes:
## Checklist


src/common/container.ts
```diff
@@ -0,0 +1,35 @@
...
+            } catch {
+                return false;
+            }
+        })
+    );
+
+    return exists.includes(true);
+}
+
+export async function detectContainerEnv(): Promise<boolean> {
```
gagik: very nit: is there a particular reason to have the internal function? seems like they could be combined

PR #341, opened 2025-07-07
chore(tests): accuracy tests for MongoDB tools exposed by MCP server MCP-39

himanshusinghs:
## **Motivation and Goal**

* **Motivation:** Consolidate MCP server tools to accommodate client soft limits, while mitigating the risk of confusing LLMs with potentially ambiguous tool schemas and descriptions.  
* **Goal:** Benchmark current tool understanding by LLMs to establish a baseline and prevent regression during consolidation.

## **Design Brief**

* **Method:** Provide LLMs (Gemini, Claude, ChatGPT, etc.) with prompts and current MCP server tool schemas.  
* **Evaluation:** Record actual tool calls made by LLMs and compare them against expected calls and parameters.
* **Reporting:** Generate a readable summary of test runs highlighting prompt, models and the achieved accuracy

## **Detailed Design**

Refer to the doc titled - `MCP Tools Accuracy Testing`

## **Current State**

* Framework implemented and integrated to test the MCP server and MongoDB tools.
* Accuracy tests for core MongoDB tool calls are written.  
* The scoring algorithm is implemented and unit-tested.  
* Supports multiple LLM providers and models.  
* Snapshots are stored on disk by default, with possibility to store in a MongoDB deployment as well.
* On each successful test run a summary is generated highlighting prompt, model and accuracy of tool calls.
* Github workflow added to trigger the test runs on a label and manual dispatch. It also attach the summary when triggered for a PR.

## For reviewers

* Please start reviewing the test cases / prompts themselves.
* Once done with the prompts review, move on to the [tool calling accuracy scorer](https://github.com/mongodb-js/mongodb-mcp-server/blob/chore/issue-307-proposal-2/tests/accuracy/sdk/accuracy-scorer.ts). I have added some [docs](https://github.com/mongodb-js/mongodb-mcp-server/blob/79cd26e41719fb06558d95ce938113cc10c08632/tests/accuracy/sdk/accuracy-scorer.ts#L4-L57) and [tests](https://github.com/mongodb-js/mongodb-mcp-server/blob/chore/issue-307-proposal-2/tests/unit/accuracy-scorer.test.ts) to help understand how it works.
* Later you can start reviewing the rest of the accuracy SDK in the folder `tests/accuracy/sdk`. Start with `describe-accuracy-test.ts` as this is where all the different parts come together and dive further into specific implementation of each parts afterwards.

Apologies for the big chunk to be reviewed here but I did not see a way around it.


.github/workflows/accuracy-tests.yml
```diff
@@ -0,0 +1,48 @@
+name: Accuracy Tests
+
+on:
+  workflow_dispatch:
```
nirinchev, line 4: Guessing we want to eventually also run them on merges to main, right?

.github/workflows/accuracy-tests.yml
```diff
@@ -0,0 +1,48 @@
...
+        run: ./scripts/run-accuracy-tests.sh
+      - name: Upload accuracy test summary
+        if: always()
+        uses: actions/upload-artifact@v4
+        with:
+          name: accuracy-test-summary
+          path: .accuracy/tests-summary.html
+      - name: Comment summary on PR
+        if: github.event_name == 'pull_request' && github.event.label.name == 'accuracy-tests'
+        uses: marocchino/sticky-pull-request-comment@v2
```
nirinchev: This needs to be a commit sha

package.json
```diff
@@ -29,18 +29,25 @@
...
+    "pre:test:accuracy": "npm run build:compile",
+    "test:accuracy": "sh ./scripts/run-accuracy-tests.sh"
   },
   "license": "Apache-2.0",
   "devDependencies": {
+    "@ai-sdk/anthropic": "^1.2.12",
+    "@ai-sdk/azure": "^1.3.23",
+    "@ai-sdk/openai": "^1.3.22",
     "@eslint/js": "^9.30.1",
+    "@himanshusinghs/google": "^1.2.11",
```
nirinchev: Hm... do we want to move this under `@mongodb-js`?

scripts/generate-test-summary.ts
```diff
@@ -0,0 +1,288 @@
...
+}
+
+function formatTokenUsage(tokensUsage: {
+    promptTokens?: number;
+    completionTokens?: number;
+    totalTokens?: number;
+}): string {
+    const total = tokensUsage.totalTokens || 0;
+    const prompt = tokensUsage.promptTokens || 0;
+    const completion = tokensUsage.completionTokens || 0;
```
nirinchev: [nit] Should we use N/A instead of 0 here? I guess we can assume that there's no world in which we actually use 0, just probably a bit more expressive.

scripts/generate-test-summary.ts
```diff
@@ -0,0 +1,288 @@
...
+function formatBaselineAccuracy(snapshot: SnapshotEntryWithBaseline): string {
+    if (!snapshot.baseline || snapshot.baseline.baselineAccuracy === undefined) {
+        return '<span class="accuracy-comparison">N/A</span>';
+    }
+
+    const baselineAccuracyText = formatAccuracy(snapshot.baseline.baselineAccuracy);
+    let comparisonClass = "accuracy-comparison";
+    let comparisonIcon = "";
+
+    if (snapshot.baseline.comparisonResult) {
```
nirinchev: [nit] should we provide a default in case we don't have a comparison? E.g. `?` for the icon?

scripts/generate-test-summary.ts
```diff
@@ -0,0 +1,288 @@
...
+import { HTML_TESTS_SUMMARY_FILE, HTML_TESTS_SUMMARY_TEMPLATE } from "../tests/accuracy/sdk/constants.js";
+import type {
+    AccuracySnapshotEntry,
+    ExpectedToolCall,
+    LLMToolCall,
+} from "../tests/accuracy/sdk/accuracy-snapshot-storage/snapshot-storage.js";
+
+interface BaselineComparison {
+    baselineAccuracy?: number;
+    comparisonResult?: "improved" | "regressed" | "same";
```
nirinchev: Why are these optional?

scripts/generate-test-summary.ts
nirinchev: Actually, this may not need to be optional, in which case we could remove the if-check.

.github/workflows/accuracy-tests.yml
```diff
@@ -0,0 +1,48 @@
...
+      pull-requests: write
+    if: |
+      github.event_name == 'workflow_dispatch' ||
+      (github.event_name == 'pull_request' && github.event.label.name == 'accuracy-tests')
+    env:
+      MDB_OPEN_AI_API_KEY: ${{ secrets.MDB_OPEN_AI_API_KEY }}
+      MDB_GEMINI_API_KEY: ${{ secrets.MDB_GEMINI_API_KEY }}
+      MDB_AZURE_OPEN_AI_API_KEY: ${{ secrets.MDB_AZURE_OPEN_AI_API_KEY }}
+      MDB_AZURE_OPEN_AI_API_URL: ${{ secrets.MDB_AZURE_OPEN_AI_API_URL }}
+      MDB_ACCURACY_MDB_URL: ${{ secrets.MDB_ACCURACY_MDB_URL }}
```
nirinchev:
```suggestion
      MDB_ACCURACY_MDB_URL: ${{ secrets.ACCURACY_MDB_CONNECTION_STRING }}
```

Also, we may want to propagate the connection string wording.

.github/workflows/accuracy-tests.yml
```diff
@@ -0,0 +1,48 @@
...
+      MDB_ACCURACY_MDB_DB: ${{ secrets.MDB_ACCURACY_MDB_DB }}
```
nirinchev:
```suggestion
      MDB_ACCURACY_MDB_DB: ${{ vars.ACCURACY_MDB_DB }}
```

.github/workflows/accuracy-tests.yml
```diff
@@ -0,0 +1,48 @@
...
+      MDB_ACCURACY_MDB_COLLECTION: ${{ secrets.MDB_ACCURACY_MDB_COLLECTION }}
```
nirinchev:
```suggestion
      MDB_ACCURACY_MDB_COLLECTION: ${{ vars.ACCURACY_MDB_COLLECTION }}
```

.github/workflows/accuracy-tests.yml
```diff
@@ -0,0 +1,48 @@
...
+    name: Run Accuracy Tests
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      pull-requests: write
+    if: |
+      github.event_name == 'workflow_dispatch' ||
+      (github.event_name == 'pull_request' && github.event.label.name == 'accuracy-tests')
+    env:
+      MDB_OPEN_AI_API_KEY: ${{ secrets.MDB_OPEN_AI_API_KEY }}
```
nirinchev:
```suggestion
      MDB_OPEN_AI_API_KEY: ${{ secrets.ACCURACY_OPEN_AI_API_KEY }}
```

scripts/generate-test-summary.ts
himanshusinghs: Answering for this and the comment above:
There are two use-cases that this tackles:
1. You may want to generate test summary and don't really bother about the comparison from historic runs. Common if you're running the tests locally.
2. The branch that you're targeting does not have any snapshots just yet. (this is only gonna happen once for the first merge but still a case)

package.json
himanshusinghs: I totally forgot about this. There's a good chance that they released a new version with my fixes. I will check and update.

.github/workflows/accuracy-tests.yml
himanshusinghs: Yea that's right - I will update this as well.

scripts/generate-test-summary.ts
nirinchev: Right, but in that case, wouldn't the baseline field itself be undefined (line 16). My expectation is that if `baseline` is set to something, the contents of the object should be required.

himanshusinghs: Ahh yea true. I must have forgotten to remove the optionals when I nested them under baseline. Will update this.

himanshusinghs: I simplified the script a little more while addressing this.

scripts/generate-test-summary.ts
himanshusinghs: Included in the latest update to the script.

package.json
himanshusinghs: Replaced with version of `@ai-sdk/google` that has the fix.

.github/workflows/accuracy-tests.yml
himanshusinghs: Done 👍

.github/workflows/accuracy-tests.yml
himanshusinghs: Also included this 👍

scripts/accuracy/generate-test-summary.ts
```diff
@@ -0,0 +1,275 @@
...
+    return template.replace(/\{\{(\w+)\}\}/g, (_, key: string) => data[key] ?? "");
+}
+
+function formatRunStatus(status: AccuracyRunStatuses) {
+    let statusClass = "chip run-status";
+    if (status === "done") {
+        statusClass += " perfect";
+    } else if (status === "in-progress") {
+        statusClass += " poor";
+    } else if (status === "failed") {
```
nirinchev: [nit] Not a CSS guru (perhaps should let @gagik review this 😅), but is it intentional that both `in-progress` and `failed` use the same class?

tests/accuracy/sdk/accuracy-result-storage/result-storage.ts
```diff
@@ -0,0 +1,116 @@
...
+    llmResponseTime: number;
+    /**
+     * A number between 0 and 1, representing how accurately the expected tools
+     * were called by LLM when responding to the provided prompts. To know more
+     * about how this number is generated, check - toolCallingAccuracy.ts */
+    toolCallingAccuracy: number;
+    /**
+     * A list of tools, along with their parameters, that are expected to be
+     * called by the LLM in test. */
+    expectedToolCalls: ExpectedToolCall[];
```
nirinchev: Is this a property of the `ModelResponse`? I'd expect this to be the same for all models, so we could move it to `PromptResult`. That being said, I'm not sure if we even need it for historical storage - it seems to be mostly useful for debugging and considering we have the commit sha, we should be able to checkout the repo and check what the expectations were.

tests/accuracy/sdk/accuracy-result-storage/get-accuracy-result-storage.ts
```diff
@@ -0,0 +1,11 @@
+import { DiskBasedResultStorage } from "./disk-storage.js";
+import { MongoDBBasedResultStorage } from "./mongodb-storage.js";
+import { AccuracyResultStorage } from "./result-storage.js";
+
+export function getAccuracyResultStorage(): AccuracyResultStorage {
```
nirinchev: Can this be moved to `result-storage.ts`?

tests/accuracy/sdk/accuracy-result-storage/mongodb-storage.ts
```diff
@@ -0,0 +1,105 @@
...
+export class MongoDBBasedResultStorage implements AccuracyResultStorage {
+    private client: MongoClient;
+    private resultCollection: Collection<AccuracyResult>;
+
+    constructor(
+        connectionString: string,
+        database: string,
+        collection: string,
+        // Omitting these as they might contain large chunk of texts
+        private readonly omittedModelResponseFields: (keyof ModelResponse)[] = ["messages", "text"]
```
nirinchev: Any reason this is a parameter instead of just a const?

tests/accuracy/sdk/accuracy-result-storage/mongodb-storage.ts
```diff
@@ -0,0 +1,105 @@
...
+        runId: string,
+        prompt: string,
+        modelResponse: ModelResponse
+    ): Promise<void> {
+        const savedModelResponse: ModelResponse = { ...modelResponse };
+        for (const field of this.omittedModelResponseFields) {
+            delete savedModelResponse[field];
+        }
+
+        await this.resultCollection.updateOne(
```
nirinchev: This whole thing can be an aggregation. It might be a bit more complex, but I'd prefer it to be atomic + it's a good exercise for writing more complex aggregations.

tests/accuracy/sdk/accuracy-result-storage/disk-storage.ts
```diff
@@ -0,0 +1,169 @@
...
+
+            return accuracyResult;
+        });
+    }
+
+    close(): Promise<void> {
+        return Promise.resolve();
+    }
+
+    private async atomicWriteResult(
```
nirinchev: This seems to be quite complex - I'm guessing, what you're guarding against here is the file being updated simultaneously from two processes. As far as I can tell, this approach still exposes us to race conditions where the file won't be corrupted, but its contents might not be correct. E.g. if I have two processes that have generated the contents of the file independently. If their writes don't conflict, you'd end up in a situation where the slower process still manages to update the file contents, even though it has never seen the update from the faster process.

My strong suggestion here would be to use a lockfile instead and just prevent any sort of file updates until the lock is acquired. A quick search seems to point to https://www.npmjs.com/package/proper-lockfile as a fairly popular package with a straightforward API.

scripts/accuracy/generate-test-summary.ts
himanshusinghs: Yea it was intentional as I wanted to highlighted in the chip color that the run could not be marked as "done" regardless of the status. And because I was getting stingy for introducing new color tokens I decided to use the existing ones.

But I have nothing against introducing more formal color tokens to denote the accuracy run status. Let me know if there is a preference, happy to accommodate the change.

tests/accuracy/sdk/accuracy-result-storage/result-storage.ts
himanshusinghs: Yea the `expectedToolCalls` will better fit in PromptResult. About the case of whether we need it or not, I think we should keep it for completeness of the historical data. Without `expectedToolCalls`, just by looking at the data, we cannot answer - "why the tool calling accuracy was x?".

While I understand that the commit SHA is there and we can check it out but by that point we can also just run the accuracy tests again :)

tests/accuracy/sdk/accuracy-result-storage/get-accuracy-result-storage.ts
himanshusinghs: That makes sense. I will do it right away.

tests/accuracy/sdk/accuracy-result-storage/mongodb-storage.ts
himanshusinghs: I was hoping to keep the SDK as configurable as possible, just in case someone else decides to use it. For that reason I kept it as a parameter.

tests/accuracy/sdk/accuracy-result-storage/mongodb-storage.ts
himanshusinghs: Sounds good - time to learn some new aggregation concepts 😄

tests/accuracy/sdk/accuracy-result-storage/disk-storage.ts
himanshusinghs: Yea it was basically to guard against simultaneous modifications by parallel test runs. The concerns are valid, I will check out the lockfile package.

tests/accuracy/sdk/accuracy-result-storage/mongodb-storage.ts
gagik: this is subjective but might be nice for these configurations to be an object options parameter in that case, then any extra configurations won't get out of hand in the future and depend on parameter ordering. 

On the other hand, I think we should only create this configurability when we actually need it.

scripts/accuracy/generate-test-summary.ts
```diff
@@ -0,0 +1,275 @@
...
+}
+
+function populateTemplate(template: string, data: Record<string, string>): string {
+    return template.replace(/\{\{(\w+)\}\}/g, (_, key: string) => data[key] ?? "");
+}
+
+function formatRunStatus(status: AccuracyRunStatuses) {
+    let statusClass = "chip run-status";
+    if (status === "done") {
+        statusClass += " perfect";
```
gagik: very nit: maybe we can use an array and `.join(' ')`

scripts/accuracy/generate-test-summary.ts
gagik: `if (status === "failed" || status === "in-progress")` (or a switch statement) might make it look more intentional. 

As far as CSS goes, there's 100 ways to do this and don't think any are better than others 😄 1 class = 1 reusable state of style seems reasonable

tests/accuracy/sdk/accuracy-result-storage/mongodb-storage.ts
nirinchev: Could try the NLQ functionality in Compass and see how that performs 😆

tests/accuracy/sdk/accuracy-result-storage/get-accuracy-result-storage.ts
himanshusinghs: Ahh doing this will make a circular dependency, disk-storage depending on result-storage which depends on disk-storage. I checked and although the resolution works fine, I am doubtful of its reliability. I think its best to keep it this untangled.

tests/accuracy/sdk/accuracy-result-storage/result-storage.ts
himanshusinghs: Moved the ExpectedToolCalls[] to PromptResult.

tests/accuracy/sdk/accuracy-result-storage/mongodb-storage.ts
himanshusinghs: That's a fair point. Currently the SDK lies within the mcp server repo itself and is not really usable as a package. I guess we can defer this configurability to a point when we pull it out as its own package.

tests/accuracy/sdk/accuracy-result-storage/disk-storage.ts
himanshusinghs: I have updated this to use `proper-lockfile` itself. I looked for more alternatives but that one seemed like the mostly used package for this usecase.

scripts/accuracy/generate-test-summary.ts
himanshusinghs: Makes sense, I have updated this one and related as suggested.

tests/accuracy/sdk/accuracy-result-storage/mongodb-storage.ts
himanshusinghs: I have moved the parameters as const for now.

tests/accuracy/sdk/accuracy-result-storage/mongodb-storage.ts
himanshusinghs: Ok finally I moved this to update pipeline as well. Learnt a bunch of new operators today. I will be adding some test cases for this as well in the follow up commit but feel free to review it already.

tests/accuracy/sdk/accuracy-result-storage/disk-storage.ts
```diff
@@ -0,0 +1,213 @@
...
+            if ((error as NodeJS.ErrnoException).code === "EEXIST") {
+                return {
+                    fileCreatedWithInitialData: false,
+                };
+            }
+            throw error;
+        }
+    }
+
+    private async atomicUpdateLink(filePath: string, linkPath: string) {
```
nirinchev: Not sure I fully understand why we need this vs using the lockfile.

tests/accuracy/sdk/accuracy-result-storage/disk-storage.ts
```diff
@@ -0,0 +1,213 @@
...
+    AccuracyResult,
+    AccuracyResultStorage,
+    AccuracyRunStatus,
+    AccuracyRunStatuses,
+    ExpectedToolCall,
+    ModelResponse,
+} from "./result-storage.js";
+
+export class DiskBasedResultStorage implements AccuracyResultStorage {
+    async getAccuracyResult(commitSHA: string, runId?: string): Promise<AccuracyResult | null> {
```
nirinchev: This needs to be called with a lock. I can see that the internal usages do that already, but since it's intended to be public, we need to lock the lockfile while reading, otherwise we risk reading corrupted data.

tests/accuracy/sdk/accuracy-result-storage/mongodb-storage.ts
nirinchev: Might have missed to push the commit here - I still see the 3 sequential updates.

tests/accuracy/sdk/accuracy-result-storage/disk-storage.ts
himanshusinghs: If you're referring to the lock held in `updateRunStatus` then that one is for the actual result file. The file, actually link, getting updated here is different than that, `latest-result.json`.

I could have acquired another lock for updating the link to `latest-result.json` but that was not really necessary because we are not working with the file data, just updating the link and `fs.rename` is atomic for that purpose.

tests/accuracy/sdk/accuracy-result-storage/mongodb-storage.ts
nirinchev: Nvm, I saw there was another function added - any reason to keep this first one then?

scripts/accuracy/generate-test-summary.ts
himanshusinghs: Made it more intentional with combined conditionals

tests/accuracy/sdk/accuracy-result-storage/mongodb-storage.ts
himanshusinghs: I added the other one for comparing the output, I have removed the older one just now and simplified the pipeline as well by removing some variables.

tests/accuracy/collection-indexes.test.ts
```diff
@@ -0,0 +1,26 @@
...
+            },
+        ],
+    };
+}
+
+describeAccuracyTests(getAvailableModels(), [
+    callsCollectionIndexes("How many indexes do I have in 'mflix.movies' namespace?"),
+    callsCollectionIndexes("List all the indexes in movies collection in mflix database"),
+    callsCollectionIndexes(
+        `Is the following query: ${JSON.stringify({ runtime: { $lt: 100 } })} on the namespace 'mflix.movies' indexed?`
```
nirinchev: Should we expect the model to call explain for this prompt instead?

tests/accuracy/sdk/describe-accuracy-tests.ts
```diff
@@ -0,0 +1,125 @@
...
+    injectConnectedAssumption?: boolean;
+
+    /**
+     * A map of tool names to their mocked implementation. When the mocked
+     * implementations are available, the testing client will prefer those over
+     * actual MCP tool calls. */
+    mockedTools?: MockedTools;
+}
+
+export function describeAccuracyTests(models: TestableModels, accuracyTestConfigs: AccuracyTestConfig[]) {
```
nirinchev: This seems to be invoked always with `getAllModels()` - I understand the desire to be flexible, but it seems quite redundant for a use case we don't know if we ever will have.

tests/accuracy/sdk/models.ts
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    getModel() {
+        return ollama(this.modelName);
+    }
+}
+
+const ALL_TESTABLE_MODELS = [new AzureOpenAIModel("gpt-4o")];
+
+export type TestableModels = ReturnType<typeof getAvailableModels>;
```
nirinchev: Is this not just `Model[]`? Seems like an unnecessarily roundabout way of defining this type.

tests/accuracy/sdk/models.ts
```diff
@@ -0,0 +1,97 @@
...
+
+export function getAvailableModels() {
```
nirinchev: Let's be explicit about the return type instead of letting ts infer it.

tests/accuracy/sdk/models.ts
```diff
@@ -0,0 +1,97 @@
+import { LanguageModelV1 } from "ai";
+import { createGoogleGenerativeAI } from "@ai-sdk/google";
+import { createAzure } from "@ai-sdk/azure";
+import { createOpenAI } from "@ai-sdk/openai";
+import { ollama } from "ollama-ai-provider";
+
+export interface Model<P extends LanguageModelV1 = LanguageModelV1> {
```
nirinchev: What does `P` stand for in the generic here? Perhaps it's best to use a more verbose name - e.g. `TLanguageModel`?

tests/accuracy/sdk/models.ts
```diff
@@ -0,0 +1,97 @@
...
+    readonly modelName: string;
+    readonly provider: string;
+    readonly displayName: string;
+    isAvailable(): boolean;
+    getModel(): P;
+}
+
+export class OpenAIModel implements Model {
```
nirinchev, line 15: None of these classes specifies the generic model argument, making me question if that is really needed.

tests/accuracy/collection-storage-size.test.ts
```diff
@@ -0,0 +1,42 @@
...
+                parameters: {
+                    database: "mflix",
+                    collection: "movies",
+                },
+            },
+        ],
+    },
+    {
+        prompt: "How much size is each collection in comics database",
+        expectedToolCalls: [
```
nirinchev: Should this expect `list-databases` as well?

tests/accuracy/count.test.ts
```diff
@@ -0,0 +1,54 @@
+import { describeAccuracyTests } from "./sdk/describe-accuracy-tests.js";
+import { getAvailableModels } from "./sdk/models.js";
+import { AccuracyTestConfig } from "./sdk/describe-accuracy-tests.js";
+
+function callsCountToolWithEmptyQuery(prompt: string, database = "mflix", collection = "movies"): AccuracyTestConfig {
```
nirinchev: These two functions are nearly identical - can we have query as optional and consolidate them?

tests/accuracy/create-collection.test.ts
```diff
@@ -0,0 +1,51 @@
...
+                parameters: {
+                    database,
+                    collection,
+                },
+            },
+        ],
+    };
+}
+
+function callsCreateCollectionWithListCollections(prompt: string, expectedToolCalls: ExpectedToolCall[]) {
```
nirinchev: This needs return type annotation.

tests/accuracy/create-collection.test.ts
```diff
@@ -0,0 +1,51 @@
...
+    return {
+        injectConnectedAssumption: true,
+        prompt: prompt,
+        mockedTools: {},
```
nirinchev: Is this not the default?

tests/accuracy/create-collection.test.ts
```diff
@@ -0,0 +1,51 @@
...
+                },
+            },
+        ],
+    };
+}
+
+function callsCreateCollectionWithListCollections(prompt: string, expectedToolCalls: ExpectedToolCall[]) {
+    return {
+        injectConnectedAssumption: true,
+        prompt: prompt,
```
nirinchev:
```suggestion
        prompt,
```

tests/accuracy/delete-many.test.ts
```diff
@@ -0,0 +1,40 @@
...
+                parameters: {
+                    database: "mflix",
+                    collection: "movies",
+                },
+            },
+        ],
+    };
+}
+
+function callsDeleteManyWithFilters(prompt: string): AccuracyTestConfig {
```
nirinchev: This is nearly identical to the function above and is hardcoding the filter - can we consolidate and expose the filter as an argument?

tests/accuracy/drop-collection.test.ts
```diff
@@ -0,0 +1,74 @@
...
+        ],
+    };
+}
+
+function callsDropCollection(prompt: string, expectedToolCalls: ExpectedToolCall[]): AccuracyTestConfig {
+    return {
+        prompt: prompt,
+        expectedToolCalls,
+    };
+}
```
nirinchev: This function doesn't make much sense to me  - it's just creating an object with its arguments.

tests/accuracy/drop-collection.test.ts
```diff
@@ -0,0 +1,74 @@
+import { describeAccuracyTests } from "./sdk/describe-accuracy-tests.js";
+import { getAvailableModels } from "./sdk/models.js";
+import { AccuracyTestConfig } from "./sdk/describe-accuracy-tests.js";
+import { ExpectedToolCall } from "./sdk/accuracy-result-storage/result-storage.js";
+
+function onlyCallsDropCollection(prompt: string): AccuracyTestConfig {
+    return {
+        prompt: prompt,
```
nirinchev:
```suggestion
        prompt,
```

tests/accuracy/drop-database.test.ts
```diff
@@ -0,0 +1,42 @@
...
+                toolName: "drop-database",
+                parameters: {
+                    database: "mflix",
+                },
+            },
+        ],
+    };
+}
+
+function callsDropDatabase(prompt: string, expectedToolCalls: ExpectedToolCall[]): AccuracyTestConfig {
```
nirinchev: Similarly - trivial function that serves no purpose.

tests/accuracy/drop-database.test.ts
```diff
@@ -0,0 +1,42 @@
+import { describeAccuracyTests } from "./sdk/describe-accuracy-tests.js";
+import { getAvailableModels } from "./sdk/models.js";
+import { AccuracyTestConfig } from "./sdk/describe-accuracy-tests.js";
+import { ExpectedToolCall } from "./sdk/accuracy-result-storage/result-storage.js";
+
+function onlyCallsDropDatabase(prompt: string): AccuracyTestConfig {
+    return {
+        prompt: prompt,
```
nirinchev:
```suggestion
        prompt,
```

tests/accuracy/explain.test.ts
```diff
@@ -0,0 +1,64 @@
...
+        },
+    });
+
+const callsExplainWithCount = (prompt: string) =>
+    callsExplain(prompt, {
+        name: "count",
+        arguments: {
+            query: { release_year: 2020 },
+        },
+    });
```
nirinchev: These are one-off functions with a bunch of hardcoded data - we don't need to explicitly specify them and should instead just inline their implementation in the describe call.

nirinchev (changes requested):
Some comments so far - haven't gone through all of it, but the comments I left for some of the accuracy tests are applicable to the rest - review the function definitions and make sure those that are clearly consolidable are merged and functions that are used just once are inlined.

Additionally, I'm seeing LLM response and conversation messages being empty for each prompt in the generated report - is this expected?

Image

Finally, the naming of the new files doesn't follow the conventions of the repo - we tend to use camelCase names. Obviously, not enforced, but would be great to be consistent there.

tests/accuracy/collection-indexes.test.ts
himanshusinghs: I think the prompt is very borderline in terms of calling collection-indexes and explain but I would expect collection-indexes to be called because of `indexed` word in the prompt.

tests/accuracy/sdk/describe-accuracy-tests.ts
himanshusinghs: Don't have any argument against that. Certainly we're always passing all models so I will move it inside.

tests/accuracy/sdk/models.ts
himanshusinghs: Yea it is, will update this.

tests/accuracy/sdk/models.ts
himanshusinghs: The P was short for ModelProvider, I will make it as such.

tests/accuracy/explain.test.ts
himanshusinghs: So for most of these tests, the idea was to be verbose via function names in what we were expecting by just having a glance at the tests. Do you see them more as noise than helping understand the test suite quickly?

tests/accuracy/sdk/describe-accuracy-tests.ts
himanshusinghs: Addressed in the latest commit.

tests/accuracy/sdk/models.ts
himanshusinghs: Addressed in the latest commit.

tests/accuracy/sdk/models.ts
himanshusinghs: Addressed in the latest commit.

tests/accuracy/sdk/models.ts
himanshusinghs: Addressed in the latest commit.

tests/accuracy/collection-indexes.test.ts
nirinchev: Honestly, this is likely not a problem for this PR, but I feel looking at the explain plan is the only surefire way to check if a query is covered by an index. I guess for a simple query like this the model can look at the indexes and try and guess, but if it was `{ runtime: { $lt: 100 }, title: { $startsWith: "foo" } }`, just listing the indexes is likely not enough. We might want to tweak the explain description though to hint to models to prefer using that if the user is asking about a query.

tests/accuracy/explain.test.ts
nirinchev: I think they live in this limbo where they have some configurable arguments, but also some hardcoded ones, making it really awkward. My suggestion would be to either go all in on the functions or only use them when they're being called with different arguments. An example of the former would be to make these functions just constants instead by incorporating the prompt:

```ts
const testSimpleCount: AccuracyTestConfig = callsExplain("Will counting documents, where release_year is 2020, from 'mflix.movies' namespace perform a collection scan?", {
        name: "count",
        arguments: {
            query: { release_year: 2020 },
        },
    });

const testMultistageAggregation = ...

describeAccuracyTests([
  testSimpleCount,
  testSimpleAggregation,
  testMultistageAggregation,
  testSimpleFind,
]);
```

Alternatively, if we want to keep the functions for only reusable code, we could do something like:

```ts
function callsExplain(prompt: string, methodName: string, methodArguments: Record<string, unknown>): AccuracyTestConfig {
  return {
    prompt,
    expectedToolCalls: [{ 
      toolName: "explain",
      parameters: {
        database: "mflix",
        collection: "movies",
        method: [{ name: methodName, arguments: methodArguments  }]
      }
    }]
  };
}

describeAccuracyTests([
  callsExplain(
    `Will fetching documents, where release_year is 2020, from 'mflix.movies' namespace perform a collection scan?`, 
    "find", 
    { query: { release_year: 2020 }}),
  callsExplain(...)
```

I'm fine either way - the only thing that I don't like is that we have a function that takes the prompt as an argument, but hardcodes the expected query, meaning that invoking it with any other prompt would not be correct.

tests/accuracy/collection-indexes.test.ts
kmruiz: I do agree that this would be more a use case for the explain plan. However, if the models that we are testing do call list-indexes, either we should improve our prompting on the tools so LLMs prefer other tools or just keep it like this.

I believe this PR should show the current behaviour we already have, not fix what is not entirely correct.

tests/accuracy/drop-collection.test.ts
kmruiz: I don't really like having empty functions, but this follows the convention on all the tests, and likely callDropsCollection is a better name that a random JSON object. A nit would be to concat always the call to dropsCollection into the expectedToolCalls and leave expectedToolCalls as an optional vararg parameter for additional tool calls.

I'm personally fine with the current scenario because the tests are easy already to understand and maintain, and I don't think this is something that won't scale properly with the amount of tests.

tests/accuracy/explain.test.ts
kmruiz: > I'm fine either way - the only thing that I don't like is that we have a function that takes the prompt as an argument, but hardcodes the expected query, meaning that invoking it with any other prompt would not be correct.

That is what we would expect from the test, right? These tests ensure that specific prompts behave as we expect: we have multiple prompts that should behave the same, if they don't, we likely broke something when we changed our tools.

I don't have strong opinions on the aesthetics if the test is easy to change and it does what it has to do. What I like of this approach is that it's not easy to change the expected outcome, so it's not easy to monkey patch.

tests/accuracy/collection-indexes.test.ts
himanshusinghs: Cool then I will include this as part of improvements that we can do for the tool schemas and that's when I will move the this prompt over to explain test. Does that sound good to you both?

tests/accuracy/explain.test.ts
himanshusinghs: > the only thing that I don't like is that we have a function that takes the prompt as an argument, but hardcodes the expected query, meaning that invoking it with any other prompt would not be correct

So the functions are called only with prompt because all those prompts are expected to end up with same tool calls and same parameters. Wherever the parameter or tool calls differ, we have different functions (find test suite for example). If we instead do just objects in the list then you end up duplicating expectedToolCalls for no good reason and for some the prompts the expectedToolCalls list is big.

I think there are valid cases where a function definition is not needed at all, such as a suite having few prompts where tool call list is not that big to become repetitive noise but then there are actual suites where its better to reduce the repetitive noise by having it configurable via functions. Perhaps we should keep the tests writing style open to both methods and not stick to one?

tests/accuracy/collection-indexes.test.ts
nirinchev: Yeah, totally fine - we don't need to make any adjustments to the tools in this PR, but this was more of a note to look into it when we're actively working to improve the accuracy.

tests/accuracy/explain.test.ts
nirinchev: Again, fine with mixing-and-matching, what I don't like is having a function like:

```ts
const callsExplainWithCount = (prompt: string) =>
    callsExplain(prompt, {
        name: "count",
        arguments: {
            query: { release_year: 2020 },
        },
    });
```

Because if I call it like `callsExplainWithCount("Does counting documents where age is greater than 5 use an index")`, it will clearly not work. I guess there are some args that we can hardcode (e.g. database/collection, or better yet - we can use constants), but for me the prompt and the expected query are closely related and taking one as an argument, while hardcoding the other is counterintuitive.

tests/accuracy/sdk/accuracy-scorer.ts
```diff
@@ -0,0 +1,114 @@
...
+
+    if (differences.length === 0) {
+        return 1;
+    }
+
+    const hasOnlyAdditions = differences.every((d) => d.type === "CREATE");
+    const hasRemovals = differences.some((d) => d.type === "REMOVE");
+    const hasChanges = differences.some((d) => d.type === "CHANGE");
+
+    if (hasOnlyAdditions && !hasRemovals && !hasChanges) {
```
nirinchev: This check seems to be redundant here - am I missing something, or does `hasOnlyAdditions` being true guarantee that `hasRemovals` and `hasChanges` are false?

tests/accuracy/sdk/accuracy-scorer.ts
```diff
@@ -0,0 +1,114 @@
...
+        return 0.75;
```
nirinchev: Is this safe to assume? If I expected `find` to be called with no filter, but the LLM hallucinated a filter, that fundamentally changes the accuracy of the result. Scoring it 0.75 seems to be too generous, but perhaps I'm missing valid cases where extra arguments don't have too big of an impact.

tests/accuracy/sdk/accuracy-scorer.ts
```diff
@@ -0,0 +1,114 @@
...
+    actualToolCalls: LLMToolCall[]
+): number {
+    if (expectedToolCalls.length === 0) {
+        return actualToolCalls.length === 0 ? 1 : 0.75;
+    }
+
+    const maxAccuracy = actualToolCalls.length > expectedToolCalls.length ? 0.75 : 1;
+
+    const individualAccuracies: number[] = [];
+    const checkedActualToolCallIndexes = new Set<number>();
```
nirinchev: Not sure I understand the need for this. I understand it on a theoretical level - we don't want to associate several expected tool calls with the same actual tool call, but on a practical level, is this something that can realistically have an impact on the outcome? E.g. I imagine if I have several expected tool calls with different arguments and they map to the same actual tool call, the score for that should be 0, because the arguments would change?

If we do remove that safeguard, we can significantly simplify the algorithm then by doing something like:

```ts
const individualAccuracy = Math.min(
  Math.max(Object.groupBy(actualToolCalls, (t) => t.name)[expectedCall.name]
    ?.map((t) => compareParams(expectedCall.parameters, t.parameters) ?? [0]),
  maxAccuracy);
```

tests/accuracy/sdk/accuracy-testing-client.ts
```diff
@@ -0,0 +1,93 @@
...
+/**
+ * AccuracyTestingClient is a bridge between actual MCP client connected to our
+ * MCP server and our Tool calling agent. Its serves the following purposes:
+ * 1. Captures actual tools provided by our MCP server
+ * 2. Translates captured MCP tools to tool definitions that can be consumed by
+ *    Tool Calling agent (Ref: `vercelTools`)
+ * 3. Allow dynamic mocking and resetting of mocks of individual tool calls.
+ * 4. Records and provides tool calls made by LLMs with their parameters.
+ */
+export class AccuracyTestingClient {
```
nirinchev: We should add return type annotations to all functions here. We should probably enable the corresponding eslint rule, but I'd suggest doing it in another PR since it's likely to complain about pre-existing issues.

tests/accuracy/collection-indexes.test.ts
himanshusinghs: Yea I will certainly loop this in when we start updating the tool.

tests/accuracy/sdk/accuracy-scorer.ts
himanshusinghs: The check is still needed. All three types can appear at the same time for an object where there are additional parameters / tool calls, some parameter value difference and missing parameters / tool calls.

I guess I will change the name to make it less confusing.

tests/accuracy/sdk/accuracy-scorer.ts
himanshusinghs: The safeguard is for the case when LLM calls the tool incorrectly, figures out the problem and then calls the tool again with possibly correct parameters (might be more than just one correction). This happened quite often in find tool where it tool LLMs at-least 2 retries to figure out the correct tool call. In that case, out of 3 actual tool calls only one really map to expected tool call. The other 2 were hallucinations / incorrect calls and for that reason it was necessary that we pair as accurately as possible the expected and the actual tool calls.

For the case you mentioned, when we expected several tool calls with the same parameters and LLM does only one tool call then yea we will grade it 0. The entire logic stand on the idea that expectedToolCalls will have the bare-minimum tool calls and parameters required to answer the prompt. So if LLMs don't make those tool calls then they certainly don't answer the prompt correctly and hence the 0 accuracy.

tests/accuracy/sdk/accuracy-scorer.ts
himanshusinghs: This is something to be thought of. I was not expecting LLM to provide an empty object when it can actually just skip it (in case of no filters), that is what I based this assumption on but there is a good chance that LLM can pass in an empty filter object in which case instead of grading the accuracy 0, we will grade it 0.75 and it seems wrong.

I guess one quick fix is to check again in the expected params if the added key is undefined or an empty object and only for that case, mark the accuracy 0.

tests/accuracy/sdk/accuracy-scorer.ts
nirinchev: How is `checkedActualToolCallIndexes` involved in the first example you shared? The way I understand it, there's one tool call expected, let's say 3 actual. We'd max it at 0.75 because there are more tool calls, then we'd evaluate each actual vs the expected, will find the 3rd one, which may be a perfect match, but we'd limit it to 0.75, which will be the final accuracy. It's very possible that I'm not fully following the algorithm here, but I don't see how `checkedActualToolCallIndexes` impacts the outcome.

tests/accuracy/sdk/accuracy-scorer.ts
nirinchev: What is a case where you can see an extra parameter being supplied as not affecting the accuracy that much?

tests/accuracy/sdk/accuracy-scorer.ts
nirinchev: How is that possible - if `differences.every((d) => d.type === "CREATE")`, this means that every difference in the array has type `CREATE`. How would it be possible then to find a difference where the type is `CHANGE`?

tests/accuracy/sdk/accuracy-scorer.ts
himanshusinghs: The `checkedActualToolCallIndexes` ensures that you don't look at the already considered actual tool call again. Its like you're establishing pairs of expected tool call with llm tool call and you would not want to pair an already paired llm tool call again with another instance of similar expected tool call.

tests/accuracy/sdk/accuracy-scorer.ts
himanshusinghs: There could be multiple, for example:
When you ask LLM to find all the documents, based on our schema, it can decide to put the sort key and parameters become `{ database: "", coll: "", sort: { "field": 1 } }`. For our case, simply `{ database: "", coll: "" }` would have been fine but because it provided `sort` we will grade it 0.75.

The same can go for projection as well and even for filters when our expected tool call don't expect filter key to be present but LLM decide to put it with an empty object. Its the same in that case.

tests/accuracy/explain.test.ts
himanshusinghs: I have removed the functions entirely and inlined the test configs.

tests/accuracy/sdk/accuracy-scorer.ts
himanshusinghs: Ahh I mixed the function calls here - yes the additional check is redundant. I will update accordingly.

himanshusinghs: Updated

himanshusinghs:
> Additionally, I'm seeing LLM response and conversation messages being empty for each prompt in the generated report - is this expected?

Yep that's expected because we don't save the conversation and text when saving to MongoDB. When you run it locally, you will get those details. If it helps I can update the description to point this out.

tests/accuracy/sdk/accuracy-scorer.ts
himanshusinghs: I added a fix for the problem here in the following commit - [8488144](https://github.com/mongodb-js/mongodb-mcp-server/pull/341/commits/8488144fd5e14104071fd37b58326af1194bf771)

tests/accuracy/sdk/parameterScorer.ts
```diff
@@ -0,0 +1,59 @@
...
+                    !(
+                        valueAtPath === null ||
+                        valueAtPath === undefined ||
+                        (typeof valueAtPath === "object" && Object.keys(valueAtPath).length === 0) ||
+                        (Array.isArray(valueAtPath) && !valueAtPath.length)
+                    )
+                );
+            });
+
+            return hasNonEmptyAdditions ? 0 : 0.75;
```
himanshusinghs: This could be more lenient and just return 1 instead of 0.75 in case of empty additions but I thought of keeping it this way to be aligned with how we see hallucinations.

Issue #321, opened 2025-06-25
`atlas-connect-cluster` command randomly fails to connect


javamatte:
As the title says, the `atlas-connect-cluster` command sometimes fails with the message:
> Error running atlas-connect-cluster: Authentication failed.

That sounds like a repeatable error (bad credentials possibly?), but if you run it 3 (or 5!?) more times it eventually connects with zero changes other than retrying.

Can you either enhance the error messaging with timeout information (this feels like a 15 sec timeout problem) or other causes of failure if you can't determine that the credentials are, in fact, incorrect?

Instructions to the LLM can work-around the issue, but it feels dumb telling it to just manually retry 5 or 6 times ignoring the error message to actually connect.

Our situation is a project with multiple clusters, so a connectionString auth isn't possible. We are using a Project-level Application (clientId, clientSecret) to auth.

Partial log of an evetually-successful attempt to switch clusters:
```
⏺ Now let me connect to the account-service to get the details of this user:

⏺ mongodb:atlas-connect-cluster (MCP)(projectId: "-redacted-", clusterName: "account-service")
  ⎿  Error: Error running atlas-connect-cluster: Authentication failed.

⏺ mongodb:atlas-connect-cluster (MCP)(projectId: "-redacted-", clusterName: "account-service")
  ⎿  Error: Error running atlas-connect-cluster: Authentication failed.

⏺ mongodb:atlas-connect-cluster (MCP)(projectId: "-redacted-", clusterName: "account-service")
  ⎿  Error: Error running atlas-connect-cluster: Authentication failed.

⏺ Let me try a different approach and see if I can get more details from the forum service about this user:

⏺ mongodb:atlas-connect-cluster (MCP)(projectId: "-redacted-", clusterName: "forum-service")

> No try to connect to account-service a few more times

⏺ mongodb:atlas-connect-cluster (MCP)(projectId: "-redacted-", clusterName: "account-service")
  ⎿  Error: Error running atlas-connect-cluster: Authentication failed.

⏺ mongodb:atlas-connect-cluster (MCP)(projectId: "-redacted-", clusterName: "account-service")
  ⎿  Error: Error running atlas-connect-cluster: Authentication failed.

⏺ mongodb:atlas-connect-cluster (MCP)(projectId: "-redacted-", clusterName: "account-service")
  ⎿  Connected to cluster "account-service"

⏺ Great! Now let me find the account details for the user...
```

lduff:
[...]

PR #343, opened 2025-07-07
fix: turn atlas-connect-cluster async

fmenezes:
fixes https://github.com/mongodb-js/mongodb-mcp-server/issues/321

## Proposed changes

turn `atlas-connect-cluster` tool into an async tool. Unfortunately, rely on atlas user creation api to connect, sometimes the propagation time of the user from control plane to the data plane can take time, there is no state to query other than check if user has access to DB.

## Checklist

### DIFF
--- a/src/tools/atlas/metadata/connectCluster.ts
+++ b/src/tools/atlas/metadata/connectCluster.ts
@@ -142,8 +142,7 @@ export class ConnectClusterTool extends AtlasToolBase {
                 this.session.connectedAtlasCluster.projectId != projectId ||
                 this.session.connectedAtlasCluster.clusterName != clusterName
             ) {
-                lastError = new Error("Cluster connection aborted");
-                break;
+                throw new Error("Cluster connection aborted");
             }
 
             try {
@@ -167,7 +166,7 @@ export class ConnectClusterTool extends AtlasToolBase {
         }
 
         if (lastError) {
-            if (this.session.connectedAtlasCluster?.projectId && this.session.connectedAtlasCluster?.username) {
+            if (this.session.connectedAtlasCluster?.projectId == projectId && this.session.connectedAtlasCluster?.clusterName == clusterName && this.session.connectedAtlasCluster?.username) {
                 void this.session.apiClient
                     .deleteDatabaseUser({
                         params: {
--- a/src/tools/atlas/metadata/connectCluster.ts
+++ b/src/tools/atlas/metadata/connectCluster.ts
@@ -223,7 +223,7 @@ export class ConnectClusterTool extends AtlasToolBase {
             content: [
                 {
                     type: "text",
-                    text: `Connecting to cluster "${clusterName}"...`,
+                    text: `Attempting to connect to cluster "${clusterName}"...`,
                 },
             ],
         };
--- a/src/tools/atlas/metadata/connectCluster.ts
+++ b/src/tools/atlas/metadata/connectCluster.ts
@@ -14,7 +14,7 @@ function sleep(ms: number): Promise<void> {
 
 export class ConnectClusterTool extends AtlasToolBase {
     protected name = "atlas-connect-cluster";
-    protected description = "Connect to / Inspect connection of MongoDB Atlas cluster";
+    protected description = "Connect to MongoDB Atlas cluster";
     protected operationType: OperationType = "metadata";
     protected argsShape = {
         projectId: z.string().describe("Atlas project ID"),
--- a/src/tools/atlas/metadata/connectCluster.ts
+++ b/src/tools/atlas/metadata/connectCluster.ts
@@ -122,11 +122,7 @@ export class ConnectClusterTool extends AtlasToolBase {
         return cn.toString();
     }
 
-    private async connectToCluster(
-        projectId: string,
-        clusterName: string,
-        connectionString: string
-    ): Promise<void> {
+    private async connectToCluster(projectId: string, clusterName: string, connectionString: string): Promise<void> {
         let lastError: Error | undefined = undefined;
 
         logger.debug(
@@ -166,7 +162,11 @@ export class ConnectClusterTool extends AtlasToolBase {
         }
 
         if (lastError) {
-            if (this.session.connectedAtlasCluster?.projectId == projectId && this.session.connectedAtlasCluster?.clusterName == clusterName && this.session.connectedAtlasCluster?.username) {
+            if (
+                this.session.connectedAtlasCluster?.projectId == projectId &&
+                this.session.connectedAtlasCluster?.clusterName == clusterName &&
+                this.session.connectedAtlasCluster?.username
+            ) {
                 void this.session.apiClient
                     .deleteDatabaseUser({
                         params: {
@@ -218,7 +218,7 @@ export class ConnectClusterTool extends AtlasToolBase {
                 default:
                     await this.session.disconnect();
                     const connectionString = await this.prepareClusterConnection(projectId, clusterName);
-                    
+
                     // try to connect for about 5 minutes asynchronously
                     void this.connectToCluster(projectId, clusterName, connectionString).catch((err: unknown) => {
                         const error = err instanceof Error ? err : new Error(String(err));
--- a/src/tools/atlas/metadata/connectCluster.ts
+++ b/src/tools/atlas/metadata/connectCluster.ts
@@ -108,7 +108,12 @@ export class ConnectClusterTool extends AtlasToolBase {
         return cn.toString();
     }
 
-    private async connectToCluster(connectionString: string, tryCount: number): Promise<void> {
+    private async connectToCluster(
+        projectId: string,
+        clusterName: string,
+        connectionString: string,
+        tryCount: number
+    ): Promise<void> {
         let lastError: Error | undefined = undefined;
 
         logger.debug(
@@ -118,7 +123,11 @@ export class ConnectClusterTool extends AtlasToolBase {
         );
 
         for (let i = 0; i < tryCount; i++) {
-            if (!this.session.connectedAtlasCluster) {
+            if (
+                !this.session.connectedAtlasCluster ||
+                this.session.connectedAtlasCluster.projectId != projectId ||
+                this.session.connectedAtlasCluster.clusterName != clusterName
+            ) {
                 lastError = new Error("Cluster connection aborted");
                 break;
             }
@@ -225,7 +234,7 @@ export class ConnectClusterTool extends AtlasToolBase {
         try {
             // First, try to connect to the cluster within the current tool call.
             // We give it 60 attempts with 500 ms delay between each, so ~30 seconds
-            await this.connectToCluster(connectionString, 60);
+            await this.connectToCluster(projectId, clusterName, connectionString, 60);
 
             return {
                 content: [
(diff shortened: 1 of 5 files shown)


src/tools/atlas/metadata/connectCluster.ts
```diff
@@ -83,9 +107,20 @@ export class ConnectClusterTool extends AtlasToolBase {
         cn.searchParams.set("authSource", "admin");
         const connectionString = cn.toString();
 
+        return connectionString;
```
nirinchev:
```suggestion
        return cn.toString();
```

src/tools/atlas/metadata/connectCluster.ts
```diff
@@ -106,14 +141,97 @@ export class ConnectClusterTool extends AtlasToolBase {
         }
 
         if (lastError) {
+            void this.session.apiClient
+                .deleteDatabaseUser({
+                    params: {
+                        path: {
+                            groupId: this.session.connectedAtlasCluster?.projectId || "",
+                            username: this.session.connectedAtlasCluster?.username || "",
```
nirinchev: If those are not set, does it make sense to make that call at all?

src/tools/atlas/metadata/connectCluster.ts
```diff
@@ -106,14 +141,97 @@ export class ConnectClusterTool extends AtlasToolBase {
...
         }
 
+        logger.debug(
+            LogId.atlasConnectSucceeded,
+            "atlas-connect-cluster",
+            `connected to cluster: ${this.session.connectedAtlasCluster?.clusterName}`
+        );
+    }
+
+    protected async execute({ projectId, clusterName }: ToolArgs<typeof this.argsShape>): Promise<CallToolResult> {
```
nirinchev, line 200: I'm not so sure about the DX changes here - any reason not to do the polling internally and only return a result once we're connected? Assuming the common use case for connecting to a cluster is to do something with it (e.g. insert/query documents) - the user will be waiting anyway, so might as well just notify them when we're done. Note that once we add support for streamable http, we'd be able to use the progress notifications mechanism to notify the user that something is going on, so they won't be just starting at a spinner for 10-15 seconds.

fmenezes: timeouts, client-side controls them no extended timeouts per tool, even our CI is intermittent, notice I added a very generous timeout (5 minutes)

src/tools/atlas/metadata/connectCluster.ts
```diff
@@ -20,9 +21,32 @@ export class ConnectClusterTool extends AtlasToolBase {
...
+        }
+
+        if (!this.session.serviceProvider) {
+            return "connecting";
+        }
+
+        await this.session.serviceProvider.runCommand("admin", {
+            ping: 1,
+        });
+        return "connected";
```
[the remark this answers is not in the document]
fmenezes: no need, I'm bubbling up the error

src/tools/atlas/metadata/connectCluster.ts
fmenezes: this is coming from https://github.com/mongodb-js/mongodb-mcp-server/issues/321

fmenezes: poking around I've found that typescripts default timeout for tool calling is 1 minute (https://github.com/modelcontextprotocol/typescript-sdk/blob/main/src/shared/protocol.ts#L53C45-L53C50)

nirinchev: Hm... looking at #321, it doesn't appear like the issue is a timeout, but rather that the original 10-ish second window we gave the db user to be connected isn't enough (assuming that the auth error is due to the user still not being created after 20 retries with 500 ms delay each).

How would you feel if we went a middle ground where we polled internally for e.g. 45 seconds and if we still didn't have a valid user, we would return something along the lines of "provisioning the user and creating to the cluster is taking a bit longer, so check back in 15-20 seconds"? That way for the majority of users, we'd still connect within the first tool call, while still providing a graceful fallback for the few where it takes a lot longer.

fmenezes: Right, I think that is acceptable

fmenezes: I'll limit it to 30 secs given we have two extra API calls that can also eat some time

fmenezes:
@nirinchev this is ready for another look

src/tools/atlas/metadata/connectCluster.ts
```diff
@@ -106,16 +143,117 @@ export class ConnectClusterTool extends AtlasToolBase {
...
-        return {
+        logger.debug(
+            LogId.atlasConnectSucceeded,
+            "atlas-connect-cluster",
+            `connected to cluster: ${this.session.connectedAtlasCluster?.clusterName}`
+        );
+    }
+
+    protected async execute({ projectId, clusterName }: ToolArgs<typeof this.argsShape>): Promise<CallToolResult> {
+        const connectingResult = {
```
nirinchev: Any reason to have this at the top of the function if it's only used at the last catch clause?

src/tools/atlas/metadata/connectCluster.ts
```diff
@@ -106,16 +143,117 @@ export class ConnectClusterTool extends AtlasToolBase {
...
+                `error querying cluster: ${error.message}`
+            );
+            // fall through to create new connection
+        }
+
+        await this.session.disconnect();
+        const connectionString = await this.prepareClusterConnection(projectId, clusterName);
+
+        try {
+            await this.connectToCluster(connectionString, 60);
```
nirinchev: Let's add some comments that could guide readers - e.g.:


```suggestion
            // First, try to connect to the cluster within the current tool call.
            // We give it 60 attempts with 500 ms delay between each, so ~30 seconds
            await this.connectToCluster(connectionString, 60);
```

src/tools/atlas/metadata/connectCluster.ts
```diff
@@ -106,16 +143,117 @@ export class ConnectClusterTool extends AtlasToolBase {
...
+            };
+        } catch (err: unknown) {
+            const error = err instanceof Error ? err : new Error(String(err));
+            logger.debug(
+                LogId.atlasConnectFailure,
+                "atlas-connect-cluster",
+                `error connecting to cluster: ${error.message}`
+            );
+
+            process.nextTick(async () => {
```
nirinchev:
```suggestion
            // We couldn't connect in ~30 seconds, likely because user creation is taking longer
            // Retry the connection with longer timeout (~5 minutes), while also returning a response
            // to the client. Many clients will have a 1 minute timeout for tool calls, so we want to
            // return well before that.
            //
            // Once we add support for streamable http, we'd want to use progress notifications here.
            process.nextTick(async () => {
```

tests/integration/tools/atlas/clusters.test.ts
```diff
@@ -183,13 +183,27 @@ describeWithAtlas("clusters", (integration) => {
...
                 const projectId = getProjectId();
 
-                const response = (await integration.mcpClient().callTool({
-                    name: "atlas-connect-cluster",
-                    arguments: { projectId, clusterName },
-                })) as CallToolResult;
-                expect(response.content).toBeArray();
-                expect(response.content).toHaveLength(1);
-                expect(response.content[0]?.text).toContain(`Connected to cluster "${clusterName}"`);
+                for (let i = 0; i < 600; i++) {
```
nirinchev: Now that we added the 30 second retry logic in the connect tool, this may be a bit excessive - worst case scenario, this will result in 5 hours of waiting for the test to fail.

nirinchev (approved):
Looks good - let's add some clarifying comments (feel free to use my suggestion or reword them) and fix the retry attempts for the "connects to cluster" test.

tests/integration/tools/atlas/clusters.test.ts
gagik: I'd expect jest has some default test timeouts. maybe we can set an explicit timeout here and turn this into a while loop or create a `waitFor` / `retry` helper

src/tools/atlas/metadata/connectCluster.ts
gagik: what is the benefit of using `process.nextTick` here? 
There's no problematic operation that'd take precedence I can think of that would justify its usage here.

src/tools/atlas/metadata/connectCluster.ts
fmenezes: This seems resolved/agreed hitting resolve, feel free to reopen

src/tools/atlas/metadata/connectCluster.ts
fmenezes: yes, it is used in two cases, the last return and a switch case early on `case "connecting"`

src/tools/atlas/metadata/connectCluster.ts
fmenezes: this is my bad, when I use to work with node we didn't have/use promises (10 yrs ago) so process.nextTick was used for async operations, now-a-days `void somePromise` is much handier and I keep forgetting

tests/integration/tools/atlas/clusters.test.ts
fmenezes: This is actually not the case, on subsequent calls we don't try for 30 secs, we know there is a background process running so we return `Attempting ...` message straight away.

src/tools/atlas/metadata/connectCluster.ts
fmenezes: done

src/tools/atlas/metadata/connectCluster.ts
fmenezes: done

fmenezes: addressed

gagik (approved):

tests/integration/tools/atlas/clusters.test.ts
fmenezes: changed to reflect what we discussed offline, now we always wait 30 secs, adjusted the test to 10 times only

src/tools/atlas/metadata/connectCluster.ts
```diff
@@ -106,14 +162,91 @@ export class ConnectClusterTool extends AtlasToolBase {
...
+                        ],
+                    };
+                case "connecting":
+                    break;
+                case "connected-to-other-cluster":
+                case "disconnected":
+                case "unknown":
+                default:
+                    await this.session.disconnect();
+                    const connectionString = await this.prepareClusterConnection(projectId, clusterName);
```
nirinchev, line 222: We should fix this, though it seems like we can move it entirely into `connectToCluster`, can't we?

nirinchev: Actually, we can't - we need to await this before we continue because it's setting the connected cluster on the session. Guess we just need to wrap it in curly braces.

nirinchev (approved):

src/tools/atlas/metadata/connectCluster.ts
fmenezes: fixed the tests and syntax

PR #349, opened 2025-07-09
chore: better connect tool guidance with Atlas MCP-48

nirinchev:
## Proposed changes

We're updating the response when trying to run a data-plane tool before we're connected. The main difference is that now we're actively hinting to the LLM that it needs to use the atlas-cluster-connect tool if it's available and providing some extra context in general. I've tested the results manually with Cursor, Claude, and VSCode and they all seem to be pretty good about picking up the hints.

This should resolve the issue reported by the IS team where the model was hallucinating connection strings instead of suggesting to connect to an Atlas cluster.

## Checklist


README.md
```diff
@@ -301,10 +301,11 @@ Operation types:
 - `delete` - Tools that delete resources, such as delete document, drop collection, etc.
 - `read` - Tools that read resources, such as find, aggregate, list clusters, etc.
 - `metadata` - Tools that read metadata, such as list databases, list collections, collection schema, etc.
+- `connect` - Tools that allow you to connect or switch the connection to a MongoDB instance. If this is disabled, you will need to provide a connection string through the config when starting the server.
```
nirinchev, line 321: Added a new tool operation type to more clearly cluster the connection tools and allow users to disable them if they want to be 100% certain they're always connected to the same cluster (this was another recommendation by the IS team).

Right now, I'm keeping that enabled when working with `--readOnly`, but am open to be convinced otherwise.

src/tools/tool.ts
```diff
@@ -1,27 +1,28 @@
...
+export type OperationType = "metadata" | "read" | "create" | "delete" | "update" | "connect";
 export type ToolCategory = "mongodb" | "atlas";
 export type TelemetryToolMetadata = {
     projectId?: string;
     orgId?: string;
 };
 
 export abstract class ToolBase {
-    protected abstract name: string;
+    public abstract name: string;
```
nirinchev, line 21: These are now public, because I wanted to avoid hardcoding the tool names

himanshusinghs (approved):
Looks good, no concerns about the new operation type!

PR #352, opened 2025-07-10
chore: revoke access tokens on server shutdown [MCP-53]

fmenezes:
## Proposed changes
 revoke atlas access tokens on server shutdown [MCP-53]

## Checklist

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -101,9 +101,9 @@ export class Server {
     }
 
     async close(): Promise<void> {
+        await this.mcpServer.close();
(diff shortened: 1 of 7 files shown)


src/server.ts
```diff
@@ -101,9 +101,9 @@ export class Server {
     }
 
     async close(): Promise<void> {
+        await this.mcpServer.close();
```
blva: if we revoke the token this should be after telemetry and session no?

blva: nvm it's session

PR #359, opened 2025-07-14
feat: add streamable http [MCP-55]

fmenezes:
## Proposed changes

Adding streamable http to MCP-42 feature branch.

Note: state management and tests will be added in a future PR.

## Checklist


README.md
```diff
@@ -278,16 +299,20 @@ The MongoDB MCP Server can be configured using multiple methods, with the follow
...
+| `indexCheck`       | false      | When set to true, enforces that query operations must use an index, rejecting queries that perform a collection scan.                                         |
+| `telemetry`        | enabled    | When set to disabled, disables telemetry collection.                                                                                                          |
+| `transport`        | stdio      | Either 'stdio' or 'http'.                                                                                                                                     |
+| `httpPort`         | 3000       | Port number.                                                                                                                                                  |
+| `httpHost`         | 127.0.0.1  | Host to bind the http server.                                                                                                                                 |
+| `logger`           | disk,mcp   | Comma separated values, possible values are `mcp`, `disk` and `stderr`.                                                                                       |
```
blva: nit: can we expand a little bit here or below in the features the logger purpose and usage examples?

src/transports/streamableHttpTransport.ts
```diff
@@ -0,0 +1,81 @@
...
+
+import { config } from "../common/config.js";
+import logger, { LogId } from "../common/logger.js";
+
+export function createHttpTransport(): StreamableHTTPServerTransport {
+    const app = express();
+    app.use(express.json());
+    app.enable("trust proxy"); // needed for reverse proxy support
+    app.use(express.urlencoded({ extended: true }));
+    app.use(express.json());
```
blva: [q] do we need to call     app.use(express.json()); twice?

src/transports/streamableHttpTransport.ts
```diff
@@ -0,0 +1,81 @@
...
+    app.post("/mcp", async (req: express.Request, res: express.Response) => {
+        try {
+            await transport.handleRequest(req, res, req.body);
+        } catch (error) {
+            logger.error(
+                LogId.streamableHttpTransportRequestFailure,
+                "streamableHttpTransport",
+                `Error handling request: ${error instanceof Error ? error.message : String(error)}`
+            );
+            res.sendStatus(400);
```
blva: nit: we have a task for this but we'll need to return JSON-RPC response error body, so this is a non-blocker, just in case its too simple to add here

src/transports/streamableHttpTransport.ts
```diff
@@ -0,0 +1,81 @@
...
+
+    const server = app.listen(config.httpPort, config.httpHost, () => {
+        logger.info(
+            LogId.streamableHttpTransportStarted,
+            "streamableHttpTransport",
+            `Server started on http://${config.httpHost}:${config.httpPort}`
+        );
+    });
+
+    transport.onclose = () => {
```
blva: [q] should we call this in index.ts?

fmenezes: no, `mcpServer.close()` should close the transport

src/transports/stdioTransport.ts
```diff
@@ -39,7 +39,7 @@ export class EJsonReadBuffer {
 //
 // This function creates a StdioServerTransport and replaces the internal readBuffer with EJsonReadBuffer
 // that uses EJson.parse instead.
-export function createEJsonTransport(): StdioServerTransport {
+export function createStdioTransport(): StdioServerTransport {
```
blva: nit: why not call `stdio.ts` if we're under transports folder? same for streamableHttp

src/transports/streamableHttpTransport.ts
```diff
@@ -0,0 +1,100 @@
...
+            await transport.handleRequest(req, res, req.body);
+        } catch (error) {
+            logger.error(
+                LogId.streamableHttpTransportRequestFailure,
+                "streamableHttpTransport",
+                `Error handling request: ${error instanceof Error ? error.message : String(error)}`
+            );
+            res.status(400).json({
+                jsonrpc: "2.0",
+                error: {
```
blva: based on the [spec](https://modelcontextprotocol.io/docs/concepts/transports#responses) we need to also add a request ID field here

src/transports/streamableHttpTransport.ts
```diff
@@ -0,0 +1,100 @@
...
+                    code: JSON_RPC_ERROR_CODE_PROCESSING_REQUEST_FAILED,
+                    message: `Error handling request: ${error instanceof Error ? error.message : String(error)}`,
```
blva: nit: should we throw the classic error message and populate any extra info in the `error.data` optional field?

src/transports/streamableHttpTransport.ts
blva: (note: feel free to address in the rpc ticket, probably worth just assigning it to you already)

blva (approved):
non-blocking comments - but need to fix linting

src/transports/streamableHttpTransport.ts
fmenezes: I see null ids everywhere in the SDK (https://github.com/modelcontextprotocol/typescript-sdk/blob/a9c907dd700a5c4b977624ac47fd8cc24c6bd473/src/server/streamableHttp.ts#L291). I'll leave it for now.

PR #361, opened 2025-07-14
feat: add streamable http [MCP-43]

fmenezes:
## Checklist


src/common/sessionStore.ts
```diff
@@ -0,0 +1,104 @@
...
+
+        session.abortTimeout.reset();
+
+        session.notificationTimeout.reset();
+    }
+
+    private sendNotification(sessionId: string): void {
+        const session = this.sessions[sessionId];
+        if (!session) {
+            return;
```
gagik, line 55: nit: maybe we should log a warning here?

src/common/sessionStore.ts
```diff
@@ -0,0 +1,104 @@
...
+        }
+        const logger = new McpLogger(session.mcpServer);
+        logger.info(
```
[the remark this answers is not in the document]
gagik: agreed, we should create a logger and store it inside the our session map

src/common/sessionStore.ts
```diff
@@ -0,0 +1,104 @@
...
+        );
+    }
+
+    setSession(sessionId: string, transport: StreamableHTTPServerTransport, mcpServer: McpServer): void {
+        if (this.sessions[sessionId]) {
+            throw new Error(`Session ${sessionId} already exists`);
+        }
+        const abortTimeout = new TimeoutManager(async () => {
+            const logger = new McpLogger(mcpServer);
+            logger.info(
```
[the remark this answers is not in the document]
gagik: agree

src/common/timeoutManager.ts
```diff
@@ -0,0 +1,63 @@
...
+            } catch (error: unknown) {
+                this.onerror?.(error);
+            }
+        }
+    }
+
+    /**
+     * Resets the timeout.
+     */
+    reset() {
```
gagik:
```suggestion
     * Restarts the timeout.
     */
    restart() {
```
Just to better differentiate from `clear` as it's not clear that the initial state has the timeout started already

src/common/timeoutManager.ts
```diff
@@ -0,0 +1,63 @@
...
+            clearTimeout(this.timeoutId);
+            this.timeoutId = undefined;
+        }
+    }
+
+    /**
+     * Runs the callback function.
+     */
+    private async runCallback() {
+        if (this.callback) {
```
gagik: `this.callback` is always defined right? this is not needed

src/common/timeoutManager.ts
```diff
@@ -0,0 +1,63 @@
...
+ * It is used to ensure that a callback function is called after a certain amount of time.
+ * If the callback function is not called after the timeout, it will be called with an error.
+ */
+export class TimeoutManager {
+    private timeoutId?: NodeJS.Timeout;
+
+    /**
+     * A callback function that is called when the timeout is reached.
+     */
+    public onerror?: (error: unknown) => void;
```
gagik: this function is never defined anywhere, what is the purpose of it? if it's meant to be used then it should be a parameter

src/common/timeoutManager.ts
```diff
@@ -0,0 +1,63 @@
+/**
+ * A class that manages timeouts for a callback function.
+ * It is used to ensure that a callback function is called after a certain amount of time.
+ * If the callback function is not called after the timeout, it will be called with an error.
```
gagik: when would a function not be called after a timeout? if we're setting a `setTimeout` then it would be called right?

src/common/timeoutManager.ts
```diff
@@ -0,0 +1,63 @@
...
+     * @param timeoutMS - The timeout in milliseconds.
+     */
+    constructor(
+        private readonly callback: () => Promise<void> | void,
+        private readonly timeoutMS: number
+    ) {
+        if (timeoutMS <= 0) {
+            throw new Error("timeoutMS must be greater than 0");
+        }
+        this.reset();
```
gagik: we should not have side-effects in constructors. I think this whole helper can be better represented as a single function.
```ts
type ManagedTimeout = {
     cancel: () => void;
     restart: () => void;
};

export function setManagedTimeout(
        callback: () => Promise<void> | void,
        timeoutMS: number,
        { onError }: { onError?: (error: unknown) => void }
) {
      const wrappedCallback = () => {
            try {
                await this.callback();
            } catch (error: unknown) {
                onError?.(error);
            }
      };
      let timeout = setTimeout(wrappedCallback, timeoutMS);
      const cancel = () => {
           cancelTimeout(timeout);
           timeout = undefined;
      };
      const restart = () => {
           cancel();
           timeout = setTimeout(wrappedCallback, timeoutMS);
      }
      return {
           cancel,
           restart
      };
}
```

src/server.ts
```diff
@@ -186,6 +189,35 @@ export class Server {
     }
 
     private async validateConfig(): Promise<void> {
+        const transport = this.userConfig.transport as string;
```
gagik, line 192: are these type casts needed?

gagik (changes requested):
Going to be focused on some technical writing today but just some initial review on first part of the changes

src/server.ts
fmenezes: That is the way I've found to validate it.

The problem here is as far as we are concerned, the possible values are predefined, so the literal types make sense, but the inputs are cli args or env vars, which are basically strings that can contain any values.

fmenezes:
@gagik I'm addressing all of these at https://github.com/mongodb-js/mongodb-mcp-server/pull/386, if you're happy with the changes, would you approve it?

thanks!

fmenezes:
@gagik I've merged the PR according to your comments, would take another look?

src/index.ts
```diff
@@ -1,58 +1,49 @@
...
+        try {
+            await transportRunner.close();
+            logger.error(LogId.serverClosed, "server", "Server closed");
+        } catch (error: unknown) {
+            logger.error(LogId.serverCloseFailure, "server", `Error closing server: ${error as string}`);
+        } finally {
+            process.exit(1);
+        }
+    }
+}
```
gagik, line 44:
```suggestion
    try {
        await transportRunner.start();
    } catch (error: unknown) {
        logger.emergency(LogId.serverStartFailure, "server", `Fatal error running server: ${error as string}`);
        await transportRunner.close();
    } finally {
        process.exit(1);
     }
}
```
nit: just to avoid jumbling too many nested try / catches. An error would propagate if something happens on close right?

src/server.ts
```diff
@@ -66,15 +67,17 @@ export class Server {
             return existingHandler(request, extra);
         });
 
-        const containerEnv = await detectContainerEnv();
-
-        if (containerEnv) {
```
gagik, line 71: do we not need to do this setting anymore?

src/server.ts
gagik: ah okay I see 
```ts
  if (transport !== "http" && transport !== "stdio") {
      throw new Error(`Invalid transport: ${String(transport)}`);
  }
```
could prevent it but yeah honestly isn't much better.

We can also use `zod` for assertions here. We can define a schema for the config and just let it assert, it'd do the error messages well. But can be done later too

src/transports/streamableHttp.ts
```diff
@@ -0,0 +1,178 @@
...
+            res.status(400).json({
+                jsonrpc: "2.0",
+                error: {
+                    code: JSON_RPC_ERROR_CODE_PROCESSING_REQUEST_FAILED,
+                    message: `failed to handle request`,
+                    data: error instanceof Error ? error.message : String(error),
+                },
+            });
+        });
+    };
```
gagik, line 36:
```suggestion
    return async (req: express.Request, res: express.Response, next: express.NextFunction) {
        try {
            await fn(req, res, next);
        } catch (error) {
            logger.error(
                LogId.streamableHttpTransportRequestFailure,
                "streamableHttpTransport",
                `Error handling request: ${error instanceof Error ? error.message : String(error)}`
            );
            res.status(400).json({
                jsonrpc: "2.0",
                error: {
                    code: JSON_RPC_ERROR_CODE_PROCESSING_REQUEST_FAILED,
                    message: `failed to handle request`,
                    data: error instanceof Error ? error.message : String(error),
                },
            });
        };
    };
```
nit: just prefer async syntax wherever possible

src/transports/streamableHttp.ts
```diff
@@ -0,0 +1,178 @@
...
+import { randomUUID } from "crypto";
+import { SessionStore } from "../common/sessionStore.js";
+
+const JSON_RPC_ERROR_CODE_PROCESSING_REQUEST_FAILED = -32000;
+const JSON_RPC_ERROR_CODE_SESSION_ID_REQUIRED = -32001;
+const JSON_RPC_ERROR_CODE_SESSION_ID_INVALID = -32002;
+const JSON_RPC_ERROR_CODE_SESSION_NOT_FOUND = -32003;
+const JSON_RPC_ERROR_CODE_INVALID_REQUEST = -32004;
+
+function promiseHandler(
```
gagik, line 17:
```suggestion
function withErrorHandling(
```
nit: I assume that's the purpose of it?

src/transports/streamableHttp.ts
```diff
@@ -0,0 +1,178 @@
...
+        super();
+        this.sessionStore = new SessionStore(this.userConfig.idleTimeoutMs, this.userConfig.notificationTimeoutMs);
+    }
+
+    async start() {
+        const app = express();
+        app.enable("trust proxy"); // needed for reverse proxy support
+        app.use(express.json());
+
+        const handleRequest = async (req: express.Request, res: express.Response) => {
```
gagik, line 53:
```suggestion
        const handleSessionRequest = async (req: express.Request, res: express.Response) => {
```

just to make it clearer this is for session-specific requests.
maybe we should move this to a private class method along with the general request handler?

src/transports/streamableHttp.ts
```diff
@@ -0,0 +1,178 @@
...
+                                "streamableHttpTransport",
+                                `Error closing session: ${error instanceof Error ? error.message : String(error)}`
+                            );
+                        }
+                    },
+                });
+
+                transport.onclose = () => {
+                    server.close().catch((error) => {
+                        logger.error(
```
gagik, line 130: nit: maybe worth moving this to a private class method? not sure what would be a good name but something to differnetiate from `handleSessionRequest` for example. Just to make this start function define the paths and individual methods to handle them

gagik (approved):
Seems reasonable. Just a couple questions and structure ideas.

src/server.ts
fmenezes: yes, for our telemetry

src/transports/streamableHttp.ts
fmenezes: The problem here is that Express.js does not work well with async middlewares. This is about making sure errors are still caught in Express.js.

src/transports/streamableHttp.ts
fmenezes: I don't think I follow? you want to move the `logger.error` or `server.close`? I want the close of http transport to trigger the server to stop listening on port and gracefully shutdown if possible.

src/server.ts
gagik: I meant I don't see it in the new code, is this being done somewhere else?

src/transports/streamableHttp.ts
gagik: but this is still an async middleware right? it returns a promise

fmenezes:
@gagik I'm merging as is and will address points in a follow up pr

src/transports/streamableHttp.ts
gagik: ah no this isn't about this line, I just meant we can consider moving the handler into its own named function as well to make the `start` function not nest too much. purely stylistic nit. I don't mind it like this either

src/index.ts
fmenezes: done in https://github.com/mongodb-js/mongodb-mcp-server/pull/389

src/transports/streamableHttp.ts
fmenezes: it does not return a promise it acts as a callback, if `fn()` which returns a promise fails I use `.catch` to treat it.

if I change to `async function ...` there is a chance of it throwing and we never respond.

fmenezes: in essence, this converts promise back to callback to make `Express.js` happy

PR #363, opened 2025-07-14
chore(tests): switch to vitest

gagik:
This is a proposal to switch to vitest in light of continued issues with ESM modules and jest. Jest support for ESM modules is still [experimental](https://jestjs.io/docs/ecmascript-modules) and it will likely require constant patching to keep it working as we end up adopting more ESM-based modules. This also applies to extra TypeScript processing.

[vitest](https://vitest.dev/) is tried and tested now and is essentially a drop-in replacement for jest (minus our minor usage of `jest-extended`, which is replaceable). 

### Pros of vitest
- Native TypeScript, ESM module support means **little to no configuration** and extra work required compared to configurating jest or chai/mocha/sinon.
- All in one package compared to chai/mocha/sinon.

### Cons of vitest
- As it is a more recent project, there's not much usage of it in the mongodb-js organization outside of small projects. I do believe that this is largely because of older age of most other codebases. vitest is used by the https://github.com/mongodb/chatbot for example.

### Possible alternatives:
- **Configuring jest**: Probably would take even more effort than this switching to vitest right now and we may run into future issues as its support is experimental.
- **Using chai/mocha/sinon**: We may have better time with this and would align with other mongodb-js codebases if that matters. Would likely again require extra setup but be less experimental than jest.

In my opinion minimal configuration is preferable and vitest is mature enough and well worth being adopted.


tests/integration/tools/mongodb/create/createIndex.test.ts
```diff
@@ -8,6 +8,7 @@ import {
     expectDefined,
 } from "../../../helpers.js";
 import { IndexDirection } from "mongodb";
+import { expect, it } from "vitest";
```
gagik, line 11: these imports can actually be avoided by opting in to use globals which would make this PR look less scary. However I'm not a fan of globals in general as they're a bit too magical and hide real dependencies.

tests/integration/helpers.ts
```diff
@@ -60,8 +60,7 @@ export function setupIntegrationTest(getUserConfig: () => UserConfig): Integrati
 
         // Mock hasValidAccessToken for tests
         if (userConfig.apiClientId && userConfig.apiClientSecret) {
-            const mockFn = jest.fn<() => Promise<boolean>>().mockResolvedValue(true);
-            // @ts-expect-error accessing private property for testing
+            const mockFn = vi.fn().mockResolvedValue(undefined);
```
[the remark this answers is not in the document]
nirinchev: Hm... this seems to be indeed a change from the old behavior.

tests/integration/tools/mongodb/create/createIndex.test.ts
nirinchev: I think using globals in this particular situation is kind of fine, but not a strong opinion.

tests/integration/tools/mongodb/create/insertMany.test.ts
```diff
@@ -7,6 +7,8 @@ import {
     validateThrowsForInvalidArguments,
     expectDefined,
 } from "../../../helpers.js";
+import { expect } from "vitest";
+import { it } from "vitest";
```
nirinchev:
```suggestion
import { expect, it } from "vitest";
```

tests/integration/tools/mongodb/metadata/listCollections.test.ts
```diff
@@ -55,10 +56,9 @@ describeWithMongoDB("listCollections tool", (integration) => {
             });
             const items2 = getResponseElements(response2.content);
             expect(items2).toHaveLength(2);
-            expect(items2.map((item) => item.text)).toIncludeSameMembers([
-                'Name: "collection-1"',
-                'Name: "collection-2"',
-            ]);
+            expect(items2.map((item) => item.text)).toEqual(
+                expect.arrayContaining(['Name: "collection-1"', 'Name: "collection-2"'])
+            );
```
nirinchev: Can this be `.strictEqual(...)`?

nirinchev (commented):
Looks reasonable - no strong feelings on vi vs the rest. Looks like the coverage reporter is slightly different though and will now report 2 lines for each missed line (now seems to be counting the closing curly brace). Probably not a blocking issue.

tests/integration/tools/mongodb/create/createIndex.test.ts
gagik: yeah another thing with globals was that every time it required eslint/typescript/etc. configuration to make it apply to some files but not others so this ended up being quicker

tests/integration/tools/mongodb/metadata/listCollections.test.ts
gagik: a lot of these `toIncludeSameMembers` tests are not order-sensitive but `toStrictEqual` is so generally no

nirinchev: Ah, true. Kind of annoying there's no built-in matcher for arrays ignoring the order.

gagik: we can add custom matcher; honestly will be great to use that kind of stuff more, will add it now

tests/integration/tools/mongodb/read/aggregate.test.ts
```diff
@@ -20,7 +21,7 @@ describeWithMongoDB("aggregate tool", (integration) => {
     validateThrowsForInvalidArguments(integration, "aggregate", [
         {},
         { database: "test", collection: "foo" },
-        { database: test, pipeline: [] },
```
gagik, line 23: this is actually best argument against globals, we had an accidental usage of jest's `test` global here.... I'm actually surprised the test was passing...

PR #371, opened 2025-07-15
chore(build): build a universal ESM and CommonJS package

gagik:
Adds ESM and CommonJS-based exports to the package.

The motivation is largely to add support for VSCode to use the library.

### Possible Alternatives

- Target `cjs` and use [gen-esm-wrapper](https://www.npmjs.com/package/gen-esm-wrapper) - might be simpler in a sense and more consistent with how we deal with it in i.e. mongosh and other packages. But because we were distributing an ESM package until now, the potential breaking change is a bit worrying.

Inspired by https://github.com/mongodb-js/mongodb-mcp-server/pull/325

Relevant reading:
- https://evertpot.com/universal-commonjs-esm-typescript-packages/
- https://dev.to/jakobjingleheimer/configuring-commonjs-es-modules-for-nodejs-12ed


scripts/update-package-info.ts
```diff
@@ -0,0 +1,26 @@
...
+    name?: string;
+    description?: string;
+}
+
+// Read package.json
+const packageJsonPath = join(import.meta.dirname, "..", "package.json");
+const packageJson = JSON.parse(readFileSync(packageJsonPath, "utf-8")) as PackageJson;
+
+// Define the packageInfo.ts content
+const packageInfoContent = `// This file was generated by scripts/update-package-info.ts - Do not edit it manually.
```
gagik: there's no good way to have `__dirname` / `import.meta.dirname` as explored in https://evertpot.com/universal-commonjs-esm-typescript-packages/ and I couldn't think of anything better either. so instead we'll generate the file on-demand. This is similar to how it is done in github.com/mongodb-js/mongosh/

package.json
```diff
@@ -14,13 +27,14 @@
   "publishConfig": {
     "access": "public"
   },
-  "type": "module",
   "scripts": {
     "prepare": "npm run build",
     "build:clean": "rm -rf dist",
-    "build:compile": "tsc --project tsconfig.build.json",
+    "build:update-package-info": "tsx scripts/update-package-info.ts",
+    "build:esm": "tsc --project tsconfig.esm.json && echo '{\"type\":\"module\"}' > dist/package.json",
```
gagik: [Source:](https://evertpot.com/universal-commonjs-esm-typescript-packages/) As you can also see that both our cjs and esm packages get a 1-line package.json file with a type property.

This is required because Node needs to know wether files with a .js extensions should be interpreted as CommonJS or ESM. It can’t figure it out after opening it.

Node will look at the nearest package.json to see if the type property was specified.

tsconfig.esm.json
```diff
@@ -0,0 +1,8 @@
+{
+  "extends": "./tsconfig.build.json",
+  "compilerOptions": {
+    "module": "esnext",
+    "moduleResolution": "bundler",
+    "outDir": "./dist"
```
gagik: I'm keeping the ESM output in `./dist` for backwards compatibility but could be nice to make this`./dist/esm`

We can also accomplish this with an extra index.js file which would re-export esm/index.js

.github/workflows/check.yml
```diff
@@ -55,4 +55,4 @@ jobs:
           rm -rf node_modules
           npm pkg set scripts.prepare="exit 0"
           npm install --omit=dev
-      - run: npx -y @modelcontextprotocol/inspector --cli --method tools/list -- node dist/index.js --connectionString "mongodb://localhost"
```
gagik: I'm setting to ESM to keep the changes minimal. Doing cjs default won't change our issues with the `oidc-plugin`, I think the only way to deal with that is to make some changes on the plugin itself to not depend on the `require` of an ESM module which is a higher Node version feature.

Generally worth checking if we're going to have any problems with `dist/index.js` going away.

I'd expect the worst being our dev environment configurations no longer being correct.

Otherwise we can also create a index.js alias to point to esm.

src/lib.ts
```diff
@@ -0,0 +1,4 @@
+export { Server, type ServerOptions } from "./server.js";
+export { Telemetry } from "./telemetry/telemetry.js";
```
gagik, line 2: we can consider exporting a single class which would make the setup easier but I'll leave that to future followups

scripts/createUniversalPackage.ts
```diff
@@ -0,0 +1,25 @@
...
+writeFileSync(esmPath, JSON.stringify({ type: "module" }));
+
+// CJS package.json
+const cjsPath = resolve(distDir, "cjs", "package.json");
+mkdirSync(resolve(distDir, "cjs"), { recursive: true });
+writeFileSync(cjsPath, JSON.stringify({ type: "commonjs" }));
+
+// Create a dist/index.js file that imports the ESM index.js file
+// To minimize breaking changes from pre-universal package time.
+const indexPath = resolve(distDir, "index.js");
```
gagik, line 24: this is just out of precaution in case some of our deployments end up looking for `dist/index.js`... can be removed though

.github/workflows/check.yml
himanshusinghs: I think this is solved, now that you have dist/index.js pointing to esm.

.github/workflows/prepare_release.yaml
```diff
@@ -32,6 +32,7 @@ jobs:
         id: bump-version
         run: |
           echo "NEW_VERSION=$(npm version ${{ inputs.version }} --no-git-tag-version)" >> $GITHUB_OUTPUT
+          npm run build:update-version
```
himanshusinghs: Ohh is this new, or were we missing something earlier?

package.json
```diff
@@ -2,26 +2,41 @@
...
+        "types": "./dist/esm/lib.d.ts",
+        "default": "./dist/esm/lib.js"
+      },
+      "require": {
+        "types": "./dist/cjs/lib.d.ts",
+        "default": "./dist/cjs/lib.js"
+      }
+    }
+  },
+  "main": "./dist/cjs/lib.js",
```
himanshusinghs, line 18: Perhaps we should add the `types` as well pointing to `"./dist/cjs/lib.d.ts"` considering that the `main` is for older tooling?

scripts/createUniversalPackage.ts
himanshusinghs: I personally think its fine to have it.

himanshusinghs (approved):
Looks perfect. Tested it locally with vscode and works as expected 🚀 

Totally unrelated but you might wanna cherry-pick the changes in [this commit](https://github.com/mongodb-js/mongodb-mcp-server/pull/325/commits/dbfb29d1f92b13a9e1262dedd6fea69a02266751) because currently we are shipping some unnecessary files in our final dist.

.github/workflows/prepare_release.yaml
himanshusinghs: nvm I just saw the update script.

src/resources/resource.ts
```diff
@@ -8,76 +8,84 @@ import { LogId } from "../common/logger.js";
...
+        this.config = server.userConfig;
 
-            for (const event of events) {
-                this.session.on(event, (...args: SessionEvents[typeof event]) => {
-                    this.reduceApply(event, (args as unknown[])[0] as PayloadOf<typeof event>);
-                    void this.triggerUpdate();
-                });
-            }
-        }
+        this.setupEventListeners();
```
gagik, line 46: this is technically a side effect in a constructor which we generally avoid but won't tackle this here.

PR #372, opened 2025-07-15
chore: Improve access list and connect experienece [MCP-5]

blva:
## Checklist


src/tools/atlas/connect/connectCluster.ts
```diff
@@ -56,6 +57,19 @@ export class ConnectClusterTool extends AtlasToolBase {
                 "atlas-connect-cluster",
                 `error querying cluster: ${error.message}`
             );
+
+            // sometimes the error can be "error querying cluster: bad auth : Authentication failed."
+            // which just means it's still propagating permissions to the cluster
+            // in that case, we want to classify this as connecting.
+            if (error.message.includes("Authentication failed")) {
```
blva: @fmenezes

fmenezes (approved):

src/common/atlas/accessListUtils.ts
```diff
@@ -0,0 +1,48 @@
+import { ApiClient } from "./apiClient.js";
+import logger, { LogId } from "../logger.js";
+import { ApiClientError } from "./apiClientError.js";
+
+export async function makeCurrentIpAccessListEntry(
+    apiClient: ApiClient,
+    projectId: string,
+    comment: string = "Added by Atlas MCP"
```
nirinchev: [nit] The product name is `MongoDB MCP Server` - should we use that here to avoid confusion?

src/common/atlas/accessListUtils.ts
```diff
@@ -0,0 +1,48 @@
...
+
+/**
+ * Ensures the current public IP is in the access list for the given Atlas project.
+ * If the IP is already present, this is a no-op.
+ * @param apiClient The Atlas API client instance
+ * @param projectId The Atlas project ID
+ */
+export async function ensureCurrentIpInAccessList(apiClient: ApiClient, projectId: string): Promise<void> {
+    // Get the current public IP
+    const entry = await makeCurrentIpAccessListEntry(apiClient, projectId, "Added by MCP pre-run access list helper");
```
nirinchev: [nit] Is that comment helpful externally? If I was a user and saw that, I wouldn't know what it means.

src/tools/atlas/create/createAccessList.ts
```diff
@@ -2,6 +2,7 @@ import { z } from "zod";
 import { CallToolResult } from "@modelcontextprotocol/sdk/types.js";
 import { AtlasToolBase } from "../atlasTool.js";
 import { ToolArgs, OperationType } from "../../tool.js";
+import { makeCurrentIpAccessListEntry } from "../../../common/atlas/accessListUtils.js";
 
 const DEFAULT_COMMENT = "Added by Atlas MCP";
```
nirinchev: This is duplicated now - should we move it to `accessListUtils`?

src/common/atlas/accessListUtils.ts
```diff
@@ -0,0 +1,48 @@
...
+            LogId.atlasIpAccessListAdded,
+            "accessListUtils",
+            `IP access list created: ${JSON.stringify(entry)}`
+        );
+    } catch (err) {
+        if (err instanceof ApiClientError && err.response?.status === 409) {
+            // 409 Conflict: entry already exists, ignore
+            return;
+        }
+        logger.debug(
```
nirinchev: This is silently ignoring the error - does it make sense that we do that? I assume that this could give users the false impression everything is working and then hit an issue with the IP address config at a later point?

nirinchev (commented):
Mostly nits/questions, but nothing blocking from me.

src/common/atlas/accessListUtils.ts
blva: thanks! updated

src/common/atlas/accessListUtils.ts
blva: I updated a bit more, but I think it's helpful to know where the new access list entry came from

src/tools/atlas/create/createAccessList.ts
blva: thanks, updated!

src/common/atlas/accessListUtils.ts
blva: Yes for now, there are different configurations where user does not have access to add access list but has accsess to proceed with the tool call, I'd like to avoid causing errors and having the logs to troubleshoot for now

blva: To be clear - I'm not sure about those scenarios but was trying to avoid causing disruption. If you think the risk is too low I can throw the error back. Thanks!

nirinchev: Yeah, not sure either, just not ideal that this function seems like a prerequisite for many features, but we're not communicating to the user that something might have gone wrong. I would say this PR is a clear improvement to the status quo, and the worst case of this failing and us logging it is no worse than what we currently have, so fine with keeping it as is. I guess we can at least bump the log level to warn so that it's a bit more prominent.

PR #379, opened 2025-07-16
feat: add session management for streamableHttp [MCP-52]

fmenezes:
## Proposed changes
add session management for streamableHttp
## Checklist

### DIFF
--- a/src/common/sessionStore.ts
+++ b/src/common/sessionStore.ts
@@ -1,0 +1,44 @@
+import { StreamableHTTPServerTransport } from "@modelcontextprotocol/sdk/server/streamableHttp.js";
+import logger, { LogId } from "./logger.js";
+
+export class SessionStore {
+    private sessions: Map<string, StreamableHTTPServerTransport> = new Map();
+
+    getSession(sessionId: string): StreamableHTTPServerTransport | undefined {
+        return this.sessions.get(sessionId);
+    }
+
+    setSession(sessionId: string, transport: StreamableHTTPServerTransport): void {
+        if (this.sessions.has(sessionId)) {
+            throw new Error(`Session ${sessionId} already exists`);
+        }
+        this.sessions.set(sessionId, transport);
+    }
+
+    async closeSession(sessionId: string, closeTransport: boolean = true): Promise<void> {
+        if (!this.sessions.has(sessionId)) {
+            throw new Error(`Session ${sessionId} not found`);
+        }
+        if (closeTransport) {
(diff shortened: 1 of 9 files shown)


andreaangiolillo (approved):
LGTM

src/common/sessionStore.ts
```diff
@@ -0,0 +1,48 @@
+import { StreamableHTTPServerTransport } from "@modelcontextprotocol/sdk/server/streamableHttp.js";
+import logger, { LogId } from "./logger.js";
+
+export class SessionStore {
+    private sessions: { [sessionId: string]: StreamableHTTPServerTransport | undefined } = {};
```
jeroenvervaeke: [q] Is there a reason why they value is nullable?
I don't see any location where the value (StreamableHTTPServerTransport) is set to `undefined`/removed

src/transports/streamableHttp.ts
```diff
@@ -1,90 +1,132 @@
...
+    private sessionStore: SessionStore = new SessionStore();
 
     async start() {
         const app = express();
         app.enable("trust proxy"); // needed for reverse proxy support
-        app.use(express.urlencoded({ extended: true }));
         app.use(express.json());
 
+        const handleRequest = async (req: express.Request, res: express.Response) => {
+            const sessionId = req.headers["mcp-session-id"] as string;
```
jeroenvervaeke: Wouldn't the better way to write this be:

```ts
const sessionId = req.headers["mcp-session-id"];
if (typeof sessionId !== 'string' || !sessionId) {
```

Or at least use `as string | undefined`, instead of `as string`, which is incorrect?

src/transports/streamableHttp.ts
```diff
@@ -1,90 +1,132 @@
...
+            await transport.handleRequest(req, res, req.body);
+        };
+
         app.post(
             "/mcp",
             promiseHandler(async (req: express.Request, res: express.Response) => {
-                const transport = new StreamableHTTPServerTransport({
-                    sessionIdGenerator: undefined,
-                });
+                const sessionId = req.headers["mcp-session-id"] as string;
```
jeroenvervaeke: Same as previous remark

src/common/sessionStore.ts
fmenezes: for the case where a key is not found

src/transports/streamableHttp.ts
fmenezes: good catch, I had an issue on my IDE before I think

PR #381, opened 2025-07-17
fix: tweak the arg shapes to improve tool accuracy

nirinchev:
## Proposed changes

The core fix here is moving away from `z.record` and using a plain `z.object({}).passthrough()`. The test changes are fixes to what seems to have been copy-paste errors where the prompt and the expectations were mismatched.

## Checklist

### DIFF
--- a/src/tools/mongodb/metadata/explain.ts
+++ b/src/tools/mongodb/metadata/explain.ts
@@ -16,7 +16,7 @@ export class ExplainTool extends MongoDBToolBase {
         ...DbOperationArgs,
         method: z
             .array(
-                z.union([
+                z.discriminatedUnion("name", [
                     z.object({
                         name: z.literal("aggregate"),
                         arguments: z.object(AggregateArgs),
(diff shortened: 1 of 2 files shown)


src/tools/mongodb/read/find.ts
```diff
@@ -8,18 +8,23 @@ import { checkIndexUsage } from "../../../helpers/indexCheck.js";
...
-        .record(z.string(), z.unknown())
+        .object({})
+        .passthrough()
         .optional()
         .describe("The projection, matching the syntax of the projection argument of db.collection.find()"),
     limit: z.number().optional().default(10).describe("The maximum number of documents to return"),
     sort: z
-        .record(z.string(), z.custom<SortDirection>())
+        .object({})
+        .catchall(z.custom<SortDirection>())
```
nirinchev, line 23: This, unfortunately, doesn't correctly set the type for `additionalProperties` in the JSON schema. This is the reason I had to update the tool description to be extra verbose.

PR #383, opened 2025-07-18
feat: add client idle timeout [MCP-57]

fmenezes:
## Checklist

### DIFF
--- a/src/common/sessionStore.ts
+++ b/src/common/sessionStore.ts
@@ -1,31 +1,106 @@
 import { StreamableHTTPServerTransport } from "@modelcontextprotocol/sdk/server/streamableHttp.js";
-import logger, { LogId } from "./logger.js";
+import { McpServer } from "@modelcontextprotocol/sdk/server/mcp.js";
+import logger, { LogId, McpLogger } from "./logger.js";
 
 export class SessionStore {
-    private sessions: { [sessionId: string]: StreamableHTTPServerTransport } = {};
+    private sessions: {
+        [sessionId: string]: {
+            mcpServer: McpServer;
+            transport: StreamableHTTPServerTransport;
+            abortController: AbortController;
+            abortTimeoutId: NodeJS.Timeout;
+            notificationTimeoutId: NodeJS.Timeout;
+        };
+    } = {};
 
-    getSession(sessionId: string): StreamableHTTPServerTransport | undefined {
-        return this.sessions[sessionId];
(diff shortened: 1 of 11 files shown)


src/common/sessionStore.ts
```diff
@@ -1,31 +1,132 @@
 import { StreamableHTTPServerTransport } from "@modelcontextprotocol/sdk/server/streamableHttp.js";
-import logger, { LogId } from "./logger.js";
+import { McpServer } from "@modelcontextprotocol/sdk/server/mcp.js";
+import logger, { LogId, McpLogger } from "./logger.js";
+
+class TimeoutManager {
```
blva: nit: can we add unit tests for this?

src/common/sessionStore.ts
```diff
@@ -1,31 +1,132 @@
...
+            void this.runCallback().finally(() => {
+                this.timeoutId = undefined;
+            });
+        }, this.timeoutMS);
+    }
+}
 
 export class SessionStore {
-    private sessions: { [sessionId: string]: StreamableHTTPServerTransport } = {};
+    private sessions: {
```
blva, line 7: [q] didn't check our telemetry PR but should we consider adding some http metrics? we could open a ticket for this, just checking about what you think. example: number of active sessions, timed out sessions

src/common/sessionStore.ts
blva: [q] I know is tiny but should we move this into it's own separate file?

blva: nit: can we add JS docs to the class and some of the methods below?

fmenezes: I didn't because this is not reused and it is private to the file, no strong opinion, can be moved

src/common/sessionStore.ts
fmenezes: does that add value? number of active sessions changes over time not sure how we would track it, we can add a time out session but bear in mind all of our telemetry is session based not server based, as it it lives inside of the MCP server not outside (http server).

src/common/sessionStore.ts
fmenezes: almost none of our code has JS docs, we should add it across the board or not add any I think

src/common/sessionStore.ts
blva: fair point - we can always check w/ analytics first hence the idea of a ticket

PR #405, opened 2025-07-28
chore: Enable proxy for OIDC and MongoDB connections

kmruiz:
## Proposed changes

By providing { useEnvironmentVariableProxies: true } the NodeDriverServiceProvider already reads the relevant environment variables and sets up the proxy internally.

This is as specified in the section `Setting up the proxy for NodeDriverServiceProvider (MongoDB + OIDC)` from the Technical Design.

## Checklist

### DIFF
--- a/src/common/session.ts
+++ b/src/common/session.ts
@@ -113,6 +113,8 @@ export class Session extends EventEmitter<{
                 w: connectOptions.writeConcern,
             },
             timeoutMS: connectOptions.timeoutMS,
+            proxy: { useEnvironmentVariableProxies: true },
+            applyProxyToOIDC: true,
         });
     }
 }
--- a/tests/unit/common/session.test.ts
+++ b/tests/unit/common/session.test.ts
@@ -56,5 +56,17 @@ describe("Session", () => {
                 }
             });
         }
+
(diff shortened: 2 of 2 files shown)


src/common/session.ts
```diff
@@ -113,6 +113,8 @@ export class Session extends EventEmitter<{
                 w: connectOptions.writeConcern,
             },
             timeoutMS: connectOptions.timeoutMS,
+            proxy: { useEnvironmentVariableProxies: true },
```
blva, line 116: do we need to call this out anywhere in the readme or a separate .md file? not sure just asking :)

blva (approved):
LGTM

src/common/session.ts
kmruiz: That's a good point. I guess we can add something in the README.md along with the public documentation we will have. Will jot something down today!

fmenezes:
@kmruiz does this need to be in a feature branch? does it need other PRs to become feature complete? (not blocking just wondering)

Note: looking at the code seems self contained enough.

kmruiz:
@fmenezes The auto-merged did it's work too early :D.

It's self-contained, it can be even released if we want to.

PR #406, opened 2025-07-28
feat: update connectionString appName param - [MCP-68]

blva:
## Proposed changes

- Refactored telemetry to use centralized device ID utility
- Extended appName format to `appName--deviceId--clientName`
- Created src/helpers/deviceId.ts utility for device ID retrieval
- Updated setAppNameParamIfMissing function to support extended format
- Modified Session and Atlas connect tools to use new format
- Added unit tests and integration tests
- Decoupled connectionString validation from connection

## Checklist

### DIFF
--- a/tests/accuracy/export.test.ts
+++ b/tests/accuracy/export.test.ts
@@ -10,7 +10,6 @@ describeAccuracyTests([
                 parameters: {
                     database: "mflix",
                     collection: "movies",
-                    exportTitle: Matcher.string(),
                     exportTarget: [
                         {
                             name: "find",
@@ -29,7 +28,6 @@ describeAccuracyTests([
                 parameters: {
                     database: "mflix",
                     collection: "movies",
-                    exportTitle: Matcher.string(),
                     exportTarget: [
                         {
                             name: "find",
@@ -52,7 +50,6 @@ describeAccuracyTests([
                 parameters: {
                     database: "mflix",
                     collection: "movies",
-                    exportTitle: Matcher.string(),
                     exportTarget: [
                         {
                             name: "find",
@@ -80,7 +77,6 @@ describeAccuracyTests([
                 parameters: {
                     database: "mflix",
                     collection: "movies",
-                    exportTitle: Matcher.string(),
                     exportTarget: [
                         {
                             name: "find",


src/telemetry/telemetry.ts
```diff
@@ -87,8 +60,6 @@ export class Telemetry {
     }
 
     public async close(): Promise<void> {
-        this.deviceIdAbortController.abort();
-        this.isBufferingEvents = false;
```
fmenezes, line 100: why are you removing `this.isBufferingEvents = false;`? don't we need it anymore?

src/helpers/connectionOptions.ts
```diff
@@ -1,20 +1,51 @@
...
+    }
+
+    // Get deviceId if not provided
+    let deviceId = components.deviceId;
+    if (!deviceId) {
+        deviceId = await getDeviceIdForConnection();
     }
 
+    // Get clientName if not provided
+    let clientName = components.clientName;
```
fmenezes: nit: style

```suggestion
    const clientName = components.clientName || "unknown";
```

src/telemetry/telemetry.ts
blva: nope that was a miss from my removal, adding it back

src/helpers/connectionOptions.ts
```diff
@@ -1,20 +1,51 @@
...
 
-    if (!searchParams.has("appName") && defaultAppName !== undefined) {
-        searchParams.set("appName", defaultAppName);
+    // Only set appName if it's not already present
+    if (searchParams.has("appName")) {
+        return connectionStringUrl.toString();
+    }
+
+    // Get deviceId if not provided
+    let deviceId = components.deviceId;
```
fmenezes: nit: style

```suggestion
    const deviceId = components.deviceId || await getDeviceIdForConnection();
```

src/helpers/connectionOptions.ts
```diff
@@ -1,20 +1,51 @@
...
     }
 
+    // Get clientName if not provided
+    let clientName = components.clientName;
+    if (!clientName) {
+        clientName = "unknown";
+    }
+
+    // Build the extended appName format: appName--deviceId--clientName
+    const extendedAppName = `${components.appName}--${deviceId}--${clientName}`;
```
fmenezes: [q] what if `components.appName` is empty? should it be `unknown` like the others?

fmenezes: I see the code we are getting `packageInfo` stuff, should we not make it a parameter and get package info straight away?

blva: the previous behaviour is that we would set it to   `appName: `${packageInfo.mcpServerName} ${packageInfo.version}`,` if it's not defined, so we always send it. I could move that logic here, is that what you mean with the packageInfo question?

fmenezes: how would appName be defined? is it really a parameter? is there an env var?

fmenezes: `components.appName` seems to be fixed as `${packageInfo.mcpServerName} ${packageInfo.version}` always

src/helpers/deviceId.ts
```diff
@@ -0,0 +1,49 @@
...
+ *
+ * @example
+ * ```typescript
+ * const deviceId = await getDeviceIdForConnection();
+ * console.log(deviceId); // Outputs the device ID or "unknown" in case of failure
+ * ```
+ */
+export async function getDeviceIdForConnection(): Promise<string> {
+    const controller = new AbortController();
+    const timeoutId = setTimeout(() => controller.abort(), DEVICE_ID_TIMEOUT);
```
fmenezes: why do we need timeout? `getDeviceId` has a timeout builtin

src/tools/atlas/connect/connectCluster.ts
```diff
@@ -120,7 +122,15 @@ export class ConnectClusterTool extends AtlasToolBase {
...
         cn.password = password;
         cn.searchParams.set("authSource", "admin");
-        return cn.toString();
+
+        const connectionStringWithAuth = cn.toString();
+        return await setAppNameParamIfMissing({
+            connectionString: connectionStringWithAuth,
+            components: {
+                appName: `${packageInfo.mcpServerName} ${packageInfo.version}`,
+                clientName: this.session.agentRunner?.name || "unknown",
```
fmenezes, line 131: seems like this is set on session initialize, did you check using http? I don't think you will get anything other than `unknown`

tests/unit/telemetry.test.ts
```diff
@@ -204,84 +203,65 @@ describe("Telemetry", () => {
...
                     expect(telemetry.getCommonProperties().device_id).toBe("unknown");
-
-                    expect(loggerSpy).toHaveBeenCalledWith(
-                        LogId.telemetryDeviceIdFailure,
-                        "telemetry",
-                        "Error: Failed to get device ID"
-                    );
                 });
 
-                it("should timeout if machine ID resolution takes too long", async () => {
```
gagik, line 261: why are we removing the timeout test? I think it is useful to have

tests/unit/telemetry.test.ts
```diff
@@ -204,84 +203,65 @@ describe("Telemetry", () => {
...
                     expect(telemetry["isBufferingEvents"]).toBe(false);
-                    expect(telemetry.getCommonProperties().device_id).toBe(hashedMachineId);
+                    expect(telemetry.getCommonProperties().device_id).toBe("test-device-id");
                 });
 
-                it("should handle machine ID resolution failure", async () => {
-                    const loggerSpy = vi.spyOn(logger, "debug");
+                it("should handle device ID resolution failure gracefully", async () => {
+                    // Mock the deviceId utility to return "unknown" for this test
+                    const { getDeviceIdForConnection } = await import("../../src/helpers/deviceId.js");
```
gagik, line 235: This is technically a different test from before. It assumes `getDeviceIdForConnection` does correctly end up returning `"unknown"`

tests/unit/telemetry.test.ts
blva: didn't remove, moved it to deviceId unit tests

gagik: ah saw this now, nevermind, ty!

src/telemetry/telemetry.ts
```diff
@@ -5,33 +5,27 @@ import logger, { LogId } from "../common/logger.js";
...
-    private deviceIdAbortController = new AbortController();
     private eventCache: EventCache;
-    private getRawMachineId: () => Promise<string>;
 
     private constructor(
         private readonly session: Session,
         private readonly userConfig: UserConfig,
         private readonly commonProperties: CommonProperties,
-        { eventCache, getRawMachineId }: { eventCache: EventCache; getRawMachineId: () => Promise<string> }
+        { eventCache }: { eventCache: EventCache }
```
gagik: nit: I think it'd still be good to pass `getDeviceIdForConnection` for explicit DI. It also helps us avoid mocking the file which is a bit of JavaScript magic

src/helpers/connectionOptions.ts
```diff
@@ -1,20 +1,42 @@
...
     const searchParams = connectionStringUrl.typedSearchParams<MongoClientOptions>();
 
-    if (!searchParams.has("appName") && defaultAppName !== undefined) {
-        searchParams.set("appName", defaultAppName);
+    // Only set appName if it's not already present
+    if (searchParams.has("appName")) {
+        return connectionStringUrl.toString();
     }
 
+    const deviceId = components.deviceId || (await getDeviceIdForConnection());
```
gagik: we shouldn't end up awaiting device ID a second time, can we organize it so this device ID gets passed from one point of entry and we never fallback to this function? `deviceId` doesn't get passed in `connectToMongoDB` right now.

src/helpers/connectionOptions.ts
```diff
@@ -1,20 +1,42 @@
...
+    components: AppNameComponents;
+}): Promise<string> {
     const connectionStringUrl = new ConnectionString(connectionString);
-
     const searchParams = connectionStringUrl.typedSearchParams<MongoClientOptions>();
 
-    if (!searchParams.has("appName") && defaultAppName !== undefined) {
-        searchParams.set("appName", defaultAppName);
+    // Only set appName if it's not already present
+    if (searchParams.has("appName")) {
```
gagik, line 28: In Compass / mongosh we only pass the device ID if it's an Atlas host to minimize visibility of this, can we also do this here?

gagik (commented):
It's important in my opinion to avoid double device ID resolution

src/helpers/deviceId.ts
```diff
@@ -0,0 +1,46 @@
...
+ * If an error occurs during retrieval, the function returns "unknown".
+ *
+ * @example
+ * ```typescript
+ * const deviceId = await getDeviceIdForConnection();
+ * console.log(deviceId); // Outputs the device ID or "unknown" in case of failure
+ * ```
+ */
+export async function getDeviceIdForConnection(): Promise<string> {
+    const controller = new AbortController();
```
nirinchev: This restructuring means that we no longer abort the device id promise when the telemetry service is closed. We probably want to bring that back.

src/telemetry/telemetry.ts
blva: done! thanks

src/helpers/connectionOptions.ts
blva: I'm confused, the point of deviceID is to get the separate users that don't authenticate, so we can identify them as unique, in that sense I think we should always keep it. This was also requested so I'm inclined to always keep it

blva:
should be ready for review

src/helpers/connectionOptions.ts
blva: I refactored this quite a bit into a singleton so that we avoid double resolution, let me know what you think and if I'm missing any gaps!

src/helpers/connectionOptions.ts
gagik: essentially the idea there was to minimize visibility of this app name setting when they connect to i.e. localhost so for users they don't see identifiers appended to the app name they wouldn't expected but it's not a big deal to keep it as is yeah

src/common/logger.ts
```diff
@@ -30,8 +31,8 @@ export const LogId = {
     telemetryEmitStart: mongoLogId(1_002_003),
     telemetryEmitSuccess: mongoLogId(1_002_004),
     telemetryMetadataError: mongoLogId(1_002_005),
-    telemetryDeviceIdFailure: mongoLogId(1_002_006),
-    telemetryDeviceIdTimeout: mongoLogId(1_002_007),
+    deviceIdResolutionError: mongoLogId(1_002_006),
```
gagik, line 34: very nit: we can still keep the `telemetry` prefix since this is still what it's related to (and we have this prefix group pattern right now)

src/helpers/deviceId.ts
```diff
@@ -0,0 +1,170 @@
+import { getDeviceId } from "@mongodb-js/device-id";
+import nodeMachineId from "node-machine-id";
+import { LogId, LoggerBase } from "../common/logger.js";
+
+export const DEVICE_ID_TIMEOUT = 3000;
```
gagik, line 5: this makes me realize it'd probably be good to export the device ID timeout default constant from `@mongodb-js/device-id`

src/helpers/deviceId.ts
```diff
@@ -0,0 +1,170 @@
...
+     * @param logger - The logger instance to use
+     * @returns The DeviceIdService instance
+     */
+    public static init(logger: LoggerBase, timeout?: number): DeviceIdService {
+        if (DeviceIdService.instance) {
+            return DeviceIdService.instance;
+        }
+        DeviceIdService.instance = new DeviceIdService(logger, timeout ?? DEVICE_ID_TIMEOUT);
+        return DeviceIdService.instance;
+    }
```
gagik: I generally prefer to avoid side-effects in constructors i.e. `this.startDeviceIdCalculation()`
Since we already have an `init` pattern, we could make this constructor private and have one `create(...)` will also call the `startDeviceIdCalculation();` after creating it.

src/helpers/deviceId.ts
```diff
@@ -0,0 +1,170 @@
...
+    /**
+     * Gets the singleton instance of DeviceIdService.
+     * @returns The DeviceIdService instance
+     */
+    public static getInstance(): DeviceIdService {
+        if (!DeviceIdService.instance) {
+            throw new Error("DeviceIdService not initialized");
+        }
+        return DeviceIdService.instance;
+    }
```
gagik: I think this singleton instance pattern discourages explicit dependency injection which makes it too easy to end up accessing this service prematurely, thus needing these checks to make sure it's initialized. I get it from perspective of only doing the resolution once so we could still keep the static instance as far as `create` goes and maybe throw an error if someone tries to re-create the service.

But other than that, the `getInstance` is only used in tests right now and I think for that it'd have been preferable to create the `DeviceIdService` inside the test and pass it to whatever is being tested instead

src/helpers/deviceId.ts
gagik: Telemetry service is example of that https://github.com/mongodb-js/mongodb-mcp-server/blob/0e28212a1d2371bb64f4442224c324e503ba573f/src/telemetry/telemetry.ts#L37

src/common/logger.ts
blva: but now resolution could fail upon connection tool since this will now be used in two places, no?

src/helpers/deviceId.ts
blva: yep, I think I didn't find a constant to export here - https://github.com/mongodb-js/devtools-shared/blob/a42be86e39d5887801d175c9485ca0ced9e4f943/packages/device-id/src/get-device-id.ts#L6

so I decided to set the timeout ourselves to make it easy to test timeout and have flexibility to change in the future if we need.

gagik: agree, was more of a note to self to follow up inside `@mongodb-js/device-id` later

blva: oh got it! yes keep them coming - appreciated 🚀

src/helpers/deviceId.ts
blva: cool! I think there's no harm in clarifying this one, but bundling it with your comment below are yo suggesting to do something like telemetry but still maintain a singleton? or just use the same approach and inject the `logger` object on `getDeviceId` calls?

gagik: either is fine I think, we can also have a late setter for the logger

src/helpers/deviceId.ts
```diff
@@ -0,0 +1,137 @@
...
+
+    /**
+     * Sets up the device ID calculation promise and abort controller.
+     */
+    private setup(): void {
+        this.abortController = new AbortController();
+        this.deviceIdPromise = this.calculateDeviceId();
+        // start the promise upon setup
+        void this.deviceIdPromise;
```
gagik: the promise is already "started" from the function being called, this doesn't do anything

src/helpers/deviceId.ts
```diff
@@ -0,0 +1,137 @@
...
+        this.instance = instance;
+
+        return instance;
+    }
+
+    /**
+     * Sets up the device ID calculation promise and abort controller.
+     */
+    private setup(): void {
+        this.abortController = new AbortController();
```
gagik: it'd be good to have the abortcontroller defined inside the `calculateDeviceId` as it's directly tied to it

src/helpers/deviceId.ts
```diff
@@ -0,0 +1,137 @@
...
+    private static instance: DeviceId | undefined = undefined;
+
+    private constructor(logger: LoggerBase, timeout: number = DEVICE_ID_TIMEOUT) {
+        this.logger = logger;
+        this.timeout = timeout;
+        this.getMachineId = (): Promise<string> => nodeMachineId.machineId(true);
+    }
+
+    public static create(logger: LoggerBase, timeout?: number): DeviceId {
+        if (this.instance) {
```
gagik, line 23: nice! yeah this is what I meant. more radically, we can also throw an error to enforce that this creation is only done ever

src/helpers/deviceId.ts
```diff
@@ -0,0 +1,137 @@
...
+    /**
+     * Internal method that performs the actual device ID calculation.
+     */
+    private async calculateDeviceId(): Promise<string> {
+        if (!this.abortController) {
+            throw new Error("Device ID calculation not started");
+        }
+
+        try {
+            const deviceId = await getDeviceId({
```
gagik: `getDeviceId` already does plenty of error handling, including the abort error. I don't think this `try/catch` will ever catch anything

src/telemetry/telemetry.ts
```diff
@@ -57,46 +53,17 @@ export class Telemetry {
...
+
+        this.commonProperties.device_id = deviceIdValue;
         this.commonProperties.is_container_env = containerEnv;
 
         this.isBufferingEvents = false;
     }
 
     public async close(): Promise<void> {
-        this.deviceIdAbortController.abort();
         this.isBufferingEvents = false;
```
gagik, line 66: would `this.deviceId.close` make sense here?

tests/unit/helpers/deviceId.test.ts
```diff
@@ -0,0 +1,160 @@
...
+
+    it("should allow aborting calculation", async () => {
+        MockGetDeviceId.mockImplementation((options) => {
+            // Simulate a long-running operation that can be aborted
+            return new Promise((resolve, reject) => {
+                const timeout = setTimeout(() => resolve("device-id"), 1000);
+                options.abortSignal?.addEventListener("abort", () => {
+                    clearTimeout(timeout);
+                    const abortError = new Error("Aborted");
+                    abortError.name = "AbortError";
```
gagik, line 67: is this a real error that can get thrown? pretty sure this is handled by the `getDeviceId` function's try/catch so would never reach our own code.

src/helpers/deviceId.ts
blva: does it automatically assign unknown upon error? If so we can remove

tests/unit/helpers/deviceId.test.ts
blva: agreed - this is an autogen test that I missed to remove

src/telemetry/telemetry.ts
blva: I moved it to index.ts so I'm closing it there as well, makes sense?

src/helpers/deviceId.ts
blva: cool! in that case I guess we'd make `get()` static so that anyone can get the device Id

gagik (approved):
Other than above, changes look good to me 🙂 feel free to go ahead as you see fit.

src/helpers/deviceId.ts
blva: note: chatted offline to actually just inject it

src/telemetry/telemetry.ts
blva: now moved to transport

src/helpers/deviceId.ts
```diff
@@ -0,0 +1,113 @@
...
+
+    private constructor(logger: LoggerBase, timeout: number = DEVICE_ID_TIMEOUT) {
+        this.logger = logger;
+        this.timeout = timeout;
+        this.getMachineId = (): Promise<string> => nodeMachineId.machineId(true);
+    }
+
+    public static create(logger: LoggerBase, timeout?: number): DeviceId {
+        if (this.instance) {
+            throw new Error("DeviceId instance already exists, use get() to retrieve the device ID");
```
gagik, line 24: if we're injecting it all the way now, we can get rid of the static `get` and just use the instance mechanism as a way of enforcing that we do not end up creating 2 device ID instances ever.
So I would throw here but just say that one should be created per program and get rid of `static get` which is currently only being used in tests.

gagik (approved):
Looks great, just one comment about some remainders of static-ness

src/helpers/deviceId.ts
gagik: nevermind, misread parts of the code

PR #413, opened 2025-07-30
chore: Reactive resource support for debugging connectivity MCP-80

kmruiz:
## Proposed changes
Resources can listen to events on the session and their arguments, and reduces the current state with the argument to provide a new state.

This way this is type safe and declarative, mimicking how tools work.

## Checklist

### DIFF
--- a/src/common/session.ts
+++ b/src/common/session.ts
@@ -14,7 +14,7 @@ export interface SessionOptions {
 }
 
 export type SessionEvents = {
-    connected: [];
+    connect: [];
     close: [];
     disconnect: [];
     "connection-error": [string];
@@ -130,6 +130,6 @@ export class Session extends EventEmitter<SessionEvents> {
             throw error;
         }
 
-        this.emit("connected");
+        this.emit("connect");
     }
 }
--- a/src/common/session.ts
+++ b/src/common/session.ts
@@ -17,6 +17,7 @@ export type SessionEvents = {
     connected: [];
     close: [];
     disconnect: [];
+    "connection-error": [string];
 };
 
 export class Session extends EventEmitter<SessionEvents> {
@@ -105,20 +106,27 @@ export class Session extends EventEmitter<SessionEvents> {
             connectionString,
             defaultAppName: `${packageInfo.mcpServerName} ${packageInfo.version}`,
         });
-        this.serviceProvider = await NodeDriverServiceProvider.connect(connectionString, {
-            productDocsLink: "https://github.com/mongodb-js/mongodb-mcp-server/",
-            productName: "MongoDB MCP",
-            readConcern: {
-                level: connectOptions.readConcern,
-            },
-            readPreference: connectOptions.readPreference,
-            writeConcern: {
-                w: connectOptions.writeConcern,
-            },
-            timeoutMS: connectOptions.timeoutMS,
-            proxy: { useEnvironmentVariableProxies: true },
-            applyProxyToOIDC: true,
-        });
+
+        try {
+            this.serviceProvider = await NodeDriverServiceProvider.connect(connectionString, {
+                productDocsLink: "https://github.com/mongodb-js/mongodb-mcp-server/",
+                productName: "MongoDB MCP",
+                readConcern: {
+                    level: connectOptions.readConcern,
+                },
+                readPreference: connectOptions.readPreference,
+                writeConcern: {
+                    w: connectOptions.writeConcern,
+                },
+                timeoutMS: connectOptions.timeoutMS,
+                proxy: { useEnvironmentVariableProxies: true },
+                applyProxyToOIDC: true,
+            });
+        } catch (error: unknown) {
+            const message = error instanceof Error ? error.message : `${error}`;
+            this.emit("connection-error", message);
+            throw error;
+        }
 
         this.emit("connected");
     }
--- a/src/common/session.ts
+++ b/src/common/session.ts
@@ -122,6 +122,8 @@ export class Session extends EventEmitter<SessionEvents> {
                 proxy: { useEnvironmentVariableProxies: true },
                 applyProxyToOIDC: true,
             });
+
+            await this.serviceProvider?.runCommand?.("admin", { hello: 1 });
         } catch (error: unknown) {
             const message = error instanceof Error ? error.message : `${error}`;
             this.emit("connection-error", message);
--- a/src/common/session.ts
+++ b/src/common/session.ts
@@ -13,10 +13,13 @@ export interface SessionOptions {
     apiClientSecret?: string;
 }
 
-export class Session extends EventEmitter<{
+export type SessionEvents = {
+    connected: [];
     close: [];
     disconnect: [];
-}> {
+};
+
+export class Session extends EventEmitter<SessionEvents> {
     sessionId?: string;
     serviceProvider?: NodeDriverServiceProvider;
     apiClient: ApiClient;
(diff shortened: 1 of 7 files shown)


src/resources/common/config.ts
```diff
@@ -0,0 +1,40 @@
...
+            description:
+                "Server configuration, supplied by the user either as environment variables or as startup arguments",
+        },
+    },
+    {
+        initial: { ...config },
+        events: [],
+    }
+) {
+    reduce(previous: UserConfig, eventName: undefined, event: undefined): UserConfig {
```
kmruiz: eventName and event are undefined because we don't listen to any event here. So essentially it's static, as right now.

src/resources/common/debug.ts
```diff
@@ -0,0 +1,50 @@
...
+    },
+    {
+        initial: { tag: "disconnected" },
+        events: ["connected", "disconnect", "close"],
+    }
+) {
+    reduce(
+        previous: ConnectionStateDebuggingInformation,
+        eventName: "connected" | "disconnect" | "close",
+        event: undefined
```
kmruiz: None of our events have parameters, so this will be always undefined (it will change in the future, as according to the TD, we will have the connection status here)

src/resources/common/debug.ts
```diff
@@ -0,0 +1,60 @@
...
+    readonly connectionStringAuthType?: "scram" | "ldap" | "kerberos" | "oidc-auth-flow" | "oidc-device-flow" | "x.509";
+    readonly oidcLoginUrl?: string;
+    readonly oidcUserCode?: string;
+    readonly errorReason?: string;
+};
+
+export class DebugResource extends ReactiveResource(
+    {
+        name: "debug",
+        uri: "config://debug",
```
himanshusinghs: nit: I don't think debug is related to config itself. Perhaps we can expose it as a dedicated debug uri with this resource being debug://connection-errors?

src/resources/resource.ts
```diff
@@ -0,0 +1,80 @@
...
+import { UserConfig } from "../common/config.js";
+import { Telemetry } from "../telemetry/telemetry.js";
+import type { SessionEvents } from "../common/session.js";
+import { ReadResourceCallback, RegisteredResource, ResourceMetadata } from "@modelcontextprotocol/sdk/server/mcp.js";
+
+type PayloadOf<K extends keyof SessionEvents> = SessionEvents[K][0];
+
+type ResourceConfiguration = { name: string; uri: string; config: ResourceMetadata };
+
+export function ReactiveResource<V, KE extends readonly (keyof SessionEvents)[]>(
```
himanshusinghs:
```suggestion
export function ReactiveResource<InitialValue, SessionEvent extends readonly (keyof SessionEvents)[]>(
```
nit: just to make it a bit verbose and easier to read in one go :)

src/resources/resource.ts
```diff
@@ -0,0 +1,80 @@
...
+                    text: this.toOutput(),
+                    mimeType: "application/json",
+                    uri: uri.href,
+                },
+            ],
+        });
+
+        private triggerUpdate() {
+            this.registeredResource?.update({});
+            this.server.mcpServer.sendResourceListChanged();
```
himanshusinghs: Two concerns on these two lines:
1. `this.registeredResource?.update({});` should probably be changed to 
`server.notification({ ...correct notification about the changed resource })`
2. I don't think we always need to send the resource list changed notification. I don't know if this notification action is heavy or not but I suppose we should not assume its going to be light unless we are sure of it.

src/common/session.ts
```diff
@@ -102,19 +106,30 @@ export class Session extends EventEmitter<{
...
+                readPreference: connectOptions.readPreference,
+                writeConcern: {
+                    w: connectOptions.writeConcern,
+                },
+                timeoutMS: connectOptions.timeoutMS,
+                proxy: { useEnvironmentVariableProxies: true },
+                applyProxyToOIDC: true,
+            });
+
+            await this.serviceProvider?.runCommand?.("admin", { hello: 1 });
```
kmruiz, line 126: NodeDriverServiceProvider.connect, despite it's name, does not in fact connect. This behaviour is likely inherited from the driver.

By forcing a ping (by using hello) we force a connection and validate that it's in fact correct.

src/common/session.ts
```diff
@@ -13,10 +13,14 @@ export interface SessionOptions {
     apiClientSecret?: string;
 }
 
-export class Session extends EventEmitter<{
+export type SessionEvents = {
+    connected: [];
```
fmenezes: nit: to keep consistent with disconnect
```suggestion
    connect: [];
```

src/resources/resource.ts
```diff
@@ -0,0 +1,83 @@
...
+import { Telemetry } from "../telemetry/telemetry.js";
+import type { SessionEvents } from "../common/session.js";
+import { ReadResourceCallback, ResourceMetadata } from "@modelcontextprotocol/sdk/server/mcp.js";
+import logger, { LogId } from "../common/logger.js";
+
+type PayloadOf<K extends keyof SessionEvents> = SessionEvents[K][0];
+
+type ResourceConfiguration = { name: string; uri: string; config: ResourceMetadata };
+
+export function ReactiveResource<Value, RelevantEvents extends readonly (keyof SessionEvents)[]>(
```
fmenezes, line 13: [q] any reason to use the function returning class pattern instead of using a class straight away?

```
export class ReactiveResource<...>
```

kmruiz: My understanding is that resource metadata is immutable and by using this pattern I feel it's more convenient to specify the configuration once and still be type-safe. I think using normal classes where everything is specified in properties makes more sense when you want to change the state. With this pattern, it's not possible, as the metadata is hidden in the closure.

src/common/session.ts
fmenezes: we have a ping on atlas cluster connect tool where we added a ping, should we drop it?

src/resources/resource.ts
```diff
@@ -0,0 +1,83 @@
...
+            ],
+        });
+
+        private async triggerUpdate() {
+            try {
+                await this.server.mcpServer.server.sendResourceUpdated({ uri });
+                this.server.mcpServer.sendResourceListChanged();
+            } catch (error: unknown) {
+                logger.warning(
+                    LogId.serverClosed,
```
himanshusinghs, line 67: This most likely needs a new log id?

src/resources/common/config.ts
```diff
@@ -0,0 +1,42 @@
...
+        },
+    },
+    {
+        initial: { ...config },
+        events: [],
+    }
+) {
+    reduce(eventName: undefined, event: undefined): UserConfig {
+        void eventName;
+        void event;
```
himanshusinghs, line 21: This could possibly be avoided by having reduce without any params?

src/resources/common/debug.ts
```diff
@@ -0,0 +1,60 @@
...
+    {
+        initial: { tag: "disconnected" } as ConnectionStateDebuggingInformation,
+        events: ["connect", "disconnect", "close", "connection-error"],
+    }
+) {
+    reduce(
+        eventName: "connect" | "disconnect" | "close" | "connection-error",
+        event: string | undefined
+    ): ConnectionStateDebuggingInformation {
+        void event;
```
himanshusinghs, line 28: Is this necessary anymore, now that you're using `event` in `connection-error` case?

src/common/session.ts
kmruiz: I see that it's used for answering the client with the connection status. I think we can keep it for now and then when we remove this logic to the new ConnectionManager (Tech Design) we can refactor and use the same code for both connections. 

How do you feel about this strategy?

PR #420, opened 2025-08-04
chore: allow logging unredacted messages

nirinchev:
## Proposed changes

This is a refactoring of the logger interface to allow consumers to opt out of the automatic redacting logic. It enables a blanket opt out or opting out for specific logger type(s).

This should only be used where the messages are safe to log or where not logging the unredacted messages makes the logs meaningless. The impetus for this change is logging by the HTTP transport, which would otherwise log things like `Server started on http://<ip-address>:3000`.

## Checklist

### DIFF
--- a/src/common/logger.ts
+++ b/src/common/logger.ts
@@ -50,44 +50,95 @@ export const LogId = {
     streamableHttpTransportCloseFailure: mongoLogId(1_006_006),
 } as const;
 
+interface LogPayload {
+    id: MongoLogId;
+    context: string;
+    message: string;
+    noRedaction?: boolean | LoggerType | LoggerType[];
+}
+
+export type LoggerType = "console" | "disk" | "mcp";
+
 export abstract class LoggerBase {
-    abstract log(level: LogLevel, id: MongoLogId, context: string, message: string): void;
+    abstract log(level: LogLevel, payload: Omit<LogPayload, "noRedaction">): void;
 
-    info(id: MongoLogId, context: string, message: string): void {
-        this.log("info", id, context, message);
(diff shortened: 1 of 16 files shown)


nirinchev:
The coverage reduction is just a result of reformatting of the logger calls (multiline vs single line).

src/common/logger.ts
```diff
@@ -50,44 +50,96 @@ export const LogId = {
     streamableHttpTransportCloseFailure: mongoLogId(1_006_006),
 } as const;
 
+interface LogPayload {
+    id: MongoLogId;
+    context: string;
+    message: string;
+    noRedaction?: boolean | LoggerType | LoggerType[];
```
blva, line 57: nit: i find `no` variables difficult to read and prone to difficult statements over the code. Not super used to a single variable having so many different types too. WDYT about the below?  
```suggestion
    redaction?: {
        enabled: boolean;
        exceptFor?: LoggerType[];
    };
```

nirinchev: Agree - I also tend to prefer positively-named variables - in this case, I chose the negative because I want all messages to be redacted by default. If we had this as `redaction`/`redacted`, the implication would be that if we don't explicitly specify `redaction: true`, messages would be unredacted.

blva: good point, I threw this into an LLM and got another suggestion
```typescript
 redaction?: {
        mode: "auto" | "force" | "skip";
        skipFor?: LoggerType[];
    };
```

but I think it still adds another complexity type w/ the modes, so feel free to ignore since I don't have a strong opinion

blva (approved):

PR #423, opened 2025-08-05
chore: refactor connections to use the new ConnectionManager to isolate long running processes like OIDC connections MCP-81

kmruiz:
## Proposed changes

This PR basically proposes consolidating all connections into a single class, ConnectionManager. After this PR, both Atlas tools and MongoDB tools use the same code to connect to MongoDB, independently of the connection type.

Also, the ConnectionManager gets an upgrade on how to handle errors and debugging information:

* The ConnectionManager keeps track of the latest state and emits events when the connection status changes.
* There is a new debug resource that contains information about the last connection error.
* The ConnectionManager infers the authentication type, to provide more information to the user when something goes wrong.
* The ConnectionManager implements the non-terminal "connecting" state for OIDC flows.

To avoid changing the whole architecture and _maybe_ support multitenancy in the future, the ConnectionManager is hidden and isolated in the Session.

* Tools still use the session, who acts as a proxy between the ConnectionManager and the tool.
* Tools still work with the service provider through the Session, and the session ensures the connection is active.

This PR is bigger than usual because the refactor touches some root infrastructure, but hopefully the unit and integration tests I've added, along with the manual testing, ensures everything works correctly.

## Checklist

### DIFF
--- a/src/common/connectionManager.ts
+++ b/src/common/connectionManager.ts
@@ -21,7 +21,7 @@ export interface ConnectionSettings extends ConnectOptions {
 
 type ConnectionTag = "connected" | "connecting" | "disconnected" | "errored";
 type OIDCConnectionAuthType = "oidc-auth-flow" | "oidc-device-flow";
-type ConnectionStringAuthType = "scram" | "ldap" | "kerberos" | OIDCConnectionAuthType | "x.509";
+export type ConnectionStringAuthType = "scram" | "ldap" | "kerberos" | OIDCConnectionAuthType | "x.509";
 
 export interface ConnectionState {
     tag: ConnectionTag;
@@ -114,7 +114,7 @@ export class ConnectionManager extends EventEmitter<ConnectionManagerEvents> {
                 tag: "connected",
                 connectedAtlasCluster: settings.atlas,
                 serviceProvider,
-                connectionStringAuthType: this.inferConnectionTypeFromSettings(settings),
+                connectionStringAuthType: ConnectionManager.inferConnectionTypeFromSettings(settings),
             });
         } catch (error: unknown) {
             const errorReason = error instanceof Error ? error.message : `${error as string}`;
@@ -154,7 +154,7 @@ export class ConnectionManager extends EventEmitter<ConnectionManagerEvents> {
         return newState;
     }
 
-    private inferConnectionTypeFromSettings(settings: ConnectionSettings): ConnectionStringAuthType {
+    static inferConnectionTypeFromSettings(settings: ConnectionSettings): ConnectionStringAuthType {
         const connString = new ConnectionString(settings.connectionString);
         const searchParams = connString.typedSearchParams<MongoClientOptions>();
 
--- a/src/common/connectionManager.ts
+++ b/src/common/connectionManager.ts
@@ -1,0 +1,179 @@
+import { ConnectOptions } from "./config.js";
+import { NodeDriverServiceProvider } from "@mongosh/service-provider-node-driver";
+import EventEmitter from "events";
+import { setAppNameParamIfMissing } from "../helpers/connectionOptions.js";
+import { packageInfo } from "./packageInfo.js";
+import ConnectionString from "mongodb-connection-string-url";
+import { MongoClientOptions } from "mongodb";
+import { ErrorCodes, MongoDBError } from "./errors.js";
+
+export interface AtlasClusterConnectionInfo {
+    username: string;
+    projectId: string;
+    clusterName: string;
+    expiryDate: Date;
+}
+
+export interface ConnectionSettings extends ConnectOptions {
+    connectionString: string;
+    atlas?: AtlasClusterConnectionInfo;
+}
+
+type ConnectionTag = "connected" | "connecting" | "disconnected" | "errored";
+type OIDCConnectionAuthType = "oidc-auth-flow" | "oidc-device-flow";
+type ConnectionStringAuthType = "scram" | "ldap" | "kerberos" | OIDCConnectionAuthType | "x.509";
+
+export interface ConnectionState {
+    tag: ConnectionTag;
+    connectionStringAuthType?: ConnectionStringAuthType;
+}
+
+export interface ConnectionStateConnected extends ConnectionState {
+    tag: "connected";
+    serviceProvider: NodeDriverServiceProvider;
+    connectedAtlasCluster?: AtlasClusterConnectionInfo;
+}
+
+export interface ConnectionStateConnecting extends ConnectionState {
+    tag: "connecting";
+    serviceProvider: NodeDriverServiceProvider;
+    oidcConnectionType: OIDCConnectionAuthType;
+    oidcLoginUrl?: string;
+    oidcUserCode?: string;
+}
+
+export interface ConnectionStateDisconnected extends ConnectionState {
+    tag: "disconnected";
+}
+
+export interface ConnectionStateErrored extends ConnectionState {
+    tag: "errored";
+    errorReason: string;
+}
+
+export type AnyConnectionState =
+    | ConnectionState
+    | ConnectionStateConnected
+    | ConnectionStateConnecting
+    | ConnectionStateDisconnected
+    | ConnectionStateErrored;
+
+export interface ConnectionManagerEvents {
+    "connection-requested": [AnyConnectionState];
+    "connection-succeeded": [ConnectionStateConnected];
+    "connection-timed-out": [ConnectionStateErrored];
+    "connection-closed": [ConnectionStateDisconnected];
+    "connection-errored": [ConnectionStateErrored];
+}
+
+export class ConnectionManager extends EventEmitter<ConnectionManagerEvents> {
+    private state: AnyConnectionState;
+
+    constructor() {
+        super();
+        this.state = { tag: "disconnected" };
+    }
+
+    async connect(settings: ConnectionSettings): Promise<AnyConnectionState> {
+        if (this.state.tag == "connected" || this.state.tag == "connecting") {
+            await this.disconnect();
+        }
+
+        let serviceProvider: NodeDriverServiceProvider;
+        try {
+            settings = { ...settings };
+            settings.connectionString = setAppNameParamIfMissing({
+                connectionString: settings.connectionString,
+                defaultAppName: `${packageInfo.mcpServerName} ${packageInfo.version}`,
+            });
+
+            serviceProvider = await NodeDriverServiceProvider.connect(settings.connectionString, {
+                productDocsLink: "https://github.com/mongodb-js/mongodb-mcp-server/",
+                productName: "MongoDB MCP",
+                readConcern: {
+                    level: settings.readConcern,
+                },
+                readPreference: settings.readPreference,
+                writeConcern: {
+                    w: settings.writeConcern,
+                },
+                timeoutMS: settings.timeoutMS,
+                proxy: { useEnvironmentVariableProxies: true },
+                applyProxyToOIDC: true,
+            });
+        } catch (error: unknown) {
+            const errorReason = error instanceof Error ? error.message : `${error as string}`;
+            this.changeState("connection-errored", { tag: "errored", errorReason });
+            throw new MongoDBError(ErrorCodes.MisconfiguredConnectionString, errorReason);
+        }
+
+        try {
+            await serviceProvider?.runCommand?.("admin", { hello: 1 });
+
+            return this.changeState("connection-succeeded", {
+                tag: "connected",
+                connectedAtlasCluster: settings.atlas,
+                serviceProvider,
+                connectionStringAuthType: this.inferConnectionTypeFromSettings(settings),
+            });
+        } catch (error: unknown) {
+            const errorReason = error instanceof Error ? error.message : `${error as string}`;
+            this.changeState("connection-errored", { tag: "errored", errorReason });
+            throw new MongoDBError(ErrorCodes.NotConnectedToMongoDB, errorReason);
+        }
+    }
+
+    async disconnect(): Promise<ConnectionStateDisconnected | ConnectionStateErrored> {
+        if (this.state.tag == "disconnected") {
+            return this.state as ConnectionStateDisconnected;
+        }
+
+        if (this.state.tag == "errored") {
+            return this.state as ConnectionStateErrored;
+        }
+
+        if (this.state.tag == "connected" || this.state.tag == "connecting") {
+            const state = this.state as ConnectionStateConnecting | ConnectionStateConnected;
+            try {
+                await state.serviceProvider?.close(true);
+            } finally {
+                this.changeState("connection-closed", { tag: "disconnected" });
+            }
+        }
+
+        return { tag: "disconnected" };
+    }
+
+    get currentConnectionState(): AnyConnectionState {
+        return this.state;
+    }
+
+    changeState<State extends AnyConnectionState>(event: keyof ConnectionManagerEvents, newState: State): State {
+        this.state = newState;
+        this.emit(event, newState);
+        return newState;
+    }
+
+    private inferConnectionTypeFromSettings(settings: ConnectionSettings): ConnectionStringAuthType {
+        const connString = new ConnectionString(settings.connectionString);
+        const searchParams = connString.typedSearchParams<MongoClientOptions>();
+
+        switch (searchParams.get("authMechanism")) {
+            case "MONGODB-OIDC": {
+                return "oidc-auth-flow"; // TODO: depending on if we don't have a --browser later it can be oidc-device-flow
+            }
+            case "MONGODB-X509":
+                return "x.509";
+            case "GSSAPI":
+                return "kerberos";
+            case "PLAIN":
+                if (searchParams.get("authSource") == "$external") {
+                    return "ldap";
+                }
+                break;
+            default:
+                return "scram";
+        }
+        return "scram";
+    }
+}
--- a/src/common/session.ts
+++ b/src/common/session.ts
@@ -1,16 +1,21 @@
-import { NodeDriverServiceProvider } from "@mongosh/service-provider-node-driver";
 import { ApiClient, ApiClientCredentials } from "./atlas/apiClient.js";
 import { Implementation } from "@modelcontextprotocol/sdk/types.js";
 import logger, { LogId } from "./logger.js";
 import EventEmitter from "events";
-import { ConnectOptions } from "./config.js";
-import { setAppNameParamIfMissing } from "../helpers/connectionOptions.js";
-import { packageInfo } from "./packageInfo.js";
+import {
+    AnyConnectionState,
+    ConnectionManager,
+    ConnectionSettings,
+    ConnectionStateConnected,
+} from "./connectionManager.js";
+import { NodeDriverServiceProvider } from "@mongosh/service-provider-node-driver";
+import { ErrorCodes, MongoDBError } from "./errors.js";
 
 export interface SessionOptions {
     apiBaseUrl: string;
     apiClientId?: string;
     apiClientSecret?: string;
+    connectionManager?: ConnectionManager;
 }
 
 export type SessionEvents = {
@@ -22,7 +27,7 @@ export type SessionEvents = {
 
 export class Session extends EventEmitter<SessionEvents> {
     sessionId?: string;
-    serviceProvider?: NodeDriverServiceProvider;
+    connectionManager: ConnectionManager;
     apiClient: ApiClient;
     agentRunner?: {
         name: string;
@@ -35,7 +40,7 @@ export class Session extends EventEmitter<SessionEvents> {
         expiryDate: Date;
     };
 
-    constructor({ apiBaseUrl, apiClientId, apiClientSecret }: SessionOptions) {
+    constructor({ apiBaseUrl, apiClientId, apiClientSecret, connectionManager }: SessionOptions) {
         super();
 
         const credentials: ApiClientCredentials | undefined =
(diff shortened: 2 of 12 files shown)


src/common/connectionManager.ts
```diff
@@ -0,0 +1,179 @@
...
+                    w: settings.writeConcern,
+                },
+                timeoutMS: settings.timeoutMS,
+                proxy: { useEnvironmentVariableProxies: true },
+                applyProxyToOIDC: true,
+            });
+        } catch (error: unknown) {
+            const errorReason = error instanceof Error ? error.message : `${error as string}`;
+            this.changeState("connection-errored", { tag: "errored", errorReason });
+            throw new MongoDBError(ErrorCodes.MisconfiguredConnectionString, errorReason);
```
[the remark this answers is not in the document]
kmruiz: In this try-catch the only validation is done is that the connection string and configuration is valid. Actual connectivity happens in the ping done with runCommand below.

src/common/connectionManager.ts
```diff
@@ -0,0 +1,184 @@
...
+    tag: "disconnected";
+}
+
+export interface ConnectionStateErrored extends ConnectionState {
+    tag: "errored";
+    errorReason: string;
+}
+
+export type AnyConnectionState =
+    | ConnectionState
```
nirinchev:
```suggestion
```

I think we can remove this which will clean up a bunch of type assertions later.

src/common/connectionManager.ts
```diff
@@ -0,0 +1,184 @@
...
+        }
+
+        return { tag: "disconnected" };
+    }
+
+    get currentConnectionState(): AnyConnectionState {
+        return this.state;
+    }
+
+    changeState<State extends AnyConnectionState>(event: keyof ConnectionManagerEvents, newState: State): State {
```
nirinchev:
```suggestion
    changeState(event: keyof ConnectionManagerEvents, newState: AnyConnectionState): AnyConnectionState {
```

We're not using the generic for anything it seems.

nirinchev: It seems like `event` and `newState` here are not independent - i.e. each event type has a corresponding state, in which case we should investigate if we can more cleanly represent this.

src/common/connectionManager.ts
```diff
@@ -0,0 +1,184 @@
...
+            }
+            case "MONGODB-X509":
+                return "x.509";
+            case "GSSAPI":
+                return "kerberos";
+            case "PLAIN":
+                if (searchParams.get("authSource") == "$external") {
+                    return "ldap";
+                }
+                break;
```
nirinchev: [nit] probably best to just `return "scram"` here rather than let it leave the switch statement.

src/common/connectionManager.ts
```diff
@@ -0,0 +1,184 @@
...
+
+    constructor() {
+        super();
+        this.state = { tag: "disconnected" };
+    }
+
+    async connect(settings: ConnectionSettings): Promise<AnyConnectionState> {
+        this.emit("connection-requested", this.state);
+
+        if (this.state.tag == "connected" || this.state.tag == "connecting") {
```
nirinchev: https://github.com/mongodb-js/mongodb-mcp-server/pull/426 😛

src/common/connectionManager.ts
```diff
@@ -0,0 +1,184 @@
...
+    connectionStringAuthType?: ConnectionStringAuthType;
+}
+
+export interface ConnectionStateConnected extends ConnectionState {
+    tag: "connected";
+    serviceProvider: NodeDriverServiceProvider;
+    connectedAtlasCluster?: AtlasClusterConnectionInfo;
+}
+
+export interface ConnectionStateConnecting extends ConnectionState {
```
nirinchev, line 37: This is not used right - assuming it's prep work for the OIDC?

src/common/session.ts
```diff
@@ -92,44 +98,33 @@ export class Session extends EventEmitter<SessionEvents> {
                 });
             this.connectedAtlasCluster = undefined;
         }
-        this.emit("disconnect");
     }
 
     async close(): Promise<void> {
         await this.disconnect();
         await this.apiClient.close();
-        this.emit("close");
```
nirinchev, line 101: Why are we no longer emitting close here?

src/common/session.ts
```diff
@@ -92,44 +98,33 @@ export class Session extends EventEmitter<SessionEvents> {
...
     }
 
-    async connectToMongoDB(connectionString: string, connectOptions: ConnectOptions): Promise<void> {
-        connectionString = setAppNameParamIfMissing({
-            connectionString,
-            defaultAppName: `${packageInfo.mcpServerName} ${packageInfo.version}`,
-        });
-
+    async connectToMongoDB(settings: ConnectionSettings): Promise<AnyConnectionState> {
```
nirinchev: Any reason to return the connection state? It seems like none of the consumers actually care about it.

src/common/session.ts
```diff
@@ -92,44 +98,33 @@ export class Session extends EventEmitter<SessionEvents> {
...
+            return await this.connectionManager.connect({ ...settings });
         } catch (error: unknown) {
-            const message = error instanceof Error ? error.message : `${error as string}`;
+            const message = error instanceof Error ? error.message : (error as string);
             this.emit("connection-error", message);
             throw error;
         }
+    }
+
+    isConnectedToMongoDB(): boolean {
```
nirinchev:
```suggestion
    get isConnectedToMongoDB(): boolean {
```

src/common/session.ts
```diff
@@ -92,44 +98,33 @@ export class Session extends EventEmitter<SessionEvents> {
...
+        return this.connectionManager.currentConnectionState.tag === "connected";
+    }
+
+    get serviceProvider(): NodeDriverServiceProvider {
+        if (this.isConnectedToMongoDB()) {
+            const state = this.connectionManager.currentConnectionState as ConnectionStateConnected;
+            return state.serviceProvider;
+        }
```
nirinchev:
```suggestion
        const state = this.connectionManager.currentConnectionState;
        if (state.tag === "connected") {
            return state.serviceProvider;
        }
```

src/tools/mongodb/connect/connect.ts
```diff
@@ -82,7 +87,7 @@ export class ConnectTool extends MongoDBToolBase {
     }
 
     private updateMetadata(): void {
-        if (this.config.connectionString || this.session.serviceProvider) {
+        if (this.session.isConnectedToMongoDB()) {
```
nirinchev: This seems to be a slight change in behavior where if we have a connection string but haven't connected yet, we'd now advertise as "connect" instead of "switch-connection". I don't think this will have practical implications for scram auth, but I imagine could be the case for OIDC where we probably can't initiate the auth flow until a user tries to interact with the MCP client.

src/tools/mongodb/mongodbTool.ts
```diff
@@ -116,8 +117,8 @@ export abstract class MongoDBToolBase extends ToolBase {
         return super.handleError(error, args);
     }
 
-    protected connectToMongoDB(connectionString: string): Promise<void> {
-        return this.session.connectToMongoDB(connectionString, this.config.connectOptions);
+    protected connectToMongoDB(connectionString: string): Promise<AnyConnectionState> {
```
nirinchev: Similarly - this seems like something that can continue being `Promise<void>`

src/common/connectionManager.ts
```diff
@@ -0,0 +1,184 @@
+import { ConnectOptions } from "./config.js";
+import { NodeDriverServiceProvider } from "@mongosh/service-provider-node-driver";
+import EventEmitter from "events";
+import { setAppNameParamIfMissing } from "../helpers/connectionOptions.js";
+import { packageInfo } from "./packageInfo.js";
+import ConnectionString from "mongodb-connection-string-url";
+import { MongoClientOptions } from "mongodb";
+import { ErrorCodes, MongoDBError } from "./errors.js";
+
+export interface AtlasClusterConnectionInfo {
```
nirinchev, line 10: Seems like we need to migrate this info from the session to the connection manager. Right now `connectCluster` doesn't appear to be providing this info in the `connectToMongoDB` call and instead sets it on the session directly.

tests/integration/common/connectionManager.test.ts
```diff
@@ -0,0 +1,155 @@
...
+
+        it("should notify that the connection was requested", () => {
+            expect(connectionManagerSpies["connection-requested"]).toHaveBeenCalledOnce();
+        });
+
+        it("should notify that the connection was successful", () => {
+            expect(connectionManagerSpies["connection-succeeded"]).toHaveBeenCalledOnce();
+        });
+
+        describe("when disconnects", () => {
```
nirinchev, line 73: We probably want to validate that the state after disconnect is what we expect it

nirinchev: This also applies to the reconnect/error suites.

tests/integration/transports/stdio.test.ts
```diff
@@ -1,8 +1,16 @@
...
 import { StdioClientTransport } from "@modelcontextprotocol/sdk/client/stdio.js";
 import { Client } from "@modelcontextprotocol/sdk/client/index.js";
+import { describeWithMongoDB } from "../tools/mongodb/mongodbHelpers.js";
+
+describeWithMongoDB("StdioRunner", (integration) => {
+    beforeEach(() => {
+        integration.mcpServer().userConfig.connectionString = integration.connectionString();
+        integration.mcpServer().session.connectionManager.changeState("connection-succeeded", {
+            tag: "connected",
+        });
```
nirinchev: Why are these needed? I would expect that the env variable would be enough for this to happen automatically.

nirinchev (commented):
Looks like a solid change - I have mostly minor/code style comments, but also a few questions that may be a bit more involved.

tests/integration/transports/stdio.test.ts
kmruiz: You are right, this is a left over from a test I was doing. Removing this in a commit.

tests/integration/common/connectionManager.test.ts
kmruiz: Added them on the next commit.

src/common/connectionManager.ts
kmruiz: Yep, it's prep work for OIDC where we will need it.

src/common/connectionManager.ts
kmruiz: The triple equals always gets me :trollface:

src/resources/common/debug.ts
```diff
@@ -10,10 +10,10 @@ type ConnectionStateDebuggingInformation = {
 
 export class DebugResource extends ReactiveResource(
     {
-        name: "debug-mongodb-connectivity",
-        uri: "debug://mongodb-connectivity",
+        name: "debug-mongodb",
+        uri: "debug://mongodb",
         config: {
-            description: "Debugging information for connectivity issues.",
+            description: "Debugging information for MongoDB connectivity issues.",
```
himanshusinghs: nit: should we expand the description and include either:
- "last recorded" connectivity issue
- connectivity issue from last failed attempt

src/tools/atlas/connect/connectCluster.ts
```diff
@@ -40,7 +40,7 @@ export class ConnectClusterTool extends AtlasToolBase {
             return "connected-to-other-cluster";
         }
 
-        if (!this.session.serviceProvider) {
+        if (!this.session.isConnectedToMongoDB()) {
```
himanshusinghs: question: would it be better to be precise here and relate the `connecting` state to the actual `connecting` state on the ConnectionManager?

src/common/connectionManager.ts
kmruiz: Tried it, but eslint complains that the switch statement is not exhaustive.

src/tools/mongodb/connect/connect.ts
himanshusinghs: Isn't that already handled by isConnectedToMongoDB() which specifically looks for the `connected` state?

src/common/connectionManager.ts
kmruiz: Fixed!, I was working on making it more typesafe and forgot to commit the latest changes :(. Thanks for noticing out!

src/tools/mongodb/mongodbTool.ts
kmruiz: I usually prefer to return the latest state instead of void whenever possible, but I think it's fair to return Promise<void> too, so I'll do the change.

src/common/session.ts
kmruiz: We do, it's in the this.disconnect method.

src/common/session.ts
kmruiz: Just used to do it, I can remove it. No strong feelings about it.

src/common/session.ts
kmruiz: To clarify, the this.disconnect will close the connection through the connection manager, that will emit the event. This way we centralise all connection event emitting in the connection manger.

src/tools/mongodb/connect/connect.ts
kmruiz: You are right, will revert to have the or condition with the connectionString.

src/common/connectionManager.ts
kmruiz: Migrated! Thanks for noticing out!

src/server.ts
```diff
@@ -38,12 +38,15 @@ export class Server {
     }
 
     async connect(transport: Transport): Promise<void> {
+        // Resources are now reactive, so we register them ASAP so they can listen to events like
+        // connection events.
+        this.registerResources();
```
himanshusinghs, line 43: There is one problem I noticed with the reactive resource (which unfortunately I could not point out in the previous PR) and that's about how MCP clients work with it. We publish both `resource/list/changed` and `resource/updated` notifications but we don't advertise either in our capabilities (on line :46 below) and we should as recommended [here](https://modelcontextprotocol.io/specification/2025-06-18/server/resources#capabilities).

While some clients refresh their resources on `resource/list/changed` notification, not all of them do that. The best way is to make sure that we advertise resource change capability, implement resource subscription logic and that should allow client to consume the reactive updates. That's the only other missing piece but certainly not a blocker for this PR.

src/common/session.ts
nirinchev: The `close` event here is different from `disconnect` which is emitted in a roundabout way through the connection manager. It doesn't appear like this is used anywhere, but we should either emit it once we've disconnected and closed the api client like before, or remove it from the SessionEvents struct.

src/common/connectionManager.ts
nirinchev: Huh... seems like an eslint issue - probably try restarting it as having a `default` case should satisfy the exhaustiveness check. To be clear my suggestion is to replace the `break` statement on line 174 with `return "scram"` and remove the `return "scram"` statement on line 181.

src/common/session.ts
himanshusinghs: but it looks like the event name has changed then. On disconnect we emit `disconnect` event not close. Might wanna check if the dependents of this event are updated to the new event now.

kmruiz: I think I prefer to emit it, even it's just informational for now. I'm not a fan of "just in case we need it" but I prefer to be consistent on all the events emitted on the flow, and not leave things out.

I'll revert this change and add the close event here.

himanshusinghs (commented):
Looks great, I left minor suggestions but nothing blocking for the PR.

src/server.ts
himanshusinghs: I have added listChanged capability on [my PR](https://github.com/mongodb-js/mongodb-mcp-server/pull/424/files#diff-8a8ae07582c9d433ec8c2e5c4310ff8901e604f4965c5b90a49117ad46c47595L43). Apparently that alone does not require an implementation but the resource subscription requires a subscription handler.

src/common/session.ts
```diff
@@ -70,20 +64,24 @@ export class Session extends EventEmitter<SessionEvents> {
     }
 
     async disconnect(): Promise<void> {
+        const currentConnection = this.connectionManager.currentConnectionState;
+        const atlasCluster =
+            currentConnection.tag === "connected" ? currentConnection.connectedAtlasCluster : undefined;
```
nirinchev:
```suggestion
        const atlasCluster = this.connectedAtlasCluster;
```

src/server.ts
kmruiz: Added the capabilities to the list, it's straightforward so I'll do it now.

src/tools/atlas/connect/connectCluster.ts
kmruiz: Not sure I understand the question. What do you mean? isConnectedToMongoDB ensures that you can query MongoDB, and while "connecting" this is not true.

src/tools/atlas/connect/connectCluster.ts
```diff
@@ -61,7 +62,10 @@ export class ConnectClusterTool extends AtlasToolBase {
         }
     }
 
-    private async prepareClusterConnection(projectId: string, clusterName: string): Promise<string> {
+    private async prepareClusterConnection(
+        projectId: string,
+        clusterName: string
+    ): Promise<[string, AtlasClusterConnectionInfo]> {
```
nirinchev: [nit] I really dislike anonymous tuples - obviously here it's somewhat simple, but would still recommend returning objects with named fields instead.

tests/integration/tools/mongodb/connect/connect.test.ts
```diff
@@ -12,10 +12,10 @@ import { beforeEach, describe, expect, it } from "vitest";
...
     (integration) => {
-        beforeEach(() => {
-            integration.mcpServer().userConfig.connectionString = integration.connectionString();
-            integration.mcpServer().session.connectionManager.changeState("connection-succeeded", {
-                tag: "connected",
+        beforeEach(async () => {
+            await integration.mcpServer().session.connectToMongoDB({
+                connectionString: integration.connectionString(),
+                ...config.connectOptions,
             });
```
nirinchev: [q] Why is this change needed? Not fundamentally opposed to it, but it does change the setup for the test and want to make sure I undertand the reasoning.

tests/integration/tools/mongodb/connect/connect.test.ts
```diff
@@ -12,8 +12,11 @@ import { beforeEach, describe, expect, it } from "vitest";
 describeWithMongoDB(
     "SwitchConnection tool",
     (integration) => {
-        beforeEach(() => {
-            integration.mcpServer().userConfig.connectionString = integration.connectionString();
+        beforeEach(async () => {
+            await integration.mcpServer().session.connectToMongoDB({
+                connectionString: integration.connectionString(),
+                ...config.connectOptions,
+            });
```
nirinchev, line 19: [q] Why is this change needed? Not fundamentally opposed, just trying to understand what's prompting it.

nirinchev (commented):
Looks good - a couple of final nits from me.

tests/integration/tools/mongodb/connect/connect.test.ts
kmruiz: Before this PR, the test was assuming that we were connected because we had a connectionString, and now we really verify with a ping that we are properly connected. It's a real integration test now because there is a real connection.

src/tools/atlas/connect/connectCluster.ts
kmruiz: It's fair. Will do the change now. The types are pretty different the tuple is small so I'm not really concerned about it (also it's a private method, so it shouldn't leak) but I don't have a strong opinion and it's literally a 1min change.

src/tools/atlas/connect/connectCluster.ts
himanshusinghs: `isConnectedToMongoDB()` returns false for anything other than connected and here we have distinct `disconnected` and `connecting` states. So I was wondering if it would be better to link the states here to the actual connection state in the connection manager.

src/tools/atlas/connect/connectCluster.ts
```diff
@@ -40,7 +41,7 @@ export class ConnectClusterTool extends AtlasToolBase {
             return "connected-to-other-cluster";
         }
 
-        if (!this.session.serviceProvider) {
+        if (!this.session.isConnectedToMongoDB) {
```
fmenezes: @kmruiz I think this is not correct we should only return connecting if our connection is in progress, can we query `connectionManager.currentConnectionState()` instead?

kmruiz: I think this is related to @himanshusinghs comment before, which I didn't understand first. Yes, I think we can check the tag to ensure we are not connected yet but connecting.

himanshusinghs: yeap this is what I meant - sorry for the ambiguous comment earlier :)

src/tools/atlas/connect/connectCluster.ts
```diff
@@ -41,7 +41,7 @@ export class ConnectClusterTool extends AtlasToolBase {
             return "connected-to-other-cluster";
         }
 
-        if (!this.session.isConnectedToMongoDB) {
+        if (this.session.connectionManager.currentConnectionState.tag !== "connected") {
```
himanshusinghs: Shouldn't it instead check?

`if (this.session.connectionManager.currentConnectionState.tag === "connecting") {`

kmruiz: I don't think so, it could be errored because the user authentication is still not propagated to the data plane. We should keep attempting until it's connected or times out.

himanshusinghs: but this looks the same as earlier :) isConnectedToMongoDB does the same thing inverted we are now doing in the change.

kmruiz: Resolved on slack!

PR #424, opened 2025-08-05
feat: adds an export tool and exported-data resource MCP-16

himanshusinghs:
#### This PR introduces:
1. a new export tool that allows users to export MongoDB collection data or query results in JSON format
2. a templated resource called `exported-data://` to access the data exports created by export tool

#### JSON Export tool provides two formats for data export:
1. `relaxed`: This exports plain json format at the possible loss of some BSON types. Useful when the data is expected to be used in tools other than MongoDB such as BI tools or when simply having conversation with LLMs
2. `canonical`: This exports json with BSON types and is expected to be used when data might get imported back again in MongoDB.

#### JSON export tool allows exporting:
1. an entire namespace
2. query results for a `filter` (possibly with `limit`, `sort`, `projection`)

#### Export tool result contains:
1. a text representation of how to access the exported data via resource URI
2. a resource_link content part for the clients who are capable of accessing resource_link
3. a local path, only if the MCP server is running behind `stdio` transport to make it easier for editor clients to access the exported file.

Exported data temporarily resides on the host machine running the mongodb-mcp-server at the configured path (configuration exposed by flag `--exportPath` or environment variable `MDB_MCP_EXPORT_PATH`), until it gets cleaned up automatically either by:
- the autocleaner running every configured milliseconds (configuration exposed by flag `--exportCleanupIntervalMs` or environment variable `MDB_MCP_EXPORT_CLEANUP_INTERVAL_MS`)
- disconnect of the client

The expiry of exported data is controlled by flag `--exportTimeoutMs` or the environment variable `MDB_MCP_EXPORT_TIMEOUT_MS`

The exported data can be accessed via the resource URI template - `exported-data://{exportName}`. Because the exported data are written to non-guessable filenames, the autocomplete API is provided to easily autocomplete the matching file names.

## Checklist

- [x] JSON export tool should allow exporting BI friendly data
- [x] JSON export tool should allow exporting data with BSON types
- [x] JSON export tool should allow exporting entire namespace
- [x] JSON export tool should allow exporting query results
- [x] Exported data should be available for limited time and automatically cleaned up during a session and on session disconnect
- [x] Exported data should be made available via resources for stable and re-usable reference
- [x] Exported data should optionally be provided via local path when MCP server is connected to editor clients
- [x] Export tool covered by test cases
- [x] Exported resource covered by test cases
- [x] Export cleanup and generation should be covered by test cases

### DIFF
--- a/tests/integration/resources/exportedData.test.ts
+++ b/tests/integration/resources/exportedData.test.ts
@@ -82,7 +82,7 @@ describeWithMongoDB(
 
                     // wait for an error from the MCP Server as it
                     // means the resource is not available anymore
-                    if (response.isError === false) {
+                    if (response.isError !== true) {
                         continue;
                     }
 


package.json
```diff
@@ -89,6 +88,7 @@
     "node-machine-id": "1.1.12",
     "oauth4webapi": "^3.6.0",
     "openapi-fetch": "^0.14.0",
+    "proper-lockfile": "^4.1.2",
```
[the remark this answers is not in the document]
himanshusinghs: This is required at the runtime.

src/common/sessionExportsManager.ts
```diff
@@ -0,0 +1,278 @@
...
+                    const doc: string = EJSON.stringify(chunk, undefined, 2, ejsonOptions);
+                    const line = `${docsTransformed > 1 ? ",\n" : ""}${doc}`;
+
+                    callback(null, line);
+                } catch (err: unknown) {
+                    callback(err as Error);
+                }
+            },
+            final: function (callback) {
+                this.push("]");
```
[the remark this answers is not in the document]
himanshusinghs: This and the one below are valid suggestions which I addressed in [commit](https://github.com/mongodb-js/mongodb-mcp-server/pull/424/commits/f215e310591797ff7a3698cc8bae5deca064bccb)

src/resources/common/exportedData.ts
```diff
@@ -0,0 +1,160 @@
+import {
+    CompleteResourceTemplateCallback,
+    ListResourcesCallback,
+    ReadResourceTemplateCallback,
+    ResourceTemplate,
+} from "@modelcontextprotocol/sdk/server/mcp.js";
+import { Server } from "../../server.js";
+import logger, { LogId } from "../../common/logger.js";
+
+export class ExportedData {
```
himanshusinghs, line 11: note: This Resource does not extend from ReactiveResource class because its templated MCP resource unlike static MCP resource (one that is built by ReactiveResource class). I found several points of independence from a "ReactiveResource" for example:
- This resource does not have a "state" that needs to be reduced
- Unlike a reactive resource, it only serves as a gateway to actual resources
- The configuration options for a templated resource are unique to it and while we can extend reactive resource to allow for these options, it wont' seem right.

src/common/session.ts
```diff
@@ -18,6 +18,8 @@ export type SessionEvents = {
     close: [];
     disconnect: [];
     "connection-error": [string];
+    "export-expired": [string];
+    "export-available": [string];
```
nirinchev: This seems wrong to me. Event emitters should not have external components emit events on their behalf.

src/common/sessionExportsManager.ts
```diff
@@ -0,0 +1,308 @@
...
+};
+
+export type SessionExportsManagerConfig = Pick<
+    UserConfig,
+    "exportPath" | "exportTimeoutMs" | "exportCleanupIntervalMs"
+>;
+
+const MAX_LOCK_RETRIES = 10;
+
+export class SessionExportsManager {
```
nirinchev: Would it make sense to have the export manager live on the session rather than the server, similarly to what Kevin is doing with the connection manager? Then tools can access it via the session rather than store it upon registration.

src/common/sessionExportsManager.ts
```diff
@@ -0,0 +1,308 @@
...
+
+    public exportNameToResourceURI(nameWithExtension: string): string {
+        this.validateExportName(nameWithExtension);
+        return `exported-data://${nameWithExtension}`;
+    }
+
+    public exportsDirectoryPath(): string {
+        // If the session is not connected, we can't cannot work with exports
+        // for that session.
+        if (!this.session.sessionId) {
```
nirinchev: Not sure when and why this session id was added, but looking at it again, it's just an ObjectId that gets set when the server is initialized. Doesn't appear to be any reason not to set it upon session initialization, in which case we'd avoid these redundant checks.

src/common/sessionExportsManager.ts
```diff
@@ -0,0 +1,308 @@
...
+            );
+        } finally {
+            this.exportsCleanupInProgress = false;
+        }
+    }
+
+    /**
+     * Small utility to validate provided export name for path traversal or no
+     * extension */
+    private validateExportName(nameWithExtension: string): void {
```
nirinchev: This can be static/free function.

src/resources/common/exportedData.ts
```diff
@@ -0,0 +1,160 @@
...
+            }
+
+            if (typeof exportName !== "string") {
+                throw new Error("Cannot retrieve exported data, exportName not provided.");
+            }
+
+            return {
+                contents: [
+                    {
+                        uri: this.server.exportsManager.exportNameToResourceURI(exportName),
```
nirinchev: [q] Is this not just `uri`?

src/tools/mongodb/read/export.ts
```diff
@@ -0,0 +1,85 @@
...
+        if (!this.exportsManager) {
+            throw new Error("Incorrect server configuration, export not possible!");
+        }
+
+        await this.exportsManager.createJSONExport({ input: findCursor, exportName, jsonExportFormat });
+        const exportedResourceURI = this.exportsManager.exportNameToResourceURI(exportName);
+        const exportedResourcePath = this.exportsManager.exportFilePath(
+            this.exportsManager.exportsDirectoryPath(),
+            exportName
+        );
```
nirinchev: The path and the uri look like something that should be returned by the `exportManager.createJSONExport` rather than being queried later.

src/common/sessionExportsManager.ts
```diff
@@ -0,0 +1,308 @@
...
+            );
+        }
+    }
+
+    public exportNameToResourceURI(nameWithExtension: string): string {
+        this.validateExportName(nameWithExtension);
+        return `exported-data://${nameWithExtension}`;
+    }
+
+    public exportsDirectoryPath(): string {
```
nirinchev: This and `exportFilePath` being public seems like a code smell to me. The way I see it, the `sessionExportsManager` should encapsulate all the file management logic, and the entire public API should be more or less surfaced through the `Export` type. That is, it should have API for listing, reading, writing, and deleting `Export`'s, but the consumers should not ever concern themselves with filesystem operations or paths.

nirinchev (commented):
Haven't reviewed it thoroughly, but leaving some comments for things that should be reworked. Happy to grab some time later today or tomorrow to go through the high-level architecture and offer some suggestions there.

src/common/session.ts
himanshusinghs: Just to confirm - are you suggesting that these events should not even be emitted on Session class or rather we should provide public methods for notifying of these events and handle emitting them exclusively in this class?

src/common/sessionExportsManager.ts
himanshusinghs: Certainly - I had it on Server only because there wasn't a pattern of accessing session internals in the codebase and I did not want to disturb that. Will move it to Session now that we are already doing the same for ConnectionManager.

src/common/session.ts
nirinchev: I would be more inclined to move the event manager as a field on the session and have consumers subscribe to the events on the event manager directly, rather than having them proxied through the session.

src/common/sessionExportsManager.ts
himanshusinghs: Just revisited this and yea it can actually be set on initialization, even for http transport

src/resources/common/exportedData.ts
himanshusinghs: nope, the exportName is the variable substituted in the uri `exported-data://{exportName}`

src/tools/mongodb/read/export.ts
himanshusinghs: That's a good point - I will make this change, right away.

src/common/sessionExportsManager.ts
```diff
@@ -0,0 +1,308 @@
...
+        try {
+            const exportNameWithExtension = this.ensureExtension(exportName, "json");
+            this.validateExportName(exportNameWithExtension);
+
+            const inputStream = input.stream();
+            const ejsonDocStream = this.docToEJSONStream(this.getEJSONOptionsForFormat(jsonExportFormat));
+            await this.withExportsLock<void>(async (exportsDirectoryPath) => {
+                const exportFilePath = path.join(exportsDirectoryPath, exportNameWithExtension);
+                const outputStream = createWriteStream(exportFilePath);
+                outputStream.write("[");
```
kmruiz: Aren't we missing a closing bracket in the outputStream? Why are we opening it and not closing it?

src/common/sessionExportsManager.ts
```diff
@@ -0,0 +1,308 @@
...
+     * Creates the session exports directory and returns the path */
+    private async ensureExportsDirectory(): Promise<string> {
+        const exportsDirectoryPath = this.exportsDirectoryPath();
+        await fs.mkdir(exportsDirectoryPath, { recursive: true });
+        return exportsDirectoryPath;
+    }
+
+    /**
+     * Acquires a lock on the session exports directory. */
+    private async withExportsLock<R>(callback: (lockedPath: string) => Promise<R>): Promise<R> {
```
kmruiz: I'm not sure we need this file lock. Usually you have a file lock when you have multiple processes competing for writing on the same file, but this is not the case.

1. Deleting will remove files that are expired already. Worst case scenario, ENOENT when deleting or reading could be ignored.
2. Files are written once, and never updated.

I would suggest to keep the lock in memory, as we do right now, and avoid using file locks which have their own set of problems.

src/common/sessionExportsManager.ts
```diff
@@ -0,0 +1,308 @@
...
+                                error instanceof Error ? error.message : String(error)
+                            );
+                        }
+                    });
+                    throw pipelineError;
+                } finally {
+                    void input.close();
+                    if (pipeSuccessful) {
+                        const resourceURI = this.exportNameToResourceURI(exportNameWithExtension);
+                        this.availableExports = [
```
kmruiz: We have availableExports and exports files separated and that can lead to inconsistencies. I would suggest to keep only one as a source of truth.

If we need to clean up files that are not available, we can always iterate the folder and remove any file not in the availableExports array.

src/common/sessionExportsManager.ts
```diff
@@ -0,0 +1,308 @@
...
+    }
+
+    private docToEJSONStream(ejsonOptions: EJSONOptions | undefined) {
+        let docsTransformed = 0;
+        return new Transform({
+            objectMode: true,
+            transform: function (chunk: unknown, encoding, callback) {
+                ++docsTransformed;
+                try {
+                    const doc: string = EJSON.stringify(chunk, undefined, 2, ejsonOptions);
```
kmruiz: Do we need the padding? (the 2). This is going to be consumed likely by the agent or another program, and if a user wants to read the documents themselves, they can always use a formatter in their editor. This increases disk size and token count.

src/common/sessionExportsManager.ts
himanshusinghs: The closing bracket gets appended in the final callback of Transform stream [here](https://github.com/mongodb-js/mongodb-mcp-server/pull/424/files/e89cf2348ee1dc70950ea172dc9f71e77d6dec9f#diff-8a49a6df9b6103e0187759d5458c83fd41211e11a26b1114fd105237694768a2R201).

src/common/sessionExportsManager.ts
himanshusinghs: There is a case which explicitly requires locking. It is when writing a file takes way more time than the configured timeout, lets imagine 30 seconds. The cleanup interval, without a lock, would remove this file while its being written and thus lead to inconsistent behaviour.

In addition to the case above, there is an improvement which gets hindered if we don't do the locking. Currently we wait for the export to be finished but in an ideal world, we should be able to return the export URI immediately and then push progress notification about the export progress and eventually the export availability. If we remove the locking, there is chance of reading partially exported data.

I can't vouch for proper-filelock's stability but we have been using it for our DiskStorage in accuracy tests and it has worked pretty reliably so far. It also provides transaction primitives such lock retries on failure (same as transaction queues) and exclusive updates which we are able to use to "confidently" update the state in memory and on disk.

src/common/sessionExportsManager.ts
himanshusinghs: That's a valid point. I don't think we need padding for our case, happy to remove it.

src/common/sessionExportsManager.ts
```diff
@@ -0,0 +1,302 @@
...
+                LogId.exportCloseError,
+                "Error while closing SessionExportManager",
+                error instanceof Error ? error.message : String(error)
+            );
+        }
+    }
+
+    public exportNameToResourceURI(nameWithExtension: string): string {
+        this.validateExportName(nameWithExtension);
+        return `exported-data://${nameWithExtension}`;
```
addaleax:
```suggestion
        return `exported-data://${encodeURIComponent(nameWithExtension)}`;
```

and the reverse when decoding?

src/resources/common/exportedData.ts
nirinchev: Yes, but `this.server.exportsManager.exportNameToResourceURI(exportName)` would just put the `exported-data://` prefix there, right, eventually giving us the original uri?

src/common/sessionExportsManager.ts
kmruiz: Storing the state of the export in memory (in-progress, ready) should be enough for this use case.

File locks are complicated because they are persistent: what happens if the process dies while the lock is acquired? In a dev environment it's fine because you can delete the file, but in a user environment this can be more complicated.

Also, file locks are exclusive (only 1 writer or only 1 reader) which also limits the usability of the tool to either read a resource only if no export is in progress or only writing 1 export at a time.

To be resilient, we want a RW "lock" per export: an export can only be modified or deleted by one process, and can only be read if no one is writing into it, but everyone can read at the same time.

himanshusinghs: Hey! Sorry I did not update this thread but yesterday I took time to re-think purely from in-memory POV and your suggestion made sense while also handling the case I mentioned above, it was just a matter of tweaking how we run the export cleanup interval. the changes are in the following commit - [2002785](https://github.com/mongodb-js/mongodb-mcp-server/pull/424/commits/200278500a97e2c8046f1c916d082166d6f08218)

src/resources/common/exportedData.ts
```diff
@@ -0,0 +1,130 @@
...
+    ReadResourceTemplateCallback,
+    ResourceTemplate,
+} from "@modelcontextprotocol/sdk/server/mcp.js";
+import { Server } from "../../server.js";
+import logger, { LogId } from "../../common/logger.js";
+
+export class ExportedData {
+    private readonly name = "exported-data";
+    private readonly description = "Data files exported in the current session.";
+    private readonly uri = "exported-data://{exportName}";
```
gagik, line 14: maybe `mongodb-exported-data`? The namespace is too generic otherwise (similar argument can be made about our tool names in general)

src/common/sessionExportsManager.ts
himanshusinghs: I think this is also handled by the in-memory update that I did in the [2002785](https://github.com/mongodb-js/mongodb-mcp-server/pull/424/commits/200278500a97e2c8046f1c916d082166d6f08218)

src/common/sessionExportsManager.ts
himanshusinghs: Yep I followed on this and have thinned the exports manager interface just to `close`, `listAvailableExports`, `readExport` and `createJSONExport`. The file management happens within the manager itself and the public interface methods provide the necessary output for the rest of the code to work with, such as uri, path, etc

src/tools/mongodb/read/export.ts
himanshusinghs: implemented this in one of my recent commits

src/resources/common/exportedData.ts
himanshusinghs: Have eliminated `exportNameToResourceURI` entirely now.

src/common/session.ts
himanshusinghs: implemented this change in my recent commits

src/common/sessionExportsManager.ts
himanshusinghs: thank you for catching this 🙌, implemented this just now

src/resources/common/exportedData.ts
himanshusinghs: Yea this makes sense. I would personally like that we do it at-least for all the resources, if not for everything. Will ask if someone has an objection if I make this blanket change.

src/common/sessionExportsManager.ts
```diff
@@ -0,0 +1,278 @@
...
+
+    private async silentlyRemoveExport(exportPath: string) {
+        try {
+            await fs.unlink(exportPath);
+        } catch (error) {
+            // If the file does not exist or the containing directory itself
+            // does not exist then we can safely ignore that error anything else
+            // we need to flag.
+            if ((error as NodeJS.ErrnoException).code !== "ENOENT") {
+                this.logger.error({
```
kmruiz: This should be a warning, it doesn't require user interaction and the expected result is correct: the file does not exist anymore.

src/common/sessionExportsManager.ts
```diff
@@ -0,0 +1,278 @@
...
+    }: {
+        input: FindCursor;
+        exportName: string;
+        jsonExportFormat: JSONExportFormat;
+    }): Promise<{
+        exportURI: string;
+        exportPath: string;
+    }> {
+        try {
+            const exportNameWithExtension = validateExportName(ensureExtension(exportName, "json"));
```
kmruiz: ensureExtension adds the extension, and validateExportName ensures that the extension is there. I think validateExportName doesn't need that validation so we can remove it.

src/common/sessionExportsManager.ts
```diff
@@ -0,0 +1,278 @@
...
+            ({ createdAt }) => !isExportExpired(createdAt, this.config.exportTimeoutMs)
+        );
+    }
+
+    public async readExport(exportName: string): Promise<{
+        content: string;
+        exportURI: string;
+    }> {
+        try {
+            const exportNameWithExtension = validateExportName(exportName);
```
kmruiz: The other place where we might need the validation of the extension is here, but it's not necessary because this.availableExports only contains exports with the extension, so it's sanitised for free.

src/common/sessionExportsManager.ts
```diff
@@ -0,0 +1,278 @@
...
+        } catch (error) {
+            this.logger.error({
+                id: LogId.exportCloseError,
+                context: "Error while closing SessionExportManager",
+                message: error instanceof Error ? error.message : String(error),
+            });
+        }
+    }
+
+    public listAvailableExports(): Export[] {
```
kmruiz:
```suggestion
    get availableExports(): Export[] {
```

[nit]: Ignore me if you don't want to do this

src/tools/mongodb/read/export.ts
```diff
@@ -0,0 +1,81 @@
+import { CallToolResult } from "@modelcontextprotocol/sdk/types.js";
+import { OperationType, ToolArgs } from "../../tool.js";
+import { DbOperationArgs, MongoDBToolBase } from "../mongodbTool.js";
+import { FindArgs } from "./find.js";
+import { jsonExportFormat } from "../../../common/sessionExportsManager.js";
+import z from "zod";
+
+export class ExportTool extends MongoDBToolBase {
+    public name = "export";
+    protected description = "Export a collection data or query results in the specified json format.";
```
kmruiz:
```suggestion
    protected description = "Export a collection data or query results in the specified EJSON format.";
```

src/tools/mongodb/read/export.ts
```diff
@@ -0,0 +1,81 @@
...
+    protected argsShape = {
+        ...DbOperationArgs,
+        ...FindArgs,
+        limit: z.number().optional().describe("The maximum number of documents to return"),
+        jsonExportFormat: jsonExportFormat
+            .default("relaxed")
+            .describe(
+                [
+                    "The format to be used when exporting collection data as JSON with default being relaxed.",
```
kmruiz:
```suggestion
                    "The format to be used when exporting collection data as EJSON with default being relaxed.",
```

src/common/sessionExportsManager.ts
```diff
@@ -0,0 +1,278 @@
...
+            const exportFilePath = path.join(this.exportsDirectoryPath, exportNameWithExtension);
+
+            await fs.mkdir(this.exportsDirectoryPath, { recursive: true });
+            const inputStream = input.stream();
+            const ejsonDocStream = this.docToEJSONStream(this.getEJSONOptionsForFormat(jsonExportFormat));
+            const outputStream = createWriteStream(exportFilePath);
+            outputStream.write("[");
+            let pipeSuccessful = false;
+            try {
+                await pipeline([inputStream, ejsonDocStream, outputStream]);
```
kmruiz: Are we fine blocking the agent until the export is done? It's an unbound operation, it can take from seconds to days.

src/common/sessionExportsManager.ts
himanshusinghs: This log is only for when the error is anything other than ENOENT. Those we don't expect which is why we logging an error

src/common/sessionExportsManager.ts
```diff
@@ -0,0 +1,278 @@
...
+
+    public async createJSONExport({
+        input,
+        exportName,
+        jsonExportFormat,
+    }: {
+        input: FindCursor;
+        exportName: string;
+        jsonExportFormat: JSONExportFormat;
+    }): Promise<{
```
nirinchev: Why not return `Promise<Export>` here?

src/common/sessionExportsManager.ts
himanshusinghs: validateExportName does more than just validating the extension such as decoding and path traversal validation. And cannot remove the extension validation from validateExportName because that is used on the public API for `readExport`.

src/common/sessionExportsManager.ts
himanshusinghs: yea agreed that we don't need the validation here, there is no way to populate the available exports, other than going through createJSONExport which does the validation.

src/common/sessionExportsManager.ts
```diff
@@ -0,0 +1,278 @@
...
+import { UserConfig } from "./config.js";
+import { LoggerBase, LogId } from "./logger.js";
+
+export const jsonExportFormat = z.enum(["relaxed", "canonical"]);
+export type JSONExportFormat = z.infer<typeof jsonExportFormat>;
+
+export type Export = {
+    name: string;
+    uri: string;
+    path: string;
```
nirinchev: path and createdAt seem to only be used internally - consider restructuring the types to limit exposure of internal state to consumers.

src/common/sessionExportsManager.ts
```diff
@@ -0,0 +1,278 @@
...
+    ) {
+        super();
+        this.exportsDirectoryPath = path.join(this.config.exportsPath, sessionId);
+        this.exportsCleanupInterval = setInterval(
+            () => void this.cleanupExpiredExports(),
+            this.config.exportCleanupIntervalMs
+        );
+    }
+
+    public async close() {
```
nirinchev:
```suggestion
    public async close(): Promise<void> {
```

src/common/sessionExportsManager.ts
himanshusinghs: implemented this

src/tools/mongodb/read/export.ts
himanshusinghs: implemented this

src/common/sessionExportsManager.ts
himanshusinghs: implemented this

src/common/sessionExportsManager.ts
himanshusinghs: implemented this

src/common/sessionExportsManager.ts
himanshusinghs: No not fine really 😄 I will be implementing a change to make the tool result async. Will keep this thread open until then

src/common/sessionExportsManager.ts
himanshusinghs: talked about it on call, for now we will keep path exposed via the public APIs as we need it for clients that still cannot work with resource URIs

src/common/session.ts
```diff
@@ -10,12 +10,16 @@ import {
...
+import { ObjectId } from "bson";
+import { SessionExportsManager, SessionExportsManagerConfig } from "./sessionExportsManager.js";
+import { config } from "./config.js";
 
 export interface SessionOptions {
     apiBaseUrl: string;
     apiClientId?: string;
     apiClientSecret?: string;
     connectionManager?: ConnectionManager;
+    exportsManagerConfig?: SessionExportsManagerConfig;
```
nirinchev: I can see we are not super consistent here - (e.g. for the apiClient we provide a config, while for the connection manager, we provide an instance), but consider if it'd be better to provide the export manager instance instead of a config. The benefit of not creating it internally is that we can more easily mock it in tests.

src/common/session.ts
```diff
@@ -27,7 +31,8 @@ export type SessionEvents = {
 };
 
 export class Session extends EventEmitter<SessionEvents> {
-    sessionId?: string;
+    readonly sessionId = new ObjectId().toString();
+    readonly exportsManager: SessionExportsManager;
     connectionManager: ConnectionManager;
     apiClient: ApiClient;
```
nirinchev: [nit] these can also be readonly - feel free to drive-by fix them.

src/common/sessionExportsManager.ts
```diff
@@ -0,0 +1,278 @@
...
+    UserConfig,
+    "exportsPath" | "exportTimeoutMs" | "exportCleanupIntervalMs"
+>;
+
+type SessionExportsManagerEvents = {
+    "export-expired": [string];
+    "export-available": [string];
+};
+
+export class SessionExportsManager extends EventEmitter<SessionExportsManagerEvents> {
```
nirinchev: We can probably rename this to just `ExportsManager` - while it's indeed scoped to a session, I don't expect we'll have an unscoped exports manager in the future.

src/common/sessionExportsManager.ts
```diff
@@ -0,0 +1,278 @@
...
+    private sessionExports: Record<Export["name"], Export> = {};
+    private exportsCleanupInProgress: boolean = false;
+    private exportsCleanupInterval: NodeJS.Timeout;
+    private exportsDirectoryPath: string;
+
+    constructor(
+        readonly sessionId: string,
```
nirinchev: Do we need to store this as a field?

src/common/sessionExportsManager.ts
```diff
@@ -0,0 +1,278 @@
...
+        try {
+            const exportNameWithExtension = validateExportName(exportName);
+            const exportHandle = this.sessionExports[exportNameWithExtension];
+            if (!exportHandle) {
+                throw new Error("Requested export has either expired or does not exist!");
+            }
+
+            const { path: exportPath, uri, createdAt } = exportHandle;
+
+            if (isExportExpired(createdAt, this.config.exportTimeoutMs)) {
```
nirinchev: Do we really need this check? If the export is present in `sessionExports`, this means it's still present on disk, so we might as well return it, right?

src/common/sessionExportsManager.ts
```diff
@@ -0,0 +1,278 @@
...
+                throw new Error("Requested export has expired!");
+            }
+
+            return {
+                exportURI: uri,
+                content: await fs.readFile(exportPath, "utf8"),
```
nirinchev: There's a minor race condition here - if we're reading the file while the cleanup is triggered, we may attempt to delete it while we're reading from it. Not sure how unlink behaves in this case.

src/common/sessionExportsManager.ts
```diff
@@ -0,0 +1,278 @@
...
+            };
+        } catch (error) {
+            this.logger.error({
+                id: LogId.exportReadError,
+                context: "Error when reading export",
```
nirinchev: Can we include more information about the export we're trying to read here?

src/common/sessionExportsManager.ts
nirinchev: I think we can remove the validation from `readExport` - since we maintain the list of available exports in memory, if we do get an invalid name (e.g. path traversal or extensionless), we simply wouldn't find it in the array.

src/common/sessionExportsManager.ts
```diff
@@ -0,0 +1,278 @@
...
+            await fs.mkdir(this.exportsDirectoryPath, { recursive: true });
+            const inputStream = input.stream();
+            const ejsonDocStream = this.docToEJSONStream(this.getEJSONOptionsForFormat(jsonExportFormat));
+            const outputStream = createWriteStream(exportFilePath);
+            outputStream.write("[");
+            let pipeSuccessful = false;
+            try {
+                await pipeline([inputStream, ejsonDocStream, outputStream]);
+                pipeSuccessful = true;
+                return {
```
nirinchev: Do we have to write `]` to close the json array?

src/common/sessionExportsManager.ts
```diff
@@ -0,0 +1,278 @@
...
+                // partial and incorrect export so we remove it entirely.
+                await fs.unlink(exportFilePath).catch((error) => {
+                    if ((error as NodeJS.ErrnoException).code !== "ENOENT") {
+                        this.logger.error({
+                            id: LogId.exportCreationCleanupError,
+                            context: "Error when removing partial export",
+                            message: error instanceof Error ? error.message : String(error),
+                        });
+                    }
+                });
```
nirinchev: [nit] can we use try/catch here instead of promise continuations?

src/common/sessionExportsManager.ts
```diff
@@ -0,0 +1,278 @@
...
+            outputStream.write("[");
+            let pipeSuccessful = false;
+            try {
+                await pipeline([inputStream, ejsonDocStream, outputStream]);
+                pipeSuccessful = true;
+                return {
+                    exportURI,
+                    exportPath: exportFilePath,
+                };
+            } catch (pipelineError) {
```
nirinchev: Why do we need the extra try-catch here? Can we not move the unlinking logic inside the parent catch?

src/common/sessionExportsManager.ts
```diff
@@ -0,0 +1,278 @@
...
+        if (format === "relaxed") {
+            return {
+                relaxed: true,
+            };
+        }
+        return format === "canonical"
+            ? {
+                  relaxed: false,
+              }
+            : undefined;
```
nirinchev: I think this will be more obvious with a switch.

src/common/sessionExportsManager.ts
himanshusinghs: addressed it in this [8eeb5b2](https://github.com/mongodb-js/mongodb-mcp-server/pull/424/commits/8eeb5b295f762d4ead45126d9ad7fb5e3146dbd3)

src/resources/common/exportedData.ts
nirinchev: My understanding is the resource uris are already scoped to the MCP server, so it's not like they could conflict between different servers.

src/common/sessionExportsManager.ts
```diff
@@ -0,0 +1,297 @@
...
+    }
+
+    if (decodedName.includes("..") || decodedName.includes("/") || decodedName.includes("\\")) {
+        throw new Error("Invalid export name: path traversal hinted");
+    }
+
+    return decodedName;
+}
+
+export function isExportExpired(createdAt: number, exportTimeoutMs: number) {
```
nirinchev: https://github.com/mongodb-js/mongodb-mcp-server/pull/428 😛

src/resources/common/exportedData.ts
```diff
@@ -0,0 +1,130 @@
...
+} from "@modelcontextprotocol/sdk/server/mcp.js";
+import { Server } from "../../server.js";
+import { LogId } from "../../common/logger.js";
+
+export class ExportedData {
+    private readonly name = "exported-data";
+    private readonly description = "Data files exported in the current session.";
+    private readonly uri = "exported-data://{exportName}";
+
+    constructor(private readonly server: Server) {
```
nirinchev: [nit] Can we model this after the base tool, where we pass the session in the constructor and the server in the `register` method? I don't have strong feelings which design is better, I just want them to be as close as possible.

himanshusinghs: yeap - I have made this change in my recent commit

src/common/sessionExportsManager.ts
himanshusinghs: this happens in the final callback of the transform stream. It cannot happen outside because by the time `await pipeline` finishes, the streams are already closed.

src/common/sessionExportsManager.ts
himanshusinghs: nope, removed it

src/common/session.ts
himanshusinghs: Sure, I didn't do this because of interdependence of ExportsManager on sessionId which is in Session. I will pull the sessionId generation out of Session.

src/common/sessionExportsManager.ts
himanshusinghs: This is to account for the case when the last run of cleanup did not remove an "almost" expired export but did expire by the time it was requested.

src/common/sessionExportsManager.ts
himanshusinghs: Yeap but it should be fine because this case should be fairly uncommon and either outcomes (from below) seems good enough:
- If the file gets deleted first, before being read, the readFile will throw ENOENT as expected
- If the file gets read first and while its being read it gets deleted, the read will succeed (which I think should be fine), the same as unlink, because the file descriptor has already been opened.

src/common/sessionExportsManager.ts
himanshusinghs: yea, now that we're doing memory first export handling, we can safely remove the validation call. Will do that.

src/common/session.ts
himanshusinghs: done in my recent commit

src/common/session.ts
himanshusinghs: addressed

src/common/sessionExportsManager.ts
himanshusinghs: addressed

src/common/sessionExportsManager.ts
himanshusinghs: addressed

src/common/sessionExportsManager.ts
himanshusinghs: addressed

src/common/sessionExportsManager.ts
himanshusinghs: addressed

src/common/sessionExportsManager.ts
himanshusinghs: addressed

src/common/sessionExportsManager.ts
himanshusinghs: addressed

src/common/sessionExportsManager.ts
himanshusinghs: Implemented this in my recent changes, you might wanna re-check this.

src/resources/common/exportedData.ts
himanshusinghs: The scoping I think is controlled by clients and each might decide on implementing it differently. It should be fine if make our resources/tools more specific.

src/common/sessionExportsManager.ts
nirinchev: I guess my point is - if we already have the file and the client is requesting it, why do we artificially prevent it from accessing it? The file cleanup/expiry is something we're controlling internally and more of an implementation detail than anything else.

src/common/sessionExportsManager.ts
nirinchev: Note that on ntfs systems (notably Windows), calling `.unlink` on a file that is in use will not succeed. Instead, you'll get EPERM (or EBUSY - I always forget which is which), meaning that we'll leave the file orphaned if we don't retry the deletion.

src/common/sessionExportsManager.ts
nirinchev: Ugh... that's pretty unpleasant - can we do something so that adding the opening and the closing braces happens in the same place? Maybe in `docToEJSONStream` we can check for `docsTransformed === 0` and add the opening brace?

src/common/session.ts
```diff
@@ -52,7 +63,9 @@ export class Session extends EventEmitter<SessionEvents> {
 
         this.apiClient = new ApiClient({ baseUrl: apiBaseUrl, credentials }, logger);
 
-        this.connectionManager = connectionManager ?? new ConnectionManager();
+        this.sessionId = sessionId;
```
nirinchev: [nit] As far as I can tell, there's nothing special about the session id here that would require coordination between components. So in theory, we could just generate a random id for the export manager and just use that?

src/common/exportsManager.ts
```diff
@@ -0,0 +1,348 @@
...
+            }
+        }
+    }
+
+    static init(sessionId: string, config: ExportsManagerConfig, logger: LoggerBase): ExportsManager {
+        const exportsDirectoryPath = path.join(config.exportsPath, sessionId);
+        const exportsManager = new ExportsManager(exportsDirectoryPath, config, logger);
+        exportsManager.init();
+        return exportsManager;
+    }
```
nirinchev: Don't have super strong feelings here, but I feel this pattern is not super idiomatic. I guess you went that route to avoid starting the cleanup in the ctor, but an alternative approach would be to move setting up the interval to the first time an export is requested. That would give us 2 benefits - first, we'd be able to use the ctor like normal and second, we wouldn't be running the cleanup logic if the user never requested a an export. So it would look something like:

```ts
constructor(...) {
  // set the class-level variables as usual
}

createJSONExport(...) {
  if (!this.exportsCleanupInterval) {
    this.exportsCleanupInterval = setInterval(...);
  }
}
```

We can also get rid of the `wasInitialized` field since it's pretty much just returning whether `exportsCleanupInterval` is undefined or not.

src/resources/common/exportedData.ts
nirinchev: It is controlled by clients, but they need to be able to associate resources with servers - e.g. even if there's a conflict between schemes, there's no way that we could serve `exported-data://name` if it comes from a different MCP server.

src/resources/common/exportedData.ts
```diff
@@ -0,0 +1,130 @@
...
+    private readonly uri = "exported-data://{exportName}";
+    private server?: Server;
+
+    constructor(private readonly session: Session) {
+        const onExportChanged = (uri: string): void => {
+            this.server?.sendResourceListChanged();
+            this.server?.sendResourceUpdated(uri);
+        };
+        this.session.exportsManager.on("export-available", onExportChanged);
+        this.session.exportsManager.on("export-expired", onExportChanged);
```
nirinchev: [nit] should we move this to `register`? It's a bit inconsequential since we always call the constructor and then immediately after `.register`, but from a design perspective, it seems premature to be subscribing for events and accessing the server before we've been registered. It would also allow us to remove the class-level server variable.

src/tools/mongodb/read/export.ts
```diff
@@ -0,0 +1,82 @@
+import { CallToolResult } from "@modelcontextprotocol/sdk/types.js";
+import { OperationType, ToolArgs } from "../../tool.js";
+import { DbOperationArgs, MongoDBToolBase } from "../mongodbTool.js";
+import { FindArgs } from "./find.js";
+import { jsonExportFormat } from "../../../common/exportsManager.js";
+import z from "zod";
+
+export class ExportTool extends MongoDBToolBase {
```
nirinchev, line 9: No strong feelings here, but do we want this to be a dedicated tool vs an option for `find`/`aggregate`? Asking mostly because right now we can't export the results from an aggregation, which seems like valuable use case as well.

src/resources/common/exportedData.ts
```diff
@@ -0,0 +1,130 @@
...
+                        uri: url.href,
+                        text: `Error reading ${url.href}: ${error instanceof Error ? error.message : String(error)}`,
+                    },
+                ],
+                isError: true,
+            };
+        }
+    };
+
+    private exportNameToDescription(exportName: string): string {
```
nirinchev: Rather than parsing the export name for a description, can we just add the description for the export from the client? We can even add it as a field that the llm can provide.

src/tools/mongodb/read/export.ts
```diff
@@ -0,0 +1,82 @@
...
+            // parse a resource_link so we provide a text result for them to
+            // understand what to do with the result.
+            {
+                type: "text",
+                text: `Data for namespace ${database}.${collection} is being exported and will be made available under resource URI - "${exportURI}".`,
+            },
+            {
+                type: "resource_link",
+                name: exportName,
+                uri: exportURI,
```
nirinchev, line 68: I'm not sure how this behaves, but as things stand, we'll be providing a `uri` that's unavailable in the response. Does this confuse models at all or are they correctly handling it?

src/tools/mongodb/read/export.ts
```diff
@@ -0,0 +1,82 @@
...
+
+        // This special case is to make it easier to work with exported data for
+        // clients that still cannot reference resources (Cursor).
+        // More information here: https://jira.mongodb.org/browse/MCP-104
+        if (this.config.transport === "stdio") {
+            toolCallContent.push({
+                type: "text",
+                text: `Optionally, when the export is finished, the exported data can also be accessed under path - "${exportPath}"`,
+            });
+        }
```
nirinchev: We should probably add another config option - e.g. `isLocalServer` or use a more sophisticated logic to determine whether the server is local or remote (e.g. if it's http but bound to `127.0.0.1`, it's also local). This is primarily for cases, such as the vscode extension where we'd be running a local server over http and would probably want to include the local path in the response.

nirinchev (commented):
Some more comments from me - it's almost ready

src/common/sessionExportsManager.ts
himanshusinghs: Yea I was feeling that so I just addressed this in my recent commit.

src/resources/common/exportedData.ts
himanshusinghs: yea makes sense - I forgot this when I was cleaning up my ctors

src/tools/mongodb/read/export.ts
himanshusinghs: I have no strong feeling either - infact the initial idea was to have export as an option to find / aggregates. Back then we did not have accuracy tests and the idea of consolidation was still being discussed which is why we went this route.

Regardless I think we can do that but I would prefer to have it as a followup in a different PR, if that's ok.

src/tools/mongodb/read/export.ts
himanshusinghs: In my tests, I did not see any "confusing" responses from the models. They correctly interpreted the response and the "in-progress" part of it. Yes the response_uri is available which means users (through client) can decide to access it but even that should be fine because we always respond with "still in progress" notification.

src/resources/common/exportedData.ts
himanshusinghs: Sure, nothing against it. Will do that.

src/common/exportsManager.ts
himanshusinghs: While the suggestion looks really good, the only reason I wouldn't wanna do that is because we'll be triggering an unrelated side-effect in createJSONExport.

As per my understanding, offloading complex initializations to dedicated methods is pretty common. And I offloaded it to a static method only because otherwise it would have been a pain to do this in every tests:

```
const exportsManager = new ExportsManager(...)
exportsManager.init();
```

src/common/session.ts
himanshusinghs: That's kinda true. I was hoping to relate ExportsManager to a session and the only thing that felt like a bridge was this sessionId but we're certainly not using it in any meaningful way even the loggers initialize with their own ids.

Guess I will just get rid of this co-dependence

src/common/sessionExportsManager.ts
himanshusinghs: We are not preventing clients from accessing a still valid file. We are just ensuring that we don't respond back with expired content.

src/common/sessionExportsManager.ts
himanshusinghs: ahh that's sad. I guess I should do some sort of memory lock or maintain read/write count.

src/tools/mongodb/read/export.ts
himanshusinghs: Will take care of this. I think I will just do the isLocalServer method.

src/common/exportsManager.ts
nirinchev: I wouldn't say the side-effect is unrelated - `createJSONExport` is the entrypoint for creating files on disk. So, initializing the cleanup loop only once we have files to clean up does make sense to me.

src/tools/mongodb/read/export.ts
nirinchev: My only concern here is that merging this PR as-is would be a somewhat partial solution (as it only offers export to find but not for aggregate). If we want to keep it as a separate tool, would it make sense to adopt a design similar to the `explain` tool where the model can provide either a find or an aggregation pipeline in the args?

src/tools/mongodb/read/export.ts
nirinchev: Perhaps I'm misunderstanding the logic here, but looking at `ExportsManager.availableExports`, it seems like we're filtering out the in-progress reports, which means that the `exportURI` we supply here would not resolve to a valid resource until the export is completed.

src/tools/mongodb/read/export.ts
himanshusinghs: addressed this [c1d0117](https://github.com/mongodb-js/mongodb-mcp-server/pull/424/commits/c1d01177fa8a382226c4e66b476d0d4efce29a5e)

src/tools/mongodb/read/export.ts
himanshusinghs: The filtering is only for the autocompletion of the export names. The autocomplete is triggered when someone chose the templated URI and then started typing name of the export. But if someone requests the resource directly by just providing the exportName / exportURI then the `readResourceCallback` will be triggered and that would return "Resource is still being generated" error content.

src/resources/common/exportedData.ts
himanshusinghs: addressed here - [5bcac97](https://github.com/mongodb-js/mongodb-mcp-server/pull/424/commits/5bcac973b8e6b109ff3a60ed29bf61bd4f61cc14)

src/tools/mongodb/read/export.ts
himanshusinghs: Yea that's certainly doable but I don't think that is a blocker for this PR. The PR when merged will at-least offer exports for `find` queries and entire namespaces and that already seems to be a good value compared to nothing that we have right now. I will be super happy to immediately follow up for the aggregations as well.

src/resources/common/exportedData.ts
himanshusinghs: addressed [2136a88](https://github.com/mongodb-js/mongodb-mcp-server/pull/424/commits/2136a88845c37e5c56a813d5776c194af28df9f3)

src/common/exportsManager.ts
himanshusinghs: Even then I don't think `createJSONExport` should concern itself with starting the cleanup loop as well, that's not what its meant for.

src/tools/mongodb/read/export.ts
nirinchev: Cool, let's merge as-is then and then we can address the aggregation as a follow-up.

src/common/exportsManager.ts
nirinchev: As a counterargument, what's the point of having a cleanup loop if you know there's nothing to clean up? Not a hill I'm willing to die on, so happy to keep it as-is, but an alternative design would be to use `setTimeout` instead of a cleanup loop - that is, everytime we create a json export, we setup a clean up callback to delete it after x minutes. In that case, it would very much be `createJSONExport` that sets that up, right? While the cleanup mechanism is different, I would say the notion of triggering/scheduling/setting up a garbage collection procedure as a side effect of creating "garbage" would not be surprpising to me.

src/common/session.ts
himanshusinghs: addressed in - [a3df32c](https://github.com/mongodb-js/mongodb-mcp-server/pull/424/commits/a3df32c79ec2e9e68858d0f1eb8cc403c5727cd6)

src/common/sessionExportsManager.ts
himanshusinghs: implemented a read-write lock to address this

src/common/sessionExportsManager.ts
himanshusinghs: Removed this in my recent commit

package.json
```diff
@@ -96,6 +96,7 @@
     "@mongodb-js/devtools-proxy-support": "^0.5.1",
     "@mongosh/service-provider-node-driver": "^3.10.2",
     "@vitest/eslint-plugin": "^1.3.4",
+    "async-rwlock": "^1.1.1",
```
nirinchev: Not a huge fan of this dependency - it's a 7-year-old npm package that until a few months ago was seeing 1k weekly downloads and the github repo has 0 stars. Obviously, that doesn't mean it's of poor quality or necessarily worse than more established packages, but with fewer eyes on it and a single maintainer, it introduces significant risk for what seems to be ~100 lines of code saved.

PR #429, opened 2025-08-07
chore: add arg-parser and put the config under test MCP-86

kmruiz:
## Proposed changes

Use mongosh argument parser to generate the connection string that is going to be used to connect to MongoDB, aligning with how mongosh does it and making sure we support additional authentication mechanisms. It removes a bunch of flags and types not necessary now with the consolidation.

There will be integration tests for each of the supported connection types when playing each of their tasks. Specific support is out of the scope of this PR.

## Checklist

### DIFF
--- a/src/common/config.ts
+++ b/src/common/config.ts
@@ -275,14 +275,10 @@ function commaSeparatedToArray<T extends string[]>(str: string | string[] | unde
     }
 
     if (str.length === 1) {
-        if (str[0]?.indexOf(",") === -1) {
-            return str.filter((e) => e.trim().length > 0) as T;
-        } else {
-            return str[0]
-                ?.split(",")
-                .map((e) => e.trim())
-                .filter((e) => e.length > 0) as T;
-        }
+        return str[0]
+            ?.split(",")
+            .map((e) => e.trim())
+            .filter((e) => e.length > 0) as T;
     }
 
     return str as T;
--- a/src/common/config.ts
+++ b/src/common/config.ts
@@ -273,9 +273,12 @@ function commaSeparatedToArray<T extends string[]>(str: string | string[] | unde
 
     if (str.length === 1) {
         if (str[0]?.indexOf(",") === -1) {
-            return str as T;
+            return str.filter((e) => e.trim().length > 0) as T;
         } else {
-            return str[0]?.split(",").map((e) => e.trim()) as T;
+            return str[0]
+                ?.split(",")
+                .map((e) => e.trim())
+                .filter((e) => e.length > 0) as T;
         }
     }
 
@@ -315,7 +318,8 @@ export function setupUserConfig({
         throw new Error(`Invalid telemetry: ${telemetry}`);
     }
 
-    if (userConfig.httpPort < 1 || userConfig.httpPort > 65535) {
+    const httpPort = +userConfig.httpPort;
+    if (httpPort < 1 || httpPort > 65535 || isNaN(httpPort)) {
         throw new Error(`Invalid httpPort: ${userConfig.httpPort}`);
     }
 
--- a/src/common/config.ts
+++ b/src/common/config.ts
@@ -66,7 +66,6 @@ const OPTIONS = {
         "tlsAllowInvalidCertificates",
         "tlsAllowInvalidHostnames",
         "tlsFIPSMode",
-        "tlsUseSystemCA",
         "version",
     ],
     array: ["disabledTools", "loggers"],
--- a/src/common/config.ts
+++ b/src/common/config.ts
@@ -2,7 +2,7 @@ import path from "path";
 import os from "os";
 import argv from "yargs-parser";
 import type { CliOptions } from "@mongosh/arg-parser";
-import { ReadConcernLevel, ReadPreferenceMode, W } from "mongodb";
+import type { ConnectionInfo } from "@mongosh/arg-parser";
 
 // From: https://github.com/mongodb-js/mongosh/blob/main/packages/cli-repl/src/arg-parser.ts
 const OPTIONS = {
@@ -98,13 +98,6 @@ function isConnectionSpecifier(arg: string | undefined): boolean {
     );
 }
 
-export interface ConnectOptions {
-    readConcern: ReadConcernLevel;
-    readPreference: ReadPreferenceMode;
-    writeConcern: W;
-    timeoutMS: number;
-}
-
 // If we decide to support non-string config options, we'll need to extend the mechanism for parsing
 // env variables.
 export interface UserConfig extends CliOptions {
@@ -117,7 +110,6 @@ export interface UserConfig extends CliOptions {
     exportTimeoutMs: number;
     exportCleanupIntervalMs: number;
     connectionString?: string;
-    connectOptions: ConnectOptions;
     disabledTools: Array<string>;
     readOnly?: boolean;
     indexCheck?: boolean;
@@ -135,12 +127,6 @@ const defaults: UserConfig = {
     exportsPath: getExportsPath(),
     exportTimeoutMs: 300000, // 5 minutes
     exportCleanupIntervalMs: 120000, // 2 minutes
-    connectOptions: {
-        readConcern: "local",
-        readPreference: "secondaryPreferred",
-        writeConcern: "majority",
-        timeoutMS: 30_000,
-    },
     disabledTools: [],
     telemetry: "enabled",
     readOnly: false,
@@ -164,6 +150,19 @@ function getLocalDataPath(): string {
         ? path.join(process.env.LOCALAPPDATA || process.env.APPDATA || os.homedir(), "mongodb")
         : path.join(os.homedir(), ".mongodb");
 }
+
+export const defaultDriverOptions: ConnectionInfo["driverOptions"] = {
+    readConcern: {
+        level: "local",
+    },
+    readPreference: "secondaryPreferred",
+    writeConcern: {
+        w: "majority",
+    },
+    timeoutMS: 30_000,
+    proxy: { useEnvironmentVariableProxies: true },
+    applyProxyToOIDC: true,
+};
 
 function getLogPath(): string {
     const logPath = path.join(getLocalDataPath(), "mongodb-mcp", ".app-logs");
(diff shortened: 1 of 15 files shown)


src/common/config.ts
```diff
@@ -1,26 +1,112 @@
 import path from "path";
 import os from "os";
 import argv from "yargs-parser";
+import type { CliOptions, ConnectionInfo } from "@mongosh/arg-parser";
+import { generateConnectionInfoFromCliArgs } from "@mongosh/arg-parser";
 
-import { ReadConcernLevel, ReadPreferenceMode, W } from "mongodb";
+// From: https://github.com/mongodb-js/mongosh/blob/main/packages/cli-repl/src/arg-parser.ts
+const OPTIONS = {
```
himanshusinghs, line 8: Would it be an option to import it from the args-parser package instead?

src/common/config.ts
```diff
@@ -129,9 +227,87 @@ function SNAKE_CASE_toCamelCase(str: string): string {
...
+    const programArgs = args.slice(2);
+    const parsed = argv(programArgs, OPTIONS) as unknown as CliOptions &
+        UserConfig & {
+            _?: string[];
+        };
+
+    const positionalArguments = parsed._ ?? [];
+    if (!parsed.nodb && isConnectionSpecifier(positionalArguments[0])) {
+        parsed.connectionSpecifier = positionalArguments.shift();
+    }
```
himanshusinghs, line 258: nit: would be nice to add a comment what the intention here

src/index.ts
```diff
@@ -78,3 +104,39 @@ main().catch((error: unknown) => {
...
+        console.error("Error details:");
+        console.error(fipsError);
+        process.exit(1);
+    }
+}
+
+function assertHelpMode(): void | never {
+    if (config.help) {
+        console.log("For usage information refer to the README.md:");
+        console.log("https://github.com/mongodb-js/mongodb-mcp-server?tab=readme-ov-file#quick-start");
```
himanshusinghs, line 128: I think these should be console.error or console.warn. We shouldn't be logging to stdio as that is [reserved](https://modelcontextprotocol.io/specification/2025-06-18/basic/transports#stdio) for transport.

src/common/config.ts
```diff
@@ -1,26 +1,112 @@
...
-    readConcern: ReadConcernLevel;
-    readPreference: ReadPreferenceMode;
-    writeConcern: W;
-    timeoutMS: number;
+function isConnectionSpecifier(arg: string | undefined): boolean {
+    return (
+        arg !== undefined &&
+        (arg.startsWith("mongodb://") ||
+            arg.startsWith("mongodb+srv://") ||
+            !(arg.endsWith(".js") || arg.endsWith(".mongodb")))
```
himanshusinghs, line 97: I doubt that an arg ending with .js or .mongodb be a valid value for us?

src/common/config.ts
```diff
@@ -129,9 +227,87 @@ function SNAKE_CASE_toCamelCase(str: string): string {
...
-// Reads the cli args and parses them into a UserConfig object.
-function getCliConfig(): Partial<UserConfig> {
-    return argv(process.argv.slice(2), {
-        array: ["disabledTools", "loggers"],
-    }) as unknown as Partial<UserConfig>;
+// Right now we have arguments that are not compatible with the format used in mongosh.
+// An example is using --connectionString and positional arguments.
+// We will consolidate them in a way where the mongosh format takes precedence.
+// We will warn users that previous configuration is deprecated in favour of
+// whatever is in mongosh.
```
himanshusinghs, line 244: Is the deprecation logic planned for this PR?

src/common/config.ts
kmruiz: I don't think so:

1. args-parser does not expose it, we would need to update it, which is fine but
2. we don't support all arguments (eval for example) so we would still need to filter or do some post processing.

src/common/config.ts
kmruiz: Nope. I have another task in M5 for that.

src/index.ts
kmruiz: It's reserved once the transport is started, but this is done before the transport is started. This is for developers using --help to understand the arguments in their console, it's not expected to be used in a mcp.json file or similar.

src/common/config.ts
kmruiz: It's not a valid value, it's rejecting them.

src/common/config.ts
himanshusinghs: Yea but I still think its just a pretty long list to copy here instead of something like this?

`Omit<Options, "eval" | "or" | "others">`

that way you at-least have one source of truth.

src/index.ts
himanshusinghs: Agreed, but then you might wanna verify the behaviour of transport on start when you already have pushed some non-json rpc messages.

src/common/config.ts
himanshusinghs: ohh you little `!`, I see you now 🙄

src/common/config.ts
kmruiz: While they are similar, we do support more options (for the legacy options and Atlas credentials) and I feel being explicit on what we do support additionally and what we don't support is better. If we depend on mongosh and we add more mongosh specific flags we would inherit them without validation.

src/index.ts
kmruiz: These functions early exit, so there wouldn't be a situation where we issue a warning and then start the transport. That's why the signature of the function is void | never.

src/index.ts
```diff
@@ -1,11 +1,37 @@
 #!/usr/bin/env node
 
+let fipsError: Error | undefined;
```
gagik: why not just throw the error right away? would be good to avoid global variables if we can

src/index.ts
```diff
@@ -78,3 +104,39 @@ main().catch((error: unknown) => {
     });
     process.exit(1);
 });
+
+function assertFIPSMode(): void | never {
```
gagik:
```suggestion
function assertFIPSMode(): void {
```
nit: same below

gagik: would be good to keep the `index.ts` file as small as possible and move these to their own file(s)

src/index.ts
kmruiz: Because in line 108 we do more checks that need to be done after this.

1. We first enable FIPS before anything.
2. Then after we load the flags, we check different configurations and then give a specific error message (for example, shared openssl vs bundled)
3. Then we show FIPS error details based on the actual configuration.

src/index.ts
kmruiz: Why? `void | never` makes explicit that this function either doesn't return anything or exits the process.

kmruiz: I personally prefer to have all initialisation functions that can exit the process in the main file so it's clear that outside this file no one should do process.exit.

gagik: I don't really feel strongly about this but I have never (pun unintended) seen `never` as a value which implies it exits the process. To me it's used to enforce i.e. that a certain field / value is never going to exist in a given state, especially with type unions

src/index.ts
gagik: > Because in line 108 we do more checks that need to be done after this.

if we error in line 13, would we need to do any additional checks?  it erroring then eventually leads to it being handled late with:
```
        if (!fipsError && !crypto.getFips()) {
            fipsError = new Error("FIPS mode not enabled despite requested.");
        }
```
if `config.tlsFIPSMode` can be set to true in other ways other than argv then I don't think we'd support the alternative ways this way. If that's intended then that's fine

src/index.ts
kmruiz: It's actually documented here 😄 

https://www.typescriptlang.org/docs/handbook/2/functions.html#never

A bit hidden to be fair.

src/index.ts
kmruiz: The relevant info is in the if starting at line 115. If you see, we access process.config.variables.node_shared_openssl. This is not really relevant for us now: https://nodejs.org/api/process.html#processconfig

But I would like to keep this as close to how mongosh does it because if at some point we build our own snapshots (similar to what mongosh does) for reliability or easiness of distribution, this will be relevant.

I don't have a strong opinion either way, if you have it I can just change it, merge everything into enableFipsIfRequested and remove the assertFips function.

src/index.ts
gagik: I read that like 3 times and never (okay this pun is more intended) noticed "or terminates execution of the program" bit 😅 fair enough, today I learnt

src/index.ts
gagik: if `process.config.variables.node_shared_openssl` is already set at this point it'd be nice to just have one function and no new globals

src/index.ts
gagik: fair enough, no strong feelings there. if the file gets too large can revisit.

src/index.ts
kmruiz: SGTM, I'll do the refactor ASAP and push the changes.

src/common/config.ts
```diff
@@ -139,9 +237,123 @@ function SNAKE_CASE_toCamelCase(str: string): string {
...
+function parseCliConfig(args: string[]): CliOptions {
+    const programArgs = args.slice(2);
+    const parsed = argv(programArgs, OPTIONS) as unknown as CliOptions &
+        UserConfig & {
+            _?: string[];
```
gagik, line 249: I am missing some context but what is this `_` representing?

src/common/config.ts
```diff
@@ -139,9 +237,123 @@ function SNAKE_CASE_toCamelCase(str: string): string {
...
+    // list, so it doesn't get misunderstood by the mongosh args-parser
+    if (!parsed.nodb && isConnectionSpecifier(positionalArguments[0])) {
+        parsed.connectionSpecifier = positionalArguments.shift();
+    }
+
+    delete parsed._;
+    return parsed;
+}
+
+function commaSeparatedToArray<T extends string[]>(str: string | string[] | undefined): T {
```
gagik, line 264: very nit: any reason for using a generic type?

src/common/config.ts
```diff
@@ -139,9 +237,123 @@ function SNAKE_CASE_toCamelCase(str: string): string {
...
+    }
+
+    if (str.length === 0) {
+        return str as T;
+    }
+
+    if (str.length === 1) {
+        if (str[0]?.indexOf(",") === -1) {
+            return str.filter((e) => e.trim().length > 0) as T;
+        } else {
```
gagik: is handling this case separately necessary? I think the case below would also work fine

src/common/config.ts
kmruiz: `_` is usually used for positional arguments: https://yargs.js.org/docs/#api-reference-parse

src/common/config.ts
kmruiz: Because it can return a type that is an array of an union, like: ('mcp' | 'disk')[]

src/common/config.ts
kmruiz: Ah yes, you are right! refactoring it!

PR #450, opened 2025-08-13
feat: support for OIDC connectivity between the MCP Server and a MongoDB Cluster MCP-84

kmruiz:
## Proposed changes

This allows users to use OIDC authentication between the MCP Server and the MongoDB Cluster. It adds a test suite to ensure that the different flows do work as expected.

There is a small change on the test suite. Now describeWithMongoDb allows to specify the mongodb version to download. This is useful if we want to test OIDC in other versions. Also, defaults testing in MongoDB enterprise, as it's like community but with more capabilities, and allows us to test OIDC.

## Checklist


tests/integration/fixtures/curl.mjs
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env node
+import fetch from "node-fetch";
+
+if (process.env.MONGOSH_E2E_TEST_CURL_ALLOW_INVALID_TLS) {
+    process.env.NODE_TLS_REJECT_UNAUTHORIZED = "0";
```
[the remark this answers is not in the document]
kmruiz: This is test code, so it's irrelevant.

src/common/connectionManager.ts
```diff
@@ -82,18 +94,39 @@ export class ConnectionManager extends EventEmitter<ConnectionManagerEvents> {
...
+                connectionSpecifier: settings.connectionString,
             });
+
+            if (connectionInfo.driverOptions.oidc) {
+                connectionInfo.driverOptions.oidc.allowedFlows ??= ["auth-code"];
+                connectionInfo.driverOptions.oidc.notifyDeviceFlow ??= this.onOidcNotifyDeviceFlow.bind(this);
+            }
+
+            connectionInfo.driverOptions.proxy ??= { useEnvironmentVariableProxies: true };
+            connectionInfo.driverOptions.applyProxyToOIDC ??= true;
```
himanshusinghs, line 118: IIRC, we were doing this already somewhere, cannot find that anymore.

src/common/connectionManager.ts
```diff
@@ -67,11 +69,21 @@ export interface ConnectionManagerEvents {
...
-    constructor() {
+    constructor(
+        private userConfig: UserConfig,
+        private driverOptions: DriverOptions,
+        private logger: CompositeLogger,
+        bus?: EventEmitter
+    ) {
         super();
 
+        this.bus = bus ?? new EventEmitter();
```
himanshusinghs, line 82: I guess this was introduced because you want to keep these events private to this class?

src/common/connectionManager.ts
```diff
@@ -105,13 +138,28 @@ export class ConnectionManager extends EventEmitter<ConnectionManagerEvents> {
         }
 
         try {
+            const connectionType = ConnectionManager.inferConnectionTypeFromSettings(this.userConfig, connectionInfo);
+            if (connectionType.startsWith("oidc")) {
+                // The error here is irrelevant, we only use this ping to force the connection
+                // Errors will be handled by the auth flow.
+                void serviceProvider?.runCommand?.("admin", { hello: 1 }).catch(() => {});
```
himanshusinghs: nit: I very much like the ease of `.catch` but just to keep consistency you might want try/catch here? Or move it to a separate function if the nesting bothers.

src/common/connectionManager.ts
```diff
@@ -157,13 +205,62 @@ export class ConnectionManager extends EventEmitter<ConnectionManagerEvents> {
         return newState;
     }
 
-    static inferConnectionTypeFromSettings(settings: ConnectionSettings): ConnectionStringAuthType {
+    private onOidcAuthFailed(error: unknown): void {
+        if (this.state.tag === "connecting" && this.state.connectionStringAuthType?.startsWith("oidc")) {
+            void this.disconnect().then(() => {
```
himanshusinghs: nit: the same suggestion here for try/catch. The entire function gets called on an event so should be safe to await the disconnect here.

src/common/connectionManager.ts
```diff
@@ -181,4 +278,32 @@ export class ConnectionManager extends EventEmitter<ConnectionManagerEvents> {
                 return "scram";
         }
     }
+
+    static async waitUntil<T extends ConnectionState>(
```
himanshusinghs: Looks like this is used only in tests, should we move it to test helpers?

src/tools/mongodb/mongodbTool.ts
```diff
@@ -76,6 +86,7 @@ export abstract class MongoDBToolBase extends ToolBase {
                                 type: "text",
                                 text: "You need to connect to a MongoDB instance before you can access its data.",
                             },
+                            ...additionalPromptForOidc,
```
himanshusinghs: I feel like this and the next instruction should be mutually exclusive as the next instruction has the potential to derail the LLM thinking path and make it call the connect tool again?

himanshusinghs (changes requested):
The core functionality looks pretty great 🙌  I left some minor feedback for code consistency and one potential problem for LLM thought process.

src/common/connectionManager.ts
kmruiz: Yes, I refactored it to do it like this, because that's how mongosh does it when OIDC is enabled.

src/common/connectionManager.ts
kmruiz: Yes, ideally we don't want to look OIDC auth details outside this class. Independently on the auth mechanism, listeners of events should listen to the generic events.

src/common/connectionManager.ts
kmruiz: The idea here is to trigger a connection and ignore if it fails. With a try catch we need to await, and this would block the MCP Server until there is a connection or worse, a timeout error, that can take seconds.

I could extract the command to a function that returns a promise, and then do void ping(), but I think it's an indirection that doesn't make very much sense. How do you feel about it?

src/tools/mongodb/mongodbTool.ts
kmruiz: So far the accuracy tests were working fine, but yeah I understand the concern. Let me see how can I make the prompting more clear for the LLM.

src/common/connectionManager.ts
kmruiz: Sounds good! I was using it for prod code but after refactoring now I'm using it for tests. Will do the fix ASAP.

src/common/connectionManager.ts
himanshusinghs: I think wrapping the try/catch in a function that just `void`s the promise will indicate the intention better?

kmruiz: Can do that, it takes a second.

src/common/connectionManager.ts
```diff
@@ -181,4 +274,24 @@ export class ConnectionManager extends EventEmitter<ConnectionManagerEvents> {
                 return "scram";
         }
     }
+
+    private async pingAndForget(serviceProvider: NodeDriverServiceProvider): Promise<void> {
+        try {
+            await serviceProvider?.runCommand?.("admin", { hello: 1 }).catch(() => {});
+        } catch (error: unknown) {
+            // we really don't care about this error
+            void error;
```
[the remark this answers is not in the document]
kmruiz: We are ignoring these errors on purpose.

src/common/connectionManager.ts
```diff
@@ -181,4 +274,24 @@ export class ConnectionManager extends EventEmitter<ConnectionManagerEvents> {
...
+        }
+    }
+
+    private async disconnectOnOidcError(error: unknown): Promise<void> {
+        try {
+            await this.disconnect();
+        } catch (error: unknown) {
+            // we really don't care about this error
+            void error;
```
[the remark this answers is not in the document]
kmruiz: We are ignoring these errors on purpose.

src/common/connectionManager.ts
```diff
@@ -157,13 +203,60 @@ export class ConnectionManager extends EventEmitter<ConnectionManagerEvents> {
         return newState;
     }
 
-    static inferConnectionTypeFromSettings(settings: ConnectionSettings): ConnectionStringAuthType {
+    private onOidcAuthFailed(error: unknown): void {
+        if (this.state.tag === "connecting" && this.state.connectionStringAuthType?.startsWith("oidc")) {
+            void this.disconnectOnOidcError(error);
+        }
```
himanshusinghs, line 209: I think we should at-least log here for debugging purposes.

src/common/connectionManager.ts
```diff
@@ -181,4 +274,24 @@ export class ConnectionManager extends EventEmitter<ConnectionManagerEvents> {
                 return "scram";
         }
     }
+
+    private async pingAndForget(serviceProvider: NodeDriverServiceProvider): Promise<void> {
+        try {
+            await serviceProvider?.runCommand?.("admin", { hello: 1 }).catch(() => {});
```
himanshusinghs:
```suggestion
            await serviceProvider?.runCommand?.("admin", { hello: 1 });
```
The .catch can be omitted now.

PR #453, opened 2025-08-13
fix: update resolution and transition

blva:
## Proposed changes

fixes transition

## Checklist

### DIFF
--- a/.github/workflows/jira-issue.yml
+++ b/.github/workflows/jira-issue.yml
@@ -101,8 +101,7 @@ jobs:
         with:
           token: ${{ secrets.JIRA_API_TOKEN }}
           issue-key: ${{ steps.find_jira.outputs.issue-key }}
-          transition-id: 1381 # Resolved
-          resolution: Fixed
+          transition-id: 61
       - name: Add closure comment
         if: steps.close_jira_ticket.outcome == 'success'
         uses: peter-evans/create-or-update-comment@71345be0265236311c031f5c7866368bd1eff043


.github/workflows/jira-issue.yml
```diff
@@ -101,8 +101,7 @@ jobs:
         with:
           token: ${{ secrets.JIRA_API_TOKEN }}
           issue-key: ${{ steps.find_jira.outputs.issue-key }}
-          transition-id: 1381 # Resolved
-          resolution: Fixed
+          transition-id: 61
```
cveticm, line 104: question for my own learning: what does this transition-id do compared to 1381?

blva: great q! apparently, there is a list of valid transitions we can do per project. here, I added the one we use in AtlasCLI, however the JIRA project is different, and so are the transitions that are valid. To find the right one I had to first do a curl

``` curl -X GET \
  --url "https://<mdb-jira-url>rest/api/2/issue/MCP-<issue-id>/transitions" \
  --header "Authorization: Bearer <token> " \
  --header "Accept: application/json"
  ```
  
  to get the available transitions and test it!

cveticm: Oh interesting! 
And how come you remove `resolution`? I know when resolving a JIRA via UI, it typically requires a resolution to be set.

blva: Yep, the request would fail w/ resolution. Ideally we should send a separate one for that, but that change should go into the transition apix action

PR #464, opened 2025-08-20
feat: adds support for exporting aggregations MCP-16

himanshusinghs:
## Proposed changes

Extends the current export tool to allow exporting aggregations as well.

### DIFF
--- a/src/tools/mongodb/read/export.ts
+++ b/src/tools/mongodb/read/export.ts
@@ -1,19 +1,33 @@
 import z from "zod";
 import { ObjectId } from "bson";
+import { AggregationCursor, FindCursor } from "mongodb";
 import { CallToolResult } from "@modelcontextprotocol/sdk/types.js";
 import { OperationType, ToolArgs } from "../../tool.js";
 import { DbOperationArgs, MongoDBToolBase } from "../mongodbTool.js";
 import { FindArgs } from "./find.js";
 import { jsonExportFormat } from "../../../common/exportsManager.js";
+import { AggregateArgs } from "./aggregate.js";
 
 export class ExportTool extends MongoDBToolBase {
     public name = "export";
     protected description = "Export a collection data or query results in the specified EJSON format.";
     protected argsShape = {
-        exportTitle: z.string().describe("A short description to uniquely identify the export."),
         ...DbOperationArgs,
-        ...FindArgs,
-        limit: z.number().optional().describe("The maximum number of documents to return"),
+        exportTitle: z.string().describe("A short description to uniquely identify the export."),
+        exportTarget: z
+            .array(
+                z.discriminatedUnion("name", [
+                    z.object({
+                        name: z.literal("find"),
+                        arguments: z.object(FindArgs),
+                    }),
+                    z.object({
+                        name: z.literal("aggregate"),
+                        arguments: z.object(AggregateArgs),
+                    }),
+                ])
+            )
+            .describe("The export target along with its arguments."),
         jsonExportFormat: jsonExportFormat
             .default("relaxed")
             .describe(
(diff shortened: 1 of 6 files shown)


tests/integration/tools/mongodb/read/export.test.ts
```diff
@@ -371,6 +414,48 @@ describeWithMongoDB(
...
+                        ],
+                    },
+                });
+                const content = response.content as CallToolResult["content"];
+                const exportURI = contentWithResourceURILink(content)?.uri as string;
+                await resourceChangedNotification(integration.mcpClient(), exportURI);
+
+                const localPathPart = contentWithExportPath(content);
+                expect(localPathPart).toBeDefined();
+                const [, localPath] = /"(.*)"/.exec(String(localPathPart?.text)) ?? [];
```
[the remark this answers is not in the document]
himanshusinghs: The parsing would anyways involve regex matching so there is no point in extracting it out to a different function.

src/tools/mongodb/read/export.ts
```diff
@@ -1,19 +1,33 @@
...
         ...DbOperationArgs,
-        ...FindArgs,
-        limit: z.number().optional().describe("The maximum number of documents to return"),
+        exportTitle: z.string().describe("A short description to uniquely identify the export."),
+        exportTarget: z
+            .array(
+                z.discriminatedUnion("name", [
+                    z.object({
+                        name: z.literal("find"),
+                        arguments: z.object(FindArgs),
```
nirinchev: [nit] I missed this during the first review, but FindArgs has a default limit of 10 - if the user is asking for an export, perhaps that limit doesn't make much sense.

himanshusinghs: Yea the default certainly doesn't make sense. I had the default removed in the earlier iteration, will do the same here. Thanks for catching this.

himanshusinghs: addressed this in [04c4e9f](https://github.com/mongodb-js/mongodb-mcp-server/pull/464/commits/04c4e9fd0786c534aafba24158c8b1f41b2ae33a)

src/tools/mongodb/read/export.ts
```diff
@@ -1,19 +1,36 @@
...
-        limit: z.number().optional().describe("The maximum number of documents to return"),
+        exportTitle: z.string().describe("A short description to uniquely identify the export."),
+        exportTarget: z
+            .array(
+                z.discriminatedUnion("name", [
+                    z.object({
+                        name: z.literal("find"),
+                        arguments: z.object({
+                            ...FindArgs,
+                            limit: limitArg,
```
nirinchev: Instead of exposing the limit arg there, can't we do something like:

```typescript
arguments: z.object({
    ...FindArgs,
    limit: FindArgs.limit.removeDefault()
})
```

src/tools/mongodb/read/export.ts
```diff
@@ -1,19 +1,36 @@
...
     public name = "export";
     protected description = "Export a collection data or query results in the specified EJSON format.";
     protected argsShape = {
-        exportTitle: z.string().describe("A short description to uniquely identify the export."),
         ...DbOperationArgs,
-        ...FindArgs,
-        limit: z.number().optional().describe("The maximum number of documents to return"),
+        exportTitle: z.string().describe("A short description to uniquely identify the export."),
+        exportTarget: z
+            .array(
```
nirinchev, line 18: I wonder if there's a way to avoid using an array here - I remember that was a workaround, but perhaps a newer version of the server/zod has addressed that.

src/tools/mongodb/read/export.ts
```diff
@@ -30,24 +44,37 @@ export class ExportTool extends MongoDBToolBase {
...
-            projection,
-            sort,
-            limit,
-            promoteValues: false,
-            bsonRegExp: true,
-        });
+        const exportTarget = target[0];
+        if (!exportTarget) {
+            throw new Error("Export target not provided. Expected one of the following: `aggregate`, `find`");
+        }
```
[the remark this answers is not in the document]
himanshusinghs: This is following the [existing semantics](https://github.com/mongodb-js/mongodb-mcp-server/blob/c4ba2c912f43e1b447b96bc9f9f80fa9a154de5f/src/tools/mongodb/metadata/explain.ts#L15-L35). While changing to the suggested approach might be beneficial, it requires some effort in changing the test setup which is not worth it right now. I will drop a ticket for this.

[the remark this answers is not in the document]
himanshusinghs: [MCP-128](https://jira.mongodb.org/browse/MCP-128)

src/tools/mongodb/read/export.ts
```diff
@@ -30,24 +47,37 @@ export class ExportTool extends MongoDBToolBase {
...
+                sort,
+                limit,
+                promoteValues: false,
+                bsonRegExp: true,
+            });
+        } else {
+            const { pipeline } = exportTarget.arguments;
+            cursor = provider.aggregate(database, collection, pipeline, {
+                promoteValues: false,
+                bsonRegExp: true,
```
kmruiz, line 79: We want to add allowDiskUse here: https://www.mongodb.com/docs/manual/reference/command/aggregate/#command-fields

allowDiskUse is required for aggregations that need more than 100MB of data in memory (which is the likely situation for exports), as it spills the exceeding into disk. It's slower, but without allowDiskUse, the whole aggregate is rejected.

nirinchev (approved):
Looks good - I wonder if we want to write some accuracy tests for this.

src/tools/mongodb/read/export.ts
himanshusinghs: Nice, didn't know this was possible.

src/tools/mongodb/read/export.ts
himanshusinghs: Got it. I will have it right away.

himanshusinghs: addressed

src/tools/mongodb/read/export.ts
himanshusinghs: addressed

src/tools/mongodb/read/export.ts
himanshusinghs: We can already use Zod's discriminated unions without wrapping them in array. I thought this was done to avoid having to expand the `validateMetadata` tests back then because that test function is currently strictly bound to work with normal parameters. I also brushed off that effort only because it was not that important at the moment and instead created a ticket [MCP-128](https://jira.mongodb.org/browse/MCP-128) to follow up on that.

PR #466, opened 2025-08-20
fix: fix expectations for export accuracy tests

nirinchev:
## Proposed changes

Looks like we had some tests failing because the model was providing the export title but we weren't expecting it.

## Checklist


himanshusinghs (approved):
Weird, I ran them locally and it seemed fine 🤔

PR #467, opened 2025-08-21
fix: minor tweaks to enable vscode integration MCP-134

nirinchev:
## Proposed changes

This tweaks slightly some of the API around streamable http runner to allow it to be consumed by the vscode extension. Notably:
- We add httpHeaders config option that allows consumers to specify required headers the server should check for in the client requests. This will be used by the vscode extension to generate a random header when it starts the server.
- Exports some more types for consumers (e.g. `StreamableHttpRunner` and `defaultUserConfig`)
- Allows running the http server on a random port and exposes the address of the server.


src/transports/streamableHttp.ts
```diff
@@ -32,6 +44,17 @@ export class StreamableHttpRunner extends TransportRunnerBase {
 
         app.enable("trust proxy"); // needed for reverse proxy support
         app.use(express.json());
+        app.use((req, res, next) => {
+            for (const [key, value] of Object.entries(this.userConfig.httpHeaders)) {
+                const header = req.headers[key];
```
himanshusinghs: We should lowercase the key before comparison.

himanshusinghs (approved):

src/helpers/deviceId.ts
```diff
@@ -1,94 +1,63 @@
...
     private timeout: number;
-    private static instance: DeviceId | undefined = undefined;
 
     private constructor(logger: LoggerBase, timeout: number = DEVICE_ID_TIMEOUT) {
         this.logger = logger;
         this.timeout = timeout;
         this.getMachineId = (): Promise<string> => nodeMachineId.machineId(true);
+        this.abortController = new AbortController();
+
+        this.deviceIdPromise = DeviceId.UnknownDeviceId;
```
gagik, line 23:
```suggestion
    private constructor({logger, timeout, deviceIdPromise, abortController}: {logger: LoggerBase, timeout: number = DEVICE_ID_TIMEOUT, deviceIdPromise: Promise<string>, abortController: AbortController}) {
        this.logger = logger;
        this.timeout = timeout;
        this.abortController = abortController;
        this.deviceIdPromise = deviceIdPromise;
```
I think this is better because it'd be good to have the Promise resolve to unknown only in the end so we can always have a state of `pending -> unknown` and not `unknown -> pending -> unknown` (even though in practice we instantly override this in `create` right now

src/helpers/deviceId.ts
```diff
@@ -1,94 +1,63 @@
...
+            getMachineId: this.getMachineId,
+            onError: (reason, error) => {
+                this.handleDeviceIdError(reason, String(error));
+            },
+            timeout: this.timeout,
+            abortSignal: this.abortController.signal,
+        });
+    }
 
+    public static create(logger: LoggerBase, timeout?: number): DeviceId {
```
gagik, line 37:
```suggestion
    public static create(logger: LoggerBase, { timeout = DEVICE_ID_TIMEOUT } : { timeout: number} = {}): DeviceId {
    	const abortController = new AbortController();
    	const deviceIdPromise = () => getDeviceId({
            getMachineId: () => nodeMachineId.machineId(true),
            onError: (reason, error) => {
                this.handleDeviceIdError(reason, String(error));
            },
            timeout: this.timeout,
            abortSignal: abortController.signal,
        });
        const instance = new DeviceId({logger, timeout, deviceIdPromise, abortController});
```
personal preference re: timeout as object because it's one of these things which we basically largely have only for test DI purposes

src/transports/streamableHttp.ts
```diff
@@ -18,8 +19,20 @@ export class StreamableHttpRunner extends TransportRunnerBase {
     private httpServer: http.Server | undefined;
     private sessionStore!: SessionStore;
 
-    constructor(userConfig: UserConfig, driverOptions: DriverOptions) {
-        super(userConfig, driverOptions);
+    public get address(): string {
```
blva: nit: `serverAddress` ?

src/helpers/deviceId.ts
nirinchev: Unfortunately, we can't supply the promise through the constructor since we need the `DeviceId` instance to construct the promise itself. Technically, the only reason to have it assigned here is to make the types nicer and ensure we don't have to write `?` all over the place when we know for sure the field is initialized at this point. So we have 3 options (with no preference which one to go for):
1. Keep it as is
2. Suppress the error that we're not initializing a required field
3. Move the getDeviceId call to the constructor (and close our eyes for the fact it's introducing a side effect)

PR #470, opened 2025-08-22
chore: enable eslint rule for consistent type import/export MCP-130

himanshusinghs:
## Proposed changes

This is part of the effort for introducing an eslint rule to only use injected config in source code. This PR enables consistent type import/export rule on which the next rule will build upon.

## Checklist

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -48,6 +48,13 @@ export default defineConfig([
         rules: {
             "@typescript-eslint/switch-exhaustiveness-check": "error",
             "@typescript-eslint/no-non-null-assertion": "error",
+            "@typescript-eslint/consistent-type-imports": ["error", { prefer: "type-imports" }],
+            "@typescript-eslint/consistent-type-exports": [
+                "error",
+                {
+                    fixMixedExportsWithInlineTypeSpecifier: false,
+                },
+            ],
             eqeqeq: "error",
             "no-self-compare": "error",
             "no-unassigned-vars": "error",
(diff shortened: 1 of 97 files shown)


eslint.config.js
```diff
@@ -48,6 +48,13 @@ export default defineConfig([
         rules: {
             "@typescript-eslint/switch-exhaustiveness-check": "error",
             "@typescript-eslint/no-non-null-assertion": "error",
+            "@typescript-eslint/consistent-type-imports": ["error", { prefer: "type-imports" }],
+            "@typescript-eslint/consistent-type-exports": [
```
nirinchev, line 52: Should we enable --verbatimModuleSyntax for the typescript compiler then?

himanshusinghs: I just read about this and it makes sense. I will enable it and check if there's any side effects.

himanshusinghs: Well there is problem with cjs build with this flag enabled. From the documentation:

> Under this flag, ECMAScript imports and exports won’t be rewritten to require calls when your settings or file extension implied a different module system. Instead, you’ll get an error. If you need to emit code that uses require and module.exports, you’ll have to use TypeScript’s module syntax that predates ES2015:

And error is what I am getting. We can disable the `verbatimModuleSyntax` just for CJS or have it disabled for both. I would prefer former just because CJS is an exclusive case and very likely intermittent until we update vscode's build.

nirinchev: I'm fine either way - I don't think it's super important, just figured we could give it a try.

himanshusinghs: then I will enable it for ESM and leave it disabled for CJS.

nirinchev (approved):

PR #471, opened 2025-08-22
fix: atlas connectCluster defaults to readOnly DB roles - MCP-125

blva:
## Proposed changes

- Defaults to readOnly DB user role unless all tools are enabled

## Checklist

### DIFF
--- a/src/common/atlas/roles.ts
+++ b/src/common/atlas/roles.ts
@@ -13,12 +13,11 @@ export function getDefaultRoleFromConfig(config: UserConfig): DatabaseUserRole {
         };
     }
 
-    // If all write tools are enabled, use readWriteAnyDatabase
+    // If any of the write tools are enabled, use readWriteAnyDatabase
     if (
-        !config.disabledTools?.includes("create") &&
-        !config.disabledTools?.includes("update") &&
-        !config.disabledTools?.includes("delete") &&
-        !config.disabledTools?.includes("metadata")
+        !config.disabledTools?.includes("create") ||
+        !config.disabledTools?.includes("update") ||
+        !config.disabledTools?.includes("delete")
     ) {
         return {
             roleName: "readWriteAnyDatabase",
--- a/src/common/atlas/roles.ts
+++ b/src/common/atlas/roles.ts
@@ -1,32 +1,33 @@
 import type { UserConfig } from "../config.js";
 import type { DatabaseUserRole } from "./openapi.js";
 
+const readWriteRole: DatabaseUserRole = {
+    roleName: "readWriteAnyDatabase",
+    databaseName: "admin",
+};
+
+const readOnlyRole: DatabaseUserRole = {
+    roleName: "readAnyDatabase",
+    databaseName: "admin",
+};
+
 /**
  * Get the default role name for the database user based on the Atlas Admin API
  * https://www.mongodb.com/docs/atlas/mongodb-users-roles-and-privileges/
  */
 export function getDefaultRoleFromConfig(config: UserConfig): DatabaseUserRole {
     if (config.readOnly) {
-        return {
-            roleName: "readAnyDatabase",
-            databaseName: "admin",
-        };
+        return readOnlyRole;
     }
 
     // If any of the write tools are enabled, use readWriteAnyDatabase
     if (
-        !config.disabledTools?.includes("create") ||
-        !config.disabledTools?.includes("update") ||
-        !config.disabledTools?.includes("delete")
+        !config.disabledTools.includes("create") ||
+        !config.disabledTools.includes("update") ||
+        !config.disabledTools.includes("delete")
     ) {
-        return {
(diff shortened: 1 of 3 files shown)


src/common/atlas/roles.ts
```diff
@@ -0,0 +1,33 @@
...
+ */
+export function getDefaultRoleFromConfig(config: UserConfig): DatabaseUserRole {
+    if (config.readOnly) {
+        return {
+            roleName: "readAnyDatabase",
+            databaseName: "admin",
+        };
+    }
+
+    // If all write tools are enabled, use readWriteAnyDatabase
```
nirinchev: This is a bit hard to follow due to the double negation, but I don't think this is the right condition to check - if any write tool is enabled, we should default to readWrite, otherwise we're risking some tools to not work as intended. For example, if I want to be able to insert new data, but not delete existing documents, I may disable delete tools, but I'll still need the readWriteAnyDatabase user.

src/tools/atlas/connect/connectCluster.ts
```diff
@@ -92,12 +84,7 @@ export class ConnectClusterTool extends AtlasToolBase {
             body: {
                 databaseName: "admin",
```
nirinchev, line 85: This is unrelated, but since you're editing this file already, can we add a description for the user? That was one of the recommendations by the security team as it would clearly indicate how that user was created and why.

src/common/atlas/roles.ts
blva: > This is a bit hard to follow due to the double negation, but I don't think this is the right condition to check - if any write tool is enabled, we should default to readWrite, otherwise we're risking some tools to not work as intended. 

Oh, got it! I thought that we were trying to avoid any writes if any of them are disabled. 

> For example, if I want to be able to insert new data, but not delete existing documents, I may disable delete tools, but I'll still need the readWriteAnyDatabase user.

Fair, I thought we wanted to be more strict, but I agree that this would just cause people to enable all tools

src/tools/atlas/connect/connectCluster.ts
blva: added!

src/common/atlas/roles.ts
```diff
@@ -0,0 +1,32 @@
...
+    if (config.readOnly) {
+        return {
+            roleName: "readAnyDatabase",
+            databaseName: "admin",
+        };
+    }
+
+    // If any of the write tools are enabled, use readWriteAnyDatabase
+    if (
+        !config.disabledTools?.includes("create") ||
```
nirinchev: `disabledTools` should never be undefined - should it? I don't believe we need the optional chaining here.

src/common/atlas/roles.ts
```diff
@@ -0,0 +1,32 @@
...
+        !config.disabledTools?.includes("update") ||
+        !config.disabledTools?.includes("delete")
+    ) {
+        return {
+            roleName: "readWriteAnyDatabase",
+            databaseName: "admin",
+        };
```
nirinchev: [nit] consider extracting the two roles to constants and referencing them here. Something like:

```typescript
const readWriteRole: DatabaseUserRole = {
   ...
};

const readOnlyRole: DatabaseUserRole = {
   ....
};

export function getDefaultRoleFromConfig(config: UserConfig): DatabaseUserRole {
  if (config.readOnly) {
    return readOnlyRole;
  }

  if (...) {
    return readWriteRole;
  }

  return readOnlyRole;
}

```

src/tools/atlas/connect/connectCluster.ts
```diff
@@ -106,6 +93,8 @@ export class ConnectClusterTool extends AtlasToolBase {
                 oidcAuthType: "NONE",
                 x509Type: "NONE",
                 deleteAfterDate: expiryDate.toISOString(),
+                description:
+                    "This role is created by MongoDB MCP to connect to the cluster via the atlas-connect-cluster tool",
```
nirinchev: How about slightly rewording this like:

> This temporary user is created by the MongoDB MCP Server to connect to the cluster. For more details, see https://...

Not sure what the final docs url will be, but there's a ticket to document the internal behavior: https://jira.mongodb.org/browse/DOCSP-53007. Might ping David and see if we can get a url, even if the actual content is not there yet.

nirinchev (approved):
Final nits from me, otherwise looks good.

src/common/atlas/roles.ts
blva: better! done

src/tools/atlas/connect/connectCluster.ts
blva: I'll follow up on that link but for now will do `This temporary user is created by the MongoDB MCP Server to connect to the cluster.`

src/common/atlas/roles.ts
blva: done!

PR #472, opened 2025-08-22
feat: MongoDB Assistant Knowledge Search Tools

nlarew:
## Proposed changes

[(EAI-1266) Knowledge Search on MCP](https://jira.mongodb.org/browse/EAI-1266)

Adds a new class of `Assistant` tools for interacting with the MongoDB Assistant (AI Chatbot + Search).

This PR scaffolds the generic `AssistantToolBase` class as well as two initial tools:

- `list_knowledge_sources` - pulls a list of all available data sources. Useful for knowing what data is available and for filtering the `search_knowledge`
- `search_knowledge` - runs a natural language search against the MongoDB Assistant knowledge base. Returns content chunks for the most relevant results.

## Checklist

### DIFF
--- a/src/tools/assistant/listKnowledgeSources.ts
+++ b/src/tools/assistant/listKnowledgeSources.ts
@@ -3,7 +3,6 @@ import { formatUntrustedData, type OperationType, type ToolCategory } from "../t
 import { AssistantToolBase } from "./assistantTool.js";
 import { LogId } from "../../common/logger.js";
 import { stringify as yamlStringify } from "yaml";
-import { SearchKnowledgeTool } from "./searchKnowledge.js";
 
 export type KnowledgeSource = {
     /** The name of the data source */
@@ -29,7 +28,7 @@ export class ListKnowledgeSourcesTool extends AssistantToolBase {
     static toolName = ListKnowledgeSourcesToolName;
     static category: ToolCategory = "assistant";
     static operationType: OperationType = "read";
-    public description = `List available data sources in the MongoDB Assistant knowledge base. Use this to explore available data sources or to find search filter parameters to use in ${SearchKnowledgeTool.toolName}.`;
+    public description = `List available data sources in the MongoDB Assistant knowledge base. Use this to explore available data sources or to find search filter parameters to use in search-knowledge.`;
     public argsShape = {};
 
     protected async execute(): Promise<CallToolResult> {
--- a/src/tools/assistant/searchKnowledge.ts
+++ b/src/tools/assistant/searchKnowledge.ts
@@ -4,7 +4,6 @@ import { type ToolArgs, type OperationType, type ToolCategory, formatUntrustedDa
 import { AssistantToolBase } from "./assistantTool.js";
 import { LogId } from "../../common/logger.js";
 import { stringify as yamlStringify } from "yaml";
-import { ListKnowledgeSourcesToolName } from "./listKnowledgeSources.js";
 
 export const SearchKnowledgeToolArgs = {
     query: z
@@ -22,7 +21,7 @@ export const SearchKnowledgeToolArgs = {
         )
         .optional()
         .describe(
-            `A list of one or more data sources to limit the search to. You can specify a specific version of a data source by providing the version label. If not provided, the latest version of all data sources will be searched. Available data sources and their versions can be listed by calling the ${ListKnowledgeSourcesToolName} tool.`
+            `A list of one or more data sources to limit the search to. You can specify a specific version of a data source by providing the version label. If not provided, the latest version of all data sources will be searched. Available data sources and their versions can be listed by calling the list-knowledge-sources tool.`
         ),
 };
 
@@ -51,7 +50,8 @@ export class SearchKnowledgeTool extends AssistantToolBase {
     static toolName = SearchKnowledgeToolName;
     static category: ToolCategory = "assistant";
     static operationType: OperationType = "read";
-    public description = "Search for information in the MongoDB Assistant knowledge base";
+    public description =
+        "Search for information in the MongoDB Assistant knowledge base. This includes official documentation, curated expert guidance, and other resources provided by MongoDB. Supports filtering by data source and version.";
     public argsShape = {
         ...SearchKnowledgeToolArgs,
     };
--- a/tests/integration/transports/stdio.test.ts
+++ b/tests/integration/transports/stdio.test.ts
@@ -32,7 +32,7 @@ describeWithMongoDB("StdioRunner", (integration) => {
             const response = await client.listTools();
             expect(response).toBeDefined();
             expect(response.tools).toBeDefined();
-            expect(response.tools).toHaveLength(22);
+            expect(response.tools).toHaveLength(24);
 
             const sortedTools = response.tools.sort((a, b) => a.name.localeCompare(b.name));
             expect(sortedTools[0]?.name).toBe("aggregate");
--- a/tests/unit/assistant/searchKnowledge.test.ts
+++ b/tests/unit/assistant/searchKnowledge.test.ts
@@ -21,7 +21,7 @@ describeWithAssistant("search-knowledge", (integration) => {
     validateToolMetadata(
         integration,
         "search-knowledge",
-        "Search for information in the MongoDB Assistant knowledge base",
+        "Search for information in the MongoDB Assistant knowledge base. This includes official documentation, curated expert guidance, and other resources provided by MongoDB. Supports filtering by data source and version.",
         "read",
         [
             {


src/tools/assistant/searchKnowledge.ts
```diff
@@ -0,0 +1,68 @@
+import { z } from "zod";
+import { CallToolResult } from "@modelcontextprotocol/sdk/types.js";
+import { ToolArgs, OperationType } from "../tool.js";
+import { AssistantToolBase } from "./assistantTool.js";
+
+export const SearchKnowledgeToolArgs = {
+    query: z.string().describe("A natural language query to search for in the knowledge base"),
+    limit: z.number().min(1).max(100).optional().default(5).describe("The maximum number of results to return"),
```
mongodben, line 14:
```suggestion
    limit: z.number().min(1).max(100).optional().default(10).describe("The maximum number of results to return"),
```

i've seen 10 chunks in a bunch of other implementations. we dont do this in our chatbot, but that's largely for latency/price

src/tools/assistant/assistantTool.ts
```diff
@@ -0,0 +1,47 @@
...
+    public register(server: Server): boolean {
+        this.server = server;
+        return super.register(server);
+    }
+
+    protected resolveTelemetryMetadata(
+        // eslint-disable-next-line @typescript-eslint/no-unused-vars
+        args: ToolArgs<typeof this.argsShape>
+    ): TelemetryToolMetadata {
+        return {};
```
nlarew, line 29: Not sure what if anything I should have here - would appreciate advice from the DevTools team on this.

src/tools/assistant/searchKnowledge.ts
nlarew: I figured this should match the default of `5` we have on our server, no? Open to changing that though.

https://github.com/mongodb/chatbot/blob/003cec9b918c8e2ff3c87696021e1d6488c95b98/packages/mongodb-chatbot-server/src/routes/content/searchContent.ts#L27

src/tools/assistant/assistantTool.ts
```diff
@@ -0,0 +1,47 @@
...
+        protected readonly config: UserConfig,
+        protected readonly telemetry: Telemetry
+    ) {
+        super(session, config, telemetry);
+        this.baseUrl = new URL(config.assistantBaseUrl);
+        const serverVersion = packageInfo.version;
+        this.requiredHeaders = {
+            "x-request-origin": "mongodb-mcp-server",
+            "user-agent": serverVersion ? `mongodb-mcp-server/v${serverVersion}` : "mongodb-mcp-server",
+        };
```
blva: can we try to use the `apiClient` to make this call? do you need specific headers or can you leverage that to make requests? Check `getIpInfo` in that file for a customized call

Also, is this page not authenticated?

src/tools/assistant/listKnowledgeSources.ts
```diff
@@ -0,0 +1,51 @@
+import { z } from "zod";
```
blva: nit: could we rename the files to camelCase for consistency?

src/tools/assistant/list_knowledge_sources.ts
```diff
@@ -0,0 +1,51 @@
...
+    public operationType: OperationType = "read";
+
+    protected async execute(): Promise<CallToolResult> {
+        const searchEndpoint = new URL("content/sources", this.baseUrl);
+        const response = await fetch(searchEndpoint, {
+            method: "GET",
+            headers: this.requiredHeaders,
+        });
+        if (!response.ok) {
+            throw new Error(`Failed to list knowledge sources: ${response.statusText}`);
```
blva: can we instead return `content` instead of throwing an exception? we've been changing this across the code. it's better to have the tool return a response but with the proper messaging

src/tools/assistant/searchKnowledge.ts
```diff
@@ -0,0 +1,68 @@
...
+        ...SearchKnowledgeToolArgs,
+    };
+    public operationType: OperationType = "read";
+
+    protected async execute(args: ToolArgs<typeof this.argsShape>): Promise<CallToolResult> {
+        const searchEndpoint = new URL("content/search", this.baseUrl);
+        const response = await fetch(searchEndpoint, {
+            method: "POST",
+            headers: new Headers({ ...this.requiredHeaders, "Content-Type": "application/json" }),
+            body: JSON.stringify(args),
```
blva: same as before, let's keep api calls under the client for proper abtraction and consistency

src/tools/tool.ts
```diff
@@ -12,7 +12,7 @@ import type { Server } from "../server.js";
 export type ToolArgs<Args extends ZodRawShape> = z.objectOutputType<Args, ZodNever>;
 
 export type OperationType = "metadata" | "read" | "create" | "delete" | "update" | "connect";
-export type ToolCategory = "mongodb" | "atlas";
+export type ToolCategory = "mongodb" | "atlas" | "assistant";
```
blva: since we're here, can we also update the readme.md? there we call out all the tool categories

src/tools/assistant/search_knowledge.ts
```diff
@@ -0,0 +1,68 @@
...
+            .passthrough(),
+    })
+    .passthrough();
+
+export const searchResponseSchema = z.object({
+    results: z.array(knowledgeChunkSchema).describe("A list of search results"),
+});
+
+export class SearchKnowledgeTool extends AssistantToolBase {
+    public name = "search_knowledge";
```
blva: nit: all out tool names use `-` instead of `_` could we update them for consistency?

src/tools/assistant/assistantTool.ts
nlarew: It looks like the `apiClient` is specifically for talking to Atlas/MMS APIs? The knowledge APIs are a distinct server so I don't think it will work for this.

> do you need specific headers

Yeah we must include, at minimum, a `user-agent` + `origin` or `x-request-origin` header

> is this page not authenticated

correct - the knowledge API is not authenticated

src/tools/assistant/assistantTool.ts
nlarew: The `TelemetryToolMetadata` type seems to track a specific Atlas org/project but the assistant is not associated with a given Atlas instance. Planning to leave this empty unless someone has a good idea of what to put here.

src/tools/assistant/list_knowledge_sources.ts
nlarew: Yes! Updated for both assistant tools.

src/tools/assistant/assistantTool.ts
```diff
@@ -0,0 +1,48 @@
...
+        // Therefore, we don't have any values to add to the telemetry metadata
+        return {};
+    }
+
+    protected handleError(
+        error: unknown,
+        args: ToolArgs<typeof this.argsShape>
+    ): Promise<CallToolResult> | CallToolResult {
+        return super.handleError(error, args);
+    }
```
nirinchev: You don't need this if you're not adding any special error handling.

src/tools/assistant/listKnowledgeSources.ts
```diff
@@ -0,0 +1,66 @@
...
+                content: [
+                    {
+                        type: "text",
+                        text: message,
+                    },
+                ],
+                isError: true,
+            };
+        }
+        const { dataSources } = listDataSourcesResponseSchema.parse(await response.json());
```
nirinchev: Can we abstract this and move it to the base tool? Essentially, expose a function like:

```ts
protected function fetchKnowledgeBase<T>(method: "GET" | "POST", path: string, body?: unknown): Promise<T>
```

that calls fetch, parses the response, logs errors, etc. Additionally, we tend to prefer working with pure typescript types rather than zod schemas for API calls. Those are easier to reason about and more succinct to type.

nlarew: Good call on moving the API call up one level of abstraction - done.

I don't follow the "no zod on API calls" logic. A network boundary is exactly where validation libraries like Zod are most useful - otherwise I have to assert the type (unsafe) or roll my own validation (clunkier than zod).

nirinchev: Right - the main thing here is that we're in control of the API, so a simple type assertion is reasonable.

nlarew: Well probably not how I'd handle it but also not my repo! Updated to remove the zod parse (and corresponding tests that check if we handle a malformed response) and replaced it with a vanilla assertion.

(Note - using `as` instead of `:` assertion to be more explicit that this is escape hatching. Both are equivalent since `response.json()` returns `Promise<any>`.)

src/tools/assistant/assistantTool.ts
```diff
@@ -0,0 +1,51 @@
...
+        return {};
+    }
+
+    protected async callAssistantApi(args: { method: "GET" | "POST"; endpoint: string; body?: unknown }) {
+        const endpoint = new URL(args.endpoint, this.baseUrl);
+        const headers = new Headers(this.requiredHeaders);
+        if (args.method === "POST") {
+            headers.set("Content-Type", "application/json");
+        }
+        return await fetch(endpoint, {
```
kmruiz: We shouldn't use Node.js fetch, but customFetch configured like in the Atlas Client:

https://github.com/mongodb-js/mongodb-mcp-server/blob/04a5ddca095f9ed5433304ed7e41a8ce5e911309/src/common/atlas/apiClient.ts#L41-L47

The reason is that the customFetch supports enterprise proxies, which are relevant for customers with specific security requirements.

src/tools/assistant/listKnowledgeSources.ts
```diff
@@ -0,0 +1,64 @@
...
+        return {
+            content: dataSources.map(({ id, type, versions }) => ({
+                type: "text",
+                text: id,
+                _meta: {
+                    type,
+                    versions,
+                },
+            })),
+        };
```
kmruiz, line 70: We should convert this into a string, similar to what we do in:

https://github.com/mongodb-js/mongodb-mcp-server/blob/04a5ddca095f9ed5433304ed7e41a8ce5e911309/src/tools/mongodb/read/find.ts#L93-L103

And also we should consider the information in the knowledge list untrusted.

src/tools/assistant/searchKnowledge.ts
```diff
@@ -0,0 +1,84 @@
...
+                        type: "text",
+                        text: message,
+                    },
+                ],
+                isError: true,
+            };
+        }
+        const { results } = (await response.json()) as SearchKnowledgeResponse;
+        return {
+            content: results.map(({ text, metadata }) => ({
```
kmruiz: Same here, convert to string and treat it as untrusted data. We have a function for this:

https://github.com/mongodb-js/mongodb-mcp-server/blob/main/src/tools/tool.ts#L293

src/tools/assistant/assistantTool.ts
blva: should we move this into apiClient? we also have telemetry events logic there, it would be good to keep all the api related logic there for now

src/tools/assistant/searchKnowledge.ts
```diff
@@ -0,0 +1,84 @@
+import { z } from "zod";
+import { CallToolResult } from "@modelcontextprotocol/sdk/types.js";
+import { ToolArgs, OperationType } from "../tool.js";
+import { AssistantToolBase } from "./assistantTool.js";
+import { LogId } from "../../common/logger.js";
+
+export const SearchKnowledgeToolArgs = {
+    query: z.string().describe("A natural language query to search for in the knowledge base"),
```
blva: nit: can we add a bit more detail here? this is what the LLM will use to understand that you want it to make a generic natural question that the user haave.

src/tools/assistant/assistantTool.ts
```diff
@@ -0,0 +1,51 @@
...
+    protected requiredHeaders: Headers;
+
+    constructor({ session, config, telemetry, elicitation }: ToolConstructorParams) {
+        super({ session, config, telemetry, elicitation });
+        this.baseUrl = new URL(config.assistantBaseUrl);
+        const serverVersion = packageInfo.version;
+        this.requiredHeaders = new Headers({
+            "x-request-origin": "mongodb-mcp-server",
+            "user-agent": serverVersion ? `mongodb-mcp-server/v${serverVersion}` : "mongodb-mcp-server",
+        });
```
blva, line 18: [q] why does this need to be initialized here at the tool level? can we also move this logic into the api layer? 

also, we should probably use the `userAgent` from the apiClient and make sure we use the same name everywhere

tests/unit/assistant/assistantHelpers.ts
```diff
@@ -0,0 +1,83 @@
...
+    };
+
+    // eslint-disable-next-line vitest/valid-describe-callback
+    return describe("assistant", testDefinition);
+}
+
+/**
+ * Mocks fetch for assistant API calls
+ */
+interface MockedAssistantAPI {
```
blva, line 28: nit: if we're mocking the api calls i'd rather call these unit tests, could you please move the tests to the unit folder and the helpers? 

if you have any staging / cloud-dev environment, we could configure integration tests as well, let me know on slack if you do as I can help setting up a GH variable for the staging environment, otherwise we could keep it with unit tests only which is fine

src/tools/assistant/assistantTool.ts
nlarew: Ok no problem - I will create a new customFetch instance in the assistant tool base because we don't need all of the extra baggage from the Atlas `ApiClient` class when working with this non-Atlas API.

src/tools/assistant/listKnowledgeSources.ts
nlarew: Looks like `formatUntrustedData` maps to a single `CallToolResult["content"]` object so I will need to convert the array of results I get from each API call into a single formatted string.

src/tools/assistant/assistantTool.ts
nlarew: > why does this need to be initialized here at the tool level? can we also move this logic into the api layer?

The "API layer" as it exists seems to be tailored to Atlas - this is a totally separate server/API. If we prefer I could create an equivalent `assistant/apiClient.ts` file but personally that seems unnecessary at this point.

> we should probably use the userAgent from the apiClient

1. Our server firewall allowlists specific user agents and right now that's configured to be `mongodb-mcp-server`
2. nit: The `apiClient` uses `User-Agent: AtlasMCP` which again seems to refer to Atlas specifically

tests/unit/assistant/assistantHelpers.ts
nlarew: Gotcha that's totally reasonable. I'll move these mocked tests to `tests/unit`. It's a bit awkward because I think I still want the integration test flow (i.e. `mcpClient.callTool({ ... })`) but I'd like to use the mocks to control the output to confirm that everything is formatted/handled appropriately.

For integration tests, we could try hitting `https://knowledge-dev.mongodb.com`. What GH variables do we need to support this?

src/tools/assistant/listKnowledgeSources.ts
nlarew: Do we want to wrap API error text in `formatUntrustedData` as well? Not sure if that's already mitigated by the `isError: true` or not.

i.e. instead of

```ts
if (!response.ok) {
    return {
        content: [
            {
                type: "text",
                text: `Failed to list knowledge sources: ${response.statusText}`,
            },
        ],
        isError: true,
    };
}
```

we would have

```ts
if (!response.ok) {
    return {
        content: formatUntrustedData("Failed to list knowledge sources", response.statusText),
        isError: true,
    };
}
```

src/tools/assistant/searchKnowledge.ts
nlarew: Updated!

src/tools/assistant/listKnowledgeSources.ts
nirinchev: I don't believe the LLM will employ any injection protection measures regardless of whether the response is marked as error-ed or not. That being said, do we expect that `statusText` would contain any instructions - this should be just the textual representation of the status code, shouldn't it? I wouldn't expect there to be a way to inject anything malicious in it.

tests/unit/assistant/assistantHelpers.ts
blva: I created an env variable called `TEST_ASSISTANT_BASE_URL` in GH actions, you can make it so your integration tests get configured to that if the env is set. 

here's an atlas example. you'd need to add the logic for using it as well
```
    - name: Run tests
        env:
          MDB_MCP_API_CLIENT_ID: ${{ secrets.TEST_ATLAS_CLIENT_ID }}
          MDB_MCP_API_CLIENT_SECRET: ${{ secrets.TEST_ATLAS_CLIENT_SECRET }}
          MDB_MCP_API_BASE_URL: ${{ vars.TEST_ATLAS_BASE_URL }}
        run: npm test -- tests/integration/tools/atlas
```

kmruiz:
The only thing missing from my side in this PR is to add some accuracy tests, to make sure that the models we test with do know when to call this tool. In the tests/accuracy folder there are a few examples that can be used as inspiration.

The idea is to write a few prompts that we expect to trigger our tool and ensure that the tool is called with the specific parameters.

src/tools/assistant/listKnowledgeSources.ts
nlarew: Yeah I think `statusCode` should only ever be something reasonable like `Ok`, `Not Found`, etc.

luislobo:
Good to see there is work being done on this area.

As I mentioned on the issue I created, I have always found myself going through the documentation, but also making sure that what I'm reading is available on the version that I have installed. It is not always the case that people are on the latest of the latest, so would there be a way to "hint" the MCP about the version you need the documentation to be?

src/tools/assistant/listKnowledgeSources.ts
```diff
@@ -0,0 +1,74 @@
...
+
+export type ListKnowledgeSourcesResponse = {
+    dataSources: KnowledgeSource[];
+};
+
+export const ListKnowledgeSourcesToolName = "list-knowledge-sources";
+
+export class ListKnowledgeSourcesTool extends AssistantToolBase {
+    public name = ListKnowledgeSourcesToolName;
+    protected description = "List available data sources in the MongoDB Assistant knowledge base";
```
wtrocki: When testing locally I was able to trigger it explicitly by "listing knowledge topics.
After extending description a little bit I was able to call this tool more accurately.

"list mongodb topics for me"
```suggestion
    protected description = "List available data sources in the MongoDB Assistant knowledge base. Covers t knowledge topics for MongoDB database";
```

nlarew:
bump

gagik:
@nlarew if you have a chance to update the PR, it'd be great to get this in. Sorry for letting this get stale.
Would be cool to pass version into context as well.

src/tools/assistant/assistantTool.ts
```diff
@@ -0,0 +1,55 @@
...
+        return {};
+    }
+
+    protected async callAssistantApi(args: {
+        method: "GET" | "POST";
+        endpoint: string;
+        body?: unknown;
+    }): Promise<Response> {
+        const endpointUrl = new URL(args.endpoint, this.baseUrl);
+        const headers = new Headers(this.requiredHeaders);
```
gagik, line 38: is there a reason to re-create the headers instead of accessing them directly?

nlarew: This creates a new headers object that we can extend (e.g. by adding a `Content-Type` on the next line) without mutating the `requiredHeaders` object

package.json
```diff
@@ -171,6 +171,8 @@
     "oauth4webapi": "^3.8.0",
     "openapi-fetch": "^0.15.0",
     "ts-levenshtein": "^1.0.7",
+    "yaml": "^2.8.1",
+    "yargs-parser": "^22.0.0",
```
gagik: yargs-parser seems unnecessary, probably a leftover from rebase

gagik:
@nlarew build seems to be failing, not sure if there's a better way to share that constant. If there's no cleaner solution I personally don't mind having it duplicated/hardcoded even if it's hypothetically not enforced.

Other than that, tried it out locally and worked great! So after fixing this, should be good to merge and go ahead.

gagik:
Awesome, LGTM, thank you for all the help getting this in!

gagik (approved):

PR #473, opened 2025-08-25
chore: adds eslint rule to restrict import from config file MCP-130

himanshusinghs:
## Proposed changes

This PR adds an ESLint rule to disallow direct import of values from `config.ts` file in the sourcecode and instead expect it to be injected wherever needed.

Followed by a small refactor for injecting `driverOptions` in the Transport implementation.

References taken from compass's [no-leafygreen-outside-compass-components rule](https://github.com/mongodb-js/compass/blob/main/configs/eslint-plugin-compass/rules/no-leafygreen-outside-compass-components.js).

## Checklist


eslint-rules/no-config-imports.js
```diff
@@ -0,0 +1,80 @@
...
+                required: ["configFilePath"],
+            },
+        ],
+        messages: {
+            noConfigImports:
+                "Value imports from config.ts are not allowed. Use dependency injection instead. Only type imports are permitted.",
+        },
+    },
+    create(context) {
+        const options = context.options[0] || {};
```
jeroenvervaeke: q: Are options guaranteed to have at least one value? (Should we check the length of options before indexing it with `[0]`

himanshusinghs: I am expecting that the schema above would enforce this requirement but I will check anyways.

himanshusinghs: Unfortunately the schema does not enforce the requirement. I added a check for this just now.

kmruiz:
I do understand the request and I know where this comes from, but the userConfig doesn't change once it's created on start time, so it's safe to reuse it. While I usually dislike singletons, I think having the userConfig as a simple import (same way we use process.env) is fine most of the time.

However, if you still feel strongly confident that this is necessary, I think it's fine to keep it.

eslint-rules/no-config-imports.js
```diff
@@ -0,0 +1,86 @@
+"use strict";
+import path from "path";
+
+// Ref: https://eslint.org/docs/latest/extend/custom-rules
+export default {
+    meta: {
+        type: "problem",
+        docs: {
+            description:
+                "Disallows value imports from config.ts, with a few exceptions, to enforce dependency injection of the config.",
```
kmruiz, line 21: This description is not accurate. This is not about config.ts, this is about only allowing a specific file to be imported from a specific subset of files. This is more about visibility of files in a scope. I think this should be reworded.

eslint-rules/no-config-imports.js
```diff
@@ -0,0 +1,86 @@
...
+                        items: { type: "string" },
+                        description: "List of file paths that are allowed to import value exports from config.ts.",
+                    },
+                },
+                required: ["configFilePath"],
+            },
+        ],
+        messages: {
+            noConfigImports:
+                "Value imports from config.ts are not allowed. Use dependency injection instead. Only type imports are permitted.",
```
kmruiz, line 27: Same comment as above. This is not about config.ts only.

eslint-rules/no-config-imports.js
himanshusinghs: So the reason behind this PR is to prevent possible problems with direct imports from config file. With VSCode integration work, the extension will be providing its own config to the MCP server which is the one that should be used by the MCP server source instead of directly relying on exports from config.ts.

I think the description is appropriate as the purpose of this rule is indeed to discourage importing value exports from config.ts anywhere in the source code but only a few places where the config injection happens.

himanshusinghs: I re-read what you mentioned. Since the rule allow for configFilePath and an allow list, its usage can basically be expanded to any import and any allow list. With that in mind, it certainly is more than just `no-config-imports`. Since there is no requirement to have it generic right now, I will make it specific to config imports.

PR #474, opened 2025-08-25
chore: add dependency to kerberos MCP-89

kmruiz:
This is necessary for clients that can use kerberos. It's registered as an optional dependency so clients that can't use kerberos because they don't have the required system dependencies can still use the MCP Server w/o kerberos support.

## Checklist


nirinchev (approved):
Is there anything we should do to handle the case where kerberos cannot be installed or will the error messages of trying to use kerberos auth bubble up nicely? If it works automagically, then let's 🚢 🇮🇹

kmruiz:
We already show a decent error message when we can't load kerberos :D  that says that kerberos is not available and the system error (that comes from Kerberos itself): https://github.com/mongodb-js/mongosh/blob/main/packages/service-provider-node-driver/src/node-driver-service-provider.ts#L372C11-L374C14

PR #475, opened 2025-08-25
chore: add more information when there is a failure connecting MCP-92

kmruiz:
## Proposed changes

This adds more information to the debug resource whenever there is an issue connecting.

## Checklist


src/common/session.ts
```diff
@@ -136,13 +137,7 @@ export class Session extends EventEmitter<SessionEvents> {
...
 
     async connectToMongoDB(settings: ConnectionSettings): Promise<void> {
-        try {
-            await this.connectionManager.connect({ ...settings });
-        } catch (error: unknown) {
-            const message = error instanceof Error ? error.message : (error as string);
-            this.emit("connection-error", message);
-            throw error;
-        }
+        await this.connectionManager.connect({ ...settings });
```
kmruiz, line 140: The error is already emitted properly from ConnectionManager, so this is useless now.

src/resources/common/debug.ts
```diff
@@ -1,13 +1,13 @@
 import { ReactiveResource } from "../resource.js";
 import type { Telemetry } from "../../telemetry/telemetry.js";
 import type { Session, UserConfig } from "../../lib.js";
+import { AtlasClusterConnectionInfo, ConnectionStateErrored } from "../../common/connectionManager.js";
 
 type ConnectionStateDebuggingInformation = {
     readonly tag: "connected" | "connecting" | "disconnected" | "errored";
     readonly connectionStringAuthType?: "scram" | "ldap" | "kerberos" | "oidc-auth-flow" | "oidc-device-flow" | "x.509";
-    readonly oidcLoginUrl?: string;
```
kmruiz, line 8: If there is an error this information will likely be stale, so don't show it to the user. Show what we know is relevant only.

src/common/connectionManager.ts
```diff
@@ -109,6 +109,7 @@ export class ConnectionManager extends EventEmitter<ConnectionManagerEvents> {
 
         let serviceProvider: NodeDriverServiceProvider;
         let connectionInfo: ConnectionInfo;
+        let connectionStringAuthType: ConnectionStringAuthType = "scram";
```
blva, line 112: [q] for my own learning why do we set to `scram` initially?

src/common/session.ts
```diff
@@ -10,6 +10,7 @@ import type {
     ConnectionManager,
     ConnectionSettings,
     ConnectionStateConnected,
+    ConnectionStateErrored,
```
blva, line 13: nit: would it be valid to keep `ConnectionStateError`?

src/common/session.ts
```diff
@@ -28,7 +29,7 @@ export type SessionEvents = {
     connect: [];
     close: [];
     disconnect: [];
-    "connection-error": [string];
+    "connection-error": [ConnectionStateErrored];
```
blva, line 32: [q] I'm a bit confused by having `connection-errored` down below and `connection-error` here, could we keep it a single one?

src/common/connectionManager.ts
himanshusinghs: I think its because the [default inferred](https://github.com/mongodb-js/mongodb-mcp-server/pull/475/files#diff-f2868ae7aecc32593eab3b565097c879aee92af8503649750b334c79df072d14L293-L294) auth type is "scram".

Another question for the same line:
Couldn't we just do:
```
const connectionStringAuthType = ConnectionManager.inferConnectionTypeFromSettings(
                this.userConfig,
                connectionInfo
            );
```

right here? Is it because it can throw for malformed strings?

kmruiz: Ah, because the default authentication type in MongoDB unless something else is specified is "scram" (used/password auth). However, this is not really relevant because later it gets overwritten but this way we have a sane default and typescript doesn't complain about uninit variables.

himanshusinghs (approved):
Looks good, left a question for better understanding.

src/common/session.ts
kmruiz: Ah you mean that the name of the event is different as the name of the payload class? Not sure I really understand.

src/common/connectionManager.ts
kmruiz: Ah, @himanshusinghs , just saw your message. If you see, connectionInfo is initialised afterwards and it can fail (for example, when parsing a URL that is invalid) so sadly it needs to be done in 2 steps.

blva: thanks for the context! :D

src/common/session.ts
blva: I think I'm just finding it confusing that we have "connection-error" -> connectionErrored here and then on line 73 "connection-errored" -> "connection-error", I think we could keep the same names or make them a bit more different. I know what the code does, it's just that in isolation it bugs my brain. but that's a nitpick anyways :D

kmruiz: Ah I get it now! Yeah in connection manager we are using past tense and in session we are using present tense. I can change the names to be all present tense, so it's more consistent.

PR #476, opened 2025-08-26
chore: update user description

blva:
## Proposed changes
- Adds the URL about the temporary user role

## Checklist

### DIFF
--- a/src/tools/atlas/connect/connectCluster.ts
+++ b/src/tools/atlas/connect/connectCluster.ts
@@ -93,7 +93,8 @@ export class ConnectClusterTool extends AtlasToolBase {
                 oidcAuthType: "NONE",
                 x509Type: "NONE",
                 deleteAfterDate: expiryDate.toISOString(),
-                description: "This temporary user is created by the MongoDB MCP Server to connect to the cluster.",
+                description:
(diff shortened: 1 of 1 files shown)


kmruiz (approved):

src/tools/atlas/connect/connectCluster.ts
```diff
@@ -93,7 +93,8 @@ export class ConnectClusterTool extends AtlasToolBase {
                 oidcAuthType: "NONE",
                 x509Type: "NONE",
                 deleteAfterDate: expiryDate.toISOString(),
-                description: "This temporary user is created by the MongoDB MCP Server to connect to the cluster.",
+                description:
+                    "Temporary user, see https://dochub.mongodb.org/core/mongodb-mcp-server-tools-considerations",
```
nirinchev: I kind of like that it was mentioning it's created by the MCP server - is it getting too long now?

blva: yes :( I'm open to suggestions. dochub url is longer

kmruiz: I believe for a user wondering if it was created by the MCP Server, opening a URL that says it was created by the MCP Server is good enough. I would add, if we want to really be more clear, MCP at the beginning at that's it, something like:

`MCP Temporary user, see ...`

blva: it fits! actually, `MDB MCP Temporary user, see.. ` fits

kmruiz (approved):

PR #481, opened 2025-08-27
chore: allows ConnectionManager to be injected via TransportRunner MCP-131

himanshusinghs:
## Proposed changes

This PR modifies the interface for our TransportRunners to allow injecting a ConnectionManager implementation. This will enable VSCode integration by allowing VSCode extension to pass its own implementation of ConnectionManager.

To allow subclassing, an abstract ConnectionManager is extracted out from existing implementation with `connect`, `disconnect` as the abstract methods.

The VSCode implementation of ConnectionManager differs slightly from our implementation of ConnectionManager (MCPConnectionManager) which is why this interface modification was called for. The `VSCodeMCPConnectionManager.connect` is not as involved as our implementation, we literally borrow the connect options from an existing connection in VSCode extension and use those to create a new instance of NodeServiceProvider which essentially shares the connection state. Additionally that implementation also implements logic to switch to different connections on demand(when active connection in extension changes).

The VSCode PR is [here](https://github.com/mongodb-js/vscode/pull/1111) for reference.

## Checklist

### DIFF
--- a/tests/integration/transports/streamableHttp.test.ts
+++ b/tests/integration/transports/streamableHttp.test.ts
@@ -2,10 +2,10 @@ import { StreamableHttpRunner } from "../../../src/transports/streamableHttp.js"
 import { Client } from "@modelcontextprotocol/sdk/client/index.js";
 import { StreamableHTTPClientTransport } from "@modelcontextprotocol/sdk/client/streamableHttp.js";
 import { describe, expect, it, beforeAll, afterAll, beforeEach } from "vitest";
-import { config, driverOptions } from "../../../src/common/config.js";
+import { config } from "../../../src/common/config.js";
 import type { LoggerType, LogLevel, LogPayload } from "../../../src/common/logger.js";
 import { LoggerBase, LogId } from "../../../src/common/logger.js";
-import { MCPConnectionManager } from "../../../src/common/connectionManager.js";
+import { createMCPConnectionManager } from "../../../src/common/connectionManager.js";
 
 describe("StreamableHttpRunner", () => {
     let runner: StreamableHttpRunner;
@@ -29,9 +29,7 @@ describe("StreamableHttpRunner", () => {
         describe(description, () => {
             beforeAll(async () => {
                 config.httpHeaders = headers;
-                runner = new StreamableHttpRunner(config, ({ logger, deviceId }) =>
-                    Promise.resolve(new MCPConnectionManager(config, driverOptions, logger, deviceId))
-                );
+                runner = new StreamableHttpRunner(config);
                 await runner.start();
             });
 
@@ -112,9 +110,7 @@ describe("StreamableHttpRunner", () => {
         try {
             for (let i = 0; i < 3; i++) {
                 config.httpPort = 0; // Use a random port for each runner
-                const runner = new StreamableHttpRunner(config, ({ logger, deviceId }) =>
-                    Promise.resolve(new MCPConnectionManager(config, driverOptions, logger, deviceId))
-                );
+                const runner = new StreamableHttpRunner(config);
                 await runner.start();
                 runners.push(runner);
             }
@@ -143,12 +139,7 @@ describe("StreamableHttpRunner", () => {
 
         it("can provide custom logger", async () => {
             const logger = new CustomLogger();
-            const runner = new StreamableHttpRunner(
-                config,
-                ({ logger, deviceId }) =>
-                    Promise.resolve(new MCPConnectionManager(config, driverOptions, logger, deviceId)),
-                [logger]
-            );
+            const runner = new StreamableHttpRunner(config, createMCPConnectionManager, [logger]);
             await runner.start();
 
             const messages = logger.messages;


src/common/connectionManager.ts
```diff
@@ -71,142 +54,18 @@ export interface ConnectionManagerEvents {
...
-            const errorReason = error instanceof Error ? error.message : `${error as string}`;
-            this.changeState("connection-errored", {
-                tag: "errored",
-                errorReason,
-                connectedAtlasCluster: settings.atlas,
-            });
-            throw new MongoDBError(ErrorCodes.NotConnectedToMongoDB, errorReason);
-        }
-    }
+export interface MCPConnectParams {
```
kmruiz: nit: I personally prefer ConnectionSettings. MCPConnectParams is not accurate (vscode will have it's own connection manager) and the naming feels to coupled to the method name that is going to receive it.

src/common/connectionManager.ts
```diff
@@ -71,142 +54,18 @@ export interface ConnectionManagerEvents {
...
+    connectionString: string;
+    atlas?: AtlasClusterConnectionInfo;
+}
 
-    async disconnect(): Promise<ConnectionStateDisconnected | ConnectionStateErrored> {
-        if (this.state.tag === "disconnected" || this.state.tag === "errored") {
-            return this.state;
-        }
+export abstract class ConnectionManager<ConnectParams extends MCPConnectParams = MCPConnectParams> {
```
kmruiz: Why are we making this class generic over MCPConnectParams? It only affects the signature of a single method and we can actually make that method generic if we need to.

I personally feel the proposed design here is problematic, because when using inheritance, changing connection managers implementations without changing the actual calling code should work (Liskov), but this is not true with the actual implementation as both connection managers use different not compatible "connect" signatures.

All this hierarchy can be simplified by simply using, if we need, visibility rules:

1. ConnectionManager implements a new public method called `setupExistingConnection(connectionStatus: ConnectedConnectionState>`
2. When we use ConnectionManager outside vscode, we can use something like Omit<ConnectionManager, 'setupExistingConnection'>
3. When we use ConnectionManager within vscode, we see the whole interface.

We can even, if we want, have a single interface and a single implementation class:

1. ConnectionManager can be renamed to something like DefaultConnectionManager, a class as is with the new method.
2. ConnectionManager, an interface that doesn't expose setupExistingConnections

kmruiz: If we feel fancy, we can even use connect as an overloaded method that can either receive ConnectionSetting or a ConnectedConnectionState so we don't need to play with the interfaces.

src/transports/base.ts
```diff
@@ -7,16 +7,21 @@ import { McpServer } from "@modelcontextprotocol/sdk/server/mcp.js";
...
+import type { ConnectionManager, MCPConnectParams } from "../common/connectionManager.js";
 import { DeviceId } from "../helpers/deviceId.js";
 
-export abstract class TransportRunnerBase {
+export type CreateConnectionManagerFn<ConnectParams extends MCPConnectParams> = (createParams: {
+    logger: CompositeLogger;
+    deviceId: DeviceId;
+}) => ConnectionManager<ConnectParams> | Promise<ConnectionManager<ConnectParams>>;
+
+export abstract class TransportRunnerBase<ConnectParams extends MCPConnectParams> {
```
kmruiz: We are leaking an implementation detail of ConnectionManager in a generic on lots of layers, that is a sign of lack of encapsulation.

kmruiz (changes requested):

tests/integration/transports/streamableHttp.test.ts
```diff
@@ -28,7 +30,10 @@ describe("StreamableHttpRunner", () => {
         describe(description, () => {
             beforeAll(async () => {
                 config.httpHeaders = headers;
-                runner = new StreamableHttpRunner(config, driverOptions);
+                runner = new StreamableHttpRunner(
```
kmruiz: The main issue with this approach is that right now, the ConnectionManager is bound to the lifecycle of the session (for example, multiple users can potentially connect to different MongoDBs).

This change breaks this assumption and makes harder to support multitenancy in the future, as it couples the lifecycle of a single ConnectionManager to the transport itself.

src/common/connectionManager.ts
himanshusinghs: Some talking points for the upcoming call:

> changing connection managers implementations without changing the actual calling code should work (Liskov)

Agreed! This is problematic with the current proposal of generic connect params but only in theory. In reality, the `ConnectionManager.connect`, when ConnectionManager is provided from outside the MCP server, is not even expected to be called from inside of the MCP. I did not modify the existing usage of `connectionManager.connect` to account for this and that's real problem. Having said that, I agree that sub class implementations should be substitutable. What I have here does not correctly encapsulate what I wanted to do.

> All this hierarchy can be simplified by simply using, if we need, visibility rules
...
We can even, if we want, have a single interface and a single implementation class

I don't think that should be the approach here. MCP server shouldn't have to bother with the concepts (setupExistingConnection) that it itself never makes use of. So having this interface within MCP server's implementation of ConnectionManager would be a code smell, raising questions down the line.


I thought about this a bit more and basically the idea that I was trying to propose, and failed to, was that the internal state of ConnectionManager can either be controlled by the MCP server itself or by an external entity (VSCode extension). When the state of ConnectionManager is managed from outside, MCP server should not be able to call `connect`, `disconnect` and similar methods on the manager and the existing code should account for that.

(Happy to explain more on the call.)

tests/integration/transports/streamableHttp.test.ts
himanshusinghs: That's not true. The ConnectionManager is still being created through a factory function on every Server setup requests. So you essentially end with the same design as before with the only change being what ConnectionManager you get depends on the factory function injected.

src/transports/base.ts
himanshusinghs: this gets handled by the refactor we agreed on

src/common/connectionManager.ts
himanshusinghs: Outcome of discussion:

- We will still have different implementation of ConnectionManager injected via the factory function into the MCP server because there are legit differences in how VSCode implements ConnectionManager functionality vs how our MCP implements them and we don't want to pull them together under same interface / implementation.

- The Abstract class shouldn't have to be generic over ConnectParams and instead we will have common interface in the abstract class which gets implemented by MCPConnectionManager and the VSCode's manager. The only difference being that VSCode's connection manager will throw on the overridden connect interface because MCP code is not expected to be able to connect from VSCode's connection manager.

src/common/mcpConnectionManager.ts
```diff
@@ -0,0 +1,252 @@
...
+        super();
+        this.bus = bus ?? new EventEmitter();
+        this.bus.on("mongodb-oidc-plugin:auth-failed", this.onOidcAuthFailed.bind(this));
+        this.bus.on("mongodb-oidc-plugin:auth-succeeded", this.onOidcAuthSucceeded.bind(this));
+        this.deviceId = deviceId;
+        this.clientName = "unknown";
+    }
+
+    async connect(connectParams: MCPConnectParams): Promise<AnyConnectionState> {
+        this._events.emit("connection-requested", this.state);
```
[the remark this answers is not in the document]
himanshusinghs: Come on Copilot, don't be like ChatGPT, connection-requested accept `AnyConnectionState` which is what `this.state` is and `connectParams` isn't.

src/common/connectionManager.ts
himanshusinghs: I moved this back to settings

src/common/connectionManager.ts
```diff
@@ -71,10 +68,40 @@ export interface ConnectionManagerEvents {
...
 
-export class ConnectionManager extends EventEmitter<ConnectionManagerEvents> {
-    private state: AnyConnectionState;
+export abstract class ConnectionManager {
+    protected clientName: string = "unknown";
+
+    protected readonly _events = new EventEmitter<ConnectionManagerEvents>();
+    readonly events: Pick<EventEmitter<ConnectionManagerEvents>, "on" | "off" | "once"> = this._events;
+
+    protected state: AnyConnectionState = { tag: "disconnected" };
```
kmruiz: nit: I prefer to do the initialisation in the constructor, it's more obvious as there is only one place where all the initialisation happens.

And because you should always call the super constructor when you inherit a class, it should still be safe.

src/transports/base.ts
```diff
@@ -7,16 +7,21 @@ import { McpServer } from "@modelcontextprotocol/sdk/server/mcp.js";
 import type { LoggerBase } from "../common/logger.js";
 import { CompositeLogger, ConsoleLogger, DiskLogger, McpLogger } from "../common/logger.js";
 import { ExportsManager } from "../common/exportsManager.js";
-import { ConnectionManager } from "../common/connectionManager.js";
+import type { ConnectionManager } from "../common/connectionManager.js";
 import { DeviceId } from "../helpers/deviceId.js";
 
+export type CreateConnectionManagerFn = (createParams: {
```
kmruiz:
```suggestion
export type ConnectionManagerFactory = (createParams: {
```

Despite the joke during our call, I actually prefer to use standard naming. I prefer to say this is a factory, as it's a factory.

kmruiz: Also, in which scenario can this be async? I can't see one. Shouldn't this be always synchronous? 

If it isn't always synchronous, I would prefer to always return a promise so it keeps the function colouring consistent.

src/common/connectionManager.ts
himanshusinghs: Its essentially the same. The initial values for these properties are known during class definition so I don't see the point of going the extra route (property declaration, property initialisation in constructor).

src/transports/base.ts
himanshusinghs: > Also, in which scenario can this be async?

In VSCode the implementation also connects to the active connection if its not already connected which is why its async.

> Despite the joke during our call, I actually prefer to use standard naming. I prefer to say this is a factory, as it's a factory.

I was actually joking on the call :) While its true that the function is a factory function, `ConnectionManagerFactory` might inadvertently convey that its a factory class. Having function types suffixed with `Fn` is common in Typescript, if you prefer I can change the type to `ConnectionManagerFactoryFn`?

himanshusinghs: Renamed to `ConnectionManagerFactoryFn` and changed the return type Promise<ConnectionManager>

src/common/connectionManager.ts
```diff
@@ -85,23 +124,15 @@ export class ConnectionManager extends EventEmitter<ConnectionManagerEvents> {
...
         super();
-
         this.bus = bus ?? new EventEmitter();
-        this.state = { tag: "disconnected" };
-
         this.bus.on("mongodb-oidc-plugin:auth-failed", this.onOidcAuthFailed.bind(this));
         this.bus.on("mongodb-oidc-plugin:auth-succeeded", this.onOidcAuthSucceeded.bind(this));
-
         this.deviceId = deviceId;
         this.clientName = "unknown";
```
nirinchev:
```suggestion
```

src/common/connectionManager.ts
kmruiz: It can get pretty complicated depending on the use case, for example: https://www.typescriptlang.org/docs/handbook/2/classes.html#initialization-order

It's advisable to use constructor because it's consistent and explicit (first parent, later child), and mixing property initialisation with constructors can get nasty.

kmruiz (approved):

src/transports/base.ts
```diff
@@ -7,16 +7,21 @@ import { McpServer } from "@modelcontextprotocol/sdk/server/mcp.js";
 import type { LoggerBase } from "../common/logger.js";
 import { CompositeLogger, ConsoleLogger, DiskLogger, McpLogger } from "../common/logger.js";
 import { ExportsManager } from "../common/exportsManager.js";
-import { ConnectionManager } from "../common/connectionManager.js";
+import type { ConnectionManager } from "../common/connectionManager.js";
 import { DeviceId } from "../helpers/deviceId.js";
 
+export type ConnectionManagerFactoryFn = (createParams: {
```
nirinchev: [naming nit] having both Factory and Fn in the name is a bit redundant.

src/common/connectionManager.ts
himanshusinghs: I don't think the initialisation order concern apply to our class implementation, properties are not overridden. But I realised that should mark `state` private to avoid out of order modification by child classes and while I am at it I am happy to apply the recommendation.

himanshusinghs: applied the suggestion in [b079a6d](https://github.com/mongodb-js/mongodb-mcp-server/pull/481/commits/b079a6deb5eec661beffd6ef031da6f53b2e7550)

src/transports/stdio.ts
```diff
@@ -55,13 +54,17 @@ export function createStdioTransport(): StdioServerTransport {
 export class StdioRunner extends TransportRunnerBase {
     private server: Server | undefined;
 
-    constructor(userConfig: UserConfig, driverOptions: DriverOptions, additionalLoggers: LoggerBase[] = []) {
-        super(userConfig, driverOptions, additionalLoggers);
+    constructor(
+        userConfig: UserConfig,
+        createConnectionManager: ConnectionManagerFactoryFn,
+        additionalLoggers: LoggerBase[] = []
```
nirinchev: Should we make a type for holding these - maybe `SystemConfig`? Also, would it make sense to provide a default for the connection manager function?

src/transports/streamableHttp.ts
```diff
@@ -1,13 +1,12 @@
 import express from "express";
 import type http from "http";
+import { randomUUID } from "crypto";
```
nirinchev, line 3: Not opposed to these reorders, though they seem a bit noisy. In a follow-up, would it make sense to add eslint/prettier rule to enforce a deterministic order for the imports?

src/lib.ts
```diff
@@ -1,7 +1,16 @@
...
-export type { LogPayload, LoggerType, LogLevel } from "./common/logger.js";
+export { type ConnectionManagerFactoryFn } from "./transports/base.js";
+export {
+    ConnectionManager,
+    type AnyConnectionState,
+    type ConnectionState,
+    type ConnectionStateConnected,
+    type ConnectionStateConnecting,
+    type ConnectionStateDisconnected,
+    type ConnectionStateErrored,
```
nirinchev: [nit] do we need these type imports? My feeling is that `AnyConnectionState` is sufficient?

tests/integration/build.test.ts
```diff
@@ -41,13 +41,17 @@ describe("Build Test", () => {
...
-            "Server",
-            "Session",
-            "Telemetry",
-            "StreamableHttpRunner",
-            "defaultUserConfig",
-            "LoggerBase",
-        ]);
+        expect(cjsKeys).toEqual(
+            expect.arrayContaining([
+                "CompositeLogger",
```
nirinchev: Why are we exporting CompositeLogger?

nirinchev (commented):
Mostly nits from me, otherwise it's good.

src/transports/base.ts
himanshusinghs: A Factory could also refer to a Factory class which is why I clarified that with an `Fn` suffix.

src/transports/streamableHttp.ts
himanshusinghs: yeap totally onboard with that - will raise a follow up after this

src/lib.ts
himanshusinghs: I will trim out some of these. Certainly need AnyConnectionState, ConnectionStateDisconnected (for disconnect's interface).

tests/integration/build.test.ts
himanshusinghs: Its used by VSCodeConnectionManager as a type, I will add the type annotation.

src/transports/stdio.ts
himanshusinghs: I don't think that's necessary. connection manager factory seems important enough on the interface to think about when initialising the transport so I'd rather keep it required.

tests/integration/build.test.ts
himanshusinghs: Realised that I could just use LoggerBase so I removed it.

src/transports/stdio.ts
nirinchev: The problem is that if you don't provide a default, you'd need to export `MCPConnectionManager`, otherwise, consumers would need to always bring their own implementation. I'm not in favor of exporting the manager; instead, we should treat the factory and loggers as extension points that are not required for the operation of the server.

himanshusinghs: I was having a different view on this. I am thinking that connection management is central to running MCP server and if consumers were to ever use it as a library, I think they should be forced to think if they want the default behaviour for connection manager or bring their own. I am also not in favor of exposing MCPConnectionManager but for default use-case we can always just expose `connectionManagerFactory`?

himanshusinghs: added a default connection manager factory - [b6ade3c](https://github.com/mongodb-js/mongodb-mcp-server/pull/481/commits/b6ade3cfc0797d2e543c5fc086695899077dce0e)

PR #482, opened 2025-08-27
fix: update atlas integration tests -- MCP-98

blva:
## Proposed changes
- this fixes the issue where we were validating the size of the response. now we just check if the expected org has been found 

## Checklist

### DIFF
--- a/tests/integration/tools/atlas/orgs.test.ts
+++ b/tests/integration/tools/atlas/orgs.test.ts
@@ -13,7 +13,6 @@ describeWithAtlas("orgs", (integration) => {
         it("returns org names", async () => {
             const response = await integration.mcpClient().callTool({ name: "atlas-list-orgs", arguments: {} });
             const elements = getResponseElements(response);
-            expect(elements).toHaveLength(2);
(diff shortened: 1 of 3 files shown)


tests/integration/tools/atlas/orgs.test.ts
```diff
@@ -13,7 +13,6 @@ describeWithAtlas("orgs", (integration) => {
         it("returns org names", async () => {
             const response = await integration.mcpClient().callTool({ name: "atlas-list-orgs", arguments: {} });
             const elements = getResponseElements(response);
-            expect(elements).toHaveLength(2);
```
kmruiz, line 16: This would still fail if length < 2, right? I think the issue of this one is that sometimes it shows 3 organisations. Maybe we can use something like:

expect(elements.length).[toBeGreaterThanOrEqual(](https://vitest.dev/api/expect.html#tobegreaterthanorequal)2)

blva: I think we don't mind the amount of elements as long as we are able to find the expected org

PR #483, opened 2025-08-27
fix: don't execute cleanup code if setup failed

nirinchev:
## Proposed changes

The atlas tests would try and cleanup some resources in afterAll/afterEach, but those would run even if beforeAll failed. This was causing some red herring errors due to calling atlas endpoints with an empty projectId, for example. While this doesn't address the underlying problems that caused the setup to fail, at least it reduces the noise.

### DIFF
--- a/tests/accuracy/export.test.ts
+++ b/tests/accuracy/export.test.ts
@@ -10,6 +10,7 @@ describeAccuracyTests([
                 parameters: {
                     database: "mflix",
                     collection: "movies",
+                    exportTitle: Matcher.string(),
                     exportTarget: [
                         {
                             name: "find",
@@ -28,6 +29,7 @@ describeAccuracyTests([
                 parameters: {
                     database: "mflix",
                     collection: "movies",
+                    exportTitle: Matcher.string(),
                     exportTarget: [
                         {
                             name: "find",
@@ -50,6 +52,7 @@ describeAccuracyTests([
                 parameters: {
                     database: "mflix",
                     collection: "movies",
+                    exportTitle: Matcher.string(),
                     exportTarget: [
                         {
                             name: "find",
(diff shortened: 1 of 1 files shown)


tests/accuracy/export.test.ts
```diff
@@ -10,6 +10,7 @@ describeAccuracyTests([
                 parameters: {
                     database: "mflix",
                     collection: "movies",
+                    exportTitle: Matcher.string(),
```
nirinchev, line 13: cc @himanshusinghs - not sure what has happened here - I think I added the titles at some point, then they disappeared 😬 But looking at the results on main, it seems the accuracy tests are failing because of this.

tests/integration/common/connectionManager.oidc.test.ts
```diff
@@ -168,17 +168,28 @@ describe.skipIf(process.platform !== "linux")("ConnectionManager OIDC Tests", as
...
+                            throw new Error("No authenticated users found. Retrying.");
+                        }
+
+                        if (!result.authInfo.authenticatedUserRoles.length) {
+                            throw new Error("No authenticated user roles found. Retrying.");
+                        }
+
+                        return result;
+                    },
+                    { timeout: 5000 }
```
nirinchev, line 191: cc @kmruiz - I expanded this a bit as the test was failing due to the `authenticatedUsers` collection being empty, not `status` itself being undefined. I also added it a bit more time since I saw it fail with the 1 second default.

Here's a run where we're only checking status: https://github.com/mongodb-js/mongodb-mcp-server/actions/runs/17266097331/job/48998448665?pr=483

Here's the one that's failing with the 1 second default: https://github.com/mongodb-js/mongodb-mcp-server/actions/runs/17267312116/job/49002419896?pr=483

PR #484, opened 2025-08-27
fix: add a ping loop for http transport clients

nirinchev:
## Proposed changes

This makes it so the server sends a ping every 30 seconds to connected clients to ensure the connection is still alive. After 3 failed pings, we close the transport.

## Checklist

### DIFF
--- a/src/transports/streamableHttp.ts
+++ b/src/transports/streamableHttp.ts
@@ -74,7 +74,7 @@ export class StreamableHttpRunner extends TransportRunnerBase {
                     jsonrpc: "2.0",
                     error: {
                         code: JSON_RPC_ERROR_CODE_SESSION_ID_INVALID,
-                        message: `session id is invalid`,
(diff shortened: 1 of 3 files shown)


src/transports/streamableHttp.ts
```diff
@@ -114,12 +114,42 @@ export class StreamableHttpRunner extends TransportRunnerBase {
...
+                            } catch (err) {
+                                this.logger.warning({
+                                    id: LogId.streamableHttpTransportKeepAliveFailure,
+                                    context: "streamableHttpTransport",
+                                    message: `Error sending ping (attempt #${failedPings + 1}): ${err instanceof Error ? err.message : String(err)}`,
+                                });
+
+                                if (++failedPings > 3) {
+                                    clearInterval(keepAliveLoop);
+                                    await transport.close();
```
nirinchev: This needs to be in a try-catch

PR #489, opened 2025-08-28
chore: add some extra log level validations

nirinchev:
## Proposed changes

This is a follow-up on https://github.com/mongodb-js/mongodb-mcp-server/pull/486 - it adds an extra test to ensure our log levels array is exhaustive and a parameter validation for the http endpoint.

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -5,7 +5,7 @@ import { AtlasTools } from "./tools/atlas/tools.js";
 import { MongoDbTools } from "./tools/mongodb/tools.js";
 import { Resources } from "./resources/resources.js";
 import type { LogLevel } from "./common/logger.js";
-import { LogId } from "./common/logger.js";
+import { LogId, McpLogger } from "./common/logger.js";
 import type { Telemetry } from "./telemetry/telemetry.js";
 import type { UserConfig } from "./common/config.js";
 import { type ServerEvent } from "./telemetry/types.js";
@@ -107,6 +107,10 @@ export class Server {
         });
 
         this.mcpServer.server.setRequestHandler(SetLevelRequestSchema, ({ params }) => {
+            if (!McpLogger.LOG_LEVELS.includes(params.level)) {
+                throw new Error(`Invalid log level: ${params.level}`);
+            }
+
             this._mcpLogLevel = params.level;
             return {};
         });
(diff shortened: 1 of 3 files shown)


blva (approved):

src/server.ts
```diff
@@ -107,6 +107,10 @@ export class Server {
         });
 
         this.mcpServer.server.setRequestHandler(SetLevelRequestSchema, ({ params }) => {
+            if (!McpLogger.LOG_LEVELS.includes(params.level)) {
+                throw new Error(`Invalid log level: ${params.level}`);
```
kmruiz, line 111: nit: Shouldn't this be a MongoDBError?

tests/unit/logger.test.ts
```diff
@@ -332,5 +333,9 @@ describe("Logger", () => {
             expect(mcpLoggerSpy).toHaveBeenCalledTimes(2);
             expect(getLastMcpLogMessage()).toContain("Alert message");
         });
+
+        it("MCPLogger.LOG_LEVELS contains all possible levels", () => {
+            expect(McpLogger.LOG_LEVELS).toEqual(LoggingMessageNotificationSchema.shape.params.shape.level.options);
+        });
```
kmruiz, line 339: nit: I believe the test should be that if we add an invalid log level it fails?

nirinchev: I think that's covered - if there's a log level in the array that's invalid, the array would differ from the possible enum values in the schema.

src/server.ts
nirinchev: `MongoDBError` in the MCP server is an error thrown when interacting with mongodb, while this one is just a simple http validation.

tests/unit/logger.test.ts
kmruiz: Yeah the thing is that it checks an implementation detail, not the contract on how the loggers work.

It's a nit, so I'll resolve it, I don't think it's relevant enough to block the PR.

nirinchev: Hm, maybe I misunderstood your original suggestion. This test is only here to catch a case where a new log level is added in the future and we don't order it correctly in the `LOG_LEVELS` array. It's here mostly because there's no obvious way to express this through the type system.

PR #490, opened 2025-08-28
chore: bump dependencies to latest

nirinchev:
## Proposed changes
Only ones left out are the major version bumps for the ai sdk related packages as those are breaking and require code changes.


kmruiz (approved):
Thanks for taking care of this! <3

PR #493, opened 2025-08-29
chore: warn about the usage of deprecated cli arguments MCP-107

kmruiz:
## Checklist

### DIFF
--- a/.github/workflows/code_health.yaml
+++ b/.github/workflows/code_health.yaml
@@ -54,7 +54,7 @@ jobs:
           MDB_MCP_API_CLIENT_ID: ${{ secrets.TEST_ATLAS_CLIENT_ID }}
           MDB_MCP_API_CLIENT_SECRET: ${{ secrets.TEST_ATLAS_CLIENT_SECRET }}
           MDB_MCP_API_BASE_URL: ${{ vars.TEST_ATLAS_BASE_URL }}
-        run: npm test -- --exclude "tests/unit/**" --exclude "tests/integration/tools/mongodb/**" --exclude "tests/integration/*.ts"
+        run: npm test -- tests/integration/tools/atlas
       - name: Upload test results
         uses: actions/upload-artifact@v4
         if: always()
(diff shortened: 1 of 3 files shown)


src/common/config.ts
```diff
@@ -264,6 +271,29 @@ function parseCliConfig(args: string[]): CliOptions {
     return parsed;
 }
 
+export function warnAboutDeprecatedCliArgs(
+    args: CliOptions &
+        UserConfig & {
+            _?: string[];
+        },
+    warn: (msg: string) => void
```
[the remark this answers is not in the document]
kmruiz: It's not a console.log.

src/common/config.ts
```diff
@@ -264,6 +271,29 @@ function parseCliConfig(args: string[]): CliOptions {
...
+    // instead of --connectionString, as it's how the mongosh works.
+    if (args.connectionString) {
+        wasWarned = true;
+        warn(
+            "The --connectionString argument is deprecated. Prefer using the first positional argument for the connection string or the MDB_MCP_CONNECTION_STRING environment variable."
+        );
+    }
+
+    if (wasWarned) {
+        warn("Refer to https://www.mongodb.com/docs/mcp-server/get-started/ for setting up the MCP Server.");
```
gagik:
```suggestion
        warn("Refer to https://www.mongodb.com/docs/mcp-server/get-started/ for setting up the MCP Server.");
```
is there a difference here between just putting them all in one if statement?

kmruiz: Ah, the only reason is that if we add more deprecated arguments (for example, LDAP arguments in the future, because LDAP is deprecated in newer server versions) we will add the flag between this if and the previous, and this warning would only show once, not per each deprecated flag.

gagik: > if we add more deprecated arguments

might make sense to do this when we actually get to that otherwise we're optimizing a future assumption.
alternatively if you feel strongly about it, could rename`wasWarned` to `usedDeprecatedArgument` so that flow is a bit clearer

kmruiz: Well we will add more deprecated arguments in the mid-term so I would keep it. I'm fine renaming the variable to make it more clear though, I think that's a great suggestion. I'll rename `wasWarned` with `usedDeprecatedArgument`.

gagik (approved):

PR #500, opened 2025-09-01
feat: add more details about atlas connect flow - MCP-124

blva:
## Proposed changes

- MCP-124
- Adds more messages about the resources created to be able to connect to an atlas cluster 

## Checklist

### DIFF
--- a/src/tools/atlas/connect/connectCluster.ts
+++ b/src/tools/atlas/connect/connectCluster.ts
@@ -62,7 +62,7 @@ export class ConnectClusterTool extends AtlasToolBase {
     private async prepareClusterConnection(
         projectId: string,
         clusterName: string
-    ): Promise<{ connectionString: string; atlas: AtlasClusterConnectionInfo }> {
+    ): Promise<{ connectionString: string; atlas: AtlasClusterConnectionInfo; userCreated: boolean }> {
         const cluster = await inspectCluster(this.session.apiClient, projectId, clusterName);
 
         if (!cluster.connectionString) {
@@ -110,7 +110,7 @@ export class ConnectClusterTool extends AtlasToolBase {
         cn.password = password;
         cn.searchParams.set("authSource", "admin");
 
-        return { connectionString: cn.toString(), atlas: connectedAtlasCluster };
+        return { connectionString: cn.toString(), atlas: connectedAtlasCluster, userCreated: true };
     }
 
     private async connectToCluster(connectionString: string, atlas: AtlasClusterConnectionInfo): Promise<void> {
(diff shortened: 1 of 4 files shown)


src/tools/atlas/connect/connectCluster.ts
```diff
@@ -110,7 +115,7 @@ export class ConnectClusterTool extends AtlasToolBase {
         cn.password = password;
         cn.searchParams.set("authSource", "admin");
 
-        return { connectionString: cn.toString(), atlas: connectedAtlasCluster };
+        return { connectionString: cn.toString(), atlas: connectedAtlasCluster, userCreated: true };
```
nirinchev: It seems like we're always returning `true` for `userCreated` - am I missing something?

src/tools/atlas/connect/connectCluster.ts
```diff
@@ -10,6 +10,11 @@ import type { AtlasClusterConnectionInfo } from "../../../common/connectionManag
 import { getDefaultRoleFromConfig } from "../../../common/atlas/roles.js";
 
 const EXPIRY_MS = 1000 * 60 * 60 * 12; // 12 hours
+const addedIpAccessListMessage =
+    "Note: Your current IP address has been added to the Atlas project's IP access list to enable secure connection.";
+
+const createdUserMessage =
+    "Note: A temporary user has been created to enable secure connection to the cluster. For more information, see https://dochub.mongodb.org/core/mongodb-mcp-server-tools-considerations";
```
nirinchev: Too lazy to test it, but during manual testing, did you experience the model correctly relaying the url to users?

nirinchev (approved):
Minor nits/questions, otherwise lgtm.

src/tools/atlas/connect/connectCluster.ts
blva: I couldn't find the LLM propagating the link. But it does show up in the tool call logs if the toggle is clicked

blva: Screenshot 2025-09-01 at 20 56 55

src/tools/atlas/connect/connectCluster.ts
blva: no, if the connection state doesn't require preparation (e.g. is already connected*) then we'd not execute this

src/tools/atlas/connect/connectCluster.ts
nirinchev: Can we try tweaking the response and seeing what the output is? E.g. we could add something like: `Note to LLM Agent: it is important to include the following link in your response to the user in case they want to get more information about the temporary user created`

src/tools/atlas/connect/connectCluster.ts
nirinchev: Right - my point is that `prepareClusterConnection` always returns `userCreated: true` (or at least it's not obvious in which case it will return `false` for userCreated. In the already connected state, we wouldn't invoke this method.

src/tools/atlas/connect/connectCluster.ts
blva: that did the trick!
Screenshot 2025-09-03 at 13 42 54

src/tools/atlas/connect/connectCluster.ts
blva: ah, got it! yes, good point since we error if something goes wrong. I can remove that from now, I think we'll need to do more work in the connectCluster anyways but this is out of scope here, so let me simplify

PR #501, opened 2025-09-02
chore: expose a hook to specify telemetry hosting mode MCP-166

nirinchev:
## Proposed changes

It's best to review this with whitespace disabled since I removed an unnecessary nesting in the telemetry tests. This hook will be used by vscode to set the "hostingMode" so that we can differentiate between standalone and vscode-hosted servers.

### DIFF
--- a/src/telemetry/telemetry.ts
+++ b/src/telemetry/telemetry.ts
@@ -14,11 +14,6 @@ type EventResult = {
 };
 
 export class Telemetry {
-    public static baseCommonProperties: CommonProperties = {
-        ...MACHINE_METADATA,
-        hosting_mode: "standalone",
-    };
-
     private isBufferingEvents: boolean = true;
     /** Resolves when the setup is complete or a timeout occurs */
     public setupPromise: Promise<[string, boolean]> | undefined;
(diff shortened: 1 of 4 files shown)


src/telemetry/types.ts
```diff
@@ -73,4 +73,5 @@ export type CommonProperties = {
     config_atlas_auth?: TelemetryBoolSet;
     config_connection_string?: TelemetryBoolSet;
     session_id?: string;
+    hosting_mode?: string;
```
blva, line 136: nit: can we set the accepted strings here?

src/telemetry/telemetry.ts
```diff
@@ -14,6 +14,8 @@ type EventResult = {
 };
 
 export class Telemetry {
+    public static hostingMode?: string;
```
blva: should we default to standalone until updated by vscode?

src/telemetry/types.ts
nirinchev: The MCP server does not know them at this point - essentially, I want this to be a field populated by the application that hosts the server and we don't know beforehand what these applications are.

src/telemetry/telemetry.ts
nirinchev: We could, though we would need to treat empty and standalone as identical for historical purposes.

himanshusinghs (approved):
Looks good 🚀

PR #502, opened 2025-09-02
chore: extend library interfaces to allow injecting a custom connection error handler MCP-132

himanshusinghs:
## Proposed changes

We're expanding the transport runner and server's interface to accept a custom connection error handler so that library consumers can provide their own mechanism of handling connection error and have an opportunity to act on them.

This will allow VSCode to provide its own ConnectionErrorHandler and thus modify the tool responses in case of connection problems.

## Checklist

### DIFF
--- a/src/tools/mongodb/mongodbTool.ts
+++ b/src/tools/mongodb/mongodbTool.ts
@@ -68,9 +68,9 @@ export abstract class MongoDBToolBase extends ToolBase {
                     });
                     if (outcome?.errorHandled) {
                         return outcome.result;
-                    } else {
-                        return super.handleError(error, args);
                     }
+
+                    return super.handleError(error, args);
                 }
                 case ErrorCodes.ForbiddenCollscan:
                     return {
--- a/src/transports/base.ts
+++ b/src/transports/base.ts
@@ -30,13 +30,13 @@ export abstract class TransportRunnerBase {
 
     protected constructor({
         userConfig,
-        createConnectionManager,
-        connectionErrorHandler,
+        createConnectionManager = createMCPConnectionManager,
+        connectionErrorHandler = defaultConnectionErrorHandler,
         additionalLoggers = [],
     }: TransportRunnerConfig) {
         this.userConfig = userConfig;
-        this.createConnectionManager = createConnectionManager ?? createMCPConnectionManager;
-        this.connectionErrorHandler = connectionErrorHandler ?? defaultConnectionErrorHandler;
+        this.createConnectionManager = createConnectionManager;
+        this.connectionErrorHandler = connectionErrorHandler;
         const loggers: LoggerBase[] = [...additionalLoggers];
         if (this.userConfig.loggers.includes("stderr")) {
             loggers.push(new ConsoleLogger());


src/common/connectionErrorHandler.ts
```diff
@@ -0,0 +1,81 @@
+import type { CallToolResult } from "@modelcontextprotocol/sdk/types.js";
```
himanshusinghs, line 1: This entire file is a cut of existing logic in MongodbToolBase.

src/server.ts
```diff
@@ -252,7 +256,10 @@ export class Server {
                     "Failed to connect to MongoDB instance using the connection string from the config: ",
                     error
                 );
-                throw new Error("Failed to connect to MongoDB instance using the connection string from the config");
+                // TODO: Confirm with team if we originally wanted to silence
+                // this error or have it crash the MCP server process.
```
himanshusinghs: Can someone confirm the expected behaviour? My gut feeling is that we should gulp the error and let the server through, eventually the tool calls will highlight this problem as part of their response.

src/server.ts
```diff
@@ -252,7 +256,10 @@ export class Server {
...
+
+                // throw new Error("Failed to connect to MongoDB instance using the connection string from the config");
```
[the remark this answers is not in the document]
himanshusinghs: Resolved on main

src/server.ts
himanshusinghs: Resolved on main

src/tools/mongodb/mongodbTool.ts
```diff
@@ -56,63 +56,22 @@ export abstract class MongoDBToolBase extends ToolBase {
...
+                    const connectionError = error as MongoDBError<
+                        ErrorCodes.NotConnectedToMongoDB | ErrorCodes.MisconfiguredConnectionString
+                    >;
+                    const outcome = this.server?.connectionErrorHandler(connectionError, {
+                        availableTools: this.server?.tools ?? [],
+                        connectionState: this.session.connectionManager.currentConnectionState,
+                    });
+                    if (outcome?.errorHandled) {
+                        return outcome.result;
+                    } else {
```
nirinchev: No need for an `else` here.

src/transports/base.ts
```diff
@@ -8,17 +8,35 @@ import type { LoggerBase } from "../common/logger.js";
...
-    ) {
+    protected constructor({
+        userConfig,
+        createConnectionManager,
+        connectionErrorHandler,
+        additionalLoggers = [],
+    }: TransportRunnerConfig) {
+        this.userConfig = userConfig;
+        this.createConnectionManager = createConnectionManager ?? createMCPConnectionManager;
+        this.connectionErrorHandler = connectionErrorHandler ?? defaultConnectionErrorHandler;
```
nirinchev: Can we have these defaults in the deconstructed config instead?

src/tools/mongodb/mongodbTool.ts
himanshusinghs: addressed

src/transports/base.ts
himanshusinghs: addressed

PR #508, opened 2025-09-03
feat: add dictionary-based redaction for secrets in logs

kmruiz:
## Proposed changes

This sets up a Keychain, a holder of known secrets. When a new secret is created it should be registered in the keychain so loggers can redact them and avoid leaking secrets into our logs.

## Checklist

### DIFF
--- a/src/common/keychain.ts
+++ b/src/common/keychain.ts
@@ -34,3 +34,7 @@ export class Keychain {
         return [...this.secrets];
     }
 }
+
+export function registerGlobalSecretToRedact(value: Secret["value"], kind: Secret["kind"]): void {
+    Keychain.root.register(value, kind);
+}
--- a/src/lib.ts
+++ b/src/lib.ts
@@ -19,3 +19,5 @@ export type {
 } from "./common/connectionErrorHandler.js";
 export { ErrorCodes } from "./common/errors.js";
 export { Telemetry } from "./telemetry/telemetry.js";
+export { Keychain, registerGlobalSecretToRedact } from "./common/keychain.js";
+export type { Secret } from "./common/keychain.js";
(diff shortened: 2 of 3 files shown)


src/types/mongodb-redact.d.ts
```diff
@@ -1,4 +1,21 @@
...
+        "mongodb uri",
+    ];
+
+    export type SecretKind = (typeof SECRET_KIND)[number];
+    export type Secret = {
+        readonly value: string;
+        readonly kind: SecretKind;
+    };
+
+    export declare function redact<T>(message: T, secrets?: Secret[] | undefined): T;
```
himanshusinghs: Not not this PR but if mongodb-redact does not provide types then would it make sense to move all these type definitions to the package itself and release a version with `.d.ts`?

src/common/config.ts
```diff
@@ -369,6 +394,7 @@ export function setupUserConfig({
         }
     }
 
+    registerKnownSecretsInRootKeychain(userConfig);
```
himanshusinghs, line 397: I think we should do this in the places where either config is injected or the first interface accepting the injected config because otherwise VSCode's injected won't get covered by this.

Also might need to expose the method on the lib itself for VSCode to register a header as secret.

himanshusinghs (commented):
Looking good 🚀. The second comment about call to `registerKnownSecretsInRootKeychain` is something you might wanna think about before merging.

src/types/mongodb-redact.d.ts
kmruiz: Actually it exports the types, I will fix the configuration here in this PR because it should be trivial.

src/common/config.ts
himanshusinghs: We discussed and decided to expose a method to register secrets for redaction and let VSCode (lib consumers) take control of the calling the method.

Issue #327, opened 2025-06-29
[Bug]: TypeError: connectionString.startsWith is not a function


ImoutoChan:
### Version

mcp/mongodb@sha256:44eed0f93b85db457e86d9a7de44aa032d6d9128aac05086194402adc4e2a9b7

### App

- [ ] Cursor
- [ ] Windsurf
- [ ] VSCode
- [ ] VSCode Insiders
- [ ] Claude Desktop
- [x] Other

### Affected Models (if applicable)

- [ ] Claude 3.5 Sonnet
- [ ] Claude 3.7 Sonnet
- [ ] GPT-4a
- [ ] o4-mini
- [ ] Other

### Bug Description

I'm trying to use docker desktop feature MCP Tools: https://docs.docker.com/ai/mcp-catalog-and-toolkit/toolkit/, 
I installed mongodb and provided it with my connection string.

However, when I run `docker mcp gateway run` I get

```
docker mcp gateway run
- Those servers are enabled: mongodb
- Using images:
  - mcp/mongodb@sha256:44eed0f93b85db457e86d9a7de44aa032d6d9128aac05086194402adc4e2a9b7
- Pulling images [mcp/mongodb]
> Images pulled in 3.912ms
- Listing MCP tools...
  - Running server mcp/mongodb with [run --rm -i --init --security-opt no-new-privileges --cpus 1 --memory 2Gb --pull never --label docker-mcp=true --label docker-mcp-tool-type=mcp --label docker-mcp-name=mongodb --label docker-mcp-transport=stdio -e MDB_MCP_CONNECTION_STRING]
  > Can't start mongodb: initializing mcp/mongodb@sha256:44eed0f93b85db457e86d9a7de44aa032d6d9128aac05086194402adc4e2a9b7: Failed to connect to MongoDB instance using the connection string from the config:  TypeError: connectionString.startsWith is not a function
    at connectionStringHasValidScheme (/usr/local/lib/node_modules/mongodb-mcp-server/node_modules/mongodb-connection-string-url/lib/index.js:9:30)
    at new ConnectionString (/usr/local/lib/node_modules/mongodb-mcp-server/node_modules/mongodb-connection-string-url/lib/index.js:85:34)
    at setAppNameParamIfMissing (file:///usr/local/lib/node_modules/mongodb-mcp-server/dist/helpers/connectionOptions.js:3:33)
[...]

PR #514, opened 2025-09-04
fix(connection): support for multiple hosts urls #327

kmruiz:
## Proposed changes

Bug: MongoDB URLs passed as environment variables that contain multiple hosts are treated as arrays instead of actual URLS.
Fix: When the environment variable contents start with a mongodb connection protocol (mongodb://, mongodb+srv://) it skips preprocessing the contents and stores it in the config as is.

## Checklist

### DIFF
--- a/src/common/config.ts
+++ b/src/common/config.ts
@@ -191,6 +191,12 @@ function parseEnvConfig(env: Record<string, unknown>): Partial<UserConfig> {
             return;
         }
         if (path.length === 0) {
+            // MongoDB URLs must not be preprocessed
+            if (value.startsWith("mongodb://") || value.startsWith("mongodb+srv://")) {
+                obj[currentField] = value;
+                return;
+            }
+
             const numberValue = Number(value);
             if (!isNaN(numberValue)) {
                 obj[currentField] = numberValue;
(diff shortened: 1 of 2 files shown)


src/common/config.ts
```diff
@@ -191,6 +191,12 @@ function parseEnvConfig(env: Record<string, unknown>): Partial<UserConfig> {
             return;
         }
         if (path.length === 0) {
+            // MongoDB URLs must not be preprocessed
+            if (value.startsWith("mongodb://") || value.startsWith("mongodb+srv://")) {
```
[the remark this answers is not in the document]
kmruiz: While environment variables are always strings, I'll add the typecheck _just in case_.

src/common/config.ts
```diff
@@ -191,6 +191,12 @@ function parseEnvConfig(env: Record<string, unknown>): Partial<UserConfig> {
             return;
         }
         if (path.length === 0) {
+            // MongoDB URLs must not be preprocessed
+            if (typeof value === "string" && (value.startsWith("mongodb://") || value.startsWith("mongodb+srv://"))) {
```
blva: [q] instead of trying to infer if this is a connection string, should we check the env variable name instead? that way we don't end up with weird situations if a customer accidentally sets an env variable incorrectly

kmruiz: I don't have a strong opinion, the issue with an allowlist of possible envvars is that if we add one we need to remember updating it here to avoid the issue. If you feel strong about it I can change it.

blva (approved):
non-blocking comment

PR #517, opened 2025-09-04
feat(cli): notify when a flag is wrong and suggest a fix MCP-121

kmruiz:
## Checklist


src/common/config.ts
```diff
@@ -280,24 +319,38 @@ function parseCliConfig(args: string[]): CliOptions {
...
-): void {
+export function warnAboutDeprecatedOrUnknownCliArgs(args: object, warn: (msg: string) => void): void {
     let usedDeprecatedArgument = false;
+    let usedInvalidArgument = false;
+
+    const knownArgs = args as UserConfig & CliOptions;
     // the first position argument should be used
     // instead of --connectionString, as it's how the mongosh works.
-    if (args.connectionString) {
+    if (knownArgs.connectionString) {
```
himanshusinghs:
```suggestion
export function warnAboutDeprecatedOrUnknownCliArgs(args: Record<string, unknown>, warn: (msg: string) => void): void {
let usedDeprecatedArgument = false;
    let usedInvalidArgument = false;

// the first position argument should be used
// instead of --connectionString, as it's how the mongosh works.
    if (args.connectionString) {
```
I think a Record might be more appropriate here compared to object as its more narrower?

himanshusinghs (approved):
Look good 💯 Left a suggestion for the type.

PR #520, opened 2025-09-04
feat(atlas-local): Added Atlas Local List Deployments tool

jeroenvervaeke:
## Proposed changes
Ticket: [\[MCP-158\] \[mongodb-mcp-server\] Implement listLocalAtlasClusters tool](https://jira.mongodb.org/browse/MCP-158)

## Checklist


src/server.ts
```diff
@@ -193,8 +193,9 @@ export class Server {
         this.telemetry.emitEvents([event]).catch(() => {});
     }
 
-    private registerTools(): void {
-        for (const toolConstructor of [...AtlasTools, ...AtlasLocalTools, ...MongoDbTools]) {
+    private async registerTools(): Promise<void> {
+        const atlasLocalTools = await BuildAtlasLocalTools();
```
blva: other tools also have clients injected, why not follow their same pattern and avoid having to await registration/building something here?

tests/integration/tools/atlas-local/listDeployments.test.ts
```diff
@@ -0,0 +1,61 @@
...
+    );
+
+    it.skipIf(isMacOSInGitHubActions)("should have the atlas-local-list-deployments tool", async () => {
+        const { tools } = await integration.mcpClient().listTools();
+        const listDeployments = tools.find((tool) => tool.name === "atlas-local-list-deployments");
+        expectDefined(listDeployments);
+    });
+
+    it.skipIf(!isMacOSInGitHubActions)(
+        "[MacOS in GitHub Actions] should not have the atlas-local-list-deployments tool",
```
blva, line 30: [q] why? can we add some details here?

src/server.ts
jeroenvervaeke: I looked into it again and found a better approach that I hadn't noticed while explaining why things didn't work. Will update

tests/integration/tools/atlas-local/listDeployments.test.ts
jeroenvervaeke: Because MacOS doesn't have docker installed in github runners, I can add a comment

src/tools/atlasLocal/read/listDeployments.ts
```diff
@@ -0,0 +1,55 @@
...
+
+export class ListDeploymentsTool extends AtlasLocalToolBase {
+    public name = "atlas-local-list-deployments";
+    protected description = "List MongoDB Atlas local deployments";
+    public operationType: OperationType = "read";
+    protected argsShape = {};
+
+    protected async execute(): Promise<CallToolResult> {
+        // Get the client
+        const client = this.client;
```
kmruiz: We usually do the injection at the session level, this way we can have one client per session (in case they need specific configuration per user). We should keep this approach as much as possible. You likely want to add your client here:

https://github.com/mongodb-js/mongodb-mcp-server/blob/main/src/transports/base.ts#L83-L91

src/tools/atlasLocal/atlasLocalTool.ts
```diff
@@ -1,9 +1,17 @@
...
-import type { ToolArgs, ToolCategory } from "../tool.js";
+import type { TelemetryToolMetadata, ToolArgs, ToolCategory } from "../tool.js";
 import { ToolBase } from "../tool.js";
+import type { ToolCallback } from "@modelcontextprotocol/sdk/server/mcp.js";
+import type AtlasLocal from "@mongodb-js-preview/atlas-local";
 
 export abstract class AtlasLocalToolBase extends ToolBase {
     public category: ToolCategory = "atlas-local";
+    // Will be injected by BuildAtlasLocalTools() in atlasLocal/tools.ts
+    public client?: AtlasLocal.Client;
```
kmruiz: This should be part of the session as mentioned earlier.

src/server.ts
kmruiz: Yes, we should not do await or any potentially slow operations here, as it will block the entire server and people might not be even using the local CLI.

If registering the tool requires some slow ops (for example, checking if docker is available), we should disable these tools by default, run the process in the background and when it's done, trigger a tool list refresh. In the connect.ts tool we have an example on how to trigger a tool list refresh.

src/tools/atlasLocal/tools.ts
```diff
@@ -1 +1,33 @@
...
+            tool.prototype.client = client;
+        });
+    }
+
+    return atlasLocalTools;
+};
+
+export const GetAtlasLocalClient = async (): Promise<AtlasLocal.Client | undefined> => {
+    try {
+        const { Client: AtlasLocalClient } = await import("@mongodb-js-preview/atlas-local");
```
kmruiz: If this is an optional dependency (it can exist or not) it should be marked as an optional dependency in the package.json. If it's not an optional dependency, the import should be done synchronously at the top of the file.

package.json
```diff
@@ -96,6 +96,7 @@
   },
   "dependencies": {
     "@modelcontextprotocol/sdk": "^1.17.4",
+    "@mongodb-js-preview/atlas-local": "^0.0.0-preview.1",
```
kmruiz: This should be an optional dependency, because on unsupported platforms it might crash.

tests/integration/tools/atlas-local/listDeployments.test.ts
jeroenvervaeke: Added a comment

src/tools/atlasLocal/read/listDeployments.ts
jeroenvervaeke: I've updated the code to inject the client into the session

src/tools/atlasLocal/atlasLocalTool.ts
jeroenvervaeke: I've updated the code to inject the client into the session

src/tools/atlasLocal/tools.ts
jeroenvervaeke: I've marked `@mongodb-js-preview/atlas-local` as an optional dependency.

package.json
jeroenvervaeke: I've marked `@mongodb-js-preview/atlas-local` as an optional dependency.

src/server.ts
jeroenvervaeke: I've already moved everything into the session. Will initialize the client in the background and then refresh the tools. Thanks for the pointers!

jeroenvervaeke: `atlas-local` tools are now loading async

src/server.ts
```diff
@@ -197,8 +200,41 @@ export class Server {
...
+            const client = AtlasLocalClient.connect();
+
+            // Set Atlas Local client
+            this.session.setAtlasLocalClient(client);
+
+            // Register Atlas Local tools
+            this.registerToolInstances(AtlasLocalTools);
+        } catch (error) {
+            // If Atlas Local tools are disabled, don't log an error
+            if (this.userConfig.disabledTools.includes("atlas-local")) {
```
kmruiz: Shouldn't this be done even before the import at line 207? Why do we want to attempt this if the tools are disabled? Or am I missing something?

src/tools/atlasLocal/atlasLocalTool.ts
```diff
@@ -1,10 +1,32 @@
...
+
+    protected async execute(): Promise<CallToolResult> {
+        // Get the client
+        const client = this.session.atlasLocalClient;
+
+        // If the client is not found, throw an error
+        // This should never happen, because the tool should have been disabled.
+        // verifyAllowed in the base class returns false if the client is not found
+        if (!client) {
+            throw new Error("Atlas Local client not found, tool should have been disabled.");
```
blva: the pattern we're leaning towards lately is to return a response to the LLM instead, so that the customer understands what's happening. 

probably worth adding a log and sending a result to the LLM -> something went wrong setting up the local atlas environment / check logs or disable the tools

src/server.ts
jeroenvervaeke: Good point, I missed this. Updated

src/tools/atlasLocal/atlasLocalTool.ts
jeroenvervaeke: I'm leaning towards keeping this error because it can only occur due to a developer error.

This error should never happen because the `atlas-local` tools are only added after we've been able to connect to the atlas local client. If we fail to connect, the client won't be set, and the tools won't be added.

I have improved the comment describing why it's ok to throw here (because it'll never happen).
Please let me know if you still prefer me to return an LLM response instead.

blva: Thnanks for the context! 

Yes, I think we should still throw an LLM response. MCP is used by non-developers and when you throw the error, this doesn't always pop up to the user or to the MCP client. If you provide a proper error you can guide users to the right experience.

examples:
*  this failed because the user doesn't have docker running. Atlas local should be disable or docker installed
*  something went wrong with the atlas local tools, check the MCP server logs and report to us

jeroenvervaeke: Updated the error handling to be more graceful

tests/integration/tools/atlas-local/listDeployments.test.ts
blva: thx!

src/server.ts
```diff
@@ -197,8 +200,41 @@ export class Server {
...
+            // Import Atlas Local client asyncronously
+            // This will fail on unsupported platforms
+            const { Client: AtlasLocalClient } = await import("@mongodb-js-preview/atlas-local");
+
+            // Connect to Atlas Local client
+            // This will fail if docker is not running
+            const client = AtlasLocalClient.connect();
+
+            // Set Atlas Local client
+            this.session.setAtlasLocalClient(client);
```
blva, line 219: for streamable http, we'll have multiple sessions that will connect to the server and manage the local atlas env, do we think this is the best place to add set client or should we set it somewhere else? if you think we need a single client, you could check the transport file `base.ts` and how we inject DeviceId (long-running operation that blocks other calls) or `Telemetry`

blva:
Could we also decouple the atlas local tests? in `code-health` we can have a `Run Atlas Local` tests and update the command in `Run Atlas` tests

package.json
```diff
@@ -121,6 +121,7 @@
     "node": "^20.19.0 || ^22.12.0 || >= 23.0.0"
   },
   "optionalDependencies": {
+    "@mongodb-js-preview/atlas-local": "^0.0.0-preview.1",
```
blva, line 124: [q] sorry for asking this again, I remember Kevin asked to make this optional, but from the npm package publishing, do you know if this will be bundled into the package or will customers need to download this separately if they want to run the npm package with local atlas?

jeroenvervaeke: It'll be downloaded like any regular package; the only difference is that if the package is unavailable or unable to install (unsupported platform), then installing the MCP server won't fail.

If `@mongodb-js-preview/atlas-local` was marked as a regular dependency and you attempt to install `mongodb-mcp-server`, the installation will fail.

blva: thanks for clarifying!

src/tools/atlasLocal/atlasLocalTool.ts
blva: thank you, resolved

src/server.ts
jeroenvervaeke: It is ok to have multiple instances of the client, it should not cause any issues

blva (approved):
Thanks for all the changes and context, LGTM as we can split the tests as a follow-up

PR #528, opened 2025-09-08
feat: Adds the atlas-list-performance-advisor base tool

kylelai1:
## Proposed changes
This PR adds the atlas-list-performance-advisor tool to the MCP server, which retrieves the following performance advisor recommendations from the admin API: index suggestions, drop index suggestions, schema suggestions, slow query logs.

This PR merges the changes into the [atlas-list-performance-advisor-tool](https://github.com/mongodb-js/mongodb-mcp-server/tree/atlas-list-performance-advisor-tool) branch.

## Testing
- Manually tested that the MCP server is able to retrieve performance advisor suggestions.

## Checklist


src/tools/atlas/read/listPerformanceAdvisor.ts
```diff
@@ -0,0 +1,94 @@
...
+    protected description = "List MongoDB Atlas performance advisor recommendations";
+    public operationType: OperationType = "read";
+    protected argsShape = {
+        projectId: z.string().describe("Atlas project ID to list performance advisor recommendations"),
+        clusterName: z.string().describe("Atlas cluster name to list performance advisor recommendations"),
+        operations: z
+            .array(z.nativeEnum(PerformanceAdvisorOperation))
+            .describe("Operations to list performance advisor recommendations"),
+        since: z.number().describe("Date to list slow query logs since").optional(),
+        processId: z.string().describe("Process ID to list slow query logs").optional(),
```
nirinchev: Is this something we expect the LLM to know how to get? As far as I can tell, you get it by calling `atlas processes list` but we don't have any tools that mirror that behavior in the MCP server.

src/tools/atlas/read/listPerformanceAdvisor.ts
```diff
@@ -0,0 +1,94 @@
...
+    public name = "atlas-list-performance-advisor";
+    protected description = "List MongoDB Atlas performance advisor recommendations";
+    public operationType: OperationType = "read";
+    protected argsShape = {
+        projectId: z.string().describe("Atlas project ID to list performance advisor recommendations"),
+        clusterName: z.string().describe("Atlas cluster name to list performance advisor recommendations"),
+        operations: z
+            .array(z.nativeEnum(PerformanceAdvisorOperation))
+            .describe("Operations to list performance advisor recommendations"),
+        since: z.number().describe("Date to list slow query logs since").optional(),
```
nirinchev: Should this be `z.date` instead? Does the LLM do a good job at converting dates to unix epoch?

src/tools/atlas/read/listPerformanceAdvisor.ts
```diff
@@ -0,0 +1,94 @@
...
+        const data: PerformanceAdvisorData = {
+            suggestedIndexes: [],
+            dropIndexSuggestions: { hiddenIndexes: [], redundantIndexes: [], unusedIndexes: [] },
+            slowQueryLogs: [],
+            schemaSuggestions: [],
+        };
+
+        // If operations is empty, get all performance advisor recommendations
+        // Otherwise, get only the specified operations
+        const operationsToExecute = operations.length === 0 ? Object.values(PerformanceAdvisorOperation) : operations;
```
nirinchev: Should we mark `operations` as optional and provide a default instead? Right now there's nothing to hint to the LLM it could provide an empty array here.

src/tools/atlas/read/listPerformanceAdvisor.ts
```diff
@@ -0,0 +1,94 @@
...
+
+        try {
+            if (operationsToExecute.includes(PerformanceAdvisorOperation.SUGGESTED_INDEXES)) {
+                const { suggestedIndexes } = await getSuggestedIndexes(this.session.apiClient, projectId, clusterName);
```
nirinchev: This is probably not super critical, but right now, all of these async operations are evaluated sequentially, which means that we need to wait for one to finish before starting the next one. Instead, it would be a good idea to run them in parallel.

src/tools/atlas/read/listPerformanceAdvisor.ts
```diff
@@ -0,0 +1,94 @@
...
+                        type: "text",
+                        text: `Error retrieving performance advisor data: ${error instanceof Error ? error.message : String(error)}`,
+                    },
+                ],
+            };
+        }
+
+        return {
+            content: [{ type: "text", text: JSON.stringify(data, null, 2) }],
+        };
```
nirinchev: We should wrap the response in `formatUntrustedData` to avoid injection attacks where someone creates a slow query that contains llm instructions. Also, it might be helpful to give hints to the llm what the different fields in the json data represent and how those can be used.

src/common/atlas/performanceAdvisorUtils.ts
```diff
@@ -0,0 +1,240 @@
...
+    };
+}
+
+interface SlowQueriesResponse {
+    slowQueries?: SlowQueryLog[];
+}
+
+interface DropIndexSuggestion {
+    accessCount?: number;
+    index?: Array<{ [key: string]: 1 | -1 }>;
```
nirinchev: Is this type definition true? Would PA not suggest dropping geo or text indexes?

src/common/atlas/performanceAdvisorUtils.ts
```diff
@@ -0,0 +1,240 @@
...
+    | "QUERIES_CONTAIN_CASE_INSENSITIVE_REGEX";
+
+type SchemaRecommedationType =
+    | "REDUCE_LOOKUP_OPS"
+    | "AVOID_UNBOUNDED_ARRAY"
+    | "REDUCE_DOCUMENT_SIZE"
+    | "REMOVE_UNNECESSARY_INDEXES"
+    | "REDUCE_NUMBER_OF_NAMESPACES"
+    | "OPTIMIZE_CASE_INSENSITIVE_REGEX_QUERIES"
+    | "OPTIMIZE_TEXT_QUERIES";
```
nirinchev: Do we need to translate these to something the LLM would have an easier time interpreting?

nirinchev (commented):
Did a quick pass - overall, looks reasonable, but I'm worried that it might not be too LLM-friendly. I suggest testing it thoroughly with different agents/models and confirming it's outputting meaningful insights.

src/common/atlas/performanceAdvisorUtils.ts
kylelai1: Let me look into this. I used the Open API definitions for the Atlas Admin API, but it would be weird to not include text indexes to drop suggestions for example (same for index creation suggestions!)

src/tools/atlas/read/listPerformanceAdvisor.ts
kylelai1: Since it's an optional, the LLM does not need to pass this in. It's also noted that it's only for slow query logs. If the processId is not passed in, we use inspect cluster to get the hostname + port that can be used for the process ID, which is handled by the performance advisor util functions that the atlas-list-performance-advisor tool calls.

I also think that when we do more manual testing in the "QA" phase of adding the performance advisor tool, we can more thoroughly test prompts and see how the LLM will prompt the user for more data.

src/tools/atlas/read/listPerformanceAdvisor.ts
kylelai1: Good point. Yes, the LLM does a good job of converting dates to unix epoch. I've manually tested just telling the LLM to get metrics for the past X hours for example, and it handled that well. 

I can change this to z.date

src/tools/atlas/read/listPerformanceAdvisor.ts
kylelai1: This is intended behavior here, so providing an empty array is ok. The LLM will ask the user which operations if we don't specify any, and it was discussed with product to just list all the PA suggestions if none were given here.

src/tools/atlas/read/listPerformanceAdvisor.ts
kylelai1: Will change this one!

src/tools/atlas/read/listPerformanceAdvisor.ts
kylelai1: Yes, I'll go ahead and make this change. I'll see how other tools use this.

src/tools/atlas/read/listPerformanceAdvisor.ts
nirinchev: If we know the model has no way of figuring out this processId, what's the point of exposing it as an argument?

src/tools/atlas/read/listPerformanceAdvisor.ts
nirinchev: Not sure I understand - right now this is a required argument, which tells the LLM to try and figure out a value for it - either by asking the user or hallucinating something. Instead, if we specify it as an optional argument with a default value, we still achieve the goals we set for ourselves in the PD, but we're also more clearly communicating the behavior of the server.

src/tools/atlas/read/listPerformanceAdvisor.ts
kylelai1: Hm, I might be approaching this incorrectly. This is something that if the user is able to provide, then it'll be passed in. Otherwise, we retrieve the process ID thru the tool. If this is something that the LLM is unable to figure out and we rely on the user to provide the `processId`, would this be convention to leave out the `processId` argument?

src/tools/atlas/read/listPerformanceAdvisor.ts
kylelai1: I see what you mean. There's nothing to suggest to the LLM that an empty array is ok to pass in here.
We can provide a default which is all operations in an array, and the LLM will be able to figure out that the default will then be all operations.

src/common/atlas/performanceAdvisorUtils.ts
kylelai1: I read some of the code where we return the index suggestions in the performance advisor, and there's a check to skip complex index types, which include geospatial, text, and hashed indexes. I've confirmed this with Frank (intel PM) to make sure.

src/common/atlas/performanceAdvisorUtils.ts
kylelai1: Added maps where we can get readable descriptions so that the LLM understands this better.

src/tools/atlas/read/listPerformanceAdvisor.ts
nirinchev: I'm just trying to wrap my head around the DX we're going for here - let's say I'm a user who's facing a slow app and I ask the agent to tell me what I could do. What is the question that I'd ask that would involve me including the `processId` and how would I get the process id to provide the model? If I need to go and use the Atlas CLI, I might as well use it for the perf recommendations, can't I?

As a rule of thumb, we'd want to expose arguments that users can feasibly know - e.g. my cluster name or the database/collection I'm interested in - or arguments that the model can deduce calling other tools exposed by the MCP server - e.g. the org/project id.

I'm not super familiar with PA - how many processes are expected to run on a given cluster and what would the slow query logs for different processes be used for? Are users generally interested in the slow query logs for all of them or for a single one? And would you say the default process id inference logic that we have is sufficient for all users (i.e. if we don't have the processId argument and always infer the processId, are we losing some functionality that would be valuable for someone)?

kylelai1: In that case, I think it would be good to remove the processId. The user wouldn't really know on their own to provide the process ID to get slow query logs, so we can remove it and have the tool grab the proper process IDs. Users are generally interested in the primary host slow queries, or slow queries in the whole cluster, so slow queries for all clusters.

My idea is that for now, we can remove the processId argument, and when we more thoroughly test the tool, we can revisit if we need to retrieve process IDs a certain way, or if there is unexpected behavior or prompting from the LLM to feed to the tool when getting the slow query logs.

src/common/atlas/cluster.ts
```diff
@@ -96,3 +97,23 @@ export async function inspectCluster(apiClient: ApiClient, projectId: string, cl
...
+    clusterName: string
+): Promise<string> {
+    try {
+        const cluster = await inspectCluster(apiClient, projectId, clusterName);
+        if (!cluster.connectionString) {
+            throw new Error("No connection string available for cluster");
+        }
+        const url = new URL(cluster.connectionString);
+        const processId = `${url.hostname}:${url.port || DEFAULT_PORT}`;
+        return processId;
```
nirinchev:
```suggestion
        return `${url.hostname}:${url.port || DEFAULT_PORT}`;
```

src/common/atlas/performanceAdvisorUtils.ts
```diff
@@ -0,0 +1,358 @@
+import { LogId } from "../logger.js";
+import type { ApiClient } from "./apiClient.js";
+import { getProcessIdFromCluster } from "./cluster.js";
+
+export enum PerformanceAdvisorOperation {
+    SUGGESTED_INDEXES = "suggestedIndexes",
+    DROP_INDEX_SUGGESTIONS = "dropIndexSuggestions",
+    SLOW_QUERY_LOGS = "slowQueryLogs",
+    SCHEMA_SUGGESTIONS = "schemaSuggestions",
+}
```
nirinchev: Let's drop this as it doesn't provide too much value and we can just use type inference to get the same functionality (I'll drop a suggestion in the tool file for how that'd look like).


```suggestion
```

src/tools/atlas/read/listPerformanceAdvisor.ts
```diff
@@ -0,0 +1,144 @@
...
+    getSchemaAdvice,
+    getSlowQueries,
+    PerformanceAdvisorOperation,
+    type PerformanceAdvisorData,
+    formatSuggestedIndexesTable,
+    formatDropIndexesTable,
+    formatSlowQueriesTable,
+    formatSchemaSuggestionsTable,
+} from "../../../common/atlas/performanceAdvisorUtils.js";
+
```
nirinchev: Let's define the zod enum here and then reuse it in schema definition


```suggestion
const PerformanceAdvisorOperationType = z.enum([
    "suggestedIndexes",
    "dropIndexSuggestions",
    "slowQueryLogs",
    "schemaSuggestions",
]);
```

src/tools/atlas/read/listPerformanceAdvisor.ts
```diff
@@ -0,0 +1,144 @@
...
+export class ListPerformanceAdvisorTool extends AtlasToolBase {
+    public name = "atlas-list-performance-advisor";
+    protected description = "List MongoDB Atlas performance advisor recommendations";
+    public operationType: OperationType = "read";
+    protected argsShape = {
+        projectId: z.string().describe("Atlas project ID to list performance advisor recommendations"),
+        clusterName: z.string().describe("Atlas cluster name to list performance advisor recommendations"),
+        operations: z
+            .array(z.nativeEnum(PerformanceAdvisorOperation))
+            .default(Object.values(PerformanceAdvisorOperation))
```
nirinchev:
```suggestion
            .array(PerformanceAdvisorOperationType)
            .default(PerformanceAdvisorOperationType.options)
```

src/tools/atlas/read/listPerformanceAdvisor.ts
```diff
@@ -0,0 +1,144 @@
...
+            suggestedIndexes: [],
+            dropIndexSuggestions: { hiddenIndexes: [], redundantIndexes: [], unusedIndexes: [] },
+            slowQueryLogs: [],
+            schemaSuggestions: [],
+        };
+
+        try {
+            const performanceAdvisorPromises = [];
+
+            if (operations.includes(PerformanceAdvisorOperation.SUGGESTED_INDEXES)) {
```
nirinchev:
```suggestion
            if (operations.includes("suggestedIndexes")) {
```

src/tools/atlas/read/listPerformanceAdvisor.ts
```diff
@@ -0,0 +1,144 @@
...
+                performanceAdvisorPromises.push(
+                    getSuggestedIndexes(this.session.apiClient, projectId, clusterName).then(({ suggestedIndexes }) => {
+                        data.suggestedIndexes = suggestedIndexes;
+                    })
+                );
+            }
+
+            if (operations.includes(PerformanceAdvisorOperation.DROP_INDEX_SUGGESTIONS)) {
```
nirinchev:
```suggestion
            if (operations.includes("dropIndexSuggestions")) {
```

src/tools/atlas/read/listPerformanceAdvisor.ts
```diff
@@ -0,0 +1,144 @@
...
+                performanceAdvisorPromises.push(
+                    getDropIndexSuggestions(this.session.apiClient, projectId, clusterName).then(
+                        ({ hiddenIndexes, redundantIndexes, unusedIndexes }) => {
+                            data.dropIndexSuggestions = { hiddenIndexes, redundantIndexes, unusedIndexes };
+                        }
+                    )
+                );
+            }
+
+            if (operations.includes(PerformanceAdvisorOperation.SLOW_QUERY_LOGS)) {
```
nirinchev:
```suggestion
            if (operations.includes("slowQueryLogs")) {
```

src/tools/atlas/read/listPerformanceAdvisor.ts
```diff
@@ -0,0 +1,144 @@
...
+                performanceAdvisorPromises.push(
+                    getSlowQueries(this.session.apiClient, projectId, clusterName, since, namespaces).then(
+                        ({ slowQueryLogs }) => {
+                            data.slowQueryLogs = slowQueryLogs;
+                        }
+                    )
+                );
+            }
+
+            if (operations.includes(PerformanceAdvisorOperation.SCHEMA_SUGGESTIONS)) {
```
nirinchev:
```suggestion
            if (operations.includes("schemaSuggestions")) {
```

src/common/atlas/performanceAdvisorUtils.ts
```diff
@@ -0,0 +1,358 @@
...
+        const response = await apiClient.listClusterSuggestedIndexes({
+            params: {
+                path: {
+                    groupId: projectId,
+                    clusterName,
+                },
+            },
+        });
+        return {
+            suggestedIndexes: (response as SuggestedIndexesResponse).content.suggestedIndexes ?? [],
```
nirinchev, line 37: This is concerning - why are we recasting the response to a different shape? This seems to indicate the definitions we've generated from the open api spec are not correct.

src/common/atlas/performanceAdvisorUtils.ts
```diff
@@ -0,0 +1,358 @@
...
+}
+
+export interface SuggestedIndex {
+    avgObjSize?: number;
+    id?: string;
+    impact?: Array<string>;
+    index?: Array<{ [key: string]: 1 | -1 }>;
+    namespace?: string;
+    weight?: number;
+}
```
nirinchev: This and the other type definitions already exist in the open api definitions - we shouldn't be redefining them here. If you want a more convenient type to work with, you could do something like

```ts
export type SuggestedIndex = components["schemas"]["PerformanceAdvisorIndex"];
```

src/common/atlas/performanceAdvisorUtils.ts
```diff
@@ -0,0 +1,358 @@
...
+        apiClient.logger.debug({
+            id: LogId.atlasPaSlowQueryLogsFailure,
+            context: "performanceAdvisorUtils",
+            message: `Failed to list slow query logs: ${err instanceof Error ? err.message : String(err)}`,
+        });
+        throw new Error(`Failed to list slow query logs: ${err instanceof Error ? err.message : String(err)}`);
+    }
+}
+
+export function formatSuggestedIndexesTable(suggestedIndexes: Array<SuggestedIndex>): string {
```
nirinchev: I know we're doing a lot of formatting in other Atlas tools, but I'm not convinced this is particularly useful - the model itself doesn't care about the format and in my testing, it doesn't really return the data in tabular format, but instead reformats it in its own way. I would suggest reducing the complexity here and returning the json as is (or maybe enhance it with some pointers for the LLM).

src/common/atlas/performanceAdvisorUtils.ts
kylelai1: Yes, that's correct. I've manually tested this with the MCP server tool, and separately by making requests to the admin api (v2), and found that the response from the open API spec is not correct. It is missing the `content` field, which I've added when accessing the `suggestedIndexes`. This applies to the 4 different PA endpoints we're calling as a part of this tool.

src/common/atlas/performanceAdvisorUtils.ts
kylelai1: Ah thanks! replacing this one and the other type definitions for the other PA endpoint responses too

src/common/atlas/performanceAdvisorUtils.ts
kylelai1: Changed this back to returning a JSON for simplicity! The LLM should be able to read the raw JSON format since it isn't a human, so we don't need to pretty format it.

src/common/atlas/performanceAdvisorUtils.ts
nirinchev: I wonder why the other responses from the API client don't have the same problem. @blva's team owns that part of the codebase, so perhaps something to look into?

src/common/atlas/performanceAdvisorUtils.ts
```diff
@@ -0,0 +1,216 @@
...
+};
+
+type SchemaRecommedationType =
+    | "REDUCE_LOOKUP_OPS"
+    | "AVOID_UNBOUNDED_ARRAY"
+    | "REDUCE_DOCUMENT_SIZE"
+    | "REMOVE_UNNECESSARY_INDEXES"
+    | "REDUCE_NUMBER_OF_NAMESPACES"
+    | "OPTIMIZE_CASE_INSENSITIVE_REGEX_QUERIES"
+    | "OPTIMIZE_TEXT_QUERIES";
```
nirinchev:
```suggestion
type SchemaRecommedationType = NonNullable<components["schemas"]["SchemaAdvisorItemRecommendation"]["recommendation"]>;
```

src/common/atlas/performanceAdvisorUtils.ts
```diff
@@ -0,0 +1,216 @@
...
+    affectedNamespaces?: Array<{
+        namespace?: string | null;
+        triggers?: Array<{
+            description?: string;
+            triggerType?: SchemaTriggerType;
+        }>;
+    }>;
+    description?: string;
+    recommendation?: SchemaRecommedationType;
+}
```
nirinchev:
```suggestion
export type SchemaRecommendation = components["schemas"]["SchemaAdvisorItemRecommendation"];
```

src/common/atlas/performanceAdvisorUtils.ts
```diff
@@ -0,0 +1,216 @@
...
+export type DropIndexSuggestion = components["schemas"]["DropIndexSuggestionsIndex"];
+
+export type SlowQueryLogMetrics = components["schemas"]["PerformanceAdvisorSlowQueryMetrics"];
+
+export type SlowQueryLog = components["schemas"]["PerformanceAdvisorSlowQuery"];
+
+interface SuggestedIndexesResponse {
+    content: {
+        suggestedIndexes?: Array<SuggestedIndex>;
+    };
```
nirinchev:
```suggestion
    content: components["schemas"]["PerformanceAdvisorResponse"];
```

src/common/atlas/performanceAdvisorUtils.ts
```diff
@@ -0,0 +1,216 @@
...
+}
+
+interface DropIndexesResponse {
+    content: {
+        hiddenIndexes?: Array<DropIndexSuggestion>;
+        redundantIndexes?: Array<DropIndexSuggestion>;
+        unusedIndexes?: Array<DropIndexSuggestion>;
+    };
```
nirinchev:
```suggestion
    content: components["schemas"]["DropIndexSuggestionsResponse"];
```

src/common/atlas/performanceAdvisorUtils.ts
```diff
@@ -0,0 +1,216 @@
...
+}
+
+interface SchemaAdviceResponse {
+    content: {
+        recommendations?: Array<SchemaRecommendation>;
+    };
```
nirinchev:
```suggestion
    content: components["schemas"]["SchemaAdvisorResponse"];
```

src/common/atlas/performanceAdvisorUtils.ts
```diff
@@ -0,0 +1,216 @@
...
+}
+
+interface SlowQueriesResponse {
+    slowQueries?: Array<SlowQueryLog>;
+}
```
nirinchev:
```suggestion
```

src/common/atlas/performanceAdvisorUtils.ts
```diff
@@ -0,0 +1,216 @@
...
+
+export type SchemaTriggerType =
+    | "PERCENT_QUERIES_USE_LOOKUP"
+    | "NUMBER_OF_QUERIES_USE_LOOKUP"
+    | "DOCS_CONTAIN_UNBOUNDED_ARRAY"
+    | "NUMBER_OF_NAMESPACES"
+    | "DOC_SIZE_TOO_LARGE"
+    | "NUM_INDEXES"
+    | "QUERIES_CONTAIN_CASE_INSENSITIVE_REGEX";
```
nirinchev:
```suggestion
export type SchemaTriggerType = NonNullable<components["schemas"]["SchemaAdvisorTriggerDetails"]["triggerType"]>;
```

src/common/atlas/performanceAdvisorUtils.ts
```diff
@@ -0,0 +1,216 @@
...
+        triggers?: Array<{
+            description?: string;
+            triggerType?: SchemaTriggerType;
+        }>;
+    }>;
+    description?: string;
+    recommendation?: SchemaRecommedationType;
+}
+
+export interface PerformanceAdvisorData {
```
nirinchev: This is not being used here and there's no obvious reason to export it. We can move the type to `listPerformanceAdvisor` since it's just a bag of values we're using to aggregate the responses.

src/common/atlas/performanceAdvisorUtils.ts
```diff
@@ -0,0 +1,216 @@
...
+                    processId,
+                },
+                query: {
+                    ...(since && { since: since.getTime() }),
+                    ...(namespaces && { namespaces: namespaces }),
+                },
+            },
+        });
+
+        return { slowQueryLogs: (response as SlowQueriesResponse).slowQueries ?? [] };
```
nirinchev:
```suggestion
        return { slowQueryLogs: response.slowQueries ?? [] };
```

This doesn't seem to have the `content` wrapper around it?

src/tools/atlas/read/listPerformanceAdvisor.ts
```diff
@@ -0,0 +1,132 @@
...
+                    })
+                );
+            }
+
+            if (operations.includes("dropIndexSuggestions")) {
+                performanceAdvisorPromises.push(
+                    getDropIndexSuggestions(this.session.apiClient, projectId, clusterName).then(
+                        ({ hiddenIndexes, redundantIndexes, unusedIndexes }) => {
+                            data.dropIndexSuggestions = { hiddenIndexes, redundantIndexes, unusedIndexes };
+                        }
```
nirinchev: No need to deconstruct the suggestions here it seems:


```suggestion
                        (dropIndexSuggestions) => {
                            data.dropIndexSuggestions = dropIndexSuggestions;
                        }
```

src/common/atlas/performanceAdvisorUtils.ts
nirinchev: Additionally, all the fields should be optional.

src/tools/atlas/read/listPerformanceAdvisor.ts
```diff
@@ -0,0 +1,132 @@
...
+} from "../../../common/atlas/performanceAdvisorUtils.js";
+
+const PerformanceAdvisorOperationType = z.enum([
+    "suggestedIndexes",
+    "dropIndexSuggestions",
+    "slowQueryLogs",
+    "schemaSuggestions",
+]);
+export class ListPerformanceAdvisorTool extends AtlasToolBase {
+    public name = "atlas-list-performance-advisor";
```
nirinchev: The name is a bit awkward - "list performance advisor" doesn't read naturally. I would suggest renaming it to either `atlas-get-performance-advice` or just `atlas-performance-advisor`.

src/tools/atlas/read/listPerformanceAdvisor.ts
```diff
@@ -0,0 +1,132 @@
...
+    public operationType: OperationType = "read";
+    protected argsShape = {
+        projectId: z.string().describe("Atlas project ID to list performance advisor recommendations"),
+        clusterName: z.string().describe("Atlas cluster name to list performance advisor recommendations"),
+        operations: z
+            .array(PerformanceAdvisorOperationType)
+            .default(PerformanceAdvisorOperationType.options)
+            .describe("Operations to list performance advisor recommendations"),
+        since: z.date().describe("Date to list slow query logs since").optional(),
+        namespaces: z.array(z.string()).describe("Namespaces to list slow query logs").optional(),
```
nirinchev: We should probably clarify in the description that this is only relevant if `slowQueryLogs` is included in `operations`.

src/tools/atlas/read/listPerformanceAdvisor.ts
```diff
@@ -0,0 +1,132 @@
...
+
+const PerformanceAdvisorOperationType = z.enum([
+    "suggestedIndexes",
+    "dropIndexSuggestions",
+    "slowQueryLogs",
+    "schemaSuggestions",
+]);
+export class ListPerformanceAdvisorTool extends AtlasToolBase {
+    public name = "atlas-list-performance-advisor";
+    protected description = "List MongoDB Atlas performance advisor recommendations";
```
nirinchev: Do we want to include some hints for the model in which cases it should call this tool?

src/common/atlas/performanceAdvisorUtils.ts
```diff
@@ -0,0 +1,216 @@
...
+export type SchemaTriggerType =
+    | "PERCENT_QUERIES_USE_LOOKUP"
+    | "NUMBER_OF_QUERIES_USE_LOOKUP"
+    | "DOCS_CONTAIN_UNBOUNDED_ARRAY"
+    | "NUMBER_OF_NAMESPACES"
+    | "DOC_SIZE_TOO_LARGE"
+    | "NUM_INDEXES"
+    | "QUERIES_CONTAIN_CASE_INSENSITIVE_REGEX";
+
+export const SCHEMA_TRIGGER_DESCRIPTIONS: Record<SchemaTriggerType, string> = {
```
nirinchev: These don't appear to be used now.

nirinchev (commented):
Looking better. I've left more suggestions for further deduplicating the types. The only major thing that remains is figuring out why the openapi definitions are wrong and what it would take to fix them.

src/common/atlas/performanceAdvisorUtils.ts
kylelai1: Yes, the slow query logs response does not have a `content` wrapper around it. This is also based on testing the PA admin api endpoint manually.

src/tools/atlas/read/listPerformanceAdvisor.ts
kylelai1: I think the reasoning behind the "list performance advisor" name was to make it apparent that it is a read only tool where it lists the recommendations. I agree that the name can be tuned. For now, let's keep the name as is, and then I can work with PM/design to confirm with them a new name in a future PR.

kmruiz: Why would require this a future PR or require syncing with design? The class name is private to us, and changing the public name to the LLM shouldn't have any negative effect for the agent, but it can be a pain for users that disabled this tool specifically (they have to disable/enable it again). The sooner we fix this, the better.

I think @nirinchev suggestions are fine, so any of these work for me. If we want to be closer to the naming conventions we have, that are action based, we could also use `atlas-get-performance-advisor-recommendations` or along the lines.

src/tools/atlas/read/listPerformanceAdvisor.ts
```diff
@@ -0,0 +1,146 @@
...
+        operations: z
+            .array(PerformanceAdvisorOperationType)
+            .default(PerformanceAdvisorOperationType.options)
+            .describe("Operations to list performance advisor recommendations"),
+        since: z.date().describe("Date to list slow query logs since").optional(),
+        namespaces: z
+            .array(z.string())
+            .describe("Namespaces to list slow query logs. Only relevant for the slowQueryLogs operation.")
+            .optional(),
+    };
```
kmruiz: I think this requires some second thoughts.

When the user is connected to an Atlas Cluster, we have already the projectId and clusterName, which should be enough. I think we should leverage this as a default. This would make projectId and clusterName optional.

You can retrieve the current connected cluster from the session:

this.session.connectedAtlasCluster

src/tools/atlas/read/listPerformanceAdvisor.ts
```diff
@@ -0,0 +1,146 @@
...
+            suggestedIndexes: [],
+            dropIndexSuggestions: { hiddenIndexes: [], redundantIndexes: [], unusedIndexes: [] },
+            slowQueryLogs: [],
+            schemaSuggestions: [],
+        };
+
+        try {
+            const performanceAdvisorPromises = [];
+
+            if (operations.includes("suggestedIndexes")) {
```
kmruiz: We shouldn't mutate an array and wait for promises in another array. This is going to be misleading in the future and error prone. The issue we have can be solved by using Promise.all already.

```
const [ suggestedIndexes, dropIndexSuggestions ] = (await Promise.all([
   operations.includes("suggestedIndexes") && getSuggestedIndexes(...),
   operations.includes("dropIndexSuggestions") && getdropIndexSuggestions(...)
]))
```

src/tools/atlas/read/listPerformanceAdvisor.ts
kylelai1: Let's change it to `atlas-get-performance-advisor` then. It's customer facing when the user asks for available tools, but I think it's fine to change as it's private to us as you said.

src/tools/atlas/read/listPerformanceAdvisor.ts
kylelai1: If the user has an atlas cluster already connected, the LLM should be able to provide the `projectId` and `clusterName` already without any additional user input. If there hasn't been a specified cluster, then the LLM will prompt the user to provide the `projectId` and `clusterName`.

if we were to make the `projectId` and `clusterName` default, then we'd need to build it in to the tool to retrieve this from the connected cluster as you said.

Do we find that the LLM prompting the user for this is something we want to avoid, and that we want to instead have the tool handle as much of that as possible? Even if the `projectId` and `clusterName` are not optionals, the LLM will still provide these if there was already a previous connected cluster.

src/tools/atlas/read/listPerformanceAdvisor.ts
kylelai1: Good suggestion. I'll make this change

kylelai1: Clean this up. Also cleaned up some of the logic building the response the tool returns to the LLM.

src/tools/atlas/read/listPerformanceAdvisor.ts
nirinchev: I don't think there's an explicit expectation that the user should be connected to the cluster (via a mongodb connection) before they're able to ask for performance advice. Additionally, many other Atlas tools follow the same pattern of requiring a cluster name and project id, so I think leaving things as they are makes sense.

src/tools/atlas/read/getPerformanceAdvisor.ts
```diff
@@ -0,0 +1,98 @@
...
+        "Get MongoDB Atlas performance advisor recommendations, which includes the operations: suggested indexes, drop index suggestions, slow query logs, and schema suggestions";
+    public operationType: OperationType = "read";
+    protected argsShape = {
+        projectId: z.string().describe("Atlas project ID to get performance advisor recommendations"),
+        clusterName: z.string().describe("Atlas cluster name to get performance advisor recommendations"),
+        operations: z
+            .array(PerformanceAdvisorOperationType)
+            .default(PerformanceAdvisorOperationType.options)
+            .describe("Operations to get performance advisor recommendations"),
+        since: z.date().describe("Date to get slow query logs since").optional(),
```
nirinchev: This is also only relevant for the slow query logs.

src/tools/atlas/read/getPerformanceAdvisor.ts
```diff
@@ -0,0 +1,98 @@
...
+                        : { recommendations: [] },
+                ]);
+
+            const performanceAdvisorData = [
+                suggestedIndexesResult?.suggestedIndexes?.length > 0
+                    ? `## Suggested Indexes\n${JSON.stringify(suggestedIndexesResult.suggestedIndexes)}`
+                    : "No suggested indexes found.",
+                dropIndexSuggestionsResult
+                    ? `## Drop Index Suggestions\n${JSON.stringify(dropIndexSuggestionsResult)}`
+                    : "No drop index suggestions found.",
```
nirinchev, line 73: The way this is currently structured, this will always be truthy - if `operations` doesn't include `dropIndexSuggestions`, we return an object, which will evaluate to `true`.

src/tools/atlas/read/getPerformanceAdvisor.ts
```diff
@@ -0,0 +1,98 @@
...
+        operations,
+        since,
+        namespaces,
+    }: ToolArgs<typeof this.argsShape>): Promise<CallToolResult> {
+        try {
+            const [suggestedIndexesResult, dropIndexSuggestionsResult, slowQueryLogsResult, schemaSuggestionsResult] =
+                await Promise.all([
+                    operations.includes("suggestedIndexes")
+                        ? getSuggestedIndexes(this.session.apiClient, projectId, clusterName)
+                        : { suggestedIndexes: [] },
```
nirinchev: Can we just return `undefined` here instead of constructing an empty object?

nirinchev (approved):
Looks good! Final nits from me. We should investigate the `.content` issue, but I'd be fine if that's resolved in a follow-up PR.

src/tools/atlas/read/listPerformanceAdvisor.ts
kmruiz: Arguably this is the first tool that is actually relevant for a single cluster where the user is connected.

However, I am not going to die on this hill, but at least I would suggest to use the AtlasArgs types we have in args.ts, instead of Zod primitives, to align to what other Atlas tools do.

Edit:

The Atlas connect tool does not provide the projectId of the cluster, so my guess is that we trust the agent to provide it from another place after somehow listing projects and clusters.

I think this could be avoided and simplified making these args optional, so a prompt as easy as: give me the performance suggestions of the current cluster" would work even if rhe user never queried for other projects, action that might not have permissions to do.

kylelai1: If there is already a connected cluster, then a simple prompt should be able to be understood by the LLM without prompting for the projectID and clusterName again. We can revisit this during QA when we manually test the tool with a connected cluster or without a connected cluster to see if it will prompt the user for the projectId and clusterName again.

Looks like AtlasArgs was not created when I started work for this ticket, so thanks for pointing that out! I'll make that change.

src/tools/atlas/read/getPerformanceAdvisor.ts
kylelai1: Added that we use this for the `slowQueryLogs` operation only.

src/tools/atlas/read/getPerformanceAdvisor.ts
kylelai1: Took the suggestion below to use `undefined` so this will no longer always evaluate to true

src/tools/atlas/read/getPerformanceAdvisor.ts
kylelai1: Changed to return `undefined`.

nirinchev (approved):
Hm... Looks like you'd need to update the base branch to latest main, otherwise this PR brings in a bunch of stuff, making it harder to review.

src/common/atlas/performanceAdvisorUtils.ts
blva: I don't see content in `ApiPerformanceAdvisorResponseView` (schema name `PerformanceAdvisorResponse`) shape in [MMS](https://github.com/10gen/mms/blob/63083b09411fb4883fe25b236a938ed4f0281d49/server/src/main/com/xgen/svc/mms/api/view/monitoring/performanceadvisor/ApiPerformanceAdvisorResponseView.java#L20). 

Why are we adding this interface and why do we need this content field? I think you should be able to use the openapi specs as they are.

```
interface SuggestedIndexesResponse {
    content: components["schemas"]["PerformanceAdvisorResponse"];
}
```

You don't need to import ` content: components["schemas"]["PerformanceAdvisorResponse"];` here, you should probably use `PerformanceAdvisorResponse` from `openapi.d.ts`, right?

PR #536, opened 2025-09-09
fix: add guards against possible memory overflow in find and aggregate tools MCP-21

himanshusinghs:
## Proposed changes
This PR targets find and aggregate tool and does the following to `hopefully` guard against the memory overflow possibility:
1. Adds a configurable limit to restrict the maximum number of documents fetched per query / aggregation. This defaults to 100 and can be configured through CLI args / env variables. This limit will be applied in addition to find tool's provided limit (default to 10) and the smaller one will be considered always.
2. Replaces the use of `Cursor.toArray` with a dedicated cursor iterator that keeps track of bytes consumed in memory by the retrieved documents and cuts off the cursor iteration when there is a possibility of overflow. The overflow is based on configured `maxBytesPerQuery` parameter which defaults to 16MB. The tools that use the cursor iterator such as find and aggregate also expose `responseBytesLimit` as one of the parameter that defaults to 1MB to allow LLMs to invoke the tool with a bigger limit if needed, capped by the configured maxBytesPerQuery.

Some key decisions that were taken on a call and not documented anywhere:
- [x] We will use hard limit cap to limit the number of documents retrieved by the query and the aggregation.
- [x] We will avoid using `.toArray` and replace that with a cursor restricted to a configured limit.
- [x] We will introduce a concept of configurable max bytes per query and by default will keep it relatively lower (1MB) and will restrict the cursor iteration to be within this limit.

## Note for reviewers

Please review with whitespaces turned off.

## Checklist


src/tools/mongodb/read/find.ts
```diff
@@ -45,22 +48,57 @@ export class FindTool extends MongoDBToolBase {
...
+            // Check if find operation uses an index if enabled
+            if (this.config.indexCheck) {
+                await checkIndexUsage(provider, database, collection, "find", async () => {
+                    return provider
+                        .find(database, collection, filter, { projection, limit, sort })
+                        .explain("queryPlanner");
+                });
+            }
+
+            const appliedLimit = Math.min(limit, this.config.maxDocumentsPerQuery);
```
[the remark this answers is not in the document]
himanshusinghs: I have addressed this in a different way. My last commit removes the default limit and ensures that we either use the provided limit (if a number) or the configured max limit.

src/common/config.ts
```diff
@@ -180,6 +182,8 @@ export const defaultUserConfig: UserConfig = {
     idleTimeoutMs: 600000, // 10 minutes
     notificationTimeoutMs: 540000, // 9 minutes
     httpHeaders: {},
+    maxDocumentsPerQuery: 10, // By default, we only fetch a maximum 10 documents per query / aggregation
+    maxBytesPerQuery: 1 * 1000 * 1000, // By default, we only return ~1 mb of data per query / aggregation
```
blva: nice! the names you chose are intuitive imo

src/tools/mongodb/read/aggregate.ts
```diff
@@ -25,27 +30,50 @@ export class AggregateTool extends MongoDBToolBase {
...
+                        .aggregate(database, collection, pipeline, {}, { writeConcern: undefined })
+                        .explain("queryPlanner");
+                });
+            }
 
-        const documents = await provider.aggregate(database, collection, pipeline).toArray();
+            const cappedResultsPipeline = [...pipeline, { $limit: this.config.maxDocumentsPerQuery }];
+            aggregationCursor = provider
+                .aggregate(database, collection, cappedResultsPipeline)
+                .batchSize(this.config.maxDocumentsPerQuery);
```
blva: [q] now that this is a default config, how can a user turn this off if they want to go back to the old behaviour (unlimited queries)?

src/tools/mongodb/read/find.ts
```diff
@@ -45,22 +48,60 @@ export class FindTool extends MongoDBToolBase {
...
+            // Check if find operation uses an index if enabled
+            if (this.config.indexCheck) {
+                await checkIndexUsage(provider, database, collection, "find", async () => {
+                    return provider
+                        .find(database, collection, filter, { projection, limit, sort })
+                        .explain("queryPlanner");
+                });
+            }
+
+            const limitOnFindCursor = Math.min(limit ?? Number.POSITIVE_INFINITY, this.config.maxDocumentsPerQuery);
```
blva: [q] what's the purpose of this if you compare this with infinity?

src/tools/mongodb/read/find.ts
```diff
@@ -45,22 +48,60 @@ export class FindTool extends MongoDBToolBase {
         limit,
         sort,
     }: ToolArgs<typeof this.argsShape>): Promise<CallToolResult> {
-        const provider = await this.ensureConnected();
+        let findCursor: FindCursor<unknown> | undefined;
+        try {
```
blva, line 52: [q] why is ensureConnected moved to the try /catch? I'm guessing that we handle any potential exception in the parent tool `handleError`

src/tools/mongodb/read/aggregate.ts
himanshusinghs: Ahh I missed this point, sorry about that, will loop it in.

src/tools/mongodb/read/find.ts
himanshusinghs: So the infinity is compared only when limit is not provided (`undefined`) or is provided but the value is `null`. And both are not valid values which is why we want the configured limit to be applied and that's why the comparison.

I guess that's a sign for me to a more readable statement, if possible?

src/tools/mongodb/read/find.ts
himanshusinghs: Yea that's true, and if you notice we don't really catch the error in this try finally block. Its done only to ensure that we close the opened cursor if any in the finally block.

src/common/config.ts
gagik: very nit and I get confused by this every time but if its bytes wouldn't this be `1 * 1024 * 1024`

gagik: also, isn't 1 megabyte very little? 16 MB is the limit of the maximum size for a single document in [MongoDB](https://www.mongodb.com/docs/manual/reference/limits/).

src/helpers/iterateCursor.ts
```diff
@@ -0,0 +1,39 @@
...
+import type { AggregationCursor, FindCursor } from "mongodb";
+
+/**
+ * This function attempts to put a guard rail against accidental memory overflow
+ * on the MCP server.
+ *
+ * The cursor is iterated until we can predict that fetching next doc won't
+ * exceed the maxBytesPerQuery limit.
+ */
+export async function iterateCursorUntilMaxBytes(
```
gagik: this could benefit from an abort controller so we can abort a potentially stuck or heavy loop

src/common/config.ts
himanshusinghs: technically yes :) I will make the change

himanshusinghs: So the original proposal was to have 16MB as the default limit but then we also made an assumption that given the MCP use case where we apply a max document limit of 10 docs by default, the cases where a single document would blow off the 1 MB limit could be considered uncommon. If you think that's not so uncommon, we can certainly revisit this assumption and come to an agreement on some other number.

gagik: > we also made an assumption that given the MCP use case where we apply a max document limit of 10 docs by default

If I ask for 100 documents, will this (memory-based) limit kick in? I just don't want people to run into this when they want to just get bunch of data in one go. From perspective of token usage I can see 1 MB as reasonable if all of it does get fed into the LLM. 

That said, I think 16 MB would a more reasonable limit imo. We don't want people with large documents to spend hours trying to figure out why the MCP server returns nothing. If returned 1 document it'd at least be a hint. Less of an issue if it's easy for the agent to omit this limit.

src/tools/mongodb/read/aggregate.ts
himanshusinghs: addressed in [97ad191](https://github.com/mongodb-js/mongodb-mcp-server/pull/536/commits/97ad191a3a0ed22678f6cddd9628f3acc791ceb6)

himanshusinghs: The solution is as per our discussion, negative or zero value is treated as sign to turn off the feature.

src/tools/mongodb/read/find.ts
himanshusinghs: I have tried making this conditional path more clearer in [97ad191](https://github.com/mongodb-js/mongodb-mcp-server/pull/536/commits/97ad191a3a0ed22678f6cddd9628f3acc791ceb6)

src/common/config.ts
himanshusinghs: > If I ask for 100 documents, will this (memory-based) limit kick in?

There is always a good chance of that happening. Imagine asking 100 documents but the first 2 documents already fills the 1MB memory limit.

> If returned 1 document it'd at least be a hint

I agree that we should at-least be able to return 1 document if that alone crosses the limit but that could be solved differently without changing the default limit of 1MB. We can just choose not to be too strict [here](https://github.com/mongodb-js/mongodb-mcp-server/pull/536/files#diff-e6fd4fb08be6ff8a11dc2c7a05f10b38f660cc54e451306120d342715d47a747R34-R37) and let the document overreaching the limit go through?

src/helpers/iterateCursor.ts
himanshusinghs: I tied the cursor iteration to the client request timeouts. Let me know if you had something else in mind about this.

Additionally I realised that our tool call execution live well after the client request timeouts. This is partially because the SDK doesn't handle it at their level and instead provide an AbortSignal for consumers to work with. I have create MCP-200 to track this.

gagik: > I tied the cursor iteration to the client request timeouts

yeah that makes sense

src/common/config.ts
gagik: To me, asking for 100 documents and getting 2 documents with no explanation is a regression or at least a breaking change. Can we at least provide an explanation back to the client?

I think a low limit will become noticeable for some people.

My other problem is that if the agent runs into this (i.e. "Oh it returned only 2 documents.. why is that?") it has no way to inside runtime override this config, thus leading to a weird and confusing state where the user needs to manually investigate, find this config, disable and restart the MCP server.

I'm wondering if instead this should be an argument on the tool call itself instead of a global configuration. Or at least we should have a more generous global configuration (say, 128mb) and less generous tool-level configuration (like 1mb, which still has an upper limit of the global configuration value)

src/tools/mongodb/read/find.ts
```diff
@@ -18,7 +21,7 @@ export const FindArgs = {
         .passthrough()
         .optional()
         .describe("The projection, matching the syntax of the projection argument of db.collection.find()"),
-    limit: z.number().optional().default(10).describe("The maximum number of documents to return"),
+    limit: z.number().optional().describe("The maximum number of documents to return"),
```
gagik: I actually think the 10 document default is a good thing and likely reason why people haven't complained about this. To me this is preferrable to a memory limit default as it's human parsable and contextually easier to define by the user's request. 

Now the essentially default could be 1000x more documents.

himanshusinghs: We're still maintaining 10 document default through config option - `maxDocumentsPerQuery`

gagik: but the LLM no longer knows about it and it's a limit not a default.

gagik: I don't think global settings for this make sense. I want 1 document sometimes and other times I want 1000. 10 is definetely too few for a global imit.

src/common/config.ts
himanshusinghs: > To me, asking for 100 documents and getting 2 documents with no explanation is a regression or at least a breaking change. Can we at least provide an explanation back to the client?

We do provide an explanation [here](https://github.com/mongodb-js/mongodb-mcp-server/pull/536/files#diff-24001908950f42c25dffa1ca5b44ddd222efb6c968d8b829f70eed19c1284397R65-R69). I can tweak the explanation to be more specific about why the documents were limited.

> My other problem is that if the agent runs into this (i.e. "Oh it returned only 2 documents.. why is that?") it has no way to inside runtime override this config, thus leading to a weird and confusing state where the user needs to manually investigate, find this config, disable and restart the MCP server.

The agent knows from the response that the final output was limited and there is even a hint to run export tool if the entire query results are required. This is also reflected in our [accuracy tests](https://github.com/mongodb-js/mongodb-mcp-server/blob/819ed01657a23c9a8952340670934b676e207818/tests/accuracy/find.test.ts#L115-L151).

> I'm wondering if instead this should be an argument on the tool call itself instead of a global configuration. Or at least we should have a more generous global configuration (say, 128mb) and less generous tool-level configuration (like 1mb, which still has an upper limit of the global configuration value)

I don't think we should be approaching it this way given that the entire motivation was to avoid having memory overflow problems in the MCP server. If there is a requirement to have larger data sets returned from the MCP server then it should be an informed decision reflected as overridden config.

gagik: Didn't realize there's an explanation thanks that is better! That said the agent can't iterate on it and the user must intervene so this isn't reflected.

> avoid having memory overflow problems

The worry in my head was from the MCP server using 10 GBs of memory not 1 Megabyte. I can only see the argument now from perspective of token use, memory usage is cheap. I think that's why I find the limit to be too extreme but it's fine we can stick with 1MB and iterate in the future if needed. 
For the export tool I don't see much benefit to be this conservative with memory use. 

> If there is a requirement to have larger data sets returned from the MCP server then it should be an informed decision reflected as overridden config.

I agree with this with regards to memory but setting 10 document limit as a global configuration is not a good default IMO and quite a breaking feature, maybe at least 100 or 1000.

src/tools/mongodb/read/find.ts
himanshusinghs: discussed over call - I will bring back the smaller default of 10 on the find tool and expand the configurable `maxDocumentsPerQuery` to be 100

src/common/config.ts
himanshusinghs: discussed over call - I will tweak the explanation to specify what kind of limit was applied (maxDocumentsPerQuery or maxBytesPerQuery)

kmruiz: One thing to clarify, MiB and MB are different units:

MB: 1000 * 1000 -> https://en.wikipedia.org/wiki/Megabyte  
MiB: 1024 * 1024 -> https://en.wikipedia.org/wiki/Mebibyte

Which one to use is kind of irrelevant for our use case, And agree with the improvements suggested overall.

nirinchev:
Going to let Kevin and Gagik review the actual code, but do the limits sound reasonable? If we're preventing OOM situations, 1 MB sounds incredibly conservative - can we instead base it off the available memory of the machine - e.g. using `v8.getHeapStatistics`?

himanshusinghs:
The memory limit is to applied per tool response, so letting one tool call take almost entire available heap size seems too lenient, particularly given the case that we now support multiple clients connecting to the same server.

I still think 16MB default through configuration and 1 MB on tool interface, (configurable by LLM and capped to configured limit) seems pretty reasonable.

src/helpers/iterateCursor.ts
```diff
@@ -0,0 +1,54 @@
...
+import type { AggregationCursor, FindCursor } from "mongodb";
+
+/**
+ * This function attempts to put a guard rail against accidental memory overflow
+ * on the MCP server.
+ *
+ * The cursor is iterated until we can predict that fetching next doc won't
+ * exceed the maxBytesPerQuery limit.
+ */
+export async function iterateCursorUntilMaxBytes({
```
kmruiz: 2 nits:

1. I think if we are iterating the cursor and collecting, collectUntilMemoryLimit or something like that.
2. I would like to have it better typed. For example, we can use something like (simplified):

```
export async function collectUntilMemoryLimit<T = unknown>(cursor: FindCursor<T>): Promise<T[]>
```

src/tools/mongodb/read/find.ts
```diff
@@ -37,30 +40,82 @@ export class FindTool extends MongoDBToolBase {
...
+            }
+
+            return {
+                content: formatUntrustedData(
+                    messageDescription,
+                    documents.length > 0 ? EJSON.stringify(documents) : undefined
+                ),
+            };
+        } finally {
+            await findCursor?.close();
```
kmruiz: I'm wondering if close is a fallible function. I think we should use void instead of await, or adding a try catch, to make sure if close fails, it doesn't do anything weird.

kmruiz (approved):

kmruiz:
Yes, the memory limit is per tool call, and 1MB of results is a big chunk of results, and even more when some models are pretty aggressively using projections to reduce the context they use when reading the results. I think the question here would be: why would we need more memory?

1. In the HTTP Transport server will likely run multiple tools for multiple users at the same time and we don't want noisy neighbors to overload the system. 
1. 1MB can hold a lot of data. To give you an example, in the sample_mflix.movies collection from our sample data, a single movie (with descriptions and summaries) bsonsize is 1.8KB. It means that we can fit around 550 movies in a single query result.
1. Most agents still have context windows of 128k to 256k tokens and 1MB of JSON is already a big hit on the agent context, which makes that much content pretty useless because of the loss of accuracy.

If users want more accurate results, they should instruct the agent to delegate some of the work to an aggregation pipeline as in any other application.

src/tools/mongodb/read/find.ts
himanshusinghs: I have addressed this in [955b7d8](https://github.com/mongodb-js/mongodb-mcp-server/pull/536/commits/955b7d84a4cf07e4447a9ee3fcfab2b1488b76f6)

src/helpers/iterateCursor.ts
himanshusinghs: Addressed both in [25e0367](https://github.com/mongodb-js/mongodb-mcp-server/pull/536/commits/25e0367480ac5e9eabcaebed0c920875625c01e4)

src/tools/mongodb/read/find.ts
himanshusinghs: brought back the smaller default of 10 on the tool interface and generous default of 100 on the configuration in [25e0367](https://github.com/mongodb-js/mongodb-mcp-server/pull/536/commits/25e0367480ac5e9eabcaebed0c920875625c01e4)

src/common/config.ts
himanshusinghs: I have brought back the default limit of 10 on the tool parameters and also exposed the responseBytesLimit on the tool parameters with our 1MB default, capped to 16MBs by the configuration (maxBytesPerQuery)

src/helpers/collectCursorUntilMaxBytes.ts
```diff
@@ -0,0 +1,103 @@
...
+export async function collectCursorUntilMaxBytesLimit<T = unknown>({
+    cursor,
+    toolResponseBytesLimit,
+    configuredMaxBytesPerQuery,
+    abortSignal,
+}: {
+    cursor: FindCursor<T> | AggregationCursor<T>;
+    toolResponseBytesLimit: number | undefined | null;
+    configuredMaxBytesPerQuery: unknown;
+    abortSignal?: AbortSignal;
```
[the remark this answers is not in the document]
himanshusinghs: `configuredMaxBytesPerQuery` would really be unknown because the config variables are not transformed when parsed.

PR #539, opened 2025-09-10
chore: fix release channel detection script

himanshusinghs:
## Proposed changes
The earlier script was not for bash, maybe for windows CMD / Powershell (based on the usage of Write-Output command) so I translated the script to bash in this PR.

## Checklist

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -47,6 +47,7 @@ jobs:
         id: npm-tag
         shell: bash
         run: |
+          set -e
           VERSION="${{ steps.get-version.outputs.VERSION }}"
 
           # Extract the release channel (latest, alpha, beta, rc)
@@ -61,7 +62,8 @@ jobs:
               RELEASE_CHANNEL="latest"
             fi
           else
-            RELEASE_CHANNEL="latest"
+            echo "::error title=Invalid Version::Encountered unexpected version ${{ steps.get-version.outputs.VERSION }}, cannot proceed!"
+            exit 1
           fi
 
           echo "RELEASE_CHANNEL=${RELEASE_CHANNEL}" >> "$GITHUB_OUTPUT"
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -50,9 +50,9 @@ jobs:
           VERSION="${{ steps.get-version.outputs.VERSION }}"
 
           # Extract the release channel (latest, alpha, beta, rc)
-          if [[ $VERSION =~ ^v?(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)(-(.+))?$ ]]; then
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yaml
```diff
@@ -45,15 +45,26 @@ jobs:
...
         run: |
-          $regex = "^v?(?<major>0|[1-9]\d*)\.(?<minor>0|[1-9]\d*)\.(?<patch>0|[1-9]\d*)(?:-(?<prerelease>(?:[^.]*)))?"
-          $version = "${{ steps.get-version.outputs.VERSION }}"
-          $releaseChannel = $groups["prerelease"][0].value
+          VERSION="${{ steps.get-version.outputs.VERSION }}"
+
+          # Extract the release channel (latest, alpha, beta, rc)
+          if [[ $VERSION =~ ^v?(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)(-(.+))?$ ]]; then
+            if [[ -n "${BASH_REMATCH[5]}" ]]; then
+              CAPTURED_CHANNEL="${BASH_REMATCH[5]}"
```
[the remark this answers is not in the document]
himanshusinghs: Addressed in [e9836fd](https://github.com/mongodb-js/mongodb-mcp-server/pull/539/commits/e9836fd7e6658ec8395a01f6ff154d99eabadfa5)

.github/workflows/publish.yaml
```diff
@@ -45,15 +45,26 @@ jobs:
...
+            if [[ -n "${BASH_REMATCH[2]}" ]]; then
+              CAPTURED_CHANNEL="${BASH_REMATCH[2]}"
+              # The captured channel might have more dots, cases like
+              # v1.2.3-alpha.1 For such cases we only want the channel relevant
+              # part which is alpha.
+              RELEASE_CHANNEL="${CAPTURED_CHANNEL%%.*}"
+            else
+              RELEASE_CHANNEL="latest"
+            fi
+          else
```
blva, line 64: if it doesn't match the regex, do we want to release? or should we fail?

blva (approved):
minor comment

.github/workflows/publish.yaml
himanshusinghs: Ahh that's a good point. With the code we have right now, we would be releasing to latest channel but I would say we should fail instead because very likely either the release name is wrong or the regex is wrong and for both we need to take action. Swallowing the error by releasing to latest might be the wrong idea.

If you agree, then I will update the script to exit with a non-zero exit code.

blva: sgtm! thanks

himanshusinghs: addressed [9ae5727](https://github.com/mongodb-js/mongodb-mcp-server/pull/539/commits/9ae572732b6bd553a96579810b0ba90a238a3ce6)

PR #542, opened 2025-09-10
fix: add argument validation - MCP-188

blva:
## Proposed changes

MCP-188

- adds validation to the atlas tool arguments and unit tests
- updates the argShape format to match other mongodb tools 

Future improvements
- improve args to reuse them across tools
- add integration test coverage with `validateThrowsForInvalidArguments`, a PR is in progress but separated from this one

### Test
Screenshot 2025-09-10 at 22 31 21

## Checklist

### DIFF
--- a/tests/unit/args.test.ts
+++ b/tests/unit/args.test.ts
@@ -152,19 +152,10 @@ describe("Tool args", () => {
                 expect(() => schema.parse(longName)).toThrow("Cluster name must be 64 characters or less");
 
                 // Invalid characters
-                expect(() => schema.parse("cluster@name")).toThrow(
-                    "Cluster name " + ALLOWED_CLUSTER_NAME_CHARACTERS_ERROR
-                );
-                expect(() => schema.parse("cluster name")).toThrow(
-                    "Cluster name " + ALLOWED_CLUSTER_NAME_CHARACTERS_ERROR
-                );
-                expect(() => schema.parse("cluster.name")).toThrow(
-                    "Cluster name " + ALLOWED_CLUSTER_NAME_CHARACTERS_ERROR
-                );
-
-                expect(() => schema.parse("cluster/name")).toThrow(
-                    "Cluster name " + ALLOWED_CLUSTER_NAME_CHARACTERS_ERROR
-                );
+                expect(() => schema.parse("cluster@name")).toThrow(ALLOWED_CLUSTER_NAME_CHARACTERS_ERROR);
+                expect(() => schema.parse("cluster name")).toThrow(ALLOWED_CLUSTER_NAME_CHARACTERS_ERROR);
+                expect(() => schema.parse("cluster.name")).toThrow(ALLOWED_CLUSTER_NAME_CHARACTERS_ERROR);
+                expect(() => schema.parse("cluster/name")).toThrow(ALLOWED_CLUSTER_NAME_CHARACTERS_ERROR);
             });
 
             it("should accept exactly 64 characters", () => {
@@ -195,8 +186,8 @@ describe("Tool args", () => {
                 expect(() => schema.parse(longUsername)).toThrow("Username must be 100 characters or less");
 
                 // Invalid characters
-                expect(() => schema.parse("user@name")).toThrow("Username " + ALLOWED_USERNAME_CHARACTERS_ERROR);
-                expect(() => schema.parse("user name")).toThrow("Username " + ALLOWED_USERNAME_CHARACTERS_ERROR);
+                expect(() => schema.parse("user@name")).toThrow(ALLOWED_USERNAME_CHARACTERS_ERROR);
+                expect(() => schema.parse("user name")).toThrow(ALLOWED_USERNAME_CHARACTERS_ERROR);
             });
 
             it("should accept exactly 100 characters", () => {
@@ -394,17 +385,13 @@ describe("Tool args", () => {
             expect(() => AtlasArgs.clusterName().parse("a".repeat(65))).toThrow(
                 "Cluster name must be 64 characters or less"
             );
-            expect(() => AtlasArgs.clusterName().parse("invalid@name")).toThrow(
-                "Cluster name " + ALLOWED_CLUSTER_NAME_CHARACTERS_ERROR
-            );
+            expect(() => AtlasArgs.clusterName().parse("invalid@name")).toThrow(ALLOWED_CLUSTER_NAME_CHARACTERS_ERROR);
 
             expect(() => AtlasArgs.username().parse("")).toThrow("Username is required");
             expect(() => AtlasArgs.username().parse("a".repeat(101))).toThrow(
                 "Username must be 100 characters or less"
             );
-            expect(() => AtlasArgs.username().parse("invalid name")).toThrow(
-                "Username " + ALLOWED_USERNAME_CHARACTERS_ERROR
-            );
+            expect(() => AtlasArgs.username().parse("invalid name")).toThrow(ALLOWED_USERNAME_CHARACTERS_ERROR);
         });
     });
 });


src/tools/atlas/create/createDBUser.ts
```diff
@@ -1,41 +1,45 @@
...
+    username: AtlasArgs.username().describe("Username for the new user"),
+    // Models will generate overly simplistic passwords like SecurePassword123 or
+    // AtlasPassword123, which are easily guessable and exploitable. We're instructing
+    // the model not to try and generate anything and instead leave the field unset.
+    password: AtlasArgs.password()
+        .optional()
+        .nullable()
+        .describe(
+            "Password for the new user. IMPORTANT: If the user hasn't supplied an explicit password, leave it unset and under no circumstances try to generate a random one. A secure password will be generated by the MCP server if necessary."
+        ),
```
himanshusinghs, line 21: Question: If we never want model to set it, can we just omit it? Also since user will provide password through LLM context, that's will very leak into LLM conversations.

src/tools/args.ts
```diff
@@ -0,0 +1,55 @@
...
+            .min(1, "Region is required")
+            .max(50, "Region must be 50 characters or less")
+            .regex(/^[a-zA-Z0-9_-]+$/, "Region can only contain letters, numbers, hyphens, and underscores"),
+
+    password: (): z.ZodString =>
+        CommonArgs.string()
+            .min(1, "Password is required")
+            .max(100, "Password must be 100 characters or less")
+            .regex(/^[^/]*$/, "String cannot contain '/'")
+            .regex(/^[a-zA-Z0-9._-]+$/, "Password can only contain letters, numbers, dots, hyphens, and underscores"),
```
nirinchev: Is this true? Seems like a weird restriction not to be able to use special characters for passwords.

src/tools/args.ts
```diff
@@ -0,0 +1,55 @@
...
+
+export const AtlasArgs = {
+    objectId: (fieldName: string): z.ZodString =>
+        CommonArgs.string()
+            .min(1, `${fieldName} is required`)
+            .regex(/^[0-9a-fA-F]{24}$/, `${fieldName} must be a valid 24-character hexadecimal string`),
+
+    projectId: (): z.ZodString => AtlasArgs.objectId("projectId"),
+
+    organizationId: (): z.ZodString => AtlasArgs.objectId("organizationId"),
```
himanshusinghs: Are these always object id for every other Atlas deployment types?

src/tools/atlas/create/createDBUser.ts
```diff
@@ -1,41 +1,45 @@
...
+        .array(
+            z.object({
+                roleName: CommonArgs.string().describe("Role name"),
+                databaseName: CommonArgs.string().describe("Database name").default("admin"),
+                collectionName: CommonArgs.string().describe("Collection name").optional(),
+            })
+        )
+        .describe("Roles for the new user"),
+    clusters: z
+        .array(CommonArgs.string())
```
nirinchev: Should this be AtlasArgs.clusterName?

src/tools/atlas/read/inspectCluster.ts
```diff
@@ -1,18 +1,21 @@
...
+import { type OperationType, type ToolArgs, formatUntrustedData } from "../../tool.js";
 import { AtlasToolBase } from "../atlasTool.js";
-import type { ToolArgs, OperationType } from "../../tool.js";
-import { formatUntrustedData } from "../../tool.js";
 import type { Cluster } from "../../../common/atlas/cluster.js";
 import { inspectCluster } from "../../../common/atlas/cluster.js";
+import { AtlasArgs } from "../../args.js";
+
+export const InspectClusterArgs = {
+    projectId: AtlasArgs.projectId().describe("Atlas project ID"),
```
nirinchev, line 9: The descriptions of projectId and clusterName seem to be the same everywhere - should we just include them in the AtlasArgs helpers?

src/tools/args.ts
```diff
@@ -0,0 +1,55 @@
+import { z, type ZodString } from "zod";
+
+export const CommonArgs = {
+    string: (): ZodString =>
+        z.string().regex(/^[\x20-\x7E]*$/, "String cannot contain special characters or Unicode symbols"),
```
himanshusinghs: When I was testing this, it was matching on non-unicode strings as well, we might need to tweak this a little.

```
/^[\x20-\x7E]*$/.test('MongoDB')
```

src/tools/args.ts
```diff
@@ -0,0 +1,55 @@
...
+
+    projectId: (): z.ZodString => AtlasArgs.objectId("projectId"),
+
+    organizationId: (): z.ZodString => AtlasArgs.objectId("organizationId"),
+
+    clusterName: (): z.ZodString =>
+        CommonArgs.string()
+            .min(1, "Cluster name is required")
+            .max(64, "Cluster name must be 64 characters or less")
+            .regex(/^[^/]*$/, "String cannot contain '/'")
```
himanshusinghs: This regex was also matching on string without a `/`, probably need to tweak this as well.

himanshusinghs: Actually I realised that since you have the allowed characters regex below, you probably don't need reject regex for `/`

src/tools/args.ts
himanshusinghs: I saw the tests and then I understood that its the allow range and the error message is for when the regex don't match. Please ignore.

src/tools/args.ts
himanshusinghs: Please disregard my initial comment about `/` but I still think that since you only allow certain characters below, `/` by itself gets rejected by the same regex.

src/tools/atlas/create/createDBUser.ts
nirinchev: The password will always leak into the LLM context, regardless of whether the user specifies it or we generate it ourselves. I think it's still valuable to offer the option for the user to explicitly state what password they want, the instruction is there to avoid having the model invent one if the user didn't ask for it.

src/tools/args.ts
```diff
@@ -0,0 +1,55 @@
...
+export const CommonArgs = {
+    string: (): ZodString =>
+        z.string().regex(/^[\x20-\x7E]*$/, "String cannot contain special characters or Unicode symbols"),
+};
+
+export const AtlasArgs = {
+    objectId: (fieldName: string): z.ZodString =>
+        CommonArgs.string()
+            .min(1, `${fieldName} is required`)
+            .regex(/^[0-9a-fA-F]{24}$/, `${fieldName} must be a valid 24-character hexadecimal string`),
```
nirinchev: Q: what does z.string().regex (...).regex(...) do? Does it replace the original regex or does it test against both? If it replaces it, we should perhaps use z.string() instead of CommonArgs.string() in places where we also specify the regex to avoid implying that the checks are additive.

src/tools/args.ts
blva: yes!

src/tools/atlas/read/inspectCluster.ts
blva: Unfortunately they differ with some extra tiny context per tool, I added this as a future improvement to keep the scope of the task

src/tools/args.ts
blva: I'm using the password description of the endpoint
`            /** @description Alphanumeric string that authenticates this database user against the database specified in `databaseName`. To authenticate with SCRAM-SHA, you must specify this parameter. This parameter doesn't appear in this response. */
`

src/tools/atlas/create/createDBUser.ts
blva: yes, thanks! updated

src/tools/atlas/create/createDBUser.ts
gagik: the elicitation API should help with this in the future: https://github.com/modelcontextprotocol/typescript-sdk#eliciting-user-input
We could elicit for a password input

src/tools/args.ts
nirinchev: I tried it out and it seems like using the Atlas UI, I can create a password with a bunch of special characters, it just adds a warning that the password will be url-encoded.

src/tools/args.ts
blva: good point, did some research and we can only have one regex so I'll need to do some updates, thanks for flagging!

src/tools/atlas/create/createDBUser.ts
blva: that's awesome, I'll create a JIRA to investigate it after GA

src/tools/args.ts
blva: cool I'll make it loose

src/tools/atlas/create/createDBUser.ts
blva: created a JIRA, resolving

src/tools/args.ts
blva: updated!

src/tools/args.ts
blva: I optimized the regexes into single ones, so this doesn't exist anymore

src/tools/args.ts
```diff
@@ -0,0 +1,68 @@
...
+const ALLLOWED_REGION_CHARACTERS_REGEX = /^[a-zA-Z0-9_-]+$/;
+export const ALLOWED_REGION_CHARACTERS_ERROR = "Region can only contain letters, numbers, hyphens, and underscores";
+
+const ALLOWED_CLUSTER_NAME_CHARACTERS_REGEX = /^[a-zA-Z0-9_-]+$/;
+export const ALLOWED_CLUSTER_NAME_CHARACTERS_ERROR =
+    "Cluster names can only contain ASCII letters, numbers, and hyphens.";
+
+const ALLOWED_PROJECT_NAME_CHARACTERS_REGEX = /^[a-zA-Z0-9\s()@&+:._',-]+$/;
+export const ALLOWED_PROJECT_NAME_CHARACTERS_ERROR =
+    "Project names can't be longer than 64 characters and can only contain letters, numbers, spaces, and the following symbols: ( ) @ & + : . _ - ' ,";
```
blva, line 19: went into atlas UI and copied the exact same validation text

src/tools/args.ts
blva: @nirinchev updated this a bit, would like to get another review! thx

src/tools/args.ts
```diff
@@ -0,0 +1,68 @@
+import { z, type ZodString } from "zod";
+
+const NO_UNICODE_REGEX = /^[\x20-\x7E]*$/;
+const NO_UNICODE_ERROR = "String cannot contain special characters or Unicode symbols";
```
nirinchev: This is not exported, which implies we don't have tests for it - should we add some?

src/tools/args.ts
```diff
@@ -0,0 +1,68 @@
...
+
+const ALLOWED_NAME_CHARACTERS_REGEX = /^[a-zA-Z0-9._-]+$/;
+export const ALLOWED_CHARACTERS_ERROR = " can only contain letters, numbers, dots, hyphens, and underscores";
```
nirinchev: [nit] This has a leading space - is this intentional?

nirinchev:
```suggestion
const ALLOWED_NAME_CHARACTERS_REGEX = /^[a-zA-Z0-9._-]+$/;
export const ALLOWED_NAME_CHARACTERS_ERROR = " can only contain letters, numbers, dots, hyphens, and underscores";
```

src/tools/args.ts
```diff
@@ -0,0 +1,68 @@
...
+    "Project names can't be longer than 64 characters and can only contain letters, numbers, spaces, and the following symbols: ( ) @ & + : . _ - ' ,";
+export const CommonArgs = {
+    string: (): ZodString => z.string().regex(NO_UNICODE_REGEX, NO_UNICODE_ERROR),
+};
+
+export const AtlasArgs = {
+    objectId: (fieldName: string): z.ZodString =>
+        z
+            .string()
+            .min(1, `${fieldName} is required`)
```
nirinchev, line 26: Should this be `.length(24)`?

src/tools/args.ts
```diff
@@ -0,0 +1,68 @@
...
+
+const ALLOWED_PROJECT_NAME_CHARACTERS_REGEX = /^[a-zA-Z0-9\s()@&+:._',-]+$/;
+export const ALLOWED_PROJECT_NAME_CHARACTERS_ERROR =
+    "Project names can't be longer than 64 characters and can only contain letters, numbers, spaces, and the following symbols: ( ) @ & + : . _ - ' ,";
+export const CommonArgs = {
+    string: (): ZodString => z.string().regex(NO_UNICODE_REGEX, NO_UNICODE_ERROR),
+};
+
+export const AtlasArgs = {
+    objectId: (fieldName: string): z.ZodString =>
```
nirinchev, line 23: I wonder if `objectId` should be in CommonArgs - it doesn't really fit with the rest of the atlas args here.

src/tools/args.ts
```diff
@@ -0,0 +1,68 @@
...
+    projectId: (): z.ZodString => AtlasArgs.objectId("projectId"),
+
+    organizationId: (): z.ZodString => AtlasArgs.objectId("organizationId"),
+
+    clusterName: (): z.ZodString =>
+        z
+            .string()
+            .min(1, "Cluster name is required")
+            .max(64, "Cluster name must be 64 characters or less")
+            .regex(ALLOWED_CLUSTER_NAME_CHARACTERS_REGEX, "Cluster name " + ALLOWED_CLUSTER_NAME_CHARACTERS_ERROR),
```
nirinchev: I think this is a bit of a mistake - when you combine this strings, it will look like `Cluster name Cluster names can only contain ASCII letters, numbers, and hyphens.`. Should we remove the prefix here?

src/tools/args.ts
```diff
@@ -0,0 +1,68 @@
+import { z, type ZodString } from "zod";
+
+const NO_UNICODE_REGEX = /^[\x20-\x7E]*$/;
+const NO_UNICODE_ERROR = "String cannot contain special characters or Unicode symbols";
+
+const ALLOWED_NAME_CHARACTERS_REGEX = /^[a-zA-Z0-9._-]+$/;
```
nirinchev: Should we rename this to `ALLOWED_USERNAME_CHARACTERS_REGEX`? It seems to only be used in the username arg.

src/tools/args.ts
blva: we have tests we're just not validating the error text, will update!

src/tools/args.ts
blva: thanks for catching this, I started with trying to have a common error and now they're more individual, will fix

src/tools/args.ts
blva: fixed!

PR #544, opened 2025-09-11
feat: Allow configuration of temporary user timeout when connecting to atlas cluster

jeroenvervaeke:
## Proposed changes
Allow configuration of temporary user timeout when connecting to atlas cluster.
- config name `temporaryDatabaseUserLifetimeSeconds`
- flag name `--temporaryDatabaseUserLifetimeSeconds`
- env variable `MDB_MCP_TEMPORARY_DATABASE_USER_LIFETIME_SECONDS`

Ticket: [\[MCP-199\] \[Security\] allow users to configure expiration timeout and limit the default 12 hours to 4 hours.](https://jira.mongodb.org/browse/MCP-199)

## Checklist

### DIFF
--- a/src/common/config.ts
+++ b/src/common/config.ts
@@ -8,7 +8,7 @@ import type { Secret } from "./keychain.js";
 import levenshtein from "ts-levenshtein";
 
 // From: https://github.com/mongodb-js/mongosh/blob/main/packages/cli-repl/src/arg-parser.ts
-const MONGOSH_OPTIONS = {
+const OPTIONS = {
     string: [
         "apiBaseUrl",
         "apiClientId",
@@ -48,6 +48,7 @@ const MONGOSH_OPTIONS = {
         "tlsCertificateSelector",
         "tlsDisabledProtocols",
         "username",
+        "atlasTemporaryDatabaseUserLifetimeMs",
     ],
     boolean: [
         "apiDeprecationErrors",
@@ -92,10 +93,6 @@ const MONGOSH_OPTIONS = {
     },
 } as Readonly<Options>;
 
-const MCP_SERVER_OPTIONS = {
-    string: ["atlasTemporaryDatabaseUserLifetimeMs"],
-} as Readonly<Partial<Options>>;
-
 interface Options {
     string: string[];
     boolean: string[];
@@ -103,21 +100,6 @@ interface Options {
     alias: Record<string, string>;
     configuration: Record<string, boolean>;
 }
-
-function mergeOptions(...optionSources: Array<Partial<Options>>): Readonly<Options> {
-    return {
-        string: optionSources.flatMap((opts) => opts.string ?? []),
-        boolean: optionSources.flatMap((opts) => opts.boolean ?? []),
-        array: optionSources.flatMap((opts) => opts.array ?? []),
-        alias: Object.assign({}, ...optionSources.map((opts) => opts.alias ?? {})) as Record<string, string>,
-        configuration: Object.assign({}, ...optionSources.map((opts) => opts.configuration ?? {})) as Record<
-            string,
-            boolean
-        >,
-    };
-}
-
-const OPTIONS = mergeOptions(MONGOSH_OPTIONS, MCP_SERVER_OPTIONS);
 
 const ALL_CONFIG_KEYS = new Set(
     (OPTIONS.string as readonly string[])


src/common/config.ts
```diff
@@ -161,6 +162,7 @@ export interface UserConfig extends CliOptions {
     loggers: Array<"stderr" | "disk" | "mcp">;
     idleTimeoutMs: number;
     notificationTimeoutMs: number;
+    temporaryDatabaseUserLifetimeSeconds: number;
```
gagik: all these settings are blowing up quickly, I'm wondering if we should introduce an `atlas` object.

gagik: alternatively, it's possible for there to exist users in non-Atlas context so maybe we should add "Atlas" to this naming

src/common/config.ts
```diff
@@ -48,6 +48,7 @@ const OPTIONS = {
         "tlsCertificateSelector",
         "tlsDisabledProtocols",
         "username",
+        "temporaryDatabaseUserLifetimeSeconds",
```
gagik: this array is duplicated from mongosh and shouldn't be modified.
In the future, we should definetely export it from mongosh and import it here because this is doomed to be confusing.
Maybe let's move this to a separate file now with a comment?

jeroenvervaeke: Could you elaborate on this comment?

src/common/config.ts
jeroenvervaeke: That sounds like a good idea for a follow-up ticket.

src/common/config.ts
himanshusinghs: We don't need to (and shouldn't) modify the `OPTIONS` type as that configuration is for part of code from mongosh which we depend on for connection related work. For general config changes the user `UserConfig` modification should be enough.

src/common/config.ts
himanshusinghs: Since you eventually convert this value to milliseconds I would suggest accepting milliseconds in the config itself. That aligns with the rest of the config parameters and doesn't make user think twice before putting a value in here.

src/common/config.ts
jeroenvervaeke: I only added `temporaryDatabaseUserLifetimeSeconds` to the list because it didn't work without adding it.

src/common/config.ts
jeroenvervaeke: Updated

jeroenvervaeke: prefixed the property with `atlas`

gagik: > That sounds like a good idea for a follow-up ticket.

~This will require us to keep backwards compatibility so it's better to make a decision about it now than later.~
Actually this is going to quickly get complicated with our env etc settings so let's keep that for post-GA with a larger refactor for all settings

src/common/config.ts
gagik: that's odd, it shouldn't depend on this.
This object should be only about mongosh-related args

himanshusinghs: That's weird. I will check it out.

gagik: This `OPTIONS` object only exists as a hack because mongosh's isn't exporting it. We'd generally want to import this externally.

himanshusinghs: Ahh you're right. The OPTIONS object have been extended with our config keys as well. Can't say if that was intentional, very likely to support the args validation we do on server start, but certainly something we need to split up. I have created MCP-202 to track this, meanwhile feel free to keep it like it is.
@gagik do you have any differing thoughts?

gagik: We should definitely deal with this now. Just create a second object which extends this. It should take a minute.

jeroenvervaeke: Can this PR have a " 👍 ", or do you require any more changes on my side?

gagik: > We should definitely deal with this now. Just create a second object which extends this. It should take a minute.

We need to split the object between what is copied of over from mongosh and what is ours.
Something like
```ts
const MONGOSH_OPTIONS = [
// copy-paste from that mongosh file
]
const MCP_SERVER_OPTIONS = [
// ...
]
const OPTIONS = [...MONGOSH_OPTIONS, ...MCP_SERVER_OPTIONS]
```
(this would need to be done per field as otherwise they'd override each other)

jeroenvervaeke: I've split the options like you suggested.

jeroenvervaeke: Fixed the shallow merge issue.

src/common/config.ts
```diff
@@ -180,6 +195,7 @@ export const defaultUserConfig: UserConfig = {
     idleTimeoutMs: 600000, // 10 minutes
     notificationTimeoutMs: 540000, // 9 minutes
     httpHeaders: {},
+    atlasTemporaryDatabaseUserLifetimeMs: 14400000, // 4 hours
```
nirinchev: When representing time, I find it easier to reason about the numbers if they're expressed as multiplication of units of time - i.e. `1000 * 60 * 60 * 4` rather than the result - it's much easier to spot mistakes than making the calculation in your head.

jeroenvervaeke: Thanks for the feedback! I chose to use the plain number format here to maintain consistency with the other time values already defined in this struct, which all follow the same pattern.

jeroenvervaeke:
> The new option should probably be documented in the readme. Also, we are now entering a territory where it's much more likely that an automatically created user expires while we're still connected to the cluster. We should consider how to handle such a case - we can either recreate the user and update the connection or return an error message instructing the model to re-run the atlas-connect-cluster tool. I would suggest that you test it manually to verify what the real-world behavior is and then decide whether gracefully handling it is critical for GA.

@nirinchev 
Updated the readme

src/common/config.ts
```diff
@@ -92,6 +92,18 @@ const OPTIONS = {
     },
 } as const;
 
+const MCP_SERVER_OPTIONS = {
+    string: ["atlasTemporaryDatabaseUserLifetimeMs"],
+} as const;
+
+const OPTIONS = {
```
gagik: this is fine. alternative idea for consideration but non-blocking:
```ts
function mergeOptions(...optionSources: Array<Record<string, any>>) {
    return {
        string: optionSources.flatMap(opts => opts.string ?? []),
        boolean: optionSources.flatMap(opts => opts.boolean ?? []),
        array: optionSources.flatMap(opts => opts.array ?? []),
        alias: Object.assign({}, ...optionSources.map(opts => opts.alias ?? {})),
        configuration: Object.assign({}, ...optionSources.map(opts => opts.configuration ?? {})),
    } as const;
}

const OPTIONS = mergeOptions(MONGOSH_OPTIONS, MCP_SERVER_OPTIONS);
```

src/common/config.ts
gagik: I wouldn't mind changing the others to match as a drive-by as well

jeroenvervaeke: I've update the numbers

src/common/config.ts
jeroenvervaeke: I have accepted your suggestion.

I had to introduce a new `Options` interface to make the linter happy.

src/common/config.ts
```diff
@@ -8,7 +8,7 @@ import type { Secret } from "./keychain.js";
 import levenshtein from "ts-levenshtein";
 
 // From: https://github.com/mongodb-js/mongosh/blob/main/packages/cli-repl/src/arg-parser.ts
-const OPTIONS = {
+const MONGOSH_OPTIONS = {
```
himanshusinghs: So this is a problem. The options that you see here are not all MONGOSH options. They contain MCP related options as well.

I don't have strong opinion on fixing this right now but consider that this is only a partial solution. We still need to address it entirely, perhaps with https://jira.mongodb.org/browse/MCP-202

nirinchev:
Folks, let's hold off on the refactorings of the cli options that close to the release. I agree it's not ideal that we're intertwining mongosh and mcp arguments together, but this has been the case for a while now and besides being a code smell, hasn't really caused issues. Sorry for the back-and-forth here @jeroenvervaeke, but let's just revert all the options refactorings and go back to the state in the first commit where we just add the new option to the original `OPTIONS` object.

jeroenvervaeke:
> Folks, let's hold off on the refactorings of the cli options that close to the release. I agree it's not ideal that we're intertwining mongosh and mcp arguments together, but this has been the case for a while now and besides being a code smell, hasn't really caused issues. Sorry for the back-and-forth here @jeroenvervaeke, but let's just revert all the options refactorings and go back to the state in the first commit where we just add the new option to the original `OPTIONS` object.

Reverted those requested changes

PR #546, opened 2025-09-11
feat(atlas-local): Add Atlas Local Create Deployment tool

Luke-Sanderson:
## Proposed changes
Ticket: [\[MCP-156\] \[mongodb-mcp-server\] Implement createLocalAtlasCluster tool](https://jira.mongodb.org/browse/MCP-156)
- Add createLocalAtlasCluster tool
- Add integration tests for this new tool

## Checklist

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -75,7 +75,7 @@
         "node": "^20.19.0 || ^22.12.0 || >= 23.0.0"
       },
       "optionalDependencies": {
-        "@mongodb-js-preview/atlas-local": "^0.0.0-preview.2",
+        "@mongodb-js-preview/atlas-local": "^0.0.0-preview.3",
         "kerberos": "^2.2.2"
       }
     },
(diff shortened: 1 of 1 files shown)


Luke-Sanderson:
Test fails due to Windows gh runner. Will be addressed in [[MCP-187] 
Split Atlas Local tests from Atlas Tests ](https://jira.mongodb.org/browse/MCP-187)

src/tools/atlasLocal/create/createDeployment.ts
```diff
@@ -0,0 +1,38 @@
...
+    };
+
+    protected async executeWithAtlasLocalClient(
+        client: Client,
+        { deploymentName }: ToolArgs<typeof this.argsShape>
+    ): Promise<CallToolResult> {
+        const deploymentOptions: CreateDeploymentOptions = {
+            name: deploymentName,
+            creationSource: {
+                type: "MCPServer" as CreationSourceType,
```
jeroenvervaeke, line 22: `MCPServer` is a valid creation source. Why do we have to type `as CreationSourceType` here? Can't we drop this?

[[source]()](https://github.com/mongodb-js/atlas-local-lib-js/blob/5587387aeee9542909eb418b44a54bb42dac0fd0/index.d.ts#L44)

src/tools/atlasLocal/create/createDeployment.ts
```diff
@@ -0,0 +1,38 @@
...
+                source: "MCPServer",
+            },
+        };
+        // Create the deployment
+        await client.createDeployment(deploymentOptions);
+
+        if (deploymentName) {
```
jeroenvervaeke: Doesn't `await client.createDeployment(deploymentOptions);` return the name of the deployment?

tests/integration/tools/atlas-local/createDeployment.test.ts
```diff
@@ -0,0 +1,170 @@
...
+import { afterEach, describe, expect, it } from "vitest";
+
+const isMacOSInGitHubActions = process.platform === "darwin" && process.env.GITHUB_ACTIONS === "true";
+
+// Docker is not available on macOS in GitHub Actions
+// That's why we skip the tests on macOS in GitHub Actions
+describe("atlas-local-create-deployment", () => {
+    let deploymentNamesToCleanup: string[] = [];
+
+    afterEach(async () => {
```
jeroenvervaeke, line 18: We don't need to run this on `isMacOSInGitHubActions`, can we skip it here too?

src/tools/atlasLocal/create/createDeployment.ts
Luke-Sanderson: I got errors when trying to assign it without `as CreationSourceType`. I think the explicit cast is needed

`Type '"MCPServer"' is not assignable to type 'CreationSourceType'.ts(2322)`

tests/integration/tools/atlas-local/createDeployment.test.ts
Luke-Sanderson: I don't see a way to do the skipIf(isMacOSInGitHubActions) or similar for an afterEach. This leaves adding something like this in the afterEach function:
```
if (isMacOSInGitHubActions) {
    return;
}
````
 But since the deploymentToCleanup array will always be empty it will be the same execution time

src/tools/atlasLocal/create/createDeployment.ts
nirinchev: Does the error go away if you use `as const` instead?

tests/integration/tools/atlas-local/createDeployment.test.ts
jeroenvervaeke: Makes sense to not skip it then!

src/tools/atlasLocal/create/createDeployment.ts
Luke-Sanderson: Assuming you mean `type: "MCPServer" as const,` then no its the same error. 

I am new to typescript so may have missed something. Out of interest, what are the downsides to the `as CreationSourceType` approach or is it just because it was expected that just `"MCPServer"` without a cast should be enough?

src/tools/atlasLocal/create/createDeployment.ts
Luke-Sanderson: Changed to return deployment and extracted the name

PR #548, opened 2025-09-11
fix: add untrusted data wrapper around logs response MCP-197

nirinchev:
## Proposed changes
Part of MCP-197 - this adds the untrusted data wrapper around the logs response and updates the tests to reflect the new format.


nirinchev:
Haven't started on the export yet, will check if the LLM gets confused by the wrapper.

PR #551, opened 2025-09-12
feat(elicitation): add user consent through elicitation MCP-185

gagik:
Adds an option to require confirmation for certain tools using the new elicitation API. This is not supported by most client yet, only notably VSCode.

Clients which support it will see a confirmation option with a summary before the action is run. If the client doesn't support elicitation, the action will simply be auto-approved.

This option can be confirmed with `confirmationRequiredTools` and has a default set of `drop-database`, `drop-collection`, `delete-many`, `atlas-create-db-user`, `atlas-create-access-list` enabled. These tools also have custom messages for confirmation. If a user adds other tools they'll be met with a more generic tool confirmation message.

In VSCode one must first click "Respond" (which sets action to "accepted") and then choose a value. I decided to let there be an explcit choice of Yes / No in JSON schema instead of opting to just rely on "Respond" as it is not immediately clear that `Respond = Yes` and I imagine this vagueness in the API spec will lead to confusion across clients so it's best to have an explicit JSON schema value for confirmation.

I also went with enum string Yes / No and not boolean since the displayed value for this is more user friendly.

#### Step 1: Click "Respond"
Screenshot 2025-09-12 at 10 38 50 AM

#### Step 2: Choose "Yes"
Screenshot 2025-09-12 at 10 39 48 AM

### In Cursor (i.e. unsupported client)
The tool gets called without any confirmation.
Screenshot 2025-09-12 at 10 38 13 AM

### DIFF
--- a/src/elicitation.ts
+++ b/src/elicitation.ts
@@ -1,4 +1,4 @@
-import type { PrimitiveSchemaDefinition } from "@modelcontextprotocol/sdk/types.js";
+import type { ElicitRequest } from "@modelcontextprotocol/sdk/types.js";
 import type { McpServer } from "@modelcontextprotocol/sdk/server/mcp.js";
 
 export class Elicitation {
@@ -37,7 +37,7 @@ export class Elicitation {
      * The schema for the confirmation question.
      * TODO: In the future would be good to use Zod 4's toJSONSchema() to generate the schema.
      */
-    public static CONFIRMATION_SCHEMA: MCPElicitationSchema = {
+    public static CONFIRMATION_SCHEMA: ElicitRequest["params"]["requestedSchema"] = {
         type: "object",
         properties: {
             confirmation: {
@@ -51,9 +51,3 @@ export class Elicitation {
         required: ["confirmation"],
     };
 }
-
-export type MCPElicitationSchema = {
-    type: "object";
-    properties: Record<string, PrimitiveSchemaDefinition>;
-    required?: string[];
-};
--- a/src/tools/mongodb/delete/deleteMany.ts
+++ b/src/tools/mongodb/delete/deleteMany.ts
@@ -59,12 +59,12 @@ export class DeleteManyTool extends MongoDBToolBase {
 
     protected getConfirmationMessage({ database, collection, filter }: ToolArgs<typeof this.argsShape>): string {
         const filterDescription =
-            filter && Object.keys(filter).length > 0 ? EJSON.stringify(filter) : "All documents (No filter)";
+            filter && Object.keys(filter).length > 0
+                ? "```json\n" + `{ "filter": ${EJSON.stringify(filter)} }\n` + "```\n\n"
+                : "- **All documents** (No filter)\n\n";
         return (
             `You are about to delete documents from the \`${collection}\` collection in the \`${database}\` database:\n\n` +
-            "```json\n" +
-            `{ "filter": ${filterDescription} }\n` +
-            "```\n\n" +
+            filterDescription +
             "This operation will permanently remove all documents matching the filter.\n\n" +
             "**Do you confirm the execution of the action?**"
         );
--- a/tests/integration/elicitation.test.ts
+++ b/tests/integration/elicitation.test.ts
@@ -25,7 +25,7 @@ describe("Elicitation Integration Tests", () => {
             { elicitInput: mockElicitInput }
         );
 
-        describe("tools requiring confirmation", () => {
+        describe("tools requiring confirmation by default", () => {
             it("should request confirmation for drop-database tool and proceed when confirmed", async () => {
                 mockElicitInput.confirmYes();
 
@@ -37,19 +37,7 @@ describe("Elicitation Integration Tests", () => {
                 expect(mockElicitInput.mock).toHaveBeenCalledTimes(1);
                 expect(mockElicitInput.mock).toHaveBeenCalledWith({
                     message: expect.stringContaining("You are about to drop the `test-db` database"),
-                    requestedSchema: {
-                        type: "object",
-                        properties: {
-                            confirmation: {
-                                type: "string",
-                                title: "Would you like to confirm?",
-                                description: "Would you like to confirm?",
-                                enum: ["Yes", "No"],
-                                enumNames: ["Yes, I confirm", "No, I do not confirm"],
-                            },
-                        },
-                        required: ["confirmation"],
-                    },
+                    requestedSchema: Elicitation.CONFIRMATION_SCHEMA,
                 });
 
                 // Should attempt to execute (will fail due to no connection, but confirms flow worked)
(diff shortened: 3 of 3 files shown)


src/elicitation.ts
```diff
@@ -0,0 +1,59 @@
...
+        this.server = server;
+    }
+
+    /**
+     * Checks if the client supports elicitation capabilities.
+     * @returns True if the client supports elicitation, false otherwise.
+     */
+    public supportsElicitation(): boolean {
+        const clientCapabilities = this.server.getClientCapabilities();
+        return clientCapabilities?.elicitation !== undefined;
```
gagik, line 16: If one wants to be paranoid... there's might be a chance some client will claim it supports `elicitation` while not supporting it. Then this would likely make this tool a confusing no-op.
That said this would be client's fault and one could unblock themselves by setting confirmation required tools to nothing.

src/tools/mongodb/delete/deleteMany.ts
```diff
@@ -55,4 +56,17 @@ export class DeleteManyTool extends MongoDBToolBase {
             ],
         };
     }
+
+    protected getConfirmationMessage({ database, collection, filter }: ToolArgs<typeof this.argsShape>): string {
```
nirinchev, line 60: q1: Do we want to get confirmation for deletions even when we have a filter specified? I thought we only want to do it if you try and delete everything in the collection.
q2: Do we want to give the approximate count of documents that will be deleted in the message?

src/tools/tool.ts
```diff
@@ -237,7 +284,7 @@ export abstract class ToolBase {
             event.properties.project_id = metadata.projectId;
         }
 
-        this.telemetry.emitEvents([event]);
+        void this.telemetry.emitEvents([event]);
```
nirinchev: This should already be void - I'm guessing you need to pull main.

src/transports/base.ts
```diff
@@ -94,12 +95,15 @@ export abstract class TransportRunnerBase {
             commonProperties: this.telemetryProperties,
         });
 
+        const elicitation = new Elicitation({ server: mcpServer.server });
```
nirinchev, line 98: q: should we just construct this in the `Server` ctor instead?

src/elicitation.ts
nirinchev, line 16: Should this be `=== true` instead? If elicitation is `false`, we shouldn't use it, should we?

src/transports/base.ts
gagik: I don't mind either way but we don't really have any precedent for that, in a sense same would apply to the rest of the constructed instances here

src/elicitation.ts
gagik: elicitation is of type [object](https://modelcontextprotocol.io/specification/2025-06-18/schema#clientcapabilities-elicitation) (annoying vague, pretty sure it'd be `{}` but I assume they leave it open for future configuration) so we're not looking at a `true`/`false` here

src/tools/tool.ts
gagik: It's implicitly `void`, I don't know why our lint isn't catching it but my IDE was erroring at this.

src/tools/mongodb/delete/deleteMany.ts
gagik: > q1: Do we want to get confirmation for deletions even when we have a filter specified? I thought we only want to do it if you try and delete everything in the collection.

I don't remember the discussions around this but maybe we still should? I think with deletions it's still good to review the filter and whatnot.

>  q2: Do we want to give the approximate count of documents that will be deleted in the message?

This does seem useful but could this be an expensive operation for extremely large collections? Seems like a potentially problematic side effect.

tests/integration/elicitation.test.ts
```diff
@@ -0,0 +1,433 @@
...
+import { Telemetry } from "../../src/telemetry/telemetry.js";
+import { Keychain } from "../../src/common/keychain.js";
+import { InMemoryTransport } from "./inMemoryTransport.js";
+import { connectionErrorHandler } from "../../src/common/connectionErrorHandler.js";
+import { defaultDriverOptions, type UserConfig } from "../../src/common/config.js";
+import { defaultTestConfig } from "./helpers.js";
+import { Elicitation } from "../../src/elicitation.js";
+import { type MockClientCapabilities, createMockElicitInput } from "../utils/elicitationMocks.js";
+
+describe("Elicitation Integration Tests", () => {
```
gagik, line 8: I will change this file to properly use our integration suite... I originally moved from unit tests

src/tools/mongodb/delete/deleteMany.ts
nirinchev: I would not expect the MCP server to be used for perf critical applications, so am more inclined to think it's a worthwhile tradeoff, but don't feel strongly either way.

src/tools/tool.ts
nirinchev: Not sure why it'd error though - could it be that it hadn't seen that emitEvents is no longer a promise?

src/elicitation.ts
nirinchev: Ah... that makes sense.

src/tools/tool.ts
gagik: ah you're right, I think I had some TypeScript cache or something

src/transports/base.ts
nirinchev: I think we're typically doing that to allow injecting mocked dependencies during tests, but wasn't sure I saw that being taken advantage of here (though I skipped the tests).

gagik: I feel like having it outside is a good principle in general as it gets one to think about it if they are ever creating a new instance of the server in tests or otherwise.

src/elicitation.ts
```diff
@@ -0,0 +1,59 @@
...
+            requestedSchema: Elicitation.CONFIRMATION_SCHEMA,
+        });
+        return result.action === "accept" && result.content?.confirmation === "Yes";
+    }
+
+    /**
+     * The schema for the confirmation question.
+     * TODO: In the future would be good to use Zod 4's toJSONSchema() to generate the schema.
+     */
+    public static CONFIRMATION_SCHEMA: MCPElicitationSchema = {
```
nirinchev:
```suggestion
    public static CONFIRMATION_SCHEMA: ElicitRequest["params"]["requestedSchema"] = {
```

And then we can drop MCPElicitationSchema.

tests/integration/elicitation.test.ts
```diff
@@ -0,0 +1,321 @@
...
+
+        it("should respect custom confirmationRequiredTools configuration", async () => {
+            mockElicitInput.confirmYes();
+
+            await integration.mcpClient().callTool({
+                name: "list-databases",
+                arguments: {},
+            });
+
+            expect(mockElicitInput.mock).toHaveBeenCalledTimes(1);
```
nirinchev, line 216: Should we check that the message is the default one?

src/tools/mongodb/delete/deleteMany.ts
gagik: I do think running other DB functions for the sake of a confirmation message is a bit unexpected but probably not critically terrible yeah.
I'm going to leave it as is mainly because it'd require some changes to make the get confirmation async (which is fine and doomed to happen), so we can revisit later

PR #562, opened 2025-09-16
fix: pin yargs-parser to 21 for VSCode compatibility

himanshusinghs:
## Proposed changes
VSCode project works(not directly) with yargs-parser@21 and our project pulls in version 22. The dependency resolution in VSCode resolves yargs-parser to version 21 and thus the usage of this dependency in the imported mcp-server-code breaks.

Ideally, we wouldn't have to do this because the imported code in VSCode, theoretically, in no way relies on yargs-parser. But because of our code setup, the yargs-parser does come in the imported path eventually:
1. exporting types and vaues from config.js in lib entry point which imports yargs-parser. Ideally we should re-organise this to different files, one with schema and one with config helpers.
2. imports from config.js in our trasport/base.js which sets up default connectionManager imported from connectionManager.js which further imports config.js which imports yargs-parser.
3. imports from config.js in our config resource which imports defaultDriverOptions.

To fix this long term there are two options:
1. Re-organise files and imports such that un-necessary code does not end up in lib entry point. This will compromise a bit on DX but is more preferable, we do the same in Compass, btw. Additionally we will have to setup VSCode integration tests.
2. Use bundler to bundle the lib entry point with all the dependencies it might accidentally end up importing. Less preferred because the lib entry point might just grow without us noticing but this will work flawlessly.

## Checklist

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -29,7 +29,7 @@
         "oauth4webapi": "^3.8.0",
         "openapi-fetch": "^0.14.0",
         "ts-levenshtein": "^1.0.7",
-        "yargs-parser": "^22.0.0",
+        "yargs-parser": "^21.1.1",
         "zod": "^3.25.76"
       },
       "bin": {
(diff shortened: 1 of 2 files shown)


nirinchev (approved):

cveticm (approved):
LGTM, could you open a ticket for tracking the investigation work for long-term solutions too?

himanshusinghs:
Done - here we have it https://jira.mongodb.org/browse/MCP-214

PR #564, opened 2025-09-16
fix: Removes nullable from DBUser password MCP-212

### LINKED ISSUE #556: [Bug]: 'atlas-create-db-user' from MCP server has missing types in its parameter schema
### Version

Not sure about the version. It is being used with this config:

"command": "npx",
"args": [
	"-y",
	"mongodb-mcp-server",
	"--apiClientId",
	"xxxxxxxxxxxxxxxxxxxxxxxxxxxx",
	"--apiClientSecret",
	"xxxxxxxxxxxxxxxxxxxxxxxxxxxx"
],
"type": "stdio"

### App

- [ ] Cursor
- [ ] Windsurf
- [x] VSCode
- [ ] VSCode Insiders
- [ ] Claude Desktop
- [ ] Other

- [ ] Claude 3.5 Sonnet
- [ ] Claude 3.7 Sonnet
- [ ] GPT-4a
- [ ] o4-mini
- [x] Other

### Bug Description

When using the MCP server via Gemini CLI, I receive the following message:

Skipping tool 'atlas-create-db-user' from MCP server 'MongoDB' because it has missing types in its parameter schema. Please file an issue with the owner of the MCP server

It is not a blocking issue for the rest of the tools, but can't create db users using the  MCP. 

Image

cveticm:
## Proposed changes

Fixes #556 

On Gemini CLI, tool `atlas-create-dbusers` would be skipped due to `missing types in paramter schema`. 
This seems to be caused by setting password as both optional and nullable. 
I have replaced this with [nullish (both optional and nullable)](https://zod.dev/api?id=nullish). 
This resolves the issue.

before fix:
image

after fix:
image

## Checklist


himanshusinghs (approved):
Good catch 🚀 @nirinchev should we pull this for 1.0.0?

nirinchev:
Can we test the impact of this change with several agents - notably, we want to ensure the agent doesn't generate a password unless explicitly prompted for it. Otherwise, happy to pull it in.

cveticm:
> Can we test the impact of this change with several agents - notably, we want to ensure the agent doesn't generate a password unless explicitly prompted for it. Otherwise, happy to pull it in.

@nirinchev Tested with copilot agents GPT, grok, Claude and gemini. All exclude password or provide null value if not explicitly prompted to include password. Additionally, user provided password is used when prompted. ie, behaviour remains the same.
Let me know if there is additional agents you'd like me to test with

nirinchev (approved):

gagik (approved):

gagik:
> Tested with copilot agents GPT, grok, Claude and gemini

If this was only in VSCode/Copilot might also be worth verifying in a different _agent_ (i.e. Claude Desktop or Cursor) and not model just to make sure they interpret JSON schema correctly as well

PR #567, opened 2025-09-16
chore: add missing config properties for verification

himanshusinghs:
## Proposed changes

The idea is to have all (excpect object properties because we cannot parse them yey) the MCP config in the OPTIONS object so that the verification method can look up the config being provided and if its valid or not.

## Checklist

### DIFF
--- a/src/common/config.ts
+++ b/src/common/config.ts
@@ -94,14 +94,12 @@ const OPTIONS = {
         "greedy-arrays": true,
         "short-option-groups": false,
     },
-    object: ["httpHeaders"],
 } as Readonly<Options>;
 
 interface Options {
     string: string[];
     boolean: string[];
     array: string[];
-    object: string[];
     alias: Record<string, string>;
     configuration: Record<string, boolean>;
 }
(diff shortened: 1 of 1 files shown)


src/common/config.ts
```diff
@@ -91,19 +94,22 @@ const OPTIONS = {
...
         "short-option-groups": false,
     },
+    object: ["httpHeaders"],
 } as Readonly<Options>;
 
 interface Options {
     string: string[];
     boolean: string[];
     array: string[];
+    object: string[];
```
nirinchev: This doesn't seem right. Also, it doesn't seem like either yargs or our env variable parsing has the ability to construct an object for the headers, so not sure how this is expected to work.

himanshusinghs: Ahh that's true. I tried parsing a record like string and it did not work. I will remove this but that means httpHeaders for now are not parseable by our config. I will create a ticket for this, might become more evident when we do remote server

himanshusinghs: removed it

nirinchev (approved):

kmruiz (approved):

Issue #549, opened 2025-09-11
[Bug]: Could not filter collection by date


wael7721:
### Version

Version 0.0.3

### App

- [ ] Cursor
- [ ] Windsurf
- [x] VSCode
- [ ] VSCode Insiders
- [ ] Claude Desktop
- [x] Other

### Affected Models (if applicable)

- [x] Claude 3.5 Sonnet
- [x] Claude 3.7 Sonnet
- [ ] GPT-4a
- [ ] o4-mini
- [x] Other

### Bug Description

The MCP Failed to filter by date (Found 0 documents in the collection `hidden`) despite queries being correct

_Now let me check the date range of all evaluations to see what dates we actually have:

[MongoDB (MCP Server)]
Interesting! The data shows there ARE evaluations up to August 25, 2025. Let me now specifically count those from August 2025:_

```
{
  "collection": "hidden"
  "database": "db",
  "query": {
    "created": {
      "$gte": {
        "$date": "2025-08-01T00:00:00.000Z"
      },
      "$lt": {
        "$date": "2025-09-01T00:00:00.000Z"
      }
    }
  }
}
```

Output 
Found 0 documents in the collection `hidden`...

**Model attempted several queries but couldn't get any document back, it knows documents that match the condition exist (because it used aggregate tool with date expression operators)**

```
{
  "collection": "hidden",
  "database": "db",
  "pipeline": [
    {
      "$addFields": {
        "year": {
          "$year": "$created"
        },
        "month": {
          "$month": "$created"
        }
      }
    },
    {
      "$match": {
        "year": 2025,
        "month": 8
      }
    },
    {
      "$count": "august_count"
    }
  ]
}
```
This is puzzling. The aggregation shows a max date of August 25, 2025, but I can't find those records. Let me sort by created date descending to see the most recent ones:...

I can see from the sorted results that there are records with dates like "2025-08-25T10:56:44.046Z"...

Can someone test this to confirm?

gagik:
Hey! Thank you for raising this. 

Is the field it is filtering on inside the database of type `Date` or of type `string`? It may be coming across some mismatch there.

wael7721:
Its of type date, specifically ISODate.
when i query directly using mongodb compass it works by wrapping the date with ISODate("date_here") but the mcp cannot pass it in.

Try an attempt on a mocked data to further test this.

kmruiz:
Hi @wael7721 ,

[...]

PR #571, opened 2025-09-18
fix: Use JSON for stdio, similiar to the HTTP Transport and use EJSON deserialize when necessary MCP-207

kmruiz:
## Proposed changes

We can't leverage using EJSON as the underlying transport because it diverges from the HTTP Transport, and because not all EJSON is equivalent to it's underlying JSON. For example, take this sample into consideration:

```js
$> JSON.stringify({ long: 23098479807234 })
<= {"long":23098479807234}
$> EJSON.stringify({ long: 23098479807234 }) // relaxed = true
<= {"long":23098479807234}
$> EJSON.stringify({ long: 23098479807234 }, { relaxed: false })
<= "long":{"$numberLong":"23098479807234"}}
```

This behaviour forbids us on using `{relaxed: false}` always, as it breaks the shape of the underlying JSON. Moreover, if we use dates, it gets worse, as we lose type info in the protocol when serialising/deserializing.

The approach here is to use a custom `zod` type that leverages on EJSON.deserialize, a function that reads a plain EJSON object and transforms it to the underlying types (for example, $date to js.Date).

This is important because the Node.js driver does not work with EJSON, it works with actual JS objects (like Date), so by using this approach documents and queries using non-primitive types like BinData or Date work as expected.

It tackles the following issues:

* https://github.com/mongodb-js/mongodb-mcp-server/issues/549
* https://github.com/mongodb-js/mongodb-mcp-server/issues/211 (ensures no regression)

## Checklist

### DIFF
--- a/src/tools/args.ts
+++ b/src/tools/args.ts
@@ -78,4 +78,6 @@ function toEJSON<T extends object | undefined>(value: T): T {
     return EJSON.deserialize(value, { relaxed: false }) as T;
 }
 
-export const zEJSON = z.object({}).passthrough().transform(toEJSON);
+export function zEJSON(): z.AnyZodObject {
+    return z.object({}).passthrough().transform(toEJSON) as unknown as z.AnyZodObject;
+}
--- a/src/tools/mongodb/create/insertMany.ts
+++ b/src/tools/mongodb/create/insertMany.ts
@@ -10,7 +10,7 @@ export class InsertManyTool extends MongoDBToolBase {
     protected argsShape = {
         ...DbOperationArgs,
         documents: z
-            .array(zEJSON.describe("An individual MongoDB document"))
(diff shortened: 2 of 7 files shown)


tests/integration/tools/mongodb/read/find.test.ts
```diff
@@ -190,7 +190,7 @@ describeWithMongoDB("find tool", (integration) => {
                 arguments: {
                     database: integration.randomDbName(),
                     collection: "foo",
-                    filter: { _id: fooObject._id },
```
kmruiz, line 193: This test was misleading, because everything runs in the same process, there was no proper ObjectID Serde, it was using the real JavaScript types. This not happens in the usual scenario where the ID has to be serialized.

This change ensures that the ID is properly serialized and sent in EJSON.

PR #574, opened 2025-09-19
chore: bump `@mongosh` dependencies

gagik:
Updates the mongosh service provider. Among other things, this allows us to no longer be broken with mongosh VSCode tests and keep alignment.

Needed for https://github.com/mongodb-js/vscode/pull/1130

### DIFF
--- a/src/common/connectionManager.ts
+++ b/src/common/connectionManager.ts
@@ -223,7 +223,23 @@ export class MCPConnectionManager extends ConnectionManager {
             return this.currentConnectionState;
         }
 
-        return Promise.resolve({ tag: "disconnected" });
+        if (this.currentConnectionState.tag === "connected" || this.currentConnectionState.tag === "connecting") {
+            try {
+                if (this.currentConnectionState.tag === "connected") {
+                    await this.currentConnectionState.serviceProvider?.close();
+                }
+                if (this.currentConnectionState.tag === "connecting") {
+                    const serviceProvider = await this.currentConnectionState.serviceProvider;
+                    await serviceProvider.close();
+                }
+            } finally {
+                this.changeState("connection-close", {
+                    tag: "disconnected",
+                });
+            }
+        }
+
+        return { tag: "disconnected" };
     }
 
     private onOidcAuthFailed(error: unknown): void {


kmruiz:
I see that with this change the OIDC tests are failing with a timeout. Maybe there is a breaking change somewhere in the node provider? I'll take a look too and see if we can get it working.

gagik:
> I see that with this change the OIDC tests are failing with a timeout

~That seems to be a flake, so it's possible some change makes it more likely.~ Oh looks like it's Ubuntu-related, interesting. Will need to look into it too.

I think devtools connect bump should fix it.

kmruiz:
Yeah, OIDC tests only run on Linux (because the server only supports it in Linux) so if you see an ubuntu only issue, it's likely OIDC.

I'm debugging and it seems that the issue is that the OIDC flow never finishes, the connection manager stays on "connecting", like when the user never confirms the authentication.

package.json
```diff
@@ -101,7 +102,6 @@
     "@mongodb-js/devtools-proxy-support": "^0.5.3",
     "@mongosh/arg-parser": "^3.19.0",
     "@mongosh/service-provider-node-driver": "^3.17.0",
-    "@vitest/eslint-plugin": "^1.3.4",
```
gagik: drive by 🙈

kmruiz:
I've actually found the breaking change. I'm trying to fix it now.

It seems that there has been a change in the node provider that now opens the connection eagerly, instead of lazily, and this breaks our flow.

src/common/connectionManager.ts
```diff
@@ -130,6 +129,7 @@ export class MCPConnectionManager extends ConnectionManager {
         super();
         this.bus = bus ?? new EventEmitter();
         this.bus.on("mongodb-oidc-plugin:auth-failed", this.onOidcAuthFailed.bind(this));
+        // eslint-disable-next-line @typescript-eslint/no-misused-promises
         this.bus.on("mongodb-oidc-plugin:auth-succeeded", this.onOidcAuthSucceeded.bind(this));
```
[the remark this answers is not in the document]
kmruiz: I prefer to propagate arguments, and the promise is useless for event emitters, it's just syntactic sugar for the function so we can use await.

tests/integration/tools/mongodb/connect/connect.test.ts
```diff
@@ -67,12 +67,12 @@ describeWithMongoDB(
...
         });
 
-        describe("when the arugment connection string is invalid", () => {
+        describe("when the argument connection string is invalid", () => {
             it("returns error message", async () => {
                 const response = await integration.mcpClient().callTool({
                     name: "switch-connection",
                     arguments: {
-                        connectionString: "mongodb://localhost:12345",
+                        connectionString: "mangobd://localhost:12345",
```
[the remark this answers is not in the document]
kmruiz: It's on purpose for the test.

tests/integration/tools/mongodb/connect/connect.test.ts
```diff
@@ -125,7 +125,7 @@ describeWithMongoDB(
             it("returns error message", async () => {
                 const response = await integration.mcpClient().callTool({
                     name: "connect",
-                    arguments: { connectionString: "mongodb://localhost:12345" },
+                    arguments: { connectionString: "mangodb://localhost:12345" },
```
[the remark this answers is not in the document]
kmruiz: On purpose for the test.

tests/integration/tools/mongodb/connect/connect.test.ts
nirinchev, line 75: Why the change here? I don't suppose there's anything on port 12345 anyway.

src/common/connectionManager.ts
```diff
@@ -187,6 +189,16 @@ export class MCPConnectionManager extends ConnectionManager {
...
+
+            if (isOidcConnection) {
+                this.changeState("connection-request", {
+                    tag: "connecting",
+                    serviceProvider,
+                    connectedAtlasCluster: settings.atlas,
+                    connectionStringAuthType,
+                    oidcConnectionType: connectionStringAuthType as OIDCConnectionAuthType,
+                });
+            }
```
nirinchev: Not sure why this was moved up here - it seems like we're returning it immediately a little bit later anyway.

src/common/connectionManager.ts
```diff
@@ -199,26 +211,16 @@ export class MCPConnectionManager extends ConnectionManager {
...
-                    serviceProvider,
+                    serviceProvider: await serviceProvider,
                     connectionStringAuthType,
-                    oidcConnectionType: connectionStringAuthType as OIDCConnectionAuthType,
                 });
+            } else {
+                return this.currentConnectionState;
             }
-
-            await serviceProvider?.runCommand?.("admin", { hello: 1 });
```
nirinchev, line 202: This seems to have been removed - is this intentional?

nirinchev (commented):
Looks reasonable, though I'm not fully sure why we needed the code rearrangement in connection manager.

tests/integration/tools/mongodb/connect/connect.test.ts
kmruiz: This was testing something different: `mongodb://localhost:12345` is not an invalid URL, it's just an URL where there is no MongoDB cluster.

src/common/connectionManager.ts
kmruiz: Yep, because now the NodeDriverServiceProvider does the ping itself. It changed the actual logic and contains a few breaking changes, that's why we had to rearrenge some code. The explanation is in this commit:

https://github.com/mongodb-js/mongodb-mcp-server/pull/574/commits/213d2349e3028f3c10642efeaf7c877f536ff7fa

But basically, a summary:

NodeDriverServiceProvider.connect used to be lazy, now it's not. It tries to connect eagerly and does a ping and the Promise it returns doesn't resolve until the driver is fully connected. This is different to what we had earlier, where the connection was lazy.

It means, for example that await NDSP.connect would block _until the OIDC flow is completely finished_, which would block the agent to do anything else potentially and leaves the MCP Server in an inconsistent state (disconnected instead of connecting).

We had to rearrange and do a bit of rework, and the tests related to OIDC hadn't changed to make sure we didn't break the existing flow for users.

gagik:
Tested the new connections on VSCode and works well. Thank you @kmruiz for looking into this!

PR #576, opened 2025-09-19
chore(ci): add dependency checks and reorganize tangled dependencies

gagik:
This adds a dep check using `knip`. We use `depcheck` in Compass but this is deprecated and knip is recommend instead. This adds a basic configuration and then uses it to untangle some of our dependencies. This includes:

- Moving dependencies into dev dependencies, most notably `mongodb`! which we actually only use for types in the code itself at the moment.
- Moving certain types, consts into helper utils as they are only relevant there.

This seems like a useful feature and allows us to avoid mixing up our scripting or testing code with production.


gagik:
Ah, it seems trouble figuring out bash scripts... So some deps need to be re-added

knip.json
```diff
@@ -0,0 +1,11 @@
+{
+  "entry": [
+    "src/index.ts!",
```
gagik, line 3: FYI: for knip `!` means it's a production code it treats it differently then entry points files (i.e. that's how it complains about something being a dependency instead of a dev dependency)

package.json
```diff
@@ -91,21 +94,17 @@
...
     "typescript-eslint": "^8.41.0",
     "uuid": "^11.1.0",
-    "vitest": "^3.2.4",
-    "yaml": "^2.8.1"
+    "vitest": "^3.2.4"
   },
   "dependencies": {
     "@modelcontextprotocol/sdk": "^1.17.4",
     "@mongodb-js/device-id": "^0.3.1",
-    "@mongodb-js/devtools-connect": "^3.9.3",
```
gagik, line 100: correct me if I'm wrong but I don't think we were using `devtools-connect` anywhere...

kmruiz: I think it's a transitive dependency of the mongosh service-provider-node-driver. We can get rid of it here.

kmruiz (approved):

nirinchev (approved):

PR #580, opened 2025-09-22
fix(schema): Use sample instead of find for schema sampling

kmruiz:
## Proposed changes

Using find without sorting will return documents in natural order, which is not reliable for sampling because it depends on the last "updated" time a document was updated. By using find, we are biased towards the latest documents in a database that might not even be up to date, introducing bias.

The `$sample` works differently: it actually does an statistical sample by getting random documents from different random places within the same collection. This is more reliable on collections where not all documents have the same amount of fields.

## Checklist


himanshusinghs (commented):
Looks good, but should we use collectCursorLogic here to not exceed the memory limits?

kmruiz:
I actually believe Copilots idea and yours are good, so I'll add both a sampleSize parameter and also use the new cursor logic. This way we can just upgrade this tool to be as reliable as others.

src/tools/mongodb/metadata/collectionSchema.ts
```diff
@@ -1,19 +1,35 @@
...
     public operationType: OperationType = "metadata";
 
-    protected async execute({ database, collection }: ToolArgs<typeof DbOperationArgs>): Promise<CallToolResult> {
+    protected async execute(
+        { database, collection, sampleSize, responseBytesLimit }: ToolArgs<typeof this.argsShape>,
+        { signal }: ToolExecutionContext
+    ): Promise<CallToolResult> {
         const provider = await this.ensureConnected();
-        const documents = await provider.find(database, collection, {}, { limit: 5 }).toArray();
+        const cursor = provider.aggregate(database, collection, [{ $sample: { size: Math.min(sampleSize, this.config.maxDocumentsPerQuery) } }]);
```
nirinchev: I wonder if we want to limit the sample to `maxDocumentsPerQuery` - the way I interpreted this config option, it's dealing with the number of documents we'd be returning to the LLM, not necessarily the number of documents we're fetching internally - e.g. the LLM shouldn't care if we sample 50 or 1000 docs since it's only seeing the inferred schema anyway.

nirinchev (approved):
One question, otherwise looks good (once prettier is made happy).

src/tools/mongodb/metadata/collectionSchema.ts
kmruiz: It could be another option, I just wanted to limit in case a model gets crazy and tries to query thousands and thousands of documents for sampling. $sample is a bit more expensive than just finding, so it's just for safety.

No strong opinion here by the way, we can have a specific hardcoded option for sample in a constant.

kmruiz: Changed to a constant for the upper limit.

PR #589, opened 2025-09-23
feat: Atlas get performance advisor integration tests

kylelai1:
## Proposed changes
- Adds integration tests for the performance advisor tool.
- Gets process IDs from the standard connection string, which contains the hostnames.
- Note: Mocks out the API responses for the performance advisor for 1 test because we can't guarantee that there will be performance advisor suggestions in our testing.
- Also limits the max number of returned slow query logs from the tool, and changed wording to indicate that we are returning a sample of slow query logs. This is due to manually testing, and finding that the LLM can not interpret too large of a range of slow queries if the number increases towards 20,000 (which is the default).

## Checklist


src/common/atlas/cluster.ts
```diff
@@ -3,23 +3,45 @@ import type { ApiClient } from "./apiClient.js";
 import { LogId } from "../logger.js";
 
 const DEFAULT_PORT = "27017";
+
+function extractProcessIds(connectionString: string): Array<string> {
```
kmruiz: We already provide a way to parse connection URLs through the mongodb driver:

https://www.npmjs.com/package/mongodb-connection-string-url?activeTab=readme

You can see an example here:

https://github.com/mongodb-js/mongodb-mcp-server/blob/main/src/helpers/connectionOptions.ts#L24-L42

It's better to use this one because it's the same we use in mongosh and the driver, and it basically ensures compatibility and has been more battle tested.

I think the host:port split still needs to be done, but it doesn't depend on a regex so it should be fine.

kmruiz:
```suggestion
type AtlasProcessId = "${string}:${number}"

function extractProcessIds(connectionString: string): Array<AtlasProcessId> {
```

It is more clear what the format of the string is expected, but opaque enough so can be used without knowing it.

src/common/atlas/cluster.ts
```diff
@@ -3,23 +3,45 @@ import type { ApiClient } from "./apiClient.js";
...
+    if (!connectionString || connectionString === "") return [];
```
kmruiz: !connectionString already covers connectionString === ""

```suggestion
    if (!connectionString) return [];
```

src/common/atlas/cluster.ts
```diff
@@ -3,23 +3,45 @@ import type { ApiClient } from "./apiClient.js";
...
+    const matches = connectionString.match(/^mongodb:\/\/([^/]+)/);
+    if (!matches) {
+        return [];
+    }
+
+    // matches[1] gives us the host:port pairs
+    const hostsString = matches[1];
+    const hosts = hostsString?.split(",") ?? [];
+
+    return hosts?.map((host) => {
```
kmruiz: hosts can't be null or undefined, because of the `??` branch.

```suggestion
    return hosts.map((host) => {
```

src/common/atlas/cluster.ts
```diff
@@ -3,23 +3,45 @@ import type { ApiClient } from "./apiClient.js";
...
+    const connectionString = cluster.connectionStrings?.standardSrv || cluster.connectionStrings?.standard;
     return {
         name: cluster.name,
         instanceType: "FLEX",
         instanceSize: undefined,
         state: cluster.stateName,
         mongoDBVersion: cluster.mongoDBVersion,
-        connectionString: cluster.connectionStrings?.standardSrv || cluster.connectionStrings?.standard,
+        connectionString,
+        processIds: extractProcessIds(cluster.connectionStrings?.standard ?? ""),
```
kmruiz, line 34: Is this correct? Why are we using standard here and not standardSrv or the already extracted connectionString?

tests/integration/tools/atlas/atlasHelpers.ts
```diff
@@ -143,3 +144,61 @@ async function createProject(apiClient: ApiClient): Promise<Group & Required<Pic
...
+): Promise<void> {
+    await session.apiClient.deleteCluster({
+        params: {
+            path: {
+                groupId: projectId,
+                clusterName,
+            },
+        },
+    });
+    while (true) {
```
kmruiz: We shouldn't do this because it can block the entire MCP Server if there is an issue in Atlas or it doesn't work as we expected.

We should:

1. Always provide an upper found. It can be relatively high, like 100 iterations, but must not be unbound.
2. Provide an abort signal, so the caller can stop waiting for the deleteCluster (for example, requests timeouts can cancel this).


```suggestion
    for (let try = 0; try < MAX_ITERATIONS && !abortSignal.aborted; try++)
```

AbortSignal would be a parameter provided by the caller in deleteAndWaitCluster.

tests/integration/tools/atlas/atlasHelpers.ts
```diff
@@ -143,3 +144,61 @@ async function createProject(apiClient: ApiClient): Promise<Group & Required<Pic
...
+        try {
+            await session.apiClient.getCluster({
```
kmruiz: I think I understand what happens here (getCluster will throw an exception when the cluster is not in Atlas anymore) but I think this could be a bit more documented. Maybe we can add create a function like:

```
assertClusterIsAvailable(projectId, clusterName)
```

That returns true/false and break on that?

tests/integration/tools/atlas/atlasHelpers.ts
```diff
@@ -143,3 +144,61 @@ async function createProject(apiClient: ApiClient): Promise<Group & Required<Pic
...
+}
+
+export async function waitCluster(
+    session: Session,
+    projectId: string,
+    clusterName: string,
+    check: (cluster: ClusterDescription20240805) => boolean | Promise<boolean>,
+    pollingInterval: number = 1000
+): Promise<void> {
+    while (true) {
```
kmruiz: Same suggestion as the other while(true).

tests/integration/tools/atlas/performanceAdvisor.test.ts
```diff
@@ -0,0 +1,232 @@
+// This test file includes long running tests (>10 minutes) because we provision a real M10 cluster, which can take up to 10 minutes to provision.
```
blva, line 1: I see in some of the tests we got 
`ApiClientError: [402 Payment Required] error calling Atlas API: Payment Required; No payment information was found for group 68d40e557ad0c21953984dc6.`

Let's make sure we cover that case with details on how to setup payment. A lot of customers will hit this issue if they spin up anew project. We should share. alink on how to setup billing. Check `handleError` in `atlasTool` , I think we can wrap that error and provide a better experience!

src/common/atlas/cluster.ts
kylelai1: Yeah, this is a good find. We can use the ConnectionString class and it'll give us the host and ports.

kylelai1: Will add this!

src/common/atlas/cluster.ts
kylelai1: Yes, this is right. We are using the standard connection string because we need the port as well, and the SRV string does not give us hostnames and ports.

tests/integration/tools/atlas/atlasHelpers.ts
kylelai1: Yes, we can definitely add an upper bound for tries here. Do we need the abortSignal for now? We don't have any external triggers where we need to pass in an AbortController.

tests/integration/tools/atlas/atlasHelpers.ts
kylelai1: Yeah, we can make a separate function which should clean it up.

tests/integration/tools/atlas/performanceAdvisor.test.ts
kylelai1: Sure. What expectations do we have for people creating a new tool? Since this requires payment to run integration tests, do we expect users to do that? We're running these tests on cloud-dev, so it doesn't cost us as developers anything.

kylelai1: (Added the wrapper around the error 402 for a better description of the error)

blva: Thank you! 

This is the first tool that fails if users don't have billing setup so we'll hit such different issues, I think it's worth adding the billing setup in one of the projects manually. I can give you access there.

kylelai1: Ok sounds good. I can just add the standard mock billing details if it's on cloud-dev.

src/common/atlas/performanceAdvisorUtils.ts
```diff
@@ -112,22 +114,35 @@ export async function getSlowQueries(
...
+                        ...(namespaces && { namespaces: namespaces }),
+                        nLogs: DEFAULT_SLOW_QUERY_LOGS_LIMIT,
+                    },
                 },
-            },
-        });
+            })
+        );
+
+        const responses = await Promise.all(slowQueryPromises);
```
kmruiz: I'm not sure we want to use Promise.all here. Promise.all will return a value when all queries are successful, but if one fails, it returns a failed promise and other success values are discarded.

If we want to be more resilient, we should use [Promise.allSettled](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/allSettled). But, if we are fine with the current behaviour, I'm not against it.

kmruiz (approved):

src/tools/atlas/atlasTool.ts
```diff
@@ -38,14 +38,29 @@ For more information on setting up API keys, visit: https://www.mongodb.com/docs
                 };
             }
 
+            if (statusCode === 402) {
```
nirinchev, line 41: Are we getting 402 when a customer tries to access PA on a free/shared tier cluster?

kylelai1: No, we're getting a 402 for the integration tests when they're run with an org that has no payment set up, since an M10 is created for the new integration test. PA on a free/shared tier cluster will be a 200.

src/common/atlas/performanceAdvisorUtils.ts
kylelai1: Promise.allSettled is good. I'll use that

src/tools/atlas/atlasTool.ts
nirinchev: So are users likely to ever get the 402 response?

blva: I think one thing to add here is probably elicitation is needed for all these tools since they can add cost to the user, here's some context https://github.com/mongodb-js/mongodb-mcp-server/commit/bcbf8894915a1d65d11864007464e18d02995bc7

kylelai1: users are not likely to get the 402 response. This is purely for the integration tests.

src/tools/atlas/read/getPerformanceAdvisor.ts
```diff
@@ -57,26 +60,39 @@ export class GetPerformanceAdvisorTool extends AtlasToolBase {
...
-                    : "No drop index suggestions found.",
-                slowQueryLogsResult && slowQueryLogsResult?.slowQueryLogs?.length > 0
-                    ? `## Slow Query Logs\n${JSON.stringify(slowQueryLogsResult.slowQueryLogs)}`
-                    : "No slow query logs found.",
-                schemaSuggestionsResult && schemaSuggestionsResult?.recommendations?.length > 0
-                    ? `## Schema Suggestions\n${JSON.stringify(schemaSuggestionsResult.recommendations)}`
-                    : "No schema suggestions found.",
+                `## Suggested Indexes\n${
+                    hasSuggestedIndexes
+                        ? `Note: The "Weight" field is measured in bytes, and represents the estimated number of bytes saved in disk reads per executed read query that would be saved by implementing an index suggestion. Please convert this to MB or GB for easier readability.\n${JSON.stringify(suggestedIndexesResult.suggestedIndexes)}`
```
kylelai1: This is included from product feedback to make it apparent to the LLM that the "weight" field is in bytes, and gives the definition to relay to the LLM

src/tools/atlas/read/getPerformanceAdvisor.ts
```diff
@@ -49,34 +52,58 @@ export class GetPerformanceAdvisorTool extends AtlasToolBase {
     }: ToolArgs<typeof this.argsShape>): Promise<CallToolResult> {
         try {
             const [suggestedIndexesResult, dropIndexSuggestionsResult, slowQueryLogsResult, schemaSuggestionsResult] =
-                await Promise.all([
+                await Promise.allSettled([
```
kylelai1, line 55: Changed this to Promise.allSettled. Made sure it was type safe below as well when using the allSettled.

kylelai1:
I've just added some changes after demo-ing and testing with the product team, so I've re-requested review for the additional changes/enhancements as well.

src/tools/atlas/read/getPerformanceAdvisor.ts
```diff
@@ -49,34 +52,58 @@ export class GetPerformanceAdvisorTool extends AtlasToolBase {
...
-                    : "No drop index suggestions found.",
-                slowQueryLogsResult && slowQueryLogsResult?.slowQueryLogs?.length > 0
-                    ? `## Slow Query Logs\n${JSON.stringify(slowQueryLogsResult.slowQueryLogs)}`
-                    : "No slow query logs found.",
-                schemaSuggestionsResult && schemaSuggestionsResult?.recommendations?.length > 0
-                    ? `## Schema Suggestions\n${JSON.stringify(schemaSuggestionsResult.recommendations)}`
-                    : "No schema suggestions found.",
+                `## Suggested Indexes\n${
+                    hasSuggestedIndexes
+                        ? `Note: The "Weight" field is measured in bytes, and represents the estimated number of bytes saved in disk reads per executed read query that would be saved by implementing an index suggestion. Please convert this to MB or GB for easier readability.\n${JSON.stringify(suggestedIndexesResult.value?.suggestedIndexes)}`
```
kmruiz, line 101: Probably this calculation is something that we can do ourselves? It's trivial and we avoid the situation where different LLMs do the calculation differently or wrong.

tests/integration/tools/atlas/performanceAdvisor.test.ts
```diff
@@ -0,0 +1,233 @@
...
+                    projectId,
+                    clusterName,
+                    (cluster) => {
+                        return cluster.stateName === "IDLE";
+                    },
+                    10000,
+                    120
+                );
+
+                await sleep(10000);
```
kmruiz: Maybe we can remove this sleep if we are already waiting for the cluster to be ready?

tests/integration/tools/atlas/performanceAdvisor.test.ts
```diff
@@ -0,0 +1,233 @@
...
+            }, 1200000);
```
kmruiz: nit: Can we store this timeout in a global variable? It would be easier to understand and change.

src/tools/atlas/read/getPerformanceAdvisor.ts
kylelai1: We can do the calculation ourselves, but that would mean we'd need to determine whether we want to convert to KB or GB, which I think would depend on the number itself. If the "weight" changes, and is sometimes KB and sometimes GB, then we need to add the units to the "weight" field, which is currently typed as a number. If we want, we can just leave it as bytes, and not have any conversion being done.

tests/integration/tools/atlas/performanceAdvisor.test.ts
kylelai1: Yes, removing the sleep here!

tests/integration/tools/atlas/performanceAdvisor.test.ts
kylelai1: Done! Added underscores to improve readability.

kmruiz (approved):

PR #607, opened 2025-10-02
stop waiting for cluster deletion

blva:
## Proposed changes

- just sends the cluster deletion command and not wait for deletion to speed up tests

## Checklist


tests/integration/tools/atlas/clusters.test.ts
```diff
@@ -34,6 +29,17 @@ async function deleteAndWaitCluster(session: Session, projectId: string, cluster
     }
 }
 
+async function deleteCluster(session: Session, projectId: string, clusterName: string): Promise<void> {
```
nirinchev: Rather than creating a new function, consider adding an argument to the existing one that indicates whether to wait for the deletion or not.

PR #612, opened 2025-10-06
feat(atlas-local): Adds Atlas Local Connect Deployment tool

cveticm:
## Proposed changes

1. Implements tool atlas-local-connect-deployment

3. Adds integration tests

5. Adds accuracy tests

7. Adds hint for connecting for local deployment

## Checklist

### DIFF
--- a/src/common/connectionErrorHandler.ts
+++ b/src/common/connectionErrorHandler.ts
@@ -20,9 +20,18 @@ export const connectionErrorHandler: ConnectionErrorHandler = (error, { availabl
     // Find the first Atlas connect tool if available and suggest to the LLM to use it.
     // Note: if we ever have multiple Atlas connect tools, we may want to refine this logic to select the most appropriate one.
     const atlasConnectTool = connectTools?.find((t) => t.category === "atlas");
-    const llmConnectHint = atlasConnectTool
-        ? `Note to LLM: prefer using the "${atlasConnectTool.name}" tool to connect to an Atlas cluster over using a connection string. Make sure to ask the user to specify a cluster name they want to connect to or ask them if they want to use the "list-clusters" tool to list all their clusters. Do not invent cluster names or connection strings unless the user has explicitly specified them. If they've previously connected to MongoDB using MCP, you can ask them if they want to reconnect using the same cluster/connection.`
-        : "Note to LLM: do not invent connection strings and explicitly ask the user to provide one. If they have previously connected to MongoDB using MCP, you can ask them if they want to reconnect using the same connection string.";
+    const atlasLocalConnectTool = connectTools?.find((t) => t.category === "atlas-local");
+
+    let llmConnectHint: string;
+
+    if (atlasConnectTool) {
+        llmConnectHint = `Note to LLM: prefer using the "${atlasConnectTool.name}" tool to connect to an Atlas cluster over using a connection string. Make sure to ask the user to specify a cluster name they want to connect to or ask them if they want to use the "list-clusters" tool to list all their clusters. Do not invent cluster names or connection strings unless the user has explicitly specified them. If they've previously connected to MongoDB using MCP, you can ask them if they want to reconnect using the same cluster/connection.`;
+    } else if (atlasLocalConnectTool) {
+        llmConnectHint = `Note to LLM: prefer using the "${atlasLocalConnectTool.name}" tool for connecting to local MongoDB deployments. Ask the user to specify a deployment name or use "atlas-local-list-deployments" to show available local deployments. Do not invent deployment names unless the user has explicitly specified them. If they've previously connected to a local MongoDB deployment using MCP, you can ask them if they want to reconnect using the same deployment.`;
+    } else {
+        llmConnectHint =
+            "Note to LLM: do not invent connection strings and explicitly ask the user to provide one. If they have previously connected to MongoDB using MCP, you can ask them if they want to reconnect using the same connection string.";
+    }
 
     const connectToolsNames = connectTools?.map((t) => `"${t.name}"`).join(", ");
     const additionalPromptForConnectivity: { type: "text"; text: string }[] = [];
--- a/src/common/connectionErrorHandler.ts
+++ b/src/common/connectionErrorHandler.ts
@@ -17,21 +17,29 @@ export const connectionErrorHandler: ConnectionErrorHandler = (error, { availabl
         .filter((t) => t.operationType === "connect")
         .sort((a, b) => a.category.localeCompare(b.category)); // Sort Atlas tools before MongoDB tools
 
-    // Find the first Atlas connect tool if available and suggest to the LLM to use it.
-    // Note: if we ever have multiple Atlas connect tools, we may want to refine this logic to select the most appropriate one.
+    // Find what Atlas connect tools are available and suggest when the LLM should to use each. If no Atlas tools are found, return a suggestion for the MongoDB connect tool.
     const atlasConnectTool = connectTools?.find((t) => t.category === "atlas");
     const atlasLocalConnectTool = connectTools?.find((t) => t.category === "atlas-local");
 
-    let llmConnectHint: string;
+    const llmConnectHint = ((): string => {
+        const hints: string[] = [];
 
-    if (atlasConnectTool) {
-        llmConnectHint = `Note to LLM: prefer using the "${atlasConnectTool.name}" tool to connect to an Atlas cluster over using a connection string. Make sure to ask the user to specify a cluster name they want to connect to or ask them if they want to use the "list-clusters" tool to list all their clusters. Do not invent cluster names or connection strings unless the user has explicitly specified them. If they've previously connected to MongoDB using MCP, you can ask them if they want to reconnect using the same cluster/connection.`;
-    } else if (atlasLocalConnectTool) {
-        llmConnectHint = `Note to LLM: prefer using the "${atlasLocalConnectTool.name}" tool for connecting to local MongoDB deployments. Ask the user to specify a deployment name or use "atlas-local-list-deployments" to show available local deployments. Do not invent deployment names unless the user has explicitly specified them. If they've previously connected to a local MongoDB deployment using MCP, you can ask them if they want to reconnect using the same deployment.`;
-    } else {
-        llmConnectHint =
-            "Note to LLM: do not invent connection strings and explicitly ask the user to provide one. If they have previously connected to MongoDB using MCP, you can ask them if they want to reconnect using the same connection string.";
-    }
+        if (atlasConnectTool) {
+            hints.push(
+                `Note to LLM: prefer using the "${atlasConnectTool.name}" tool to connect to an Atlas cluster over using a connection string. Make sure to ask the user to specify a cluster name they want to connect to or ask them if they want to use the "list-clusters" tool to list all their clusters. Do not invent cluster names or connection strings unless the user has explicitly specified them. If they've previously connected to MongoDB using MCP, you can ask them if they want to reconnect using the same cluster/connection.`
+            );
+        }
+
+        if (atlasLocalConnectTool) {
+            hints.push(
+                `Note to LLM: For MongoDB Atlas Local deployments, ask the user to either provide a connection string, specify a deployment name, or use "atlas-local-list-deployments" to show available local deployments. If a deployment name is provided, prefer using the "${atlasLocalConnectTool.name}" tool. If a connection string is provided, prefer using the "connect" tool. Do not invent deployment names or connection strings unless the user has explicitly specified them. If they've previously connected to a MongoDB Atlas Local deployment using MCP, you can ask them if they want to reconnect using the same deployment.`
+            );
+        }
+
+        return hints.length > 0
+            ? hints.join("\n")
+            : "Note to LLM: do not invent connection strings and explicitly ask the user to provide one. If they have previously connected to MongoDB using MCP, you can ask them if they want to reconnect using the same connection string.";
+    })();
 
     const connectToolsNames = connectTools?.map((t) => `"${t.name}"`).join(", ");
     const additionalPromptForConnectivity: { type: "text"; text: string }[] = [];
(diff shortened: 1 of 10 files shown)


tests/integration/tools/atlas/clusters.test.ts
```diff
@@ -239,7 +239,7 @@ describeWithAtlas("clusters", (integration) => {
                         "You need to connect to a MongoDB instance before you can access its data."
                     );
                     expect(elements[1]?.text).toContain(
-                        'Please use one of the following tools: "atlas-connect-cluster", "connect" to connect to a MongoDB instance'
+                        'Please use one of the following tools: "atlas-connect-cluster", "atlas-local-connect-deployment", "connect" to connect to a MongoDB instance'
```
cveticm: added to fix test fail

tests/accuracy/connectDeployment.test.ts
```diff
@@ -0,0 +1,70 @@
...
+            {
+                toolName: "atlas-local-connect-deployment",
+                parameters: {
+                    deploymentIdOrName: "my-instance",
+                },
+            },
+        ],
+    },
+    {
+        prompt: "If and only if, the local MongoDB deployment 'local-mflix' exists, then connect to it",
```
kmruiz, line 28: Not sure this test is worth it? In this PR we don't implement the atlas-local-list-deployments tool, so essentially we are testing an impossible scenario.

src/tools/atlasLocal/connect/connectDeployment.ts
```diff
@@ -0,0 +1,34 @@
...
+    public name = "atlas-local-connect-deployment";
+    protected description = "Connect to a MongoDB Atlas Local deployment";
+    public operationType: OperationType = "connect";
+    protected argsShape = {
+        deploymentIdOrName: z.string().describe("Name or ID of the deployment to connect to"),
+    };
+
+    protected async executeWithAtlasLocalClient(
+        client: Client,
+        { deploymentIdOrName }: ToolArgs<typeof this.argsShape>
```
kmruiz: I understand that right now there is no way to fetch the list of deployments, right? At least from within the MCP Server. The user still requires the Atlas CLI?

kmruiz (changes requested):
The code itself looks fine, but there are some things to consider before merging.

Also I think this feature should be feature toggled and disabled by default until we implement `atlas-local-list-deployments`.

@blva, WDYT?

Luke-Sanderson:
> The code itself looks fine, but there are some things to consider before merging.
> 
> Also I think this feature should be feature toggled and disabled by default until we implement `atlas-local-list-deployments`.
> 
> @blva, WDYT?

I believe the `atlas-list-deployments` tool was added here https://github.com/mongodb-js/mongodb-mcp-server/pull/520

src/tools/atlasLocal/connect/connectDeployment.ts
cveticm: To clarify, the tool `atlas-local-list-deployments` does exist in the feature branch `feat-MCP-40`, which this PR is working against

kmruiz: Ah I didn't see it wasn't based on main.

Apologies!

kmruiz:
Didn't see that the base wasn't main. I see the tool now. Resolving the comment!

src/tools/atlasLocal/connect/connectDeployment.ts
```diff
@@ -0,0 +1,34 @@
...
+import type { OperationType, ToolArgs } from "../../tool.js";
+import type { Client } from "@mongodb-js-preview/atlas-local";
+import { z } from "zod";
+
+export class ConnectDeploymentTool extends AtlasLocalToolBase {
+    public name = "atlas-local-connect-deployment";
+    protected description = "Connect to a MongoDB Atlas Local deployment";
+    public operationType: OperationType = "connect";
+    protected argsShape = {
+        deploymentIdOrName: z.string().describe("Name or ID of the deployment to connect to"),
```
himanshusinghs: Instead of `z.string()` should we prefer - CommonArgs.string.

@blva wasn't the idea of this (`CommonArgs.string`) to use everywhere we want a string input? Asking because I noticed that mongodb tools still use the zod.string

tests/integration/tools/atlas-local/connectDeployment.test.ts
```diff
@@ -0,0 +1,98 @@
...
+
+// Docker is not available on macOS in GitHub Actions
+// That's why we skip the tests on macOS in GitHub Actions
+describe("atlas-local-connect-deployment", () => {
+    const integration = setupIntegrationTest(
+        () => defaultTestConfig,
+        () => defaultDriverOptions
+    );
+
+    it.skipIf(isMacOSInGitHubActions)("should have the atlas-local-connect-deployment tool", async ({ signal }) => {
```
himanshusinghs: Out of curiosity - why do we skip these tests on macos? Is there an underlying limitation of local cli deployments?

tests/integration/tools/atlas-local/connectDeployment.test.ts
```diff
@@ -0,0 +1,98 @@
...
+
+        // cleanup
+        try {
+            await integration.mcpClient().callTool({
+                name: "atlas-local-delete-deployment",
+                arguments: { deploymentName },
+            });
+        } catch (error) {
+            console.warn(`Failed to delete deployment ${deploymentName}:`, error);
+        }
```
himanshusinghs: It would be better to do this and any cleanup related logic in the tests tear down (afterEach block) because there is a good chance of this code not even executing if the expectations above it fails.

tests/integration/tools/atlas-local/connectDeployment.test.ts
cveticm: Docker is not available on MacOS in github actions,  all atlas-local integration tests skip for macos

tests/integration/tools/atlas-local/connectDeployment.test.ts
cveticm: Noted, I went for this approach since only one test required cleanup but with a cleanup block it'll be easier if we add on more cases 👍

src/tools/atlasLocal/connect/connectDeployment.ts
cveticm: Ah yes I see this change has been made to the Atlas tools since the feature branch has been created. I can open up a separate ticket for this work since the change will be required across the rest of the atlas-local tools.

src/common/connectionErrorHandler.ts
```diff
@@ -20,9 +20,18 @@ export const connectionErrorHandler: ConnectionErrorHandler = (error, { availabl
...
+
+    let llmConnectHint: string;
+
+    if (atlasConnectTool) {
+        llmConnectHint = `Note to LLM: prefer using the "${atlasConnectTool.name}" tool to connect to an Atlas cluster over using a connection string. Make sure to ask the user to specify a cluster name they want to connect to or ask them if they want to use the "list-clusters" tool to list all their clusters. Do not invent cluster names or connection strings unless the user has explicitly specified them. If they've previously connected to MongoDB using MCP, you can ask them if they want to reconnect using the same cluster/connection.`;
+    } else if (atlasLocalConnectTool) {
+        llmConnectHint = `Note to LLM: prefer using the "${atlasLocalConnectTool.name}" tool for connecting to local MongoDB deployments. Ask the user to specify a deployment name or use "atlas-local-list-deployments" to show available local deployments. Do not invent deployment names unless the user has explicitly specified them. If they've previously connected to a local MongoDB deployment using MCP, you can ask them if they want to reconnect using the same deployment.`;
```
nirinchev: Hm... I wonder if we want to be that prescriptive - my understanding is that this tool would be available as long as the user has docker installed, regardless of whether they're connecting to a local deployment or not. I guess we'd need to vibe check this in the real world.

src/tools/atlasLocal/connect/connectDeployment.ts
nirinchev: Why does this take id if we're not surfacing the id in list-deployments?

tests/integration/tools/atlas-local/connectDeployment.test.ts
```diff
@@ -0,0 +1,105 @@
...
+    setupIntegrationTest,
+    waitUntilMcpClientIsSet,
+} from "../../helpers.js";
+import { afterEach, describe, expect, it } from "vitest";
+
+const isMacOSInGitHubActions = process.platform === "darwin" && process.env.GITHUB_ACTIONS === "true";
+
+// Docker is not available on macOS in GitHub Actions
+// That's why we skip the tests on macOS in GitHub Actions
+describe("atlas-local-connect-deployment", () => {
```
nirinchev: Should we move this `.skipIf` here?

tests/integration/tools/atlas-local/connectDeployment.test.ts
```diff
@@ -0,0 +1,105 @@
...
+
+    it.skipIf(isMacOSInGitHubActions)("should have correct metadata", async ({ signal }) => {
+        await waitUntilMcpClientIsSet(integration.mcpServer(), signal);
+        const { tools } = await integration.mcpClient().listTools();
+        const connectDeployment = tools.find((tool) => tool.name === "atlas-local-connect-deployment");
+        expectDefined(connectDeployment);
+        expect(connectDeployment.inputSchema.type).toBe("object");
+        expectDefined(connectDeployment.inputSchema.properties);
+        expect(connectDeployment.inputSchema.properties).toHaveProperty("deploymentIdOrName");
+    });
```
nirinchev: Consider using the `validateToolMetadata` helper to simplify tests like this.

tests/integration/tools/atlas-local/connectDeployment.test.ts
```diff
@@ -0,0 +1,105 @@
...
+            await waitUntilMcpClientIsSet(integration.mcpServer(), signal);
+
+            const response = await integration.mcpClient().callTool({
+                name: "atlas-local-connect-deployment",
+                arguments: { deploymentIdOrName: "non-existent" },
+            });
+            const elements = getResponseElements(response.content);
+            expect(elements.length).toBeGreaterThanOrEqual(1);
+            expect(elements[0]?.text).toContain(
+                "Docker responded with status code 404: No such container: non-existent"
```
nirinchev: q: should we be translating these errors? I don't imagine customers care too deeply about docker containers.

tests/integration/tools/atlas-local/connectDeployment.test.ts
```diff
@@ -0,0 +1,105 @@
...
+
+        // Connect to the deployment
+        const response = await integration.mcpClient().callTool({
+            name: "atlas-local-connect-deployment",
+            arguments: { deploymentIdOrName: deploymentName },
+        });
+        const elements = getResponseElements(response.content);
+        expect(elements.length).toBeGreaterThanOrEqual(1);
+        expect(elements[0]?.text).toContain(
+            'Successfully connected to Atlas Local deployment "' + deploymentName + '".'
```
nirinchev:
```suggestion
            `Successfully connected to Atlas Local deployment "${deploymentName}".`
```

tests/integration/tools/atlas/clusters.test.ts
nirinchev: Hm... should this be the case across the board? I.e. on macOS GHA runners, I'd expect this tool to be unavailable, no?

tests/integration/tools/atlas-local/connectDeployment.test.ts
```diff
@@ -0,0 +1,105 @@
...
+            });
+            const elements = getResponseElements(response.content);
+            expect(elements.length).toBeGreaterThanOrEqual(1);
+            expect(elements[0]?.text).toContain(
+                "Docker responded with status code 404: No such container: non-existent"
+            );
+        }
+    );
+
+    it.skipIf(isMacOSInGitHubActions)("should connect to a deployment when calling the tool", async ({ signal }) => {
```
nirinchev: Consider creating a separate suite for this case with its own setup/cleanup. Something like:

```ts
describe("when local deployments are available", () => {
  beforeEach(() => {
    // create deployment
  });

  afterEach(() => {
    // cleanup deployment
  });

  it("should connect to a deployment...", () => {
    // test without the setup
  });
});
```

Also, might be a good idea to test when there are more than one deployments that we connect to the correct one. As things stand today, all tests will pass if we disregard the deployment id/name argument and instead connect to the first available deployment.

tests/integration/tools/atlas/clusters.test.ts
cveticm: Good point @nirinchev , i was only considering how it runs in cicd. Changing it to expect either depending on testing environment

src/tools/atlasLocal/connect/connectDeployment.ts
cveticm: opened ticket: https://jira.mongodb.org/browse/CLOUDP-349937

tests/integration/tools/atlas-local/connectDeployment.test.ts
cveticm: sounds good, added error handling to atlas local tools. This can be expanded in future to translate more errors

src/tools/atlasLocal/connect/connectDeployment.ts
cveticm: The tool can handle both id and name but I agree, since we don't expose the container id (even in cli) I changed the input name and description to avoid confusion

src/common/connectionErrorHandler.ts
cveticm: @nirinchev Thoughts on these actions?:
1. Change the logic such that both atlasConnectTool and atlasLocalConnectTool hints can be provided if both tools are available. (currently only atlasConnectTool hint is printed if the tool is avb)
2. Change atlasLocalConnectTool hint to not have preference for it - since the connectTool can also be used to connect to local deployments if given the connection string. 
```suggestion
   llmConnectHint = `Note to LLM: For Atlas Local deployments, ask the user to either provide a connection string, specify a deployment name, or use "atlas-local-list-deployments" to show available local deployments. If a deployment name is provided, prefer using the "${atlasLocalConnectTool.name}" tool. If a connection string is provided, prefer using the "${connectTool.name}" tool. Do not invent deployment names or connection strings unless the user has explicitly specified them. If they've previously connected to an Atlas Local deployment using MCP, you can ask them if they want to reconnect using the same deployment.`;
```

nirinchev: Sounds reasonable to me!

PR #621, opened 2025-10-07
feat: add ability to create vector search indexes MCP-234 MCP-243

nirinchev:
## Proposed changes

This adds the ability to create a vector search index to the `create-index` tool. This is a consolidated tool that's using a discriminated union definition when deciding what type of index to create.

For the initial release, we're feature-flagging the search capabilities, so only adding the vector search option to the union in case the feature flag is enabled, but that will be dropped in the future. We're not updating the definition regardless of whether we're connected to atlas search capable server or not.

### DIFF
--- a/tests/integration/tools/mongodb/create/createIndex.test.ts
+++ b/tests/integration/tools/mongodb/create/createIndex.test.ts
@@ -491,7 +491,7 @@ describeWithMongoDB(
                 expect(indexes[0]?.queryable).toEqual(false);
                 expect(indexes[0]?.latestDefinition).toEqual({
                     fields: [
-                        { type: "vector", path: "vector_1", numDimensions: 4, similarity: "cosine" },
+                        { type: "vector", path: "vector_1", numDimensions: 4, similarity: "euclidean" },
                         { type: "filter", path: "category" },
                     ],
                 });


src/common/connectionManager.ts
```diff
@@ -32,9 +32,33 @@ export interface ConnectionState {
...
+    constructor(
+        public serviceProvider: NodeDriverServiceProvider,
+        public connectionStringAuthType?: ConnectionStringAuthType,
+        public connectedAtlasCluster?: AtlasClusterConnectionInfo
+    ) {}
+
+    private _isSearchSupported?: boolean;
+
+    public async isSearchSupported(): Promise<boolean> {
+        if (this._isSearchSupported === undefined) {
```
himanshusinghs, line 47: A small headsup - @kmruiz has implemented a similar solution in his PR. I did have a small concern about clearing the cached state [there](https://github.com/mongodb-js/mongodb-mcp-server/pull/626#discussion_r2429001505) and that is fixed here so I feel happy about that.

src/common/session.ts
```diff
@@ -141,6 +141,15 @@ export class Session extends EventEmitter<SessionEvents> {
         return this.connectionManager.currentConnectionState.tag === "connected";
     }
 
+    get isConnectedToMongot(): Promise<boolean> {
```
himanshusinghs: q: why not call it isSearchSupported? the same as the underlying method that it calls?

src/tools/mongodb/create/createIndex.ts
```diff
@@ -1,33 +1,147 @@
...
+                    .describe("Definition for a field that will be used for pre-filtering results."),
+                z
+                    .object({
+                        type: z.literal("vector"),
+                        path: z
+                            .string()
+                            .describe(
+                                "Name of the field to index. For nested fields, use dot notation to specify path to embedded fields"
+                            ),
+                        numDimensions: z
```
himanshusinghs: Should there be a default for this?

src/tools/mongodb/create/createIndex.ts
```diff
@@ -1,33 +1,147 @@
...
+            case "vectorSearch":
+                {
+                    const isVectorSearchSupported = await this.session.isConnectedToMongot;
+                    if (!isVectorSearchSupported) {
+                        // TODO: remove hacky casts once we merge the local dev tools
+                        const isLocalAtlasAvailable =
+                            (this.server?.tools.filter((t) => t.category === ("atlas-local" as unknown as ToolCategory))
+                                .length ?? 0) > 0;
+
+                        const CTA = isLocalAtlasAvailable ? "`atlas-local` tools" : "Atlas CLI";
```
himanshusinghs, line 123: I would probably like to use the same. Thinking of moving `isLocalAtlasToolsAvailable` to Server class itself. Feel free to skip this change in your PR, I will take over in mine.

tests/accuracy/createIndex.test.ts
```diff
@@ -1,6 +1,20 @@
...
+    originalApiKey = process.env.MDB_VOYAGE_API_KEY;
+
+    // We just need a valid key when registering the tool, the actual value is not important
+    if (!originalApiKey) {
+        process.env.MDB_VOYAGE_API_KEY = "valid-key";
+    }
+});
+afterAll(() => {
+    process.env.MDB_VOYAGE_API_KEY = originalApiKey;
+});
```
himanshusinghs, line 16: Perhaps we should expand describeAccuracyTests instead of this to accept partial config. That way we don't need to have the knowhow of how process variables comes into play.

src/tools/mongodb/create/createIndex.ts
```diff
@@ -1,33 +1,147 @@
...
+                        name,
+                    },
+                ]);
+                break;
+            case "vectorSearch":
+                {
+                    const isVectorSearchSupported = await this.session.isConnectedToMongot;
+                    if (!isVectorSearchSupported) {
+                        // TODO: remove hacky casts once we merge the local dev tools
+                        const isLocalAtlasAvailable =
```
kmruiz, line 119: We are checking if a tool exists in a few places already in other places, maybe we can extract this to a function and refactor?

src/common/connectionManager.ts
nirinchev: Yeah, we'll need to converge to a single solution once we merge everything.

src/common/session.ts
nirinchev: I wanted this to be similar to `isConnectedToMongoDB`, but not opposed to renaming it.

src/common/connectionManager.ts
```diff
@@ -32,9 +32,33 @@ export interface ConnectionState {
...
+            try {
+                const dummyDatabase = `search-index-test-db-${Date.now()}`;
```
kmruiz: We don't need to generate random database or collection names to test this and this open a few authorization problems, for example, when the user only has access to specific databases/collections through RBAC. Using a hardcoded database/collection should be enough and simpler, and it won't look weird in traces for customers monitoring queries.

src/tools/mongodb/create/createIndex.ts
nirinchev: There's no default in the server docs and it'll depend on the embedding model the user wants to use.

src/common/connectionManager.ts
kmruiz: If we prefer having the search support logic in the ConnectionManager, which is also fine, I can wait and merge my PR and fix the conflicts here. I personally decided to have it somewhere else because I didn't want to have the connection manager with lots of `isBlahSupported` whenever we need to support other things like federation, encryption, etc...

src/tools/mongodb/create/createIndex.ts
nirinchev: Sure, we can do that - I didn't spend too much time polishing this because the local tools PR is not merged in main yet.

tests/accuracy/createIndex.test.ts
nirinchev: Agree - right now the accuracy tests have no way of spawning a search-enabled instance or have us provide a config. I can take a look at enhancing the harness in another PR or feel free to pick that up since they're your "baby" :P

src/tools/mongodb/create/createIndex.ts
```diff
@@ -1,33 +1,147 @@
...
+                        numDimensions: z
+                            .number()
+                            .min(1)
+                            .max(8192)
+                            .describe(
+                                "Number of vector dimensions that MongoDB Vector Search enforces at index-time and query-time"
+                            ),
+                        similarity: z
+                            .enum(["cosine", "euclidean", "dotProduct"])
+                            .default("cosine")
```
kmruiz: Default is euclidean, as per the PD

src/tools/mongodb/create/createIndex.ts
kmruiz: In the PD we assume a default of 1024, it was validated already by Gaurab. The reason is that it's the default for all Voyage models: https://docs.voyageai.com/docs/embeddings

src/common/connectionManager.ts
nirinchev: Agreed - I copied that from a previous implementation, but I saw in your other PR that you're just using `test/test`. I think your PR has the most sophisticated checks so far, so I think we should consolidate around that when it merges into main.

src/tools/mongodb/create/createIndex.ts
nirinchev: Good catch - not sure where I got cosine as the default from 🤦‍♂️

src/common/session.ts
kmruiz: mongot is a pretty internal name, I think it's clearer if we use something like Himanshu mentioned.

src/tools/mongodb/create/createIndex.ts
kmruiz: No worries, it's a PITA to remember all these names haha.

src/common/connectionManager.ts
kmruiz: As your PR is smaller, if you want I can wait for yours to be merged and I can do the cleanup, it should be fine as it's already well designed and covered, the refactors should be relatively simple.

tests/accuracy/createIndex.test.ts
himanshusinghs: Sounds good. I will take that over in a different PR then.

PR #622, opened 2025-10-07
chore: allow providing Voyage AI API key from config and tests to ensure unusable tools are not registered MCP-233 MCP-237

himanshusinghs:
## Proposed changes
Addresses both MCP-233 and MCP-237.
Apparently we always had a way to not register given certain conditions (ToolBase.verifyAllowed), I just added a test to verify the behaviour.

## Checklist


src/server.ts
```diff
@@ -31,6 +31,7 @@ export interface ServerOptions {
     telemetry: Telemetry;
     elicitation: Elicitation;
     connectionErrorHandler: ConnectionErrorHandler;
+    toolConstructors?: (new (params: ToolConstructorParams) => ToolBase)[];
```
himanshusinghs, line 34: This is only to make it easy to test tool registration behaviour.

PR #624, opened 2025-10-08
chore: use redact by default

blva:
## Checklist


src/telemetry/telemetry.ts
```diff
@@ -52,8 +53,8 @@ export class Telemetry {
         } = {}
     ): Telemetry {
         const mergedProperties = {
-            ...MACHINE_METADATA,
-            ...commonProperties,
+            ...redact(MACHINE_METADATA, session.keychain.allSecrets),
```
ciprian-tibulca: instead of redacting in various places/props, what about doing it globally for all props just before we send the events, in `telementry.sendEvents`: https://github.com/mongodb-js/mongodb-mcp-server/blob/c12de89b1edd7f9123b100d78e985aa7d6c591d9/src/telemetry/telemetry.ts#L223

PR #626, opened 2025-10-08
chore: Add new session-level service for getting embeddings of a specific collection MCP-246

kmruiz:
## Checklist

### DIFF
--- a/src/common/search/vectorSearchEmbeddings.ts
+++ b/src/common/search/vectorSearchEmbeddings.ts
@@ -1,5 +1,5 @@
 import type { NodeDriverServiceProvider } from "@mongosh/service-provider-node-driver";
-import type { Document } from "bson";
+import { BSON, type Document } from "bson";
 
 type VectorFieldIndexDefinition = {
     type: "vector";
@@ -9,16 +9,12 @@ type VectorFieldIndexDefinition = {
     similarity: "euclidean" | "cosine" | "dotProduct";
 };
 
-type EmbeddingNamespace = "${string}.${string}";
+export type EmbeddingNamespace = `${string}.${string}`;
 export class VectorSearchEmbeddings {
-    private embeddings: Map<EmbeddingNamespace, VectorFieldIndexDefinition[]>;
-
-    constructor() {
-        this.embeddings = new Map();
-    }
+    constructor(private readonly embeddings: Map<EmbeddingNamespace, VectorFieldIndexDefinition[]> = new Map()) {}
 
     cleanupEmbeddingsForNamespace({ database, collection }: { database: string; collection: string }): void {
-        const embeddingDefKey = `${database}.${collection}` as EmbeddingNamespace;
+        const embeddingDefKey: EmbeddingNamespace = `${database}.${collection}`;
         this.embeddings.delete(embeddingDefKey);
     }
 
@@ -31,7 +27,7 @@ export class VectorSearchEmbeddings {
         collection: string;
         provider: NodeDriverServiceProvider;
     }): Promise<VectorFieldIndexDefinition[] | undefined> {
-        const embeddingDefKey = `${database}.${collection}` as EmbeddingNamespace;
+        const embeddingDefKey: EmbeddingNamespace = `${database}.${collection}`;
         const definition = this.embeddings.get(embeddingDefKey);
 
         if (!definition) {
@@ -49,7 +45,71 @@ export class VectorSearchEmbeddings {
         }
     }
 
-    isVectorFieldIndexDefinition(doc: Document): doc is VectorFieldIndexDefinition {
+    async findFieldsWithWrongEmbeddings(
+        {
+            database,
+            collection,
+            provider,
+        }: {
+            database: string;
+            collection: string;
+            provider: NodeDriverServiceProvider;
+        },
+        document: Document
+    ): Promise<VectorFieldIndexDefinition[]> {
+        const embeddings = await this.embeddingsForNamespace({ database, collection, provider });
+
+        if (!embeddings) {
+            return [];
+        }
+
+        return embeddings.filter((emb) => !this.documentPassesEmbeddingValidation(emb, document));
+    }
+
+    private isVectorFieldIndexDefinition(doc: Document): doc is VectorFieldIndexDefinition {
         return doc["type"] === "vector";
+    }
+
+    private documentPassesEmbeddingValidation(definition: VectorFieldIndexDefinition, document: Document): boolean {
+        const fieldPath = definition.path.split(".");
+        let fieldRef: unknown = document;
+
+        for (const field of fieldPath) {
+            if (fieldRef && typeof fieldRef === "object" && field in fieldRef) {
+                fieldRef = (fieldRef as Record<string, unknown>)[field];
+            } else {
+                return true;
+            }
+        }
+
+        switch (definition.quantization) {
+            case "none":
+            case "scalar":
+                if (!Array.isArray(fieldRef)) {
+                    return false;
+                }
+
+                if (fieldRef.length !== definition.numDimensions) {
+                    return false;
+                }
+
+                if (typeof fieldRef[0] !== "number") {
+                    return false;
+                }
+                break;
+            case "binary":
+                if (fieldRef instanceof BSON.Binary) {
+                    try {
+                        const bits = fieldRef.toBits();
+                        return bits.length === definition.numDimensions;
+                    } catch {
+                        return false;
+                    }
+                } else {
+                    return false;
+                }
+        }
+
+        return true;
     }
 }
--- a/src/common/search/vectorSearchEmbeddings.ts
+++ b/src/common/search/vectorSearchEmbeddings.ts
@@ -1,0 +1,55 @@
+import type { NodeDriverServiceProvider } from "@mongosh/service-provider-node-driver";
+import type { Document } from "bson";
+
+type VectorFieldIndexDefinition = {
+    type: "vector";
+    path: string;
+    numDimensions: number;
+    quantization: "none" | "scalar" | "binary";
+    similarity: "euclidean" | "cosine" | "dotProduct";
+};
+
+type EmbeddingNamespace = "${string}.${string}";
+export class VectorSearchEmbeddings {
+    private embeddings: Map<EmbeddingNamespace, VectorFieldIndexDefinition[]>;
+
+    constructor() {
+        this.embeddings = new Map();
+    }
+
+    cleanupEmbeddingsForNamespace({ database, collection }: { database: string; collection: string }): void {
+        const embeddingDefKey = `${database}.${collection}` as EmbeddingNamespace;
+        this.embeddings.delete(embeddingDefKey);
+    }
+
+    async embeddingsForNamespace({
+        database,
+        collection,
+        provider,
+    }: {
+        database: string;
+        collection: string;
+        provider: NodeDriverServiceProvider;
+    }): Promise<VectorFieldIndexDefinition[] | undefined> {
+        const embeddingDefKey = `${database}.${collection}` as EmbeddingNamespace;
+        const definition = this.embeddings.get(embeddingDefKey);
+
+        if (!definition) {
+            const allSearchIndexes = await provider.getSearchIndexes(database, collection);
+            const vectorSearchIndexes = allSearchIndexes.filter((index) => index.type === "vectorSearch");
+            const vectorFields = vectorSearchIndexes
+                // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+                .flatMap<Document>((index) => (index.latestDefinition?.fields as Document) ?? [])
+                .filter((field) => this.isVectorFieldIndexDefinition(field));
+
+            this.embeddings.set(embeddingDefKey, vectorFields);
+            return vectorFields;
+        } else {
+            return definition;
+        }
+    }
+
+    isVectorFieldIndexDefinition(doc: Document): doc is VectorFieldIndexDefinition {
+        return doc["type"] === "vector";
+    }
+}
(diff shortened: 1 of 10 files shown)


src/tools/mongodb/create/insertMany.ts
```diff
@@ -23,19 +23,42 @@ export class InsertManyTool extends MongoDBToolBase {
...
 
+        const embeddingValidations = new Set(
+            ...(await Promise.all(
+                documents.flatMap((document) =>
+                    this.session.vectorSearchEmbeddings.findFieldsWithWrongEmbeddings(
+                        { database, collection, provider },
+                        document
+                    )
+                )
+            ))
```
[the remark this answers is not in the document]
kmruiz: This is wrong, as the method findFieldsWithWrongEmbeddings already caches the definition.

src/common/search/vectorSearchEmbeddingsManager.ts
```diff
@@ -0,0 +1,169 @@
...
+            if (fieldRef && typeof fieldRef === "object" && field in fieldRef) {
+                fieldRef = (fieldRef as Record<string, unknown>)[field];
+            } else {
+                return true;
+            }
+        }
+
+        switch (definition.quantization) {
+            case "none":
+                return true;
```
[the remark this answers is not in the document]
kmruiz: As we are not implementing all the proper validations per the scope of the project, and this actual validation is wrong (not only arrays are valid, also different types of BinData) we will need to trust the user on this one.

src/tools/mongodb/create/insertMany.ts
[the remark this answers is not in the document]
kmruiz: This is not right, because flatMap already flattens the array of arrays in one single array.

src/common/search/vectorSearchEmbeddings.ts
```diff
@@ -0,0 +1,186 @@
...
+        return true;
+    }
+
+    private async canListAtlasSearchIndexes(provider: NodeDriverServiceProvider): Promise<boolean> {
+        try {
+            await provider.getSearchIndexes("test", "test");
+            return true;
+        } catch {
+            return false;
+        }
```
himanshusinghs: Should we consider that atlas local deployments might take time to boot up search management service. This might become a common case as soon as we provide local-atlas tools.

src/common/search/vectorSearchEmbeddingsManager.ts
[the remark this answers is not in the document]
himanshusinghs: Its worth adding this as a comment.

src/common/search/vectorSearchEmbeddings.ts
```diff
@@ -0,0 +1,186 @@
...
+            const vectorFields = vectorSearchIndexes
+                // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+                .flatMap<Document>((index) => (index.latestDefinition?.fields as Document) ?? [])
+                .filter((field) => this.isVectorFieldIndexDefinition(field));
+
+            this.embeddings.set(embeddingDefKey, vectorFields);
+            return vectorFields;
+        } else {
+            return definition;
+        }
```
himanshusinghs:
```suggestion
        }
        
        return definition;
```

src/common/search/vectorSearchEmbeddings.ts
```diff
@@ -0,0 +1,186 @@
...
+    quantization: "none" | "scalar" | "binary";
+    similarity: "euclidean" | "cosine" | "dotProduct";
+};
+
+export type EmbeddingNamespace = `${string}.${string}`;
+export class VectorSearchEmbeddings {
+    constructor(
+        private readonly config: UserConfig,
+        private readonly embeddings: Map<EmbeddingNamespace, VectorFieldIndexDefinition[]> = new Map(),
+        private readonly atlasSearchStatus: Map<string, boolean> = new Map()
```
himanshusinghs: We probably want to listen to connection changes (here or in Session) to clear the orphan entries.

src/tools/mongodb/create/insertMany.ts
```diff
@@ -23,19 +23,42 @@ export class InsertManyTool extends MongoDBToolBase {
         documents,
     }: ToolArgs<typeof this.argsShape>): Promise<CallToolResult> {
         const provider = await this.ensureConnected();
-        const result = await provider.insertMany(database, collection, documents);
 
+        const embeddingValidations = new Set(
+            ...(await Promise.all(
```
himanshusinghs, line 28: A small note about Promise here, the only reason `findFieldsWithWrongEmbeddings` returns a Promise is because of `isAtlasSearchAvailable` (directly / indirectly).

When we're already in a tool call, is there a possibility to assume that `isAtlasSearchAvailable` is already populated and work off from that?

Just to be clear - what we have is not a problem. I see a small scope of improvement because I had to re-look what async task `findFieldsWithWrongEmbeddings` was doing.

tests/integration/tools/mongodb/create/insertMany.test.ts
```diff
@@ -93,7 +101,115 @@ describeWithMongoDB("insertMany tool", (integration) => {
...
+
+    const match = content.match(/Inserted IDs:\s(.*)/);
+    const group = match?.[1];
+    return (
+        group
+            ?.split(",")
+            .map((e) => e.trim())
+            .map((e) => ObjectId.createFromHexString(e)) ?? []
+    );
+}
```
himanshusinghs, line 213: Should we move this to helpers instead? I noticed myself sometimes re-inventing the helpers even though they exist in the codebase but only because they were in specific test files.

src/common/search/vectorSearchEmbeddings.ts
kmruiz: getSearchIndexes works (returns empty) when the service is still not active but is registered. That's why in the tests we have to try to create an index to ensure everything is working properly.

In a real world scenario, where the cluster is usually started before the MCP connects, that shouldn't happen much, because it takes just a few seconds to boot everything. Do you feel otherwise?

src/common/search/vectorSearchEmbeddings.ts
kmruiz: Yeah my assumption is that it wasn't going to change a lot, but it's a good point, I'll fix it and listen to disconnect events to clear the embeddings cache.

src/tools/mongodb/create/insertMany.ts
kmruiz: Actually that's not true right? `findFieldsWithWrongEmbeddings` do query MongoDB if the information is not in the cache.

tests/integration/tools/mongodb/create/insertMany.test.ts
kmruiz: I'm not sure we are going to test for this specific pattern outside of insertMany right? It's the only tool that inserts data and returns a result with this format.

src/common/search/vectorSearchEmbeddings.ts
himanshusinghs: I think the local-atlas tool are supposed to allow creating local deployments on the fly and that's why I was thinking that it may take a while and in between there is a possibility for agent to query already.

src/common/search/vectorSearchEmbeddings.ts
himanshusinghs: I guess we wouldn't need to do it if we just accept Nikola's method as is.

tests/integration/tools/mongodb/create/insertMany.test.ts
himanshusinghs: That's a fair point.

src/common/search/vectorSearchEmbeddings.ts
kmruiz: Yeah but we will have issues with all the tools that require atlas Search, and in that scenario, maybe it's better to let the agent retry. I can change the algorithm to consider:

1. When Atlas Search is _available_ and _ready_.
2. When Atlas Search is _available_ but not _ready_.
3. When Atlas Search is _not available_

In that case we will need to specify a "basic timeout" of how much time to wait until it's ready when we know it is available. I can work on that.

src/common/search/vectorSearchEmbeddings.ts
kmruiz: Well the embeddings cache is not in the connection status, it's here, we will still need to listen, but yeah, it will be a bit different. I'll work on rewriting parts of this.

src/tools/mongodb/create/insertMany.ts
himanshusinghs: Ahh true! I was under wrong assumption that we should be able to pre populate the cache for embeddings but we wouldn't know until there is a request for that.

src/common/search/vectorSearchEmbeddings.ts
```diff
@@ -0,0 +1,176 @@
...
+export type VectorFieldIndexDefinition = {
+    type: "vector";
+    path: string;
+    numDimensions: number;
+    quantization: "none" | "scalar" | "binary";
+    similarity: "euclidean" | "cosine" | "dotProduct";
+};
+
+export type EmbeddingNamespace = `${string}.${string}`;
+export class VectorSearchEmbeddings {
```
nirinchev: Can this be `EmbeddingsManager` or something like this?

src/common/connectionManager.ts
```diff
@@ -25,39 +25,106 @@ export interface ConnectionSettings {
...
+        );
+
+        const checkPromise = new Promise<boolean>((resolve) => {
+            void this.doCheckSearchIndexIsAvailable(resolve);
+        });
+
+        return await Promise.race([checkPromise, timeoutPromise]);
+    }
+
+    private async doCheckSearchIndexIsAvailable(resolve: (result: boolean) => void): Promise<void> {
```
nirinchev: I'm unconvinced about this - it seems way too intrusive to be creating stuff in the user's cluster just to check if search is fully initialized. I think a better approach would be to assume search is available if `getSearchIndexes` doesn't error out. Then if the user asks to create an index and that fails, they're going to get an error message and react to it. This implementation feels overly protective for what should be a fairly rare case (someone creating a local deployment and immediately starting to create search indexes).

kmruiz: Fair, simplified!

src/common/search/vectorSearchEmbeddings.ts
kmruiz: Sure, renamed.

src/common/connectionManager.ts
```diff
@@ -41,23 +42,23 @@ export class ConnectionStateConnected implements ConnectionState {
         public connectedAtlasCluster?: AtlasClusterConnectionInfo
     ) {}
 
-    private _isSearchSupported?: boolean;
+    #isSearchSupported?: boolean;
```
himanshusinghs: Not against this change. But I should point out that this is the very first occurrence of an ES private class variable in this code base. If there is no strong reason to have compile time privacy then perhaps for consistency sake we should continue using TS private keywor?

src/common/search/vectorSearchEmbeddingsManager.ts
```diff
@@ -0,0 +1,176 @@
...
+        for (const field of fieldPath) {
+            if (fieldRef && typeof fieldRef === "object" && field in fieldRef) {
+                fieldRef = (fieldRef as Record<string, unknown>)[field];
+            } else {
+                return true;
+            }
+        }
+
+        switch (definition.quantization) {
+            // Because quantization is not defined by the use
```
himanshusinghs:
```suggestion
            // Because quantization is not defined by the user
```

himanshusinghs: Unless it was meant to be something else :)

tests/integration/tools/mongodb/create/insertMany.test.ts
```diff
@@ -93,7 +101,113 @@ describeWithMongoDB("insertMany tool", (integration) => {
...
+        });
+
+        afterEach(async () => {
+            await provider.dropCollection(integration.randomDbName(), "test");
+        });
+
+        it("inserts a document when the embedding is correct", async ({ signal }) => {
+            await createVectorSearchIndexAndWait(
+                provider,
+                integration.randomDbName(),
```
himanshusinghs, line 128: Just a headsup - we're using MCPServer's internal state here for test helpers. This should instead be done using MongoClient retrieved via `integration.mongoClient()` as that was the purpose of introducing that.

Earlier there was a problem in establishing connection via test MongoClient but that's fixed now. Or feel free to skip this, I am happy to take it up in my PR as I already moved a few helpers to use MongoClient instead of provider.

src/tools/mongodb/mongodbTool.ts
```diff
@@ -82,6 +86,30 @@ export abstract class MongoDBToolBase extends ToolBase {
...
+                case ErrorCodes.AtlasSearchNotAvailable:
+                    return {
+                        content: [
+                            {
+                                text: `The connected MongoDB deployment does support vector search indexes but they are not ready yet. Try again later.`,
+                                type: "text",
+                            },
+                        ],
+                        isError: true,
+                    };
```
himanshusinghs: I don't think we're throwing this error anywhere. Maybe we can remove this and the error itself?

himanshusinghs (changes requested):

src/common/connectionManager.ts
kmruiz: In this case it is because this state was designed as an immutable ADT, and using private fields make them not enumerable, so they don't show in json.stringify or when iterating the object.

src/tools/mongodb/mongodbTool.ts
kmruiz: Yeah that's a left over, I am still cleaning a few things from removing the search availability validation.

tests/integration/tools/mongodb/create/insertMany.test.ts
kmruiz: If you know what changes to make and you are fine doing the change in your PR, I'm fine too.

PR #627, opened 2025-10-09
feat: added atlas local deployment id to telemetry

jeroenvervaeke:
## Proposed changes
Added atlas local deployment ID to telemetry.

Jira: [\[MCP-160\] \[mongodb-mcp-server\] Update existing MongoDB telemetry](https://jira.mongodb.org/browse/MCP-160)

### DIFF
--- a/src/tools/atlasLocal/atlasLocalTool.ts
+++ b/src/tools/atlasLocal/atlasLocalTool.ts
@@ -39,7 +39,7 @@ please log a ticket here: https://github.com/mongodb-js/mongodb-mcp-server/issue
         return this.executeWithAtlasLocalClient(client, ...args);
     }
 
-    protected async lookupDeploymentIdAndAddToTelemetryMetadata(client: Client, containerId: string): Promise<void> {
+    protected async lookupDeploymentId(client: Client, containerId: string): Promise<void> {
         // Don't run if telemetry is disabled
         if (this.telemetry.isTelemetryEnabled()) {
             return;
--- a/src/tools/atlasLocal/create/createDeployment.ts
+++ b/src/tools/atlasLocal/create/createDeployment.ts
@@ -27,7 +27,7 @@ export class CreateDeploymentTool extends AtlasLocalToolBase {
         const deployment = await client.createDeployment(deploymentOptions);
 
         // Lookup the deployment id and add it to the telemetry metadata
-        await this.lookupDeploymentIdAndAddToTelemetryMetadata(client, deployment.containerId);
+        await this.lookupDeploymentId(client, deployment.containerId);
 
         return {
             content: [
--- a/src/tools/atlasLocal/connect/connectDeployment.ts
+++ b/src/tools/atlasLocal/connect/connectDeployment.ts
@@ -23,7 +23,7 @@ export class ConnectDeploymentTool extends AtlasLocalToolBase {
         await this.session.connectToMongoDB({ connectionString });
 
         // Lookup the deployment id and add it to the telemetry metadata
-        await this.lookupDeploymentIdAndAddToTelemetryMetadata(client, deploymentName);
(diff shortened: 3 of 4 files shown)


src/tools/atlasLocal/atlasLocalTool.ts
```diff
@@ -38,6 +39,18 @@ please log a ticket here: https://github.com/mongodb-js/mongodb-mcp-server/issue
         return this.executeWithAtlasLocalClient(client, ...args);
     }
 
+    protected async lookupDeploymentIdAndAddToTelemetryMetadata(client: Client, containerId: string): Promise<void> {
+        // Don't run if telemetry is disabled
+        if (this.config.telemetry === "disabled") {
```
blva: could we use `isTelemetryEnabled` instead since there are more ways of disabling telemetry

src/tools/atlasLocal/atlasLocalTool.ts
```diff
@@ -38,6 +39,18 @@ please log a ticket here: https://github.com/mongodb-js/mongodb-mcp-server/issue
         return this.executeWithAtlasLocalClient(client, ...args);
     }
 
+    protected async lookupDeploymentIdAndAddToTelemetryMetadata(client: Client, containerId: string): Promise<void> {
```
blva: nit - you're not really adding it to telemetry metadata, you're just setting the deploymentId right?

src/tools/atlasLocal/create/createDeployment.ts
```diff
@@ -26,6 +26,9 @@ export class CreateDeploymentTool extends AtlasLocalToolBase {
         // Create the deployment
         const deployment = await client.createDeployment(deploymentOptions);
 
+        // Lookup the deployment id and add it to the telemetry metadata
+        await this.lookupDeploymentIdAndAddToTelemetryMetadata(client, deployment.containerId);
```
blva: double-checking: do you need to lookup the deployment Id right after creating? I could that be in the deployment result for free?

src/tools/atlasLocal/atlasLocalTool.ts
jeroenvervaeke: This was the best name I could think of. I'm open to suggestions.

src/tools/atlasLocal/atlasLocalTool.ts
jeroenvervaeke: Thanks for the suggestion, I wasn't aware of this. It would be better if we only had one source of truth that we could access from the tools.

I'll update.

src/tools/atlasLocal/create/createDeployment.ts
jeroenvervaeke: Yes, we do. `deploymentId` is not part of `createDeployment`.

`createDeployments` returns all the info based on `docker inspect` (equivalent), getting the deployment ID requires us to query the MongoDB instance inside the container, which is something we don't do by default.

src/tools/atlasLocal/atlasLocalTool.ts
jeroenvervaeke: Updated

src/tools/atlasLocal/atlasLocalTool.ts
blva: `lookupDeploymentId`, `setDeploymentId`, `updateDeploymentId` - I think you're doing only one thing, no need for the `AndAdd...`

src/tools/atlasLocal/create/createDeployment.ts
blva: thanks!

src/tools/atlasLocal/atlasLocalTool.ts
jeroenvervaeke: Thanks, I went with `lookupDeploymentId`.

PR #628, opened 2025-10-09
chore: method to detect if a cluster support search indexes and expose the same on debug resource MCP-247 MCP-248

himanshusinghs:
## Proposed changes
Adds a small helper on session to check if search indexes are supported by the connected cluster and adds this information on the debug resource as well.

## Checklist

### DIFF
--- a/tests/unit/resources/common/debug.test.ts
+++ b/tests/unit/resources/common/debug.test.ts
@@ -34,7 +34,9 @@ describe("debug resource", () => {
         debugResource.reduceApply("connect", undefined);
         const output = await debugResource.toOutput();
 
-        expect(output).toContain(`The user is connected to the MongoDB cluster.`);
+        expect(output).toContain(
+            `The user is connected to the MongoDB cluster without any support for search indexes.`
+        );
     });
 
     it("should be disconnected when a disconnect event happens", async () => {


src/common/session.ts
```diff
@@ -153,4 +153,17 @@ export class Session extends EventEmitter<SessionEvents> {
...
+    async isSearchIndexSupported(): Promise<boolean> {
+        try {
+            const dummyDatabase = `db-${Date.now()}`;
+            const dummyCollection = `coll-${Date.now()}`;
+            // If a cluster supports search indexes, the call below will succeed
+            // with a cursor otherwise will throw an Error
+            await this.serviceProvider.getSearchIndexes(dummyDatabase, dummyCollection);
+            return true;
+        } catch {
+            return false;
```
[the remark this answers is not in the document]
himanshusinghs: While that's true, the usage of this method would still have to catch the possible error and consider a boolean value. So catching all simplifies that bit.

src/resources/common/debug.ts
```diff
@@ -56,13 +56,15 @@ export class DebugResource extends ReactiveResource<
...
-    toOutput(): string {
+    async toOutput(): Promise<string> {
         let result = "";
 
         switch (this.current.tag) {
-            case "connected":
-                result += "The user is connected to the MongoDB cluster.";
+            case "connected": {
+                const searchIndexesSupported = await this.session.isSearchIndexSupported();
+                result += `The user is connected to the MongoDB cluster${searchIndexesSupported ? " with support for search indexes" : ""}.`;
```
nirinchev: Should we not clarify that search indexes are not supported here? I don't imagine the model would guess that the absence of the text implies search is unsupported.

src/resources/common/debug.ts
```diff
@@ -56,13 +56,15 @@ export class DebugResource extends ReactiveResource<
...
 
-    toOutput(): string {
+    async toOutput(): Promise<string> {
         let result = "";
 
         switch (this.current.tag) {
-            case "connected":
-                result += "The user is connected to the MongoDB cluster.";
+            case "connected": {
+                const searchIndexesSupported = await this.session.isSearchIndexSupported();
```
gagik, line 64: I don't mind these additions but this seems like a bit of random information to include. What will be the deciding factor in what output we'll show here? I imagine there's a ton of info which could be relevant in that sense.

himanshusinghs: So far the idea is to have anything connection related that might help Agent debug a problem with tool calls / connectivity problems.

src/resources/common/debug.ts
himanshusinghs: Added a clarifying line for the unsupported case. Thanks for the suggestion 🙌

src/resources/common/debug.ts
himanshusinghs: In this case - we're expecting Agent to use debug resource at some point when its attempting to do search related work like creating search index. Although the tool call would respond with the appropriate error but there is a good chance of agent pulling debug resource information so we should have search related metadata here as well.

PR #632, opened 2025-10-09
feat: Add Atlas Local Tools  

Luke-Sanderson:
## Proposed changes
Jira: [MCP-40](https://jira.mongodb.org/browse/MCP-40)
## Checklist

### DIFF
--- a/tests/accuracy/listDeployments.test.ts
+++ b/tests/accuracy/listDeployments.test.ts
@@ -2,7 +2,7 @@ import { describeAccuracyTests } from "./sdk/describeAccuracyTests.js";
 
 describeAccuracyTests([
     {
-        prompt: "What MongoDB clusters do I have running?",
+        prompt: "What local MongoDB clusters do I have running?",
         expectedToolCalls: [
             {
                 toolName: "atlas-local-list-deployments",
@@ -11,7 +11,7 @@ describeAccuracyTests([
         ],
     },
     {
-        prompt: "What MongoDB instances do I have running?",
+        prompt: "What local MongoDB instances do I have running?",
         expectedToolCalls: [
             {
                 toolName: "atlas-local-list-deployments",
@@ -20,7 +20,7 @@ describeAccuracyTests([
         ],
     },
     {
-        prompt: "How many MongoDB clusters are running?",
+        prompt: "How many local MongoDB clusters are running?",
         expectedToolCalls: [
             {
                 toolName: "atlas-local-list-deployments",


vitest.config.ts
```diff
@@ -10,6 +10,10 @@ const vitestDefaultExcludes = [
     "**/{karma,rollup,webpack,vite,vitest,jest,ava,babel,nyc,cypress,tsup,build,eslint,prettier}.config.*",
 ];
 
+if (process.env.SKIP_ATLAS_LOCAL_TESTS === "true") {
+    vitestDefaultExcludes.push("**/atlas-local/**");
+}
```
nirinchev, line 19: [nit] Can we apply the same exclusion mechanism for the atlas tests. I know that those are magically excluded by virtue of not providing the API keys, but it's kind of annoying we have several different ways of doing things.

src/common/session.ts
```diff
@@ -99,6 +101,10 @@ export class Session extends EventEmitter<SessionEvents> {
         this.connectionManager.setClientName(this.mcpClient.name || "unknown");
     }
 
+    setAtlasLocalClient(atlasLocalClient: Client): void {
+        this.atlasLocalClient = atlasLocalClient;
+    }
```
nirinchev: `atlasLocalClient` is already a public field - this method seems redundant.

Luke-Sanderson:
@nirinchev @kmruiz @himanshusinghs All PRs to the feature branch has been reviewed but we would like someone from DevTools to review before merging to main please :)

src/tools/atlasLocal/connect/connectDeployment.ts
```diff
@@ -0,0 +1,37 @@
...
+        { deploymentName }: ToolArgs<typeof this.argsShape>
+    ): Promise<CallToolResult> {
+        // Get the connection string for the deployment
+        const connectionString = await client.getConnectionString(deploymentName);
+
+        // Connect to the deployment
+        await this.session.connectToMongoDB({ connectionString });
+
+        // Lookup the deployment id and add it to the telemetry metadata
+        await this.lookupDeploymentId(client, deploymentName);
```
nirinchev: This feels a bit awkward and error-prone. Can we move this to `resolveTelemetryMetadata` similarly to what we do in the atlas tools?

src/tools/atlasLocal/atlasLocalTool.ts
```diff
@@ -0,0 +1,88 @@
+import type { CallToolResult } from "@modelcontextprotocol/sdk/types.js";
+import type { TelemetryToolMetadata, ToolArgs, ToolCategory } from "../tool.js";
+import { ToolBase } from "../tool.js";
+import type { ToolCallback } from "@modelcontextprotocol/sdk/server/mcp.js";
+import type { Client } from "@mongodb-js-preview/atlas-local";
+
+export abstract class AtlasLocalToolBase extends ToolBase {
+    public category: ToolCategory = "atlas-local";
+    protected deploymentId?: string;
```
nirinchev: We shouldn't be storing state from previous invocations - it's very flimsy and a potential footgun.

src/tools/atlasLocal/create/createDeployment.ts
```diff
@@ -0,0 +1,41 @@
...
+    };
+
+    protected async executeWithAtlasLocalClient(
+        client: Client,
+        { deploymentName }: ToolArgs<typeof this.argsShape>
+    ): Promise<CallToolResult> {
+        const deploymentOptions: CreateDeploymentOptions = {
+            name: deploymentName,
+            creationSource: {
+                type: "MCPServer" as CreationSourceType,
```
nirinchev:
```suggestion
                type: CreationSourceType.MCPServer,
```

src/tools/atlasLocal/create/createDeployment.ts
```diff
@@ -0,0 +1,41 @@
...
+    public operationType: OperationType = "create";
+    protected argsShape = {
+        deploymentName: CommonArgs.string().describe("Name of the deployment to create").optional(),
+    };
+
+    protected async executeWithAtlasLocalClient(
+        client: Client,
+        { deploymentName }: ToolArgs<typeof this.argsShape>
+    ): Promise<CallToolResult> {
+        const deploymentOptions: CreateDeploymentOptions = {
```
nirinchev, line 19: We should pass `doNotTrack` (if it means what I think it means) equal to `isTelemetryEnabled`.

src/tools/atlasLocal/read/listDeployments.ts
```diff
@@ -0,0 +1,46 @@
...
+
+    protected async executeWithAtlasLocalClient(client: Client): Promise<CallToolResult> {
+        // List the deployments
+        const deployments = await client.listDeployments();
+
+        // Format the deployments
+        return this.formatDeploymentsTable(deployments);
+    }
+
+    private formatDeploymentsTable(deployments: Deployment[]): CallToolResult {
```
nirinchev, line 22: We should avoid these formattings - the agent ignores them anyway and formats the data however it wants. We should just return the raw json with potentially some hints for the LLM if we discover that the fields are not consistently interpreted correctly.

src/tools/atlasLocal/atlasLocalTool.ts
```diff
@@ -0,0 +1,88 @@
...
+
+    protected verifyAllowed(): boolean {
+        return this.session.atlasLocalClient !== undefined && super.verifyAllowed();
```
nirinchev, line 14:
```suggestion
        return this.session.atlasLocalClient && super.verifyAllowed();
```

src/tools/atlasLocal/atlasLocalTool.ts
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    protected async execute(...args: Parameters<ToolCallback<typeof this.argsShape>>): Promise<CallToolResult> {
+        // Get the client
```
nirinchev:
```suggestion
```

This comment is not adding much value.

src/tools/atlasLocal/atlasLocalTool.ts
```diff
@@ -0,0 +1,88 @@
...
+        const client = this.session.atlasLocalClient;
+
+        // If the client is not found, throw an error
+        // This should never happen:
+        // - atlas-local tools are only added after the client is set
+        //   this means that if we were unable to get the client, the tool will not be registered
+        // - in case the tool was registered by accident
+        //   verifyAllowed in the base class would still return false preventing the tool from being registered,
```
nirinchev: > verifyAllowed in the base class would still return false

The "base class" here implies `ToolBase` but you likely mean the verifyAllowed in this class.

src/server.ts
```diff
@@ -216,6 +220,45 @@ export class Server {
...
+            // Import Atlas Local client asyncronously
+            // This will fail on unsupported platforms
+            const { Client: AtlasLocalClient } = await import("@mongodb-js-preview/atlas-local");
+
+            // Connect to Atlas Local client
+            // This will fail if docker is not running
+            const client = AtlasLocalClient.connect();
+
+            // Set Atlas Local client
+            this.session.setAtlasLocalClient(client);
```
nirinchev: We should probably do this in `TransportRunnerBase` and cache the client rather than re-do it for every connection. Then we could pass down the client to the server at initialization time and register the tools all the same time.

src/server.ts
```diff
@@ -216,6 +220,45 @@ export class Server {
...
+                if (tool.register(this)) {
+                    this.tools.push(tool);
+                }
+            }
+        } catch (error) {
+            console.warn(
+                "Failed to initialize Atlas Local client, atlas-local tools will be disabled (error: ",
+                error,
+                ")"
+            );
```
nirinchev: We should probably handle the two expected errors - failure to import due to unsupported platform and docker not being available and emit more meaningful logs.

src/tools/atlasLocal/atlasLocalTool.ts
```diff
@@ -0,0 +1,88 @@
...
+    protected abstract executeWithAtlasLocalClient(
+        client: Client,
+        ...args: Parameters<ToolCallback<typeof this.argsShape>>
+    ): Promise<CallToolResult>;
+
+    protected handleError(
+        error: unknown,
+        args: ToolArgs<typeof this.argsShape>
+    ): Promise<CallToolResult> | CallToolResult {
+        // Error Handling for expected Atlas Local errors go here
```
nirinchev, line 82: We should also handle the case the docker daemon is not running. This can happen if I started the MCP server and at a later point, shut down docker.

tests/integration/helpers.ts
```diff
@@ -376,6 +377,37 @@ export function waitUntil<T extends ConnectionState>(
     });
 }
 
+export function waitUntilMcpClientIsSet(
```
nirinchev: This name is misleading - we're waiting until the atlas local client is available - not until the mcp client is. That being said, if we refactored the initialization of the local client, this is likely to be unnecessary.

src/tools/atlasLocal/create/createDeployment.ts
Luke-Sanderson: This was [suggested here too](https://github.com/mongodb-js/mongodb-mcp-server/pull/546#discussion_r2343467636). iirc this could not be done because of the way the napi creates a typescript binding from a rust enum.

nirinchev: This suggestion is different from the original one though. Here I'm suggesting using the enum as an enum as opposed to a string.

Luke-Sanderson: Ah right, yes in that case the issue is that the atlas-local dependency is optional so you can only import it as a type rather than directly

src/tools/atlasLocal/atlasLocalTool.ts
Luke-Sanderson:
```
src/tools/atlasLocal/atlasLocalTool.ts:12:9 - error TS2322: Type 'boolean | undefined' is not assignable to type 'boolean'.
  Type 'undefined' is not assignable to type 'boolean'.

12         return this.session.atlasLocalClient && super.verifyAllowed();
           ~~~~~~
```
Keeping `!== undefined` unless there is a reason for changing to a different check

src/tools/atlasLocal/create/createDeployment.ts
nirinchev: I see, this is super awkward and we should do something to resolve this. If importing the package directly is not allowed, we should ensure we have a lint rule for that to avoid accidental imports. Additionally, we should probably convert these enums into union types so that we don't need to import the package to use them.

src/tools/atlasLocal/atlasLocalTool.ts
nirinchev: Ah, true, that works seamlessly in if-checks, but not when returning. Fine with keeping it as is.

vitest.config.ts
Luke-Sanderson: Makes sense to be explicit, added

src/common/session.ts
Luke-Sanderson: Removed the method

src/tools/atlasLocal/atlasLocalTool.ts
Luke-Sanderson: Removed "base class" here

src/tools/atlasLocal/create/createDeployment.ts
Luke-Sanderson: Passed in `!isTelemetryEnabled`

src/tools/atlasLocal/atlasLocalTool.ts
Luke-Sanderson: Removed

tests/integration/helpers.ts
Luke-Sanderson: Changed to `waitUntilAtlasLocalClientIsSet`

src/tools/atlasLocal/create/createDeployment.ts
Luke-Sanderson: Ah yes string union type that is the name. We looked into using them since that is the ideal type but napi-rs doesn't support them from what we found

src/tools/atlasLocal/connect/connectDeployment.ts
Luke-Sanderson: refactored to be more similar to atlas tools

src/tools/atlasLocal/atlasLocalTool.ts
Luke-Sanderson: Removed. Only the create tool now stores the deploymentId so we can associate the tool call with the created deployment in the telemetry

src/tools/atlasLocal/atlasLocalTool.ts
Luke-Sanderson: Added

src/server.ts
Luke-Sanderson: Split the errors and added more meaningful logging

src/tools/atlasLocal/create/createDeployment.ts
nirinchev: It looks like it's supported, just not for enums. I briefly tested a few approaches and it seems like removing the rust enum and annotating the field with the options for the union type gives us the desirable typescript definitions. This obviously removes some type safety from the rust implementation, which we'll need to add, but I think it's worth it to ensure we have a cleaner public API that only exports types and no implementation. This is roughly what I had to change:

```rs
#[napi(object)]
#[derive(PartialEq, Debug)]
pub struct CreationSource {
  #[napi(js_name = "type", ts_type = "\"AtlasCLI\" | \"Container\" | \"MCPServer\" | \"Other\"")]
  pub source_type: String,
  pub source: String,
}
```

replaces the implementation in https://github.com/mongodb-js/atlas-local-lib-js/blob/a26dda2c78f9047f82e6c87ab7832026dd313884/src/models/list_deployments.rs#L75-L90. We'd need to then validate `source_type` is one of the permitted values, but that should be straightforward enough.

What this generated was the following types:

```ts
export interface CreationSource {
  type: "AtlasCLI" | "Container" | "MCPServer" | "Other"
  source: string
}
```

jeroenvervaeke: I've [update](https://github.com/mongodb-js/atlas-local-lib-js/commit/b7b73a0c515e0c96e696ee6bc8e8beda12f02511) the enum generation process in `atlas-local-lib-js`, so now the generated enums are string unions.

Example:
```ts
export type CreationSourceType =  'AtlasCLI'|
'Container'|
'MCPServer'|
'Other';
```

src/tools/atlasLocal/read/listDeployments.ts
jeroenvervaeke: Created a followup ticket: [\[MCP-268\] \[MCP\] Improve output of atlas and atlas-local tools](https://jira.mongodb.org/browse/MCP-268)

src/server.ts
jeroenvervaeke: I've moved this logic into `TransportRunnerBase`

src/common/atlasLocal.ts
```diff
@@ -0,0 +1,35 @@
+import type { Client } from "@mongodb-js/atlas-local";
+
+export type AtlasLocalClientFactoryFn = () => Promise<Client | undefined>;
+
+export const defaultCreateAtlasLocalClient: AtlasLocalClientFactoryFn = async () => {
+    try {
+        // Import Atlas Local client asyncronously
+        // This will fail on unsupported platforms
+        // also measure the time it takes to import the client
```
nirinchev: This doesn't appear to be the case anymore.

src/common/atlasLocal.ts
```diff
@@ -0,0 +1,35 @@
...
+        const { Client: AtlasLocalClient } = await import("@mongodb-js/atlas-local");
+
+        try {
+            // Connect to Atlas Local client
+            // This will fail if docker is not running
+            const client = AtlasLocalClient.connect();
+
+            // Set Atlas Local client
+            return client;
```
nirinchev:
```suggestion
            return AtlasLocalClient.connect();
```

src/tools/atlasLocal/atlasLocalTool.ts
```diff
@@ -0,0 +1,152 @@
...
+        }
+
+        return this.executeWithAtlasLocalClient(client, ...args);
+    }
+
+    protected async lookupDeploymentId(client: Client, containerId: string): Promise<string | undefined> {
+        try {
+            // Lookup and return the deployment id for telemetry metadata.
+            const deploymentId = await client.getDeploymentId(containerId);
+            return deploymentId;
```
nirinchev:
```suggestion
            return await client.getDeploymentId(containerId);
```

src/tools/atlasLocal/atlasLocalTool.ts
```diff
@@ -0,0 +1,152 @@
...
+        if (!parsedResult.success) {
+            this.session.logger.debug({
+                id: LogId.telemetryMetadataError,
+                context: "tool",
+                message: `Error parsing tool arguments: ${parsedResult.error.message}`,
+            });
+            return toolMetadata;
+        }
+
+        const resultDeploymentId = result._meta?.[AtlasLocalToolMetadataDeploymentIdKey];
```
nirinchev, line 126: Might be a good idea to add a comment explaining what's going on here.

src/tools/atlasLocal/atlasLocalTool.ts
```diff
@@ -0,0 +1,152 @@
...
+        // Extract deploymentName using type guard and lookup deployment ID
+        if (
+            resultDeploymentId === undefined &&
+            "deploymentName" in data &&
+            typeof data.deploymentName === "string" &&
+            data.deploymentName.trim() !== ""
+        ) {
+            const deploymentId = await this.lookupDeploymentId(client, data.deploymentName);
+            toolMetadata.atlasLocaldeploymentId = deploymentId;
+        }
```
nirinchev:
```suggestion
        } else if (
            "deploymentName" in parsedResult.data &&
            typeof parsedResult.data.deploymentName === "string" &&
            parsedResult.data.deploymentName.trim() !== ""
        ) {
            const deploymentId = await this.lookupDeploymentId(client, parsedResult.data.deploymentName);
            toolMetadata.atlasLocaldeploymentId = deploymentId;
        }
```

src/tools/tool.ts
```diff
@@ -277,25 +278,26 @@ export abstract class ToolBase {
...
+    ): TelemetryToolMetadata | Promise<TelemetryToolMetadata>;
 
     /**
      * Creates and emits a tool telemetry event
      * @param startTime - Start time in milliseconds
      * @param result - Whether the command succeeded or failed
      * @param args - The arguments passed to the tool
      */
-    private emitToolEvent(
+    private async emitToolEvent(
```
nirinchev: I'm not sure I like this change. `emitToolEvent` should be synchronous and never wait on async work. By making it async, we're delaying responding to users for the sake of collecting data, which is never great.

src/tools/atlasLocal/read/listDeployments.ts
nirinchev: While I think we should rework existing atlas tools, this shouldn't be a tied to the formatting of responses from new tools. Since the local tools are brand new, it's easier to just return json from them instead of merging this PR with table formatting only to rework it tomorrow.

src/tools/atlasLocal/create/createDeployment.ts
```diff
@@ -0,0 +1,45 @@
...
+                type: "MCPServer",
+                source: "MCPServer",
+            },
+            doNotTrack: !this.telemetry.isTelemetryEnabled(),
+        };
+        // Create the deployment
+        const deployment = await client.createDeployment(deploymentOptions);
+
+        // Capture deployment ID for telemetry
+        const createdDeploymentId = await this.lookupDeploymentId(client, deployment.containerId);
```
nirinchev: `resolveTelemetryMetadata` will already to this - why are we doing it here?

src/server.ts
```diff
@@ -221,7 +222,7 @@ export class Server {
     }
 
     private registerTools(): void {
-        for (const toolConstructor of this.toolConstructors) {
+        for (const toolConstructor of [...this.toolConstructors, ...AtlasLocalTools]) {
```
nirinchev: Why are we adding the atlas local tools here instead of in the ctor?

src/common/atlasLocal.ts
jeroenvervaeke: I've updated the comment (removed the last line)

src/common/atlasLocal.ts
jeroenvervaeke: Updated

src/tools/atlasLocal/atlasLocalTool.ts
jeroenvervaeke: Updated

src/tools/atlasLocal/atlasLocalTool.ts
jeroenvervaeke: Added comments

src/tools/atlasLocal/atlasLocalTool.ts
jeroenvervaeke: I did not accept this change, instead removed both branches and went with a solution closer to the original solution, but instead of setting local state the telemetry is returned in `_metadata`. This also allowed me to make `emitToolEvent` sync again.

src/tools/tool.ts
jeroenvervaeke: Updated, `emitToolEvent` is sync again

src/tools/atlasLocal/create/createDeployment.ts
jeroenvervaeke: Updated, `resolveTelemetryMetadata` is not doing this anymore

src/server.ts
jeroenvervaeke: Small oversight, this code has changed since the merge, I missed that. Now `AtlasLocalTools` is in the ctor

src/tools/atlasLocal/read/listDeployments.ts
jeroenvervaeke: Updated

src/tools/atlasLocal/read/listDeployments.ts
```diff
@@ -0,0 +1,44 @@
...
+        const deploymentsJson = deployments.map((deployment) => {
+            return {
+                name: deployment.name,
+                state: deployment.state,
+                mongodbVersion: deployment.mongodbVersion,
+            };
+        });
+
+        return {
+            content: formatUntrustedData("Deployments", JSON.stringify(deploymentsJson)),
```
nirinchev:
```suggestion
            content: formatUntrustedData(`Found ${deployments.length} deployments`, JSON.stringify(deploymentsJson)),
```

src/tools/atlasLocal/atlasLocalTool.ts
```diff
@@ -0,0 +1,155 @@
...
+                    },
+                ],
+                isError: true,
+            };
+        }
+
+        return this.executeWithAtlasLocalClient(client, ...args);
+    }
+
+    protected async lookupDeploymentId(client: Client, containerId: string): Promise<string | undefined> {
```
nirinchev:
```suggestion
    private async lookupDeploymentId(client: Client, containerId: string): Promise<string | undefined> {
```

This doesn't seem to be used outside of this file

src/tools/atlasLocal/atlasLocalTool.ts
```diff
@@ -0,0 +1,155 @@
...
+        const parsedResult = argsShape.safeParse(args[0]);
+
+        if (!parsedResult.success) {
+            this.session.logger.debug({
+                id: LogId.telemetryMetadataError,
+                context: "tool",
+                message: `Error parsing tool arguments: ${parsedResult.error.message}`,
+            });
+            return toolMetadata;
+        }
```
nirinchev:
```suggestion
```

We don't seem to need any of that now.

tests/accuracy/connectDeployment.test.ts
```diff
@@ -0,0 +1,70 @@
...
+        prompt: "If and only if, the local MongoDB deployment 'local-mflix' exists, then connect to it",
+        mockedTools: {
+            "atlas-local-list-deployments": (): CallToolResult => ({
+                content: [
+                    { type: "text", text: "Found 1 deployment:" },
+                    {
+                        type: "text",
+                        text: "Deployment Name | State | MongoDB Version\n----------------|----------------|----------------\nlocal-mflix | Running | 6.0",
+                    },
+                ],
```
nirinchev: This mock no longer matches the format returned by the list-deployments tool.

tests/accuracy/connectDeployment.test.ts
```diff
@@ -0,0 +1,70 @@
...
+        },
+        expectedToolCalls: [
+            {
+                toolName: "atlas-local-list-deployments",
+                parameters: {},
+            },
+            {
+                toolName: "atlas-local-connect-deployment",
+                parameters: {
+                    deploymentIdOrName: "local-mflix",
```
nirinchev: This argument has been renamed - I would expect the accuracy tests to be failing.

tests/accuracy/createDeployment.test.ts
```diff
@@ -0,0 +1,82 @@
...
+                parameters: {
+                    deploymentName: "new-database",
+                },
+            },
+        ],
+    },
+    {
+        prompt: "If and only if, the local MongoDB deployment 'existing-database' does not exist, then create it",
+        mockedTools: {
+            "atlas-local-list-deployments": (): CallToolResult => ({
```
nirinchev, line 65: This mock needs to be updated

tests/accuracy/deleteDeployment.test.ts
```diff
@@ -0,0 +1,102 @@
...
+                parameters: {
+                    deploymentName: "my-instance",
+                },
+            },
+        ],
+    },
+    {
+        prompt: "Delete all my local MongoDB instances",
+        mockedTools: {
+            "atlas-local-list-deployments": (): CallToolResult => ({
```
nirinchev, line 31: Mock needs updating (and below as well)

tests/accuracy/listDeployments.test.ts
```diff
@@ -0,0 +1,31 @@
+import { describeAccuracyTests } from "./sdk/describeAccuracyTests.js";
+
+describeAccuracyTests([
+    {
+        prompt: "What MongoDB clusters do I have running?",
```
nirinchev: Is this really calling the local tool here? I would expect it to call `atlas-list-clusters`.

tests/integration/tools/atlas-local/connectDeployment.test.ts
```diff
@@ -0,0 +1,105 @@
...
+
+    it("should connect to correct deployment when calling the tool", async () => {
+        // Connect to the deployment
+        const response = await integration.mcpClient().callTool({
+            name: "atlas-local-connect-deployment",
+            arguments: { deploymentName },
+        });
+        const elements = getResponseElements(response.content);
+        expect(elements.length).toBeGreaterThanOrEqual(1);
+        expect(elements[0]?.text).toContain(`Successfully connected to Atlas Local deployment "${deploymentName}".`);
```
nirinchev, line 79: Do we want to test that this is indeed connect - that is, we can call some data-plane tools and insert/read data from the deployment?

tests/integration/tools/atlas-local/connectDeployment.test.ts
```diff
@@ -0,0 +1,105 @@
...
+    defaultDriverOptions,
+    defaultTestConfig,
+    expectDefined,
+    getResponseElements,
+    setupIntegrationTest,
+    validateToolMetadata,
+} from "../../helpers.js";
+import { afterEach, describe, expect, it } from "vitest";
+
+const isMacOSInGitHubActions = process.platform === "darwin" && process.env.GITHUB_ACTIONS === "true";
```
nirinchev: Can we move this to a helper? Or even better, create a `describeWithAtlasLocal` similar to what we do for MongoDB and Atlas integration tests.

tests/integration/tools/atlas-local/connectDeployment.test.ts
```diff
@@ -0,0 +1,105 @@
...
+const integration = setupIntegrationTest(
```
nirinchev: This should be called in a `describe` since it sets up before/after hooks - they'll work outside, but produce super difficult to trace output if they fail.

tests/integration/tools/atlas-local/createDeployment.test.ts
```diff
@@ -0,0 +1,167 @@
...
+            }
+        }
+        deploymentNamesToCleanup = [];
+    });
+    const integration = setupIntegrationTest(
+        () => defaultTestConfig,
+        () => defaultDriverOptions
+    );
+
+    it.skipIf(isMacOSInGitHubActions)("should have the atlas-local-create-deployment tool", async () => {
```
nirinchev: These `skipIf`-s should probably go on the describe (or the helper if we create one).

tests/integration/tools/atlas-local/createDeployment.test.ts
```diff
@@ -0,0 +1,167 @@
...
+        });
+
+        // Check that deployment exists after creation
+        const afterResponse = await integration.mcpClient().callTool({
+            name: "atlas-local-list-deployments",
+            arguments: {},
+        });
+
+        const afterElements = getResponseElements(afterResponse.content);
+        expect(afterElements.length).toBeGreaterThanOrEqual(1);
```
nirinchev:
```suggestion
        expect(afterElements).toHaveLength(2);
```

Here we know for certain we should have at least 1 deployments, so 2 elements.

tests/integration/tools/atlas-local/createDeployment.test.ts
```diff
@@ -0,0 +1,167 @@
...
+                arguments: { deploymentName },
+            });
+
+            // Try to create the same deployment again
+            const response = await integration.mcpClient().callTool({
+                name: "atlas-local-create-deployment",
+                arguments: { deploymentName },
+            });
+            const elements = getResponseElements(response.content);
+            expect(elements.length).toBeGreaterThanOrEqual(1);
```
nirinchev: We could be more prescriptive about the expectation here as well. If this is throwing an error, it should have a single element, right?

tests/integration/tools/atlas-local/createDeployment.test.ts
```diff
@@ -0,0 +1,167 @@
...
+            // Create a deployment
+            const deploymentName = `test-deployment-${Date.now()}`;
+            deploymentNamesToCleanup.push(deploymentName);
+            await integration.mcpClient().callTool({
+                name: "atlas-local-create-deployment",
+                arguments: { deploymentName },
+            });
+
+            // Try to create the same deployment again
+            const response = await integration.mcpClient().callTool({
```
nirinchev: We should add an assertion that response.isError is true.

tests/integration/tools/atlas-local/deleteDeployment.test.ts
```diff
@@ -0,0 +1,92 @@
...
+        expect(beforeElements.length).toBeGreaterThanOrEqual(1);
+        expect(beforeElements[1]?.text ?? "").toContain(deploymentName);
+
+        // Delete the deployment
+        await integration.mcpClient().callTool({
+            name: "atlas-local-delete-deployment",
+            arguments: { deploymentName },
+        });
+
+        // Count the number of deployments after deleting the deployment
```
nirinchev: This comment doesn't appear to be correct.

tests/accuracy/connectDeployment.test.ts
jeroenvervaeke: Updated

tests/accuracy/connectDeployment.test.ts
jeroenvervaeke: Updated, accuracy tests are succeeding now: https://github.com/mongodb-js/mongodb-mcp-server/actions/runs/18680547550/job/53260492021

tests/accuracy/createDeployment.test.ts
jeroenvervaeke: Updated, accuracy tests are succeeding now: https://github.com/mongodb-js/mongodb-mcp-server/actions/runs/18680547550/job/53260492021

tests/accuracy/deleteDeployment.test.ts
jeroenvervaeke: Updated, accuracy tests are succeeding now: https://github.com/mongodb-js/mongodb-mcp-server/actions/runs/18680547550/job/53260492021

tests/accuracy/listDeployments.test.ts
jeroenvervaeke: Updated, accuracy tests are succeeding now: https://github.com/mongodb-js/mongodb-mcp-server/actions/runs/18680547550/job/53260492021

tests/integration/tools/atlas-local/createDeployment.test.ts
jeroenvervaeke: Added `describeWithAtlasLocal` and `describeWithAtlasLocalDisabled` helpers.

tests/integration/tools/atlas-local/connectDeployment.test.ts
jeroenvervaeke: Added test which inserts data to verify connection is working

tests/integration/tools/atlas-local/connectDeployment.test.ts
jeroenvervaeke: Added `describeWithAtlasLocal` and `describeWithAtlasLocalDisabled` helpers.

tests/integration/tools/atlas-local/createDeployment.test.ts
jeroenvervaeke: Updated

tests/integration/tools/atlas-local/connectDeployment.test.ts
jeroenvervaeke: Added `describeWithAtlasLocal` and `describeWithAtlasLocalDisabled` helpers.

tests/integration/tools/atlas-local/deleteDeployment.test.ts
jeroenvervaeke: Updated

package.json
```diff
@@ -122,6 +122,7 @@
     "node": "^20.19.0 || ^22.12.0 || >= 23.0.0"
   },
   "optionalDependencies": {
+    "@mongodb-js/atlas-local": "^1.0.2",
```
gagik, line 125: is this meant to be an optional dependency?

tests/integration/tools/atlas-local/createDeployment.test.ts
jeroenvervaeke: Added a comment

tests/integration/tools/atlas-local/createDeployment.test.ts
jeroenvervaeke: Updated

package.json
jeroenvervaeke: Yes it is, the import might not succeed on all platforms.

jeroenvervaeke:
> We should run the accuracy tests once we've fixed the mocks to check that they're indeed passing - I'd be very surprised if they do with the current wording of the prompts.

I ran the accuracy tests and all the `atlas-local` accuracy tests succeed with 100% success rate: https://github.com/mongodb-js/mongodb-mcp-server/actions/runs/18687565982

PR #642, opened 2025-10-13
feat: adds support for dropping vector search indexes in drop-index tool MCP-239

himanshusinghs:
## Proposed changes
Adds support for dropping vector search indexes in drop-index tool. The functionality is behind a feature flag.

- Tests have been added for basic tool functionality and elicitation
- listSearchIndexes tool and its tests have been modified to re-use some of the existing code.

## Checklist


src/tools/mongodb/delete/dropSearchIndex.ts
```diff
@@ -0,0 +1,79 @@
...
+    ): Promise<CallToolResult> | CallToolResult {
+        if (error instanceof Error && "codeName" in error && error.codeName === "SearchNotEnabled") {
+            return {
+                content: [
+                    {
+                        text: "This MongoDB cluster does not support Search Indexes. Make sure you are using an Atlas Cluster, either remotely in Atlas or using the Atlas Local image, or your cluster supports MongoDB Search.",
+                        type: "text",
+                        isError: true,
+                    },
+                ],
```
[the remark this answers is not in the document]
himanshusinghs: Fixed and added to listSearchIndexes tool as well.

tests/integration/tools/mongodb/delete/dropSearchIndex.test.ts
```diff
@@ -0,0 +1,154 @@
...
+                expect(content).toEqual(
+                    'Index with name "non-existent" does not exist in the provided namespace "any.foo".'
+                );
+            });
+        });
+
+        describe("when attempting to delete an existing index", () => {
+            let moviesCollection: Collection;
+            beforeEach(async ({ signal }) => {
+                await integration.connectMcpClient();
```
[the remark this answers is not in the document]
himanshusinghs: Removed

tests/integration/tools/mongodb/mongodbClusterProcess.ts
```diff
@@ -27,7 +27,8 @@ export class MongoDBClusterProcess {
 
             return new MongoDBClusterProcess(
                 () => runningContainer.stop(),
-                () => `mongodb://${runningContainer.getHost()}:${runningContainer.getMappedPort(27017)}`
+                () =>
+                    `mongodb://${runningContainer.getHost()}:${runningContainer.getMappedPort(27017)}/?directConnection=true`
```
kmruiz, line 32: Why do we need the direct connection here? It shouldn't be necessary.

tests/integration/helpers.ts
```diff
@@ -417,3 +421,110 @@ export function getDataFromUntrustedContent(content: string): string {
...
+        } catch (error) {
+            if (shouldRetryOnError && (await shouldRetryOnError(error))) {
+                await sleep(retryTimeout);
+                continue;
+            }
+            throw error;
+        }
+    }
+}
+
```
kmruiz: Can't this be replaced with [vi.waitFor](https://vitest.dev/api/vi.html#vi-waitfor)?

tests/integration/helpers.ts
```diff
@@ -417,3 +421,110 @@ export function getDataFromUntrustedContent(content: string): string {
...
+export async function waitUntilSearchManagementServiceIsReady(
```
kmruiz: Why aren't we using [waitUntilSearchIsReady](https://github.com/mongodb-js/mongodb-mcp-server/blob/main/tests/integration/tools/mongodb/mongodbHelpers.ts#L286)

kmruiz: Probably you should take a look at [mongodbHelpers.ts](https://github.com/mongodb-js/mongodb-mcp-server/blob/main/tests/integration/tools/mongodb/mongodbHelpers.ts) and see what can be reused and refactor.

src/helpers/searchErrorHandler.ts
```diff
@@ -0,0 +1,24 @@
+import type { CallToolResult } from "@modelcontextprotocol/sdk/types.js";
+import { type DbOperationArgs, MongoDBToolBase } from "../tools/mongodb/mongodbTool.js";
+import type { ToolArgs } from "../tools/tool.js";
+
+export abstract class MongoDBToolWithSearchErrorHandler extends MongoDBToolBase {
```
kmruiz: Not sure we want this. In https://github.com/mongodb-js/mongodb-mcp-server/pull/626 we are using the MongoDBToolBase, and it would be best to have all errors centralised in one single place.

tests/integration/tools/mongodb/mongodbClusterProcess.ts
himanshusinghs: Unfortunately it is. Without it the connection just keeps hanging because driver tries to resolve involved clusters and since the internal ports are never exposed the driver is unable to connect. BTW this is also what is recommended by the CLI when you kick start a deployment.

tests/integration/helpers.ts
himanshusinghs: It should be. I had a feeling that this existed, thanks for pointing it out.

src/helpers/searchErrorHandler.ts
himanshusinghs: I just now stumbled upon it while I was going through your PR. Certainly that's more preferable.

tests/integration/helpers.ts
himanshusinghs: These are the same methods. I just renamed / refactored them to re-use parts of the logic.

tests/integration/tools/mongodb/mongodbClusterProcess.ts
himanshusinghs: I see that in your PR you're using the service provider from server instance itself and because our connection code handles it internally (in generateConnectionInfoFromCliArgs) it works fine but for tests I think it would be better to continue using the test mongo client provided by the MdbIntegration.

tests/integration/helpers.ts
himanshusinghs: Replaced in [7db2402](https://github.com/mongodb-js/mongodb-mcp-server/pull/642/commits/7db2402464f6cc079d5bc097ab024945b629f28e)

src/tools/mongodb/delete/dropIndex.ts
```diff
@@ -35,6 +55,35 @@ export class DropIndexTool extends MongoDBToolBase {
...
+                    indexName,
+                    namespace: `${database}.${collection}`,
+                })
+            ),
+        };
+    }
+
     protected getConfirmationMessage({ database, collection, indexName }: ToolArgs<typeof this.argsShape>): string {
         return (
             `You are about to drop the \`${indexName}\` index from the \`${database}.${collection}\` namespace:\n\n` +
```
kmruiz: Shouldn't we add in the elicitation message if the index is a normal or a search one?

himanshusinghs: added in [ebe41fa](https://github.com/mongodb-js/mongodb-mcp-server/pull/642/commits/ebe41fa339f51d746d9abcab7c18646fc6010673)

PR #644, opened 2025-10-13
feat: introduces drop-index tool for regular index

himanshusinghs:
## Proposed changes
Adds a drop-index tool categorised under delete operation and marked as one requiring elicitation.
1. The tool has been marked to require user confirmation before execution.
2. Tests are added to confirm the negative and positive behaviour

## Checklist

### DIFF
--- a/src/tools/mongodb/delete/dropIndex.ts
+++ b/src/tools/mongodb/delete/dropIndex.ts
@@ -25,9 +25,10 @@ export class DropIndexTool extends MongoDBToolBase {
 
         return {
             content: formatUntrustedData(
-                `${result.ok ? "Successfully dropped" : "Failed to drop"} the index.`,
+                `${result.ok ? "Successfully dropped" : "Failed to drop"} the index from the provided namespace.`,
                 JSON.stringify({
                     indexName,
+                    namespace: `${database}.${collection}`,
                 })
             ),
             isError: result.ok ? undefined : true,
--- a/tests/integration/tools/mongodb/delete/dropIndex.test.ts
+++ b/tests/integration/tools/mongodb/delete/dropIndex.test.ts
@@ -106,9 +106,9 @@ describeWithMongoDB("drop-index tool", (integration) => {
             });
             expect(response.isError).toBe(undefined);
             const content = getResponseContent(response.content);
-            expect(content).toContain(`Successfully dropped the index.`);
+            expect(content).toContain(`Successfully dropped the index from the provided namespace.`);
             const data = getDataFromUntrustedContent(content);
-            expect(JSON.parse(data)).toMatchObject({ indexName });
+            expect(JSON.parse(data)).toMatchObject({ indexName, namespace: "mflix.movies" });
         });
     });
 });


src/tools/mongodb/delete/dropIndex.ts
```diff
@@ -0,0 +1,44 @@
...
+import { DbOperationArgs, MongoDBToolBase } from "../mongodbTool.js";
+import type { ToolArgs, OperationType } from "../../tool.js";
+import { CommonArgs } from "../../args.js";
+
+export class DropIndexTool extends MongoDBToolBase {
+    public name = "drop-index";
+    protected description = "Drop an index for the provided database and collection.";
+    protected argsShape = {
+        ...DbOperationArgs,
+        indexName: CommonArgs.string().nonempty().describe("The name of the index to be dropped."),
```
nirinchev: This doesn't seem right - index names can be unicode:

Image

src/tools/mongodb/delete/dropIndex.ts
```diff
@@ -0,0 +1,44 @@
...
+        const provider = await this.ensureConnected();
+        const result = await provider.runCommand(database, {
+            dropIndexes: collection,
+            index: indexName,
+        });
+
+        return {
+            content: [
+                {
+                    text: `${result.ok ? "Successfully dropped" : "Failed to drop"} the index with name "${indexName}" from the provided namespace "${database}.${collection}".`,
```
nirinchev: Should this be wrapped in untrusted content?

tests/integration/tools/mongodb/delete/dropIndex.test.ts
```diff
@@ -0,0 +1,143 @@
...
+    });
+
+    afterEach(async () => {
+        try {
+            await moviesCollection.dropIndex(indexName);
+        } catch (error) {
+            if (error instanceof Error && !error.message.includes("index not found with name")) {
+                throw error;
+            }
+        }
```
nirinchev: Why do this if we're dropping the collection anyway?

tests/integration/tools/mongodb/delete/dropIndex.test.ts
```diff
@@ -0,0 +1,143 @@
...
+            expect(response.isError).toBe(undefined);
+            const content = getResponseContent(response.content);
+            expect(content).toEqual(
+                `Successfully dropped the index with name "${indexName}" from the provided namespace "mflix.movies".`
+            );
+        });
+    });
+});
+
+describe("drop-search-index tool - when invoked via an elicitation enabled client", () => {
```
nirinchev:
```suggestion
describe("drop-index tool - when invoked via an elicitation enabled client", () => {
```

tests/integration/tools/mongodb/delete/dropIndex.test.ts
```diff
@@ -0,0 +1,143 @@
...
+    const mockElicitInput = createMockElicitInput();
+    const integration = setupIntegrationTest(
+        () => defaultTestConfig,
+        () => defaultDriverOptions,
+        { elicitInput: mockElicitInput }
+    );
+
+    it("should ask for confirmation before proceeding with tool call", async () => {
```
nirinchev, line 146: Should we add a test that if the request is rejected, we don't actually drop the index?

src/tools/mongodb/delete/dropIndex.ts
himanshusinghs: Interesting. Just noticed that rest of the tools don't enforce these checks. I do wonder though how practical are such index names?

tests/integration/tools/mongodb/delete/dropIndex.test.ts
himanshusinghs: Ahh I did not know that dropping the collection would drop the indexes as well :) I will make the change.

src/tools/mongodb/delete/dropIndex.ts
himanshusinghs: Yea makes sense. I will split the static header and dynamic data.

src/tools/mongodb/delete/dropIndex.ts
himanshusinghs: Removed the unicode check

src/tools/mongodb/delete/dropIndex.ts
himanshusinghs: Fixed this in recent commit.

tests/integration/tools/mongodb/delete/dropIndex.test.ts
himanshusinghs: Fixed this in recent commit.

tests/integration/tools/mongodb/delete/dropIndex.test.ts
himanshusinghs: Fixed this in recent commit.

tests/integration/tools/mongodb/delete/dropIndex.test.ts
himanshusinghs: Fixed this in recent commit.

PR #645, opened 2025-10-13
feat: atlas-get-performance-advisor tool: tweak language for slow queries

kylelai1:
## Proposed changes
From a product request: Adjusts the language when getting the slow queries for the performance advisor tool. Conveys the language to the LLM that the max limit for number of slow queries is due to the MCP server tool only (and not across Atlas).

## Checklist

Screenshot 2025-10-13 at 6 15 31 PM

### DIFF
--- a/src/tools/atlas/read/getPerformanceAdvisor.ts
+++ b/src/tools/atlas/read/getPerformanceAdvisor.ts
@@ -9,6 +9,8 @@ import {
     getSchemaAdvice,
     getSlowQueries,
     DEFAULT_SLOW_QUERY_LOGS_LIMIT,
+    SUGGESTED_INDEXES_COPY,
+    SLOW_QUERY_LOGS_COPY,
 } from "../../../common/atlas/performanceAdvisorUtils.js";
 import { AtlasArgs } from "../../args.js";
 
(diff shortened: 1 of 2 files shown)


src/tools/atlas/read/getPerformanceAdvisor.ts
```diff
@@ -102,7 +102,7 @@ export class GetPerformanceAdvisorTool extends AtlasToolBase {
                         : "No suggested indexes found."
                 }`,
                 `## Drop Index Suggestions\n${hasDropIndexSuggestions ? JSON.stringify(dropIndexSuggestionsResult.value) : "No drop index suggestions found."}`,
-                `## Slow Query Logs\n${hasSlowQueryLogs ? JSON.stringify(slowQueryLogsResult.value?.slowQueryLogs) : "No slow query logs found."}`,
+                `## Slow Query Logs\n${hasSlowQueryLogs ? `Please notify the user that the slow query logs are limited to the most recent ${DEFAULT_SLOW_QUERY_LOGS_LIMIT} slow query logs. This is a limitation of the MCP server tool only.\n${JSON.stringify(slowQueryLogsResult.value?.slowQueryLogs)}` : "No slow query logs found."}`,
```
nirinchev: q: should we add this clarification only if the number of query logs we've extracted is lower than that limit?  And should we include a CTA in the response? Something like "Please notify the user that the slow query logs are limited to the most recent ${DEFAULT_SLOW_QUERY_LOGS_LIMIT} slow query logs. This is done to optimize the performance of the MCP server - they can view the complete list of logs in the Atlas UI" or something similar?

kylelai1: Adding a CTA to the Atlas UI for specifically the performance advisor is difficult because we need to get the metrics route, which is not as simple as using a group ID and cluster ID. Would need to check with InTel on implementation here as well, since the metrics route is different for flex, dedicated, sharded, etc. PM has decided not to add this step.

I think we should keep the clarification for the limit of slow query logs always so the user is aware of the limit for other cases and not just the current tool call.

If we can't add a CTA, then maybe we can just add that last sentence referencing the Atlas UI? Instead of a specific link to the performance advisor for the cluster, maybe we can link to some docs about accessing the performance advisor in the Atlas UI? What do you think?

nirinchev: Yeah, that was what I meant by CTA - if we can get a link, that'd be great, but otherwise, just telling them they can see more logs if they go to their atlas UI is enough.

kylelai1: Added in a CTA for the tool response.

PR #650, opened 2025-10-14
chore: use describeWithMongoDB to setup test suites instead of custom wiring

himanshusinghs:
## Proposed changes
In some places we were doing the same wiring as the one that is already done in `describeWithMongoDB` and the reason for that was we needed some extra configuration to be passed down to `setupIntegrationTest` or to `setupMongoDBIntegrationTest`. This PR modifies `describeWithMongoDB` to accept the additional parameters and pass them down to `setupIntegrationTest` and `setupMongoDBIntegrationTest` and unify the usage in tests to use `describeWithMongoDB` instead of manual wiring.

The motivation is to have one place to have checks like skipping test on specific environments.

## Notes for reviewers
Review individual commits.
- The first commit is only a refactor, changes the describeWithMongoDB interface to accept test suite config instead of individual function parameters.
- The second commit adds elicitation related options to the interface of describeWithMongoDB and adapts elicitation tests to use describeWithMongoDB.

## Checklist

### DIFF
--- a/tests/integration/common/connectionManager.oidc.test.ts
+++ b/tests/integration/common/connectionManager.oidc.test.ts
@@ -3,12 +3,7 @@ import { describe, beforeEach, afterAll, it, expect, vi } from "vitest";
 import semver from "semver";
 import process from "process";
 import type { MongoDBIntegrationTestCase } from "../tools/mongodb/mongodbHelpers.js";
-import {
-    describeWithMongoDB,
-    isCommunityServer,
-    getServerVersion,
-    defaultTestSuiteConfig,
-} from "../tools/mongodb/mongodbHelpers.js";
+import { describeWithMongoDB, isCommunityServer, getServerVersion } from "../tools/mongodb/mongodbHelpers.js";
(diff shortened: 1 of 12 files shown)


tests/integration/common/connectionManager.oidc.test.ts
```diff
@@ -137,14 +142,20 @@ describe.skipIf(process.platform !== "linux")("ConnectionManager OIDC Tests", as
...
-            () => oidcConfig,
-            () => ({
-                ...setupDriverConfig({
-                    config: oidcConfig,
-                    defaults: {},
-                }),
-            }),
-            { runner: true, downloadOptions: { enterprise: true, version: mongodbVersion }, serverArgs }
+            {
+                ...defaultTestSuiteConfig,
```
nirinchev: Rather than pass the default test config always, what we can do is have the config argument be `Partial<TestSuiteConfig>` and have consumers only provide what they want to customize. Then when we run it we merge the default with the customized and call the corresponding API.

PR #653, opened 2025-10-15
chore: update atlas tools output to json - MCP-264

blva:
## Proposed changes
- updates the atlas tool responses to return json instead of tables

## Checklist

### DIFF
--- a/src/tools/atlas/read/listAlerts.ts
+++ b/src/tools/atlas/read/listAlerts.ts
@@ -31,8 +31,8 @@ export class ListAlertsTool extends AtlasToolBase {
         const alerts = data.results.map((alert) => ({
             id: alert.id,
             status: alert.status,
-            created: alert.created ? new Date(alert.created).toISOString() : null,
-            updated: alert.updated ? new Date(alert.updated).toISOString() : null,
+            created: alert.created ? new Date(alert.created).toISOString() : "N/A",
+            updated: alert.updated ? new Date(alert.updated).toISOString() : "N/A",
             eventTypeName: alert.eventTypeName,
             acknowledgementComment: alert.acknowledgementComment ?? "N/A",
         }));


src/tools/atlas/read/inspectCluster.ts
```diff
@@ -25,13 +25,17 @@ export class InspectClusterTool extends AtlasToolBase {
     }
 
     private formatOutput(formattedCluster: Cluster): CallToolResult {
+        const clusterDetails = {
+            name: formattedCluster.name || "Unknown",
```
jeroenvervaeke, line 29: In other commands we return `null`/`undefined`, could we do that for all tools?

src/tools/atlas/read/listAlerts.ts
```diff
@@ -28,22 +28,20 @@ export class ListAlertsTool extends AtlasToolBase {
...
-                    return `${alert.id} | ${alert.status} | ${created} | ${updated} | ${alert.eventTypeName} | ${comment}`;
-                })
-                .join("\n");
+        const alerts = data.results.map((alert) => ({
+            id: alert.id,
+            status: alert.status,
+            created: alert.created ? new Date(alert.created).toISOString() : null,
+            updated: alert.updated ? new Date(alert.updated).toISOString() : null,
+            eventTypeName: alert.eventTypeName,
+            acknowledgementComment: alert.acknowledgementComment ?? "N/A",
```
jeroenvervaeke, line 37: Same remark, can we avoid mixing `null` and `N/A`?

src/tools/atlas/read/inspectCluster.ts
gagik: That's an interesting point. having these human-readable make it easier to parse from both human and LLM perspective I'd think. If there's both `null` and `undefined` cases which mean different things then it's worth keeping or handling them differently but otherwise I don't have strong feelings about it.

blva: ack, I'll keep "N/A" like the previous implementation everywhere then! thanks for flagging

src/tools/atlas/read/inspectCluster.ts
```diff
@@ -25,13 +25,17 @@ export class InspectClusterTool extends AtlasToolBase {
...
+            instanceType: formattedCluster.instanceType,
+            instanceSize: formattedCluster.instanceSize || "N/A",
```
himanshusinghs, line 31: Not sure if this is supposed to be a number but if yes then you might wanna use `??` instead of `||` because the latter works on any falsy values, including 0.

src/tools/atlas/read/inspectCluster.ts
fmenezes: for context on the initial atlas tools I've added N/A as it was a better fit for markdown tables, for json _I personally think_ `null` would make more sense (or omitting the field)

fmenezes:
Everything seems fine, one side note why not add `structuredContent` field? (ref https://modelcontextprotocol.io/specification/2025-06-18/server/tools#structured-content)

src/tools/atlas/read/inspectCluster.ts
blva: thanks for flagging but it's text

src/tools/atlas/read/inspectCluster.ts
blva: I think I can group this feedback and `structuredContent` into upcoming changes and get rid of the tables for now, so i'll merge

PR #656, opened 2025-10-16
chore: disable code health runs on forks and add dependabot cooldown

himanshusinghs:
## Proposed changes
This PR adds a 7 days of cooldown for dependabot before pulling updates for npm dependencies.

Additionally we are disabling code health checks on forks.

## Checklist

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -8,9 +8,6 @@ updates:
       default-days: 7
       include:
         - "*"
-      exclude:
-        - "@mongodb-js/*"
-        - "@mongosh/*"
     ignore:
       # We are ignoring major updates on yargs-parser because yargs-parser@22
       # does not play nicely when bundled using webpack. Our VSCode extension


.github/dependabot.yml
```diff
@@ -4,6 +4,13 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
+    cooldown:
+      default-days: 7
+      include:
+        - "*"
+      exclude:
+        - "@mongodb-js/*"
```
kmruiz: These dependencies are not safe either as they don't have the cooldown. We should not make any exclusion.

himanshusinghs: The assumption (while it does not hold right now) is that they will also have a cooldown in their dependabot config.

If you feel strongly about it I am happy to include them. It was more of a convenience to have internal package PRs always ready.

himanshusinghs: Based on our discussion I have removed the exclusion for now. If this ever becomes a bother, we can add exclusion later.

PR #662, opened 2025-10-16
chore: When querying with vectorSearch use the generated embeddings MCP-245

kmruiz:
## Proposed changes

This PR introduces the `EmbeddingsProvider`, a new component responsible for generating embeddings by accessing external APIs like VoyageAI. This provider is designed for exclusive use within the `VectorSearchEmbeddingsManager`. While it currently only supports Voyage models, it is structured to make adding new models straightforward.

It exposes contracts as Zod schemas, which can be used within the tool schema. This approach creates a single source of truth for our contracts that is also useful for agents. Quantization is not yet supported, but users can provide their required data types. By default, the provider uses a _float_ data type and _none_ for quantization.

## Checklist

### DIFF
--- a/src/tools/mongodb/read/aggregate.ts
+++ b/src/tools/mongodb/read/aggregate.ts
@@ -174,10 +174,15 @@ export class AggregateTool extends MongoDBToolBase {
         }
 
         for (const stage of pipeline) {
+            // This validates that in readOnly mode or "write" operations are disabled, we can't use $out or $merge.
+            // This is really important because aggregates are the only "multi-faceted" tool in the MQL, where you
+            // can both read and write.
             if ((stage.$out || stage.$merge) && writeStageForbiddenError) {
                 throw new MongoDBError(ErrorCodes.ForbiddenWriteOperation, writeStageForbiddenError);
             }
 
+            // This ensure that you can't use $vectorSearch if the cluster does not support MongoDB Search
+            // either in Atlas or in a local cluster.
             if (stage.$vectorSearch && !isSearchSupported) {
                 throw new MongoDBError(
                     ErrorCodes.AtlasSearchNotSupported,


package.json
```diff
@@ -103,6 +102,7 @@
     "@mongodb-js/devtools-proxy-support": "^0.5.3",
     "@mongosh/arg-parser": "^3.19.0",
     "@mongosh/service-provider-node-driver": "^3.17.0",
+    "ai": "^5.0.72",
```
gagik, line 106: is there a reason ai is a production dependency now?

kmruiz: Yes, it is used to call the voyage API, it's in the `VoyageEmbeddingsProvider`. The voyage-ai-provider is used through the `ai` sdk, not directly.

gagik: ahh, makes sense

src/common/search/embeddingsProvider.ts
```diff
@@ -0,0 +1,84 @@
...
+});
+
+type VoyageModels = z.infer<typeof zVoyageModels>;
+type VoyageEmbeddingParameters = z.infer<typeof zVoyageEmbeddingParameters> & EmbeddingParameters;
+
+class VoyageEmbeddingsProvider implements EmbeddingsProvider<VoyageModels, VoyageEmbeddingParameters> {
+    private readonly voyage: VoyageProvider;
+
+    constructor({ voyageApiKey }: UserConfig, providedFetch?: typeof fetch) {
+        assert(voyageApiKey, "voyageApiKey does not exist. This is likely a bug.");
```
gagik: Maybe `Voyage API key` since there's multiple ways to pass this (similar to how we do for connection string)

src/common/search/vectorSearchEmbeddingsManager.ts
```diff
@@ -216,6 +219,51 @@ export class VectorSearchEmbeddingsManager {
...
+        inputType,
+    }: {
+        database: string;
+        collection: string;
+        path: string;
+        rawValues: string[];
+        embeddingParameters: SupportedEmbeddingParameters;
+        inputType: EmbeddingParameters["inputType"];
+    }): Promise<unknown[]> {
+        const provider = await this.assertAtlasSearchIsAvailable();
```
gagik: I realize this is from old code but this function isn't asserting that atlas search is available, right? it's just retrieving it, we're asserting it after the fact.

src/tools/mongodb/read/aggregate.ts
```diff
@@ -118,14 +173,17 @@ export class AggregateTool extends MongoDBToolBase {
                 "When 'create', 'update', or 'delete' operations are disabled, you can not run pipelines with $out or $merge stages.";
         }
 
-        if (!writeStageForbiddenError) {
-            return;
-        }
-
         for (const stage of pipeline) {
-            if (stage.$out || stage.$merge) {
```
gagik, line 126: probably worth having a comment here as it's hard to figure out the reasoning for this case.
Shouldn't we always throw if there's `writeStageForbiddenError`? I'm lacking some context on the expectations here.

src/common/search/embeddingsProvider.ts
kmruiz: Sounds good! Changing it rn!

src/common/search/vectorSearchEmbeddingsManager.ts
kmruiz: Yeah the name is a bit misleading, I'll change it.

src/tools/mongodb/read/aggregate.ts
kmruiz: So the idea is here is that we validate 2 scenarios:

1. The agent tries to use $out and $merge in readOnly mode, where writeStageForbiddenError != null
2. The agent tries to use $vectorSearch when Search is not available, where we throw an error.

gagik: got it, thanks! might be worth documenting that as a comment for future reference

gagik (approved):

src/tools/mongodb/read/aggregate.ts
kmruiz: Doing it right now!

src/common/search/embeddingsProvider.ts
```diff
@@ -0,0 +1,87 @@
...
+        });
+
+        return embeddings;
+    }
+}
+
+export function getEmbeddingsProvider(
+    userConfig: UserConfig
+): EmbeddingsProvider<VoyageModels, VoyageEmbeddingParameters> | undefined {
+    if (VoyageEmbeddingsProvider.isConfiguredIn(userConfig)) {
```
gagik, line 79: is this function / assertion not redundant? we do the same check inside constructor

gagik: ah, I see how this is avoiding this error. Then maybe we can make the arguments of the constructor just { voyageApiKey: string } and not userConfig and pass it here after this check. That way we can avoid having an invalid state. I suppose we can still assert it's not an empty string.


Might be easier to parse if we drop this function and just write the `!!userConfig.voyageApiKey` here

kmruiz: Yes, but it's on purpose.

The idea is that getEmbeddingProvider, depending on the configuration, can return a different EmbeddingProvider (for example, voyage or azure foundry or cohere). Having the same assertion in the constructor makes sure that nobody instantiates it wrongly if the configuration is not set up.

Does it make sense?

gagik: got it, yeah no strong feelings there if there's reasoning for it

PR #673, opened 2025-10-21
chore: Support atlas connect via private and private endpoint connection strings

cveticm:
## Proposed changes
Adds new optional input for `atlas-connect-cluster` `connectionType`. This allows users to connect via standard, private, or private endpoint connection strings.

### Manual Testing
Although I was unable to successfully set up a PE to test the connection in a timely manner, I have verified that the PE connection string obtained in this tool matches that which is provided when connection via the Atlas UI.
`mongodb+srv://<user>:<password>@<cluster-name>-pl-<...>.mongodb-dev.net/?authSource=admin`

_JIRA ticket_: [MCP-259](https://jira.mongodb.org/browse/MCP-259)

## Checklist

### DIFF
--- a/src/tools/atlas/read/inspectCluster.ts
+++ b/src/tools/atlas/read/inspectCluster.ts
@@ -28,7 +28,7 @@ export class InspectClusterTool extends AtlasToolBase {
         return {
             content: formatUntrustedData(
                 "Cluster details:",
-                `Cluster Name | Cluster Type | Tier | State | MongoDB Version | Standard Connection String
+                `Cluster Name | Cluster Type | Tier | State | MongoDB Version | Connection String
 ----------------|----------------|----------------|----------------|----------------|----------------
 ${formattedCluster.name || "Unknown"} | ${formattedCluster.instanceType} | ${formattedCluster.instanceSize || "N/A"} | ${formattedCluster.state || "UNKNOWN"} | ${formattedCluster.mongoDBVersion || "N/A"} | ${formattedCluster.connectionStrings?.standardSrv || formattedCluster.connectionStrings?.standard || "N/A"}`
             ),


src/tools/atlas/connect/connectCluster.ts
```diff
@@ -69,12 +72,17 @@ export class ConnectClusterTool extends AtlasToolBase {
...
         const cluster = await inspectCluster(this.session.apiClient, projectId, clusterName);
 
-        if (!cluster.connectionString) {
-            throw new Error("Connection string not available");
+        if (cluster.connectionStrings === undefined) {
+            throw new Error("Connection strings not available");
+        }
+        const connectionString = getConnectionString(cluster.connectionStrings, connectionType);
+        if (connectionString === undefined) {
+            throw new Error(`Connection string for connection type "${connectionType}" not available`);
```
gagik: is there something we could instruct or recommend them / the model to do in this case?

src/tools/atlas/read/inspectCluster.ts
```diff
@@ -30,7 +30,7 @@ export class InspectClusterTool extends AtlasToolBase {
                 "Cluster details:",
                 `Cluster Name | Cluster Type | Tier | State | MongoDB Version | Connection String
 ----------------|----------------|----------------|----------------|----------------|----------------
-${formattedCluster.name || "Unknown"} | ${formattedCluster.instanceType} | ${formattedCluster.instanceSize || "N/A"} | ${formattedCluster.state || "UNKNOWN"} | ${formattedCluster.mongoDBVersion || "N/A"} | ${formattedCluster.connectionString || "N/A"}`
+${formattedCluster.name || "Unknown"} | ${formattedCluster.instanceType} | ${formattedCluster.instanceSize || "N/A"} | ${formattedCluster.state || "UNKNOWN"} | ${formattedCluster.mongoDBVersion || "N/A"} | ${formattedCluster.connectionStrings?.standardSrv || formattedCluster.connectionStrings?.standard || "N/A"}`
```
gagik, line 33: is there a case where a private connection string would be relevant here?

cveticm: I'd be reluctant to add it to the table to keep it more light weight and more user friendly to users who are not yet familiar with private connections. If there is future customer desire for it, then I would reconsider.

But I will change the column title to explicit this is the standard connection string for clarity.

src/tools/atlas/connect/connectCluster.ts
cveticm: Good point. I'll expand the error message to point to atlas docs on connecting to clusters

src/tools/atlas/read/inspectCluster.ts
gagik: Fair enough, I think it's fine to leave this to be "Connection string". "Standard connection string" is probably confusing for most users

cveticm: Reverting to `Connection String` 👍

PR #675, opened 2025-10-22
chore: more accuracy test tweaks

nirinchev:
## Proposed changes

This tweaks a few more accuracy tests so that the expectations are more loosely defined and don't flag valid tool calls as inaccurate.


blva (approved):
LGTM but would be good to trigger the accuracy test in the branch before merging!

nirinchev:
Those changes don't fix all the false negatives, but I did run them locally and improved the accuracy of some tests. The annoying part is that those are quite unpredictable so we need multiple runs to identify the flakes.

PR #679, opened 2025-10-23
chore: add MCP registry publishing

gagik:
According to instructions in https://github.com/modelcontextprotocol/registry/blob/main/docs/guides/publishing/publish-server.md

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -15,11 +15,6 @@ jobs:
       RELEASE_CHANNEL: ${{ steps.npm-tag.outputs.RELEASE_CHANNEL }}
     steps:
       - uses: GitHubSecurityLab/actions-permissions/monitor@v1
-      - uses: mongodb-js/devtools-shared/actions/setup-bot-token@main
-        id: app-token
-        with:
-          app-id: ${{ vars.DEVTOOLS_BOT_APP_ID }}
-          private-key: ${{ secrets.DEVTOOLS_BOT_PRIVATE_KEY }}
       - uses: actions/checkout@v5
         with:
           fetch-depth: 0
@@ -85,6 +80,11 @@ jobs:
     if: needs.check.outputs.VERSION_EXISTS == 'false'
     steps:
       - uses: GitHubSecurityLab/actions-permissions/monitor@v1
+      - uses: mongodb-js/devtools-shared/actions/setup-bot-token@main
+        id: app-token
+        with:
+          app-id: ${{ vars.DEVTOOLS_BOT_APP_ID }}
+          private-key: ${{ secrets.DEVTOOLS_BOT_PRIVATE_KEY }}
       - uses: actions/checkout@v5
       - uses: actions/setup-node@v6
         with:
--- a/src/common/config.ts
+++ b/src/common/config.ts
@@ -226,7 +226,7 @@ export const UserConfigSchema = z.object({
         .record(z.string())
         .default({})
         .describe(
-            "Custom HTTP headers to include in responses from the HTTP server (only used when transport is 'http'). Useful for adding CORS headers, authentication tokens, or other custom headers required by your client application."
+            "Header that the HTTP server will validate when making requests (only used when transport is 'http')."
         ),
     idleTimeoutMs: z
         .number()


scripts/generateArguments.ts
```diff
@@ -0,0 +1,289 @@
...
+        for (const pkg of serverJson.packages) {
+            pkg.environmentVariables = envVarsArray as EnvironmentVariable[];
+            pkg.packageArguments = packageArguments;
+            pkg.version = version;
+
+            // Update OCI identifier version tag if this is an OCI package
+            if (pkg.registryType === "oci" && pkg.identifier) {
+                // Replace the version tag in the OCI identifier (e.g., docker.io/mongodb/mongodb-mcp-server:1.0.0)
+                pkg.identifier = pkg.identifier.replace(/:[^:]+$/, `:${version}`);
+            }
```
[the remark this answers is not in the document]
gagik: I don't think this matters for a script

.github/workflows/publish.yml
```diff
@@ -95,6 +96,22 @@ jobs:
...
+        run: |
+          VERSION="${{ needs.check.outputs.VERSION }}"
+          VERSION="${VERSION#v}"
+          npm run generate:arguments
+
+      - name: Install MCP Publisher
+        run: brew install mcp-publisher
+
+      - name: Login to MCP Registry
+        run: mcp-publisher login github-oidc
```
gagik: I am not too sure if this is going to work but seems worth trying

scripts/generateArguments.ts
```diff
@@ -0,0 +1,289 @@
...
+        result[key] = {
+            description,
+            defaultValue,
+        };
+    }
+
+    return result;
+}
+
+function parseOptionsFromConfig(): ParsedOptions {
```
gagik: need to remove this...

.github/workflows/publish.yml
```diff
@@ -95,6 +96,22 @@ jobs:
...
           NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+
+      - name: Update server.json version and arguments
+        run: |
+          VERSION="${{ needs.check.outputs.VERSION }}"
+          VERSION="${VERSION#v}"
+          npm run generate:arguments
+
+      - name: Install MCP Publisher
+        run: brew install mcp-publisher
```
nirinchev: This runs on `ubuntu-latest`, so I don't expect homebrew to work.

server.json
```diff
@@ -0,0 +1,670 @@
+{
+  "$schema": "https://static.modelcontextprotocol.io/schemas/2025-10-17/server.schema.json",
+  "name": "io.github.mongodb-js/mongodb-mcp-server",
```
gagik, line 3: we can either use `io.github.mongodb-js` and use GitHub Auth or `mongodb.com` and use DNS auth. the latter depends on a lot of external approvals I imagine so sticking to GitHub

scripts/generateArguments.ts
```diff
@@ -0,0 +1,247 @@
+#!/usr/bin/env tsx
+
+/**
+ * This script generates environment variable definitions and updates:
+ * - server.json environmentVariables arrays
+ * - TODO: README.md configuration table
```
gagik, line 6: this needed some extra handling but will be great to have in the future.

src/common/config.ts
```diff
@@ -157,38 +159,118 @@ function isConnectionSpecifier(arg: string | undefined): boolean {
...
+    connectionString: z
+        .string()
+        .optional()
+        .describe(
+            "MongoDB connection string for direct database connections. Optional, if not set, you'll need to call the connect tool before interacting with MongoDB data."
+        ),
+    loggers: z
+        .array(z.enum(["stderr", "disk", "mcp"]))
+        .default(["disk", "mcp"])
+        .describe("Comma separated values, possible values are 'mcp', 'disk' and 'stderr'."),
```
nirinchev, line 181: The description doesn't seem to match the type.

src/common/config.ts
```diff
@@ -157,38 +159,118 @@ function isConnectionSpecifier(arg: string | undefined): boolean {
...
+            "When set to true, enforces that query operations must use an index, rejecting queries that perform a collection scan."
+        ),
+    telemetry: z
+        .enum(["enabled", "disabled"])
+        .default("enabled")
+        .describe("When set to disabled, disables telemetry collection."),
+    transport: z.enum(["stdio", "http"]).default("stdio").describe("Either 'stdio' or 'http'."),
+    httpPort: z.number().default(3000).describe("Port number."),
+    httpHost: z.string().default("127.0.0.1").describe("Host to bind the http server."),
+    httpHeaders: z.record(z.string()).describe("HTTP headers for the server."),
```
nirinchev: Are these descriptions important? If so, we should probably improve the wording here to clarify what these options configure and when they'd be relevant.

server.json
gagik: we can also use an auth token, though OIDC publishing will also be great to adopt for NPM in the future: https://github.blog/changelog/2025-07-31-npm-trusted-publishing-with-oidc-is-generally-available/

src/common/config.ts
```diff
@@ -157,38 +159,118 @@ function isConnectionSpecifier(arg: string | undefined): boolean {
...
+        .number()
+        .default(120_000)
+        .describe("Time in milliseconds between export cleanup cycles that remove expired export files."),
+    atlasTemporaryDatabaseUserLifetimeMs: z
+        .number()
+        .default(14_400_000)
+        .describe(
+            "Time in milliseconds that temporary database users created when connecting to MongoDB Atlas clusters will remain active before being automatically deleted."
+        ),
+    voyageApiKey: z
```
nirinchev, line 266: Is this optional?

src/common/config.ts
```diff
@@ -157,38 +159,117 @@ function isConnectionSpecifier(arg: string | undefined): boolean {
...
-    notificationTimeoutMs: number;
-    maxDocumentsPerQuery: number;
-    maxBytesPerQuery: number;
-    atlasTemporaryDatabaseUserLifetimeMs: number;
-    voyageApiKey: string;
-    disableEmbeddingsValidation: boolean;
-    vectorSearchDimensions: number;
-    vectorSearchSimilarityFunction: Similarity;
-}
+export const UserConfigSchema = z.object({
```
gagik, line 162: I want to turn this into a single source of truth for our arguments. This should be useful in the future, including for runtime validation.

server.json
```diff
@@ -0,0 +1,670 @@
...
+          "name": "MDB_MCP_VOYAGE_API_KEY",
+          "description": "API key for Voyage AI embeddings service (required for vector search operations with text-to-embedding conversion).",
+          "isRequired": false,
+          "format": "string",
+          "isSecret": true
+        }
+      ],
+      "packageArguments": [
+        {
+          "type": "positional",
```
nirinchev: These are not positional, are they? Or am I misunderstanding what `positional` means here?

gagik: ah you're right, I thought `positional` + `value_hint` is what this was but it's actually `named`

src/common/config.ts
gagik: these were added originally because I was also trying to get the script to update the README.md table but I'll remove them and follow up later.

src/common/config.ts
gagik: Yeah these are copied over from README.md but the schema were created with (eventual) validation in mind. Definitely a confusing setup the way it is, so will remove all the descriptions for now since I dropped my bonus goal to get this to handle README.md

src/common/config.ts
gagik: ah wait, they're also used by the server.json file. This was one of the only fields which we didn't have in our README.md, made it slightly better; should we hide it or keep it in README?

src/common/config.ts
gagik: So... I think for the time being this is fine, this isn't something we're consuming yet and as far as user-facing description and program-facing validation, it still makes sense.
In the future we should adopt https://zod.dev/metadata and have multiple fields to define this. This needs a Zod version bump otherwise I'd use it now

src/common/config.ts
nirinchev: I think we can keep it, just gotta improve the descriptions a little to make it clear in which cases they're used.

.github/workflows/publish.yml
```diff
@@ -15,6 +15,11 @@ jobs:
       RELEASE_CHANNEL: ${{ steps.npm-tag.outputs.RELEASE_CHANNEL }}
     steps:
       - uses: GitHubSecurityLab/actions-permissions/monitor@v1
+      - uses: mongodb-js/devtools-shared/actions/setup-bot-token@main
```
nirinchev: Is this in the wrong job?

.github/workflows/publish.yml
```diff
@@ -75,6 +80,7 @@ jobs:
     environment: Production
     permissions:
       contents: write
+      id-token: write # Required for OIDC authentication with MCP Registry
```
nirinchev: Do we need this now that we're using the app token?

.github/workflows/publish.yml
```diff
@@ -95,6 +101,23 @@ jobs:
         run: npm publish --tag ${{ needs.check.outputs.RELEASE_CHANNEL }}
         env:
           NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+
+      - name: Update server.json version and arguments
```
nirinchev: Can we move this at least temporarily as the last step? I expect there may be rough edges in the first couple of runs, so don't want this failing to impact the other steps.

src/common/config.ts
```diff
@@ -216,9 +216,19 @@ export const UserConfigSchema = z.object({
...
+        .default(3000)
+        .describe("Port number for the HTTP server (only used when transport is 'http')."),
+    httpHost: z
+        .string()
+        .default("127.0.0.1")
+        .describe("Host address to bind the HTTP server to (only used when transport is 'http')."),
+    httpHeaders: z
+        .record(z.string())
+        .describe(
+            "Custom HTTP headers to include in responses from the HTTP server (only used when transport is 'http'). Useful for adding CORS headers, authentication tokens, or other custom headers required by your client application."
```
gagik: is this accurate? my interpretation of when one would use this @nirinchev

.github/workflows/publish.yml
gagik: is it? seems to be the way we get the token everywhere

nirinchev: You're adding this to the `check` job, but seems like you're trying to use it in the `publish` job. Step outputs cannot be passed across jobs.

src/common/config.ts
nirinchev: No, these are headers that the http server will validate when making requests. This is e.g. what we do for the auth header by the vscode extension.

PR #682, opened 2025-10-24
chore: add a temporary OIDC-based registry publish workflow

gagik:
Since the token-based auth seems to be either buggy or unsupported at the moment (see https://github.com/modelcontextprotocol/registry/issues/698), this adds a separate workflow for us to iterate on with regards to OIDC auth.

I tried this locally and I don't have the permissions for it but either a) in actions context the permissions can apply to our CI or b) someone else in our team might have all the necessary permissions for publishing as a temporary solution.

### DIFF
--- a/.github/workflows/registry-publish.yml
+++ b/.github/workflows/registry-publish.yml
@@ -16,16 +16,6 @@ jobs:
     steps:
       - uses: GitHubSecurityLab/actions-permissions/monitor@v1
       - uses: actions/checkout@v5
-      - uses: actions/setup-node@v6
-        with:
-          node-version-file: package.json
-          registry-url: "https://registry.npmjs.org"
-          cache: "npm"
-
-      - name: Build package
-        run: |
-          npm ci
-          npm run build
 
       - name: Install MCP Publisher
         run: |


.github/workflows/registry-publish.yml
```diff
@@ -0,0 +1,38 @@
...
+          cache: "npm"
+
+      - name: Build package
+        run: |
+          npm ci
+          npm run build
+
+      - name: Install MCP Publisher
+        run: |
+          curl -L "https://github.com/modelcontextprotocol/registry/releases/latest/download/mcp-publisher_$(uname -s | tr '[:upper:]' '[:lower:]')_$(uname -m | sed 's/x86_64/amd64/;s/aarch64/arm64/').tar.gz" | tar xz mcp-publisher
```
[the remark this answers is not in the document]
gagik: I don't think this is true at all lol

.github/workflows/registry-publish.yml
```diff
@@ -0,0 +1,35 @@
...
+      - uses: actions/setup-node@v6
+        with:
+          node-version-file: package.json
+          registry-url: "https://registry.npmjs.org"
+          cache: "npm"
+
+      - name: Build package
+        run: |
+          npm ci
+          npm run build
```
nirinchev: Doesn't sound like we need to setup node or build at all, do we?

gagik: kept in just in case but fair enough

PR #684, opened 2025-10-24
chore: retry the whole test with fixtures in case of error

kmruiz:
## Proposed changes

This adds a retry mechanism to OIDC tests in case the fail. It reuses the retry mechanism from vitest, which should also run fixtures.

## Checklist

### DIFF
--- a/tests/integration/common/connectionManager.oidc.test.ts
+++ b/tests/integration/common/connectionManager.oidc.test.ts
@@ -13,7 +13,8 @@ import type { OIDCMockProviderConfig } from "@mongodb-js/oidc-mock-provider";
 import { OIDCMockProvider } from "@mongodb-js/oidc-mock-provider";
 import { type TestConnectionManager } from "../../utils/index.js";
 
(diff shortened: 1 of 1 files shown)


tests/integration/common/connectionManager.oidc.test.ts
```diff
@@ -107,13 +107,15 @@ describe.skipIf(process.platform !== "linux")("ConnectionManager OIDC Tests", as
             (integration) => {
                 function oidcIt(name: string, cb: Parameters<OidcIt>[1]): void {
                     /* eslint-disable vitest/expect-expect */
-                    it(name, { timeout: DEFAULT_TIMEOUT }, async (context) => {
+                    it(name, { timeout: DEFAULT_TIMEOUT, retry: 5 }, async (context) => {
```
gagik: is there some wait time for this? would it make sense to introduce some wait on retries?

kmruiz: vitest doesn't support wait time :( https://vitest.dev/api/#test-api-reference

PR #686, opened 2025-10-24
feat: add ability to specify enabled preview features

nirinchev:
## Proposed changes

This change adds a `previewFeatures` config option that users can provide to enable non-GA functionality. For now, the only option is `vectorSearch`.

### DIFF
--- a/src/common/config.ts
+++ b/src/common/config.ts
@@ -559,13 +559,13 @@ export function setupUserConfig({
 }: {
     cli: string[];
     env: Record<string, unknown>;
-    defaults: Partial<UserConfig>;
+    defaults: UserConfig;
 }): UserConfig {
     const userConfig = {
         ...defaults,
         ...parseEnvConfig(env),
         ...parseCliConfig(cli),
-    } as UserConfig;
+    } satisfies UserConfig;
 
     userConfig.disabledTools = commaSeparatedToArray(userConfig.disabledTools);
     userConfig.loggers = commaSeparatedToArray(userConfig.loggers);


gagik (approved):
Looks good to me.

README.md
```diff
@@ -490,6 +491,18 @@ You can disable telemetry using:
 
 > **💡 Platform Note:** For Windows users, see [Environment Variables](#environment-variables) for platform-specific instructions.
 
+#### Opting into Preview Features
+
+The MongoDB MCP Server may offer functionality that is still in development and may change in future releases. These features are considered "preview features" and are not enabled by default. Generally, these features are well tested, but may not offer the complete functionality we intend to provide in the final release or we'd like to gather feedback before making them generally available. To enable one or more preview features, use the `previewFeatures` configuration option.
+
+- For **environment variable** configuration, use a comma-separated string: `export MDB_MCP_PREVIEW_FEATURES="vectorSearch,dedicatedClusters"`.
+- For **command-line argument** configuration, use a space-separated string: `--previewFeatures vectorSearch dedicatedClusters`.
```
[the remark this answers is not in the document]
gagik: that could be misleading, yeah. but also understand it's nice to have an example of multiple things so it's fine to keep it as is.

src/tools/mongodb/delete/dropIndex.ts
```diff
@@ -2,15 +2,15 @@ import z from "zod";
...
+import { type ToolArgs, type OperationType, formatUntrustedData } from "../../tool.js";
 
 export class DropIndexTool extends MongoDBToolBase {
     public name = "drop-index";
     protected description = "Drop an index for the provided database and collection.";
     protected argsShape = {
         ...DbOperationArgs,
         indexName: z.string().nonempty().describe("The name of the index to be dropped."),
-        type: this.isFeatureFlagEnabled(FeatureFlags.VectorSearch)
+        type: this.isFeatureEnabled("vectorSearch")
```
blva, line 13: the ultimate nit: why not keep `vectorSearch` as a const against typos or even maintain`isVectorSerachFeatureEnabled` at the same level as `isFeatureEnabled`? 

another nit: btw I usually prefer to keeping a tiny file with this logic as it might grow and we could also add some unit tests just while we're here to keep it safe from future changes.

src/common/config.ts
```diff
@@ -556,7 +561,7 @@ export function setupUserConfig({
     env: Record<string, unknown>;
     defaults: Partial<UserConfig>;
 }): UserConfig {
-    const userConfig: UserConfig = {
+    const userConfig = {
```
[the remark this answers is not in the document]
blva: [q] why are we removing this?

blva (approved):
non-blocking comments

src/tools/mongodb/delete/dropIndex.ts
nirinchev: "vectorSearch" is a const in a union type. So when calling `isFeatureEnabled` you can only provide one of the preview features we've defined in the `previewFeatures` enum in config.ts.

Not sure I fully understood the second half of your suggestion - what part of the logic would you suggest we move to a separate file? For context, the enum values are currently defined in `config.ts` along with the `PreviewFeatures` union type, while `isFeatureEnabled` is defined on the base `tool` type. We could move it to `config.ts` if that makes more sense, but I wanted to keep the user config as close to a data bag as possible, which is why I opted to define the `isFeatureEnabled` helper in the base tool.

src/common/config.ts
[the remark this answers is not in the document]
nirinchev: It's already type asserted a few lines down, so I figured it's a bit redundant - `const foo: Foo = { } as Foo` is a bit repetitive. For context, we're not losing on type safety because we have the `as UserConfig` assertion anyway. That being said, since this is multiline, I can see how it might not be super obvious, so happy to bring it back if it's clearer.

src/tools/mongodb/delete/dropIndex.ts
gagik: To add to this: enums in JavaScript are pretty terrible and it's best to avoid them. 
https://jsdevspace.substack.com/p/shrink-typescript-code-replace-enums
Especially because of "erasable syntax" that can allow us to run TypeScript in Node directly:
https://www.totaltypescript.com/erasable-syntax-only
Probably something to introduce as a lint rule soon.

src/common/config.ts
gagik:
```suggestion
    } satisfies UserConfig;
```
nit: but would ensure this actually matches

PR #688, opened 2025-10-24
feat: add support for automatic embeddings for the insert many tool MCP-236

gagik:
Adds automatic generation using the Voyage API key, with a system to minimize requests being made in bulk.

### DIFF
--- a/src/common/search/vectorSearchEmbeddingsManager.ts
+++ b/src/common/search/vectorSearchEmbeddingsManager.ts
@@ -108,13 +108,14 @@ export class VectorSearchEmbeddingsManager {
         { database, collection }: { database: string; collection: string },
         documents: Document[]
     ): Promise<void> {
-        const embeddingValidationResults = await Promise.all(
-            documents.map((document) => this.findFieldsWithWrongEmbeddings({ database, collection }, document))
-        );
-        const embeddingValidations = new Set(embeddingValidationResults.flat());
+        const embeddingValidationResults = (
+            await Promise.all(
+                documents.map((document) => this.findFieldsWithWrongEmbeddings({ database, collection }, document))
+            )
+        ).flat();
 
-        if (embeddingValidations.size > 0) {
-            const embeddingValidationMessages = Array.from(embeddingValidations).map(
+        if (embeddingValidationResults.length > 0) {
+            const embeddingValidationMessages = embeddingValidationResults.map(
                 (validation) =>
                     `- Field ${validation.path} is an embedding with ${validation.expectedNumDimensions} dimensions and ${validation.expectedQuantization}` +
                     ` quantization, and the provided value is not compatible. Actual dimensions: ${validation.actualNumDimensions}, ` +
@@ -293,7 +294,7 @@ export class VectorSearchEmbeddingsManager {
         rawValues: string[];
         embeddingParameters: SupportedEmbeddingParameters;
         inputType: EmbeddingParameters["inputType"];
-    }): Promise<number[][]> {
+    }): Promise<unknown[][]> {
         const provider = await this.atlasSearchEnabledProvider();
         if (!provider) {
             throw new MongoDBError(
--- a/src/common/search/embeddingsProvider.ts
+++ b/src/common/search/embeddingsProvider.ts
@@ -7,7 +7,7 @@ import { createFetch } from "@mongodb-js/devtools-proxy-support";
 import { z } from "zod";
 
 type EmbeddingsInput = string;
-type Embeddings = number[];
+type Embeddings = number[] | unknown[];
 export type EmbeddingParameters = {
     inputType: "query" | "document";
 };
--- a/src/tools/mongodb/read/aggregate.ts
+++ b/src/tools/mongodb/read/aggregate.ts
@@ -298,7 +298,7 @@ export class AggregateTool extends MongoDBToolBase {
                 // $vectorSearch.queryVector can be a BSON.Binary: that it's not either number or an array.
                 // It's not exactly valid from the LLM perspective (they can't provide binaries).
                 // That's why we overwrite the stage in an untyped way, as what we expose and what LLMs can use is different.
-                vectorSearchStage.queryVector = embeddings;
+                vectorSearchStage.queryVector = embeddings as string | number[];
             }
         }
 
--- a/tests/unit/common/search/vectorSearchEmbeddingsManager.test.ts
+++ b/tests/unit/common/search/vectorSearchEmbeddingsManager.test.ts
@@ -390,6 +390,149 @@ describe("VectorSearchEmbeddingsManager", () => {
         });
     });
 
+    describe("assertFieldsHaveCorrectEmbeddings", () => {
+        it("does not throw for invalid documents when validation is disabled", async () => {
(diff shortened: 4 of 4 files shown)


src/tools/mongodb/create/insertMany.ts
```diff
@@ -21,37 +37,22 @@ export class InsertManyTool extends MongoDBToolBase {
...
-                    "There were errors when inserting documents. No document was inserted.",
-                    ...embeddingValidationMessages
-                ),
-                isError: true,
-            };
-        }
+        await this.session.vectorSearchEmbeddingsManager.assertFieldsHaveCorrectEmbeddings(
+            { database, collection },
+            documents
+        );
```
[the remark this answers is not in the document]
gagik: well the embeddings generation (when it happens) is relevant for asserting if the embeddings are correct, mr robot

tests/integration/tools/mongodb/create/insertMany.test.ts
```diff
@@ -181,8 +191,416 @@ describeWithMongoDB(
...
+                expect(insertedIds).toHaveLength(1);
+
+                const doc = await collection.findOne({ _id: insertedIds[0] });
+                expect((doc?.title as Record<string, unknown>)?.text).toBe("The Matrix");
+                expect((doc?.title as Record<string, unknown>)?.embeddings).not.toBeDefined();
+                expect((doc?.["title.embeddings"] as unknown as number[]).length).toBe(1024);
+            });
+
+            it("returns an error when input field does not have a vector search index", async () => {
+                await createVectorSearchIndexAndWait(integration.mongoClient(), integration.randomDbName(), "test", [
```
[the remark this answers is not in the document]
gagik: this is a confusing name to me but i cannot be bothered to track all of these down

tests/integration/tools/mongodb/create/insertMany.test.ts
```diff
@@ -110,10 +111,12 @@ describeWithMongoDB(
     "insertMany tool when search is enabled",
     (integration) => {
         let collection: Collection;
+        let database: string;
 
         beforeEach(async () => {
             await integration.connectMcpClient();
-            collection = await integration.mongoClient().db(integration.randomDbName()).createCollection("test");
+            database = integration.randomDbName();
```
gagik, line 118: I regret doing this because there's too many calls of this function but I'd not have expected `integration.randomDbName()` to always return the same value so switched to storing in a variable

CONTRIBUTING.md
```diff
@@ -76,6 +76,10 @@ npm test -- path/to/test/file.test.ts
 npm test -- path/to/directory
 ```
 
+#### Accuracy Tests and colima
```
gagik, line 79: I don't know if this is considered niche but I find colima to be a nice simple way to run docker on Mac and if this existed I'd not have lost many painful moments of figuring this bit out

src/tools/mongodb/create/insertMany.ts
```diff
@@ -12,46 +23,44 @@ export class InsertManyTool extends MongoDBToolBase {
         documents: z
```
gagik, line 23: okay so I did my best to prompt engineer it to generate embeddings by supplying plain text to the vector search indexed fields as strings inside `documents`.
However, in my experience models get confused about this idea of "the schema says it's a vector but I need to ignore this and supply a string" and I also think relying on the description of the documents field for this is less scalable.
So, the best way I found to force the model to provide these vector index raw strings is to put it as a required input as part of embeddingParameters. It seems to do a good job at mapping the 2 arrays and respective indexes. It does sometimes provide a string in place of the original document's field in the input, but this gets overriden so seems quite fine.

Open to suggestions if I may be overlooking something here.

gagik:
I'll look into adding a few more tests I guess to keep coveralls happy

src/tools/mongodb/create/insertMany.ts
```diff
@@ -12,46 +23,44 @@ export class InsertManyTool extends MongoDBToolBase {
...
+        documents = await this.replaceRawValuesWithEmbeddingsIfNecessary({
+            database,
+            collection,
+            documents,
+            embeddingParameters,
+        });
+
+        await this.session.vectorSearchEmbeddingsManager.assertFieldsHaveCorrectEmbeddings(
+            { database, collection },
+            documents
```
blva, line 62: [q] do we need to run this if embeddings are not enabled?

gagik: this was existing behavior so I kept it as we _would_ want to prevent the model from adding artbirary data to a vector search indexed field.
that said, I'm wondering if the insert many call would have failed at DB-level anyhow; not sure if that's redundant then.
cc: @kmruiz

kmruiz: So @gagik is completely right. It's to prevent the agent adding a raw field in a place where the server expects an embedding. Currently, the server does not reject these invalid values (unless the user specifies a JSON Schema) and it would break the VS Index and the behaviour can be pretty inconsistent depending on how it breaks.

src/common/search/vectorSearchEmbeddingsManager.ts
```diff
@@ -239,21 +266,34 @@ export class VectorSearchEmbeddingsManager {
...
+    public async generateEmbeddings({
+        rawValues,
+        embeddingParameters,
+        inputType,
+    }: {
         rawValues: string[];
         embeddingParameters: SupportedEmbeddingParameters;
         inputType: EmbeddingParameters["inputType"];
-    }): Promise<unknown[]> {
+    }): Promise<number[][]> {
```
kmruiz: Sadly, this is not correct. An Embedding can be also other types, like object[] (for different types of quantisations). That's why it was an unknown[].

src/tools/mongodb/create/insertMany.ts
kmruiz: I think this should work fine, we have the accuracy tests if something goes wary in the future and we can safely tweak the prompt.

kmruiz (approved):

src/common/search/vectorSearchEmbeddingsManager.ts
gagik: sg, I'll update the `EmbeddingsProvider` type for this as well in this case

src/common/search/vectorSearchEmbeddingsManager.ts
```diff
@@ -103,7 +104,33 @@ export class VectorSearchEmbeddingsManager {
...
 
-    async findFieldsWithWrongEmbeddings(
+    async assertFieldsHaveCorrectEmbeddings(
+        { database, collection }: { database: string; collection: string },
+        documents: Document[]
+    ): Promise<void> {
+        const embeddingValidationResults = await Promise.all(
+            documents.map((document) => this.findFieldsWithWrongEmbeddings({ database, collection }, document))
+        );
+        const embeddingValidations = new Set(embeddingValidationResults.flat());
```
gagik: fyi @kmruiz this set isn't doing anything because it looks as object reference equality instead of structural which is pretty meaingless for us.
I removed it for now as it feels like not the biggest deal to have duplicates here but we can follow-up if needed.

PR #689, opened 2025-10-24
chore: support for pre-filters in $vectorSearch MCP-240

kmruiz:
## Proposed changes

Adds support pre-filters in the aggregate tool, when using the $vectorSearch stage. We are also adding more matchers for better accuracy tests, namely:

- `Matcher.caseInsensitiveString`: for checking if a string is equal to another ignoring the case. Some LLMs can change the casing of values, and we don't have control over that unless the user specifically prompts for it, so for our tests, we will assume they are correct.
- `Matcher.not`: negates a matcher. For example, to ensure an array does not contain a specific value.
- `Matcher.arrayOrSingle`: Matches either `[ value ]` or `value`. This is important because MQL queries sometimes support both values. 

## Checklist

### DIFF
--- a/src/common/search/embeddingsProvider.ts
+++ b/src/common/search/embeddingsProvider.ts
@@ -27,13 +27,32 @@ export const zVoyageModels = z
     .enum(["voyage-3-large", "voyage-3.5", "voyage-3.5-lite", "voyage-code-3"])
     .default("voyage-3-large");
 
+// Zod does not undestand JS boxed numbers (like Int32) as integer literals,
+// so we preprocess them to unwrap them so Zod understands them.
+function unboxNumber(v: unknown): number {
+    if (v && typeof v === "object" && typeof v.valueOf === "function") {
+        const n = Number(v.valueOf());
+        if (!Number.isNaN(n)) return n;
+    }
+    return v as number;
+}
+
 export const zVoyageEmbeddingParameters = z.object({
     outputDimension: z
-        .union([z.literal(256), z.literal(512), z.literal(1024), z.literal(2048), z.literal(4096)])
+        .preprocess(
+            unboxNumber,
+            z.union([z.literal(256), z.literal(512), z.literal(1024), z.literal(2048), z.literal(4096)])
+        )
         .optional()
         .default(1024),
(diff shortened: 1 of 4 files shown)


src/common/search/embeddingsProvider.ts
```diff
@@ -27,14 +27,33 @@ export const zVoyageModels = z
...
+    return v as number;
+}
+
 export const zVoyageEmbeddingParameters = z.object({
     outputDimension: z
-        .union([z.literal(256), z.literal(512), z.literal(1024), z.literal(2048), z.literal(4096)])
+        .preprocess(
+            unboxNumber,
+            z.union([z.literal(256), z.literal(512), z.literal(1024), z.literal(2048), z.literal(4096)])
+        )
```
himanshusinghs, line 45: Is it SDK that is doing the boxing of Number to Int32, etc? wondering if that is something we should be doing everywhere else as well.

src/tools/mongodb/read/aggregate.ts
```diff
@@ -59,11 +59,21 @@ const VectorSearchStage = z.object({
...
 
 export const AggregateArgs = {
-    pipeline: z
-        .array(z.union([AnyStage, VectorSearchStage]))
-        .describe(
-            "An array of aggregation stages to execute. $vectorSearch can only appear as the first stage of the aggregation pipeline or as the first stage of a $unionWith subpipeline. When using $vectorSearch, unless the user explicitly asks for the embeddings, $unset any embedding field to avoid reaching context limits."
-        ),
+    pipeline: z.array(z.union([AnyStage, VectorSearchStage])).describe(
+        `An array of aggregation stages to execute.  
+\`$vectorSearch\` **MUST** be the first stage of the pipeline, or the first stage of a \`$unionWith\` subpipeline.
```
himanshusinghs, line 64: Unless proven otherwise by accuracy tests I fear this sounding more like $vectorSearch always has to be the first stage regardless of the pipeline needing one or not.

Would it make sense to frame it like -
`If $vectorSearch is to be used it **MUST** ... rest of the content`

himanshusinghs (approved):
Looks good. I left minor concerns/questions.

PR #692, opened 2025-10-27
feat: add ability to create atlas search indexes MCP-275

nirinchev:
## Proposed changes

This adds support for creating atlas search indexes. Dropping and listing is already supported, so needed no changes.

Notable omissions from the schema, that we will want to evaluate based on user feedback and internal testing:
* `analyzers` - based on consultation with the Search team
* `mappings.dynamic.typeSet` - preview feature
* `searchAnalyzer` - based on consultation with the Search team
* `storedSource` - based on testing of model performance
* `synonyms` - for simplicity
* `typeSets` - for simplicity
* `mappings.fields.*` - the fields definition only includes the type and uses `passthrough` for the remaining fields. We'll have to evaluate its performance and if necessary, define the complete json schema.

### DIFF
--- a/tests/accuracy/createIndex.test.ts
+++ b/tests/accuracy/createIndex.test.ts
@@ -80,7 +80,7 @@ describeAccuracyTests(
                                     {
                                         type: "vector",
                                         path: "plotSummary",
-                                        numDimensions: 1024,
+                                        numDimensions: "1024",
                                     },
                                 ],
                             },
@@ -106,17 +106,19 @@ describeAccuracyTests(
                                     {
                                         type: "vector",
                                         path: "plotSummary",
-                                        numDimensions: Matcher.number(
-                                            (value) => value % 2 === 0 && value >= 256 && value <= 8192
-                                        ),
+                                        numDimensions: Matcher.string((value) => {
+                                            const intValue = parseInt(value);
+                                            return intValue % 2 === 0 && intValue >= 256 && intValue <= 8192;
+                                        }),
                                         similarity: Matcher.anyOf(Matcher.undefined, Matcher.string()),
                                     },
                                     {
                                         type: "vector",
                                         path: "genre",
-                                        numDimensions: Matcher.number(
-                                            (value) => value % 2 === 0 && value >= 256 && value <= 8192
-                                        ),
+                                        numDimensions: Matcher.string((value) => {
+                                            const intValue = parseInt(value);
+                                            return intValue % 2 === 0 && intValue >= 256 && intValue <= 8192;
+                                        }),
                                         similarity: Matcher.anyOf(Matcher.undefined, Matcher.string()),
                                     },
                                 ],
@@ -143,7 +145,7 @@ describeAccuracyTests(
                                     {
                                         type: "vector",
                                         path: "plotSummary",
-                                        numDimensions: 1024,
+                                        numDimensions: "1024",
                                     },
                                     {
                                         type: "filter",
--- a/tests/accuracy/sdk/matcher.ts
+++ b/tests/accuracy/sdk/matcher.ts
@@ -32,8 +32,8 @@ export abstract class Matcher {
         return new BooleanMatcher(expected);
     }
 
-    public static string(): Matcher {
-        return new StringMatcher();
+    public static string(additionalFilter: (value: string) => boolean = () => true): Matcher {
+        return new StringMatcher(additionalFilter);
     }
 
     public static caseInsensitiveString(text: string): Matcher {
@@ -153,8 +153,11 @@ class BooleanMatcher extends Matcher {
 }
 
 class StringMatcher extends Matcher {
+    constructor(private additionalFilter: (value: string) => boolean = () => true) {
+        super();
+    }
     public match(actual: unknown): number {
-        return typeof actual === "string" ? 1 : 0;
+        return typeof actual === "string" && this.additionalFilter(actual) ? 1 : 0;
     }
 }
 
--- a/tests/integration/tools/mongodb/create/createIndex.test.ts
+++ b/tests/integration/tools/mongodb/create/createIndex.test.ts
@@ -948,7 +948,7 @@ describeWithMongoDB(
     {
         getUserConfig: () => ({
             ...defaultTestConfig,
-            previewFeatures: ["vectorSearch"],
+            previewFeatures: ["search"],
         }),
         downloadOptions: {
             search: true,


src/tools/mongodb/create/createIndex.ts
```diff
@@ -6,61 +6,121 @@ import type { IndexDirection } from "mongodb";
 import { quantizationEnum, similarityEnum } from "../../../common/search/vectorSearchEmbeddingsManager.js";
 
 export class CreateIndexTool extends MongoDBToolBase {
-    private vectorSearchIndexDefinition = z.object({
```
nirinchev, line 10: This is best viewed with the "Hide whitespace" option - it's just prettier reformatting the indents.

tests/accuracy/sdk/accuracyTestingClient.ts
```diff
@@ -91,7 +91,7 @@ export class AccuracyTestingClient {
             return [`--${key}`, value];
         });
 
-        const args = [MCP_SERVER_CLI_SCRIPT, "--connectionString", mdbConnectionString, ...additionalArgs];
+        const args = [MCP_SERVER_CLI_SCRIPT, mdbConnectionString, ...additionalArgs];
```
nirinchev, line 95: `--connectionString` is deprecated so using the positional argument.

tests/integration/tools/mongodb/delete/dropIndex.test.ts
```diff
@@ -362,35 +376,39 @@ describe.each([{ vectorSearchEnabled: false }, { vectorSearchEnabled: true }])(
...
+                                    namespace: `${database}.${collection}`,
                                 });
-                            }
-                        );
+                            });
+                        });
 
-                        describe("and attempting to delete an existing index", () => {
+                        describe.each([
+                            { description: "search", indexName: getSearchIndexName() },
```
nirinchev: This should pass in the function rather than evaluating it

src/tools/mongodb/create/createIndex.ts
```diff
@@ -6,61 +6,121 @@ import type { IndexDirection } from "mongodb";
...
+                .refine((fields) => fields.some((f) => f.type === "vector"), {
+                    message: "At least one vector field must be defined",
+                })
+                .describe(
+                    "Definitions for the vector and filter fields to index, one definition per document. You must specify `vector` for fields that contain vector embeddings and `filter` for additional fields to filter on. At least one vector-type field definition is required."
+                ),
+        })
+        .describe("Definition for a Vector Search index.");
+
+    private atlasSearchIndexDefinition = z
```
kmruiz, line 69: Why aren't we supporting custom analyzers?

https://www.mongodb.com/docs/atlas/atlas-search/analyzers/custom/

src/tools/mongodb/create/createIndex.ts
```diff
@@ -6,61 +6,121 @@ import type { IndexDirection } from "mongodb";
...
+        .object({
+            type: z.literal("search"),
+            analyzer: z
```
kmruiz, line 72: Probably this should be an enum of the analyzers.

src/tools/mongodb/create/createIndex.ts
```diff
@@ -6,61 +6,121 @@ import type { IndexDirection } from "mongodb";
...
+                .string()
+                .optional()
+                .default("lucene.standard")
+                .describe(
+                    "The analyzer to use for the index. Can be one of the built-in lucene analyzers (`lucene.standard`, `lucene.simple`, `lucene.whitespace`, `lucene.keyword`), a language-specific analyzer, such as `lucene.cjk` or `lucene.czech`, or a custom analyzer defined in the Atlas UI."
+                ),
+            mappings: z
+                .object({
```
kmruiz, line 80: Lack support of:

* numPartitions
* searchAnalyzer vs analyze
* custom analyzers
* storedSources
* synonyms
* typeSets

kmruiz: https://www.mongodb.com/docs/atlas/atlas-search/index-definitions/?deployment-type=atlas&interface=driver&language=nodejs#std-label-ref-index-definitions

We could say that custom analyzers are not that important, but storedSources is actually relevant most of the times.

src/tools/mongodb/create/createIndex.ts
```diff
@@ -6,61 +6,121 @@ import type { IndexDirection } from "mongodb";
...
+                    dynamic: z
+                        .boolean()
```
kmruiz, line 82: Dynamic can be an object of typeSets.

src/tools/mongodb/create/createIndex.ts
```diff
@@ -6,61 +6,121 @@ import type { IndexDirection } from "mongodb";
...
+                        .default(false)
+                        .describe(
+                            "Enables or disables dynamic mapping of fields for this index. If set to true, Atlas Search recursively indexes all dynamically indexable fields. If set to false, you must specify individual fields to index using mappings.fields."
+                        ),
+                    fields: z
+                        .record(
+                            z.string().describe("The field name"),
+                            z
+                                .object({
+                                    type: z
```
kmruiz, line 93: Objects will require additional fields depending on the type. I know passthrough will keep them, but we should document them so the agent knows which ones to use and how. For example, autocomplete supports defining a custom analyzer, how to tokenize (which is really important) and similarity functions.

src/tools/mongodb/create/createIndex.ts
nirinchev: The args shape is based on the POC the search team did for index support and I was going off of the assumption that they've selected the fields that they see the most value in exposing to LLMs. I realize there's a lot more configuration that's possible, I'm just not sure how much of that is stuff we expect agents to configure vs an actual human who wants to fine-tune the index.

src/tools/mongodb/create/createIndex.ts
nirinchev: This is in preview though, so I don't expect there's sufficient docs or training data for general-purpose models to accurately choose which one to use.

src/tools/mongodb/create/createIndex.ts
nirinchev: The exact shape is extremely complex to represent in a json schema. I'm worried that being overly specific will result in this being more harmful than helpful, especially if we expect the majority of the use cases to revolve around just specifying the type.

src/tools/mongodb/create/createIndex.ts
kmruiz: A POC to see the feasibility to create search indexes and production code are likely to have different requirements.

src/tools/mongodb/create/createIndex.ts
kmruiz: The preview is for vector search though, FTS is explicitly out of scope of the vector search project.

src/tools/mongodb/create/createIndex.ts
kmruiz: Yep, the schema is complicated, it has a lot of options that are not compatible even between them. We should have proper documentation of which ones we want to expose and which ones not, something that we haven't discussed yet because supporting the most used bits of Atlas Search is already a substantial effort.

src/tools/mongodb/create/createIndex.ts
nirinchev: The `typeSets` functionality is in preview.

src/tools/mongodb/create/createIndex.ts
nirinchev: Discussing with the search team, `searchAnalyzer` and `analyzers` is not something we need to support based on current usage patterns. We can add support in the future if we see customer demand for it.

src/tools/mongodb/create/createIndex.ts
nirinchev: * `typeSets` is still in preview, so I'm leaving it out for now.
* `numPartitions` was added
* `searchAnalyzer` and `analyzers` are out of scope per search team's recommendation

I'm leaving the following out of this PR for now as they have more complex schema and have been more error prone when testing:
* `synonyms`
* `storedSources`

We can add them as we move toward GA and build out a more comprehensive testing suite.

src/tools/mongodb/create/createIndex.ts
nirinchev: Leaving this out of scope for now and we can decide how much detail to provide as we build out a more comprehensive testing suite.

PR #693, opened 2025-10-27
chore: check that a vector search index exists with indexCheck

kmruiz:
## Checklist

### DIFF
--- a/src/tools/mongodb/read/aggregate.ts
+++ b/src/tools/mongodb/read/aggregate.ts
@@ -90,21 +90,25 @@ export class AggregateTool extends MongoDBToolBase {
 
             // Check if aggregate operation uses an index if enabled
             if (this.config.indexCheck) {
-                const usesVectorSearchIndex = await this.isVectorSearchIndexUsed({ database, collection, pipeline });
+                const [usesVectorSearchIndex, indexName] = await this.isVectorSearchIndexUsed({
+                    database,
+                    collection,
+                    pipeline,
+                });
                 switch (usesVectorSearchIndex) {
-                    case "no-vector-search-query":
+                    case "not-vector-search-query":
                         await checkIndexUsage(provider, database, collection, "aggregate", async () => {
                             return provider
                                 .aggregate(database, collection, pipeline, {}, { writeConcern: undefined })
                                 .explain("queryPlanner");
                         });
                         break;
-                    case false:
+                    case "non-existent-index":
                         throw new MongoDBError(
                             ErrorCodes.AtlasVectorSearchIndexNotFound,
-                            "Could not find provided vector search index."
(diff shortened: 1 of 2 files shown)


src/tools/mongodb/read/aggregate.ts
```diff
@@ -279,6 +281,41 @@ export class AggregateTool extends MongoDBToolBase {
...
+            return "no-vector-search-query";
+        }
+
+        const indexExists = await this.session.vectorSearchEmbeddingsManager.indexExists({
+            database,
+            collection,
+            indexName,
+        });
+
+        return indexExists;
```
himanshusinghs:
```suggestion
        return await this.session.vectorSearchEmbeddingsManager.indexExists({
            database,
            collection,
            indexName,
        });
```

src/tools/mongodb/read/aggregate.ts
```diff
@@ -100,11 +90,23 @@ export class AggregateTool extends MongoDBToolBase {
...
+                        await checkIndexUsage(provider, database, collection, "aggregate", async () => {
+                            return provider
+                                .aggregate(database, collection, pipeline, {}, { writeConcern: undefined })
+                                .explain("queryPlanner");
+                        });
+                        break;
+                    case false:
+                        throw new MongoDBError(
+                            ErrorCodes.AtlasVectorSearchIndexNotFound,
+                            "Could not find provided vector search index."
```
nirinchev: Should we include the index name here?

src/tools/mongodb/read/aggregate.ts
```diff
@@ -279,6 +281,41 @@ export class AggregateTool extends MongoDBToolBase {
...
 
+    private async isVectorSearchIndexUsed({
+        database,
+        collection,
+        pipeline,
+    }: {
+        database: string;
+        collection: string;
+        pipeline: Document[];
+    }): Promise<boolean | "no-vector-search-query"> {
```
nirinchev: [non-blocking nit] I generally dislike mixing types like that. It's rather confusing without reading the method implementation to discern in what cases `false` is different from `no-vector-search-query`. I know it's a bit more verbose, but I would suggest rewording this to something like (only an example, feel free to choose better names if you don't like these):

```
checkVectorSearchIndexUsage(...): Promise<"valid-index-used" | "not-vector-search-query" | "non-existent-index-used">
```

That way, it's more obvious what the three options are and when to use what.

kmruiz:
> I would expect we'd get an error from the server if we try to use `$vectorSearch` 

https://www.mongodb.com/docs/atlas/atlas-vector-search/vector-search-stage/#fields

> MongoDB Vector Search doesn't return results if you misspell the index name or if the specified index doesn't already exist on the cluster.

🥲

PR #695, opened 2025-10-27
chore: update release process documentation

gagik:
This uses the [wording from the mongosh release process](https://github.com/mongodb-js/mongosh/blob/c3c5abe9a826cf6bd2dd56a5701dfa59d3952c81/packages/build/README.md#L26) to better direct our process with Jira and publicizing our release.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -166,7 +166,7 @@ To release a new version of the MCP server, follow these steps:
    - NPM: https://www.npmjs.com/package/mongodb-mcp-server
    - Docker: https://hub.docker.com/r/mongodb/mongodb-mcp-server
    - MCP Registry: `curl "https://registry.modelcontextprotocol.io/v0.1/servers/io.github.mongodb-js%2Fmongodb-mcp-server/versions/latest"`
-7. Close the Jira ticket for the release. 
+7. Close the Jira ticket for the release.
 8. Go to the [Releases](https://jira.mongodb.org/projects/MCP?selectedItem=com.atlassian.jira.jira-projects-plugin%3Arelease-page&status=released-unreleased) section the and rename the `vNext` to the new version number and mark it as Released. Create a new `vNext` for the next release.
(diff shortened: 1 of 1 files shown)


himanshusinghs (approved):

CONTRIBUTING.md
```diff
@@ -143,31 +143,26 @@ When adding new tools to the MCP server:
...
 
-Our release process is automated using GitHub Actions workflows:
+To release a new version of the MCP server, follow these steps:
 
-### Version Bumping
-
-1. To create a new version, go to the GitHub repository Actions tab
-2. Select the "Version Bump" workflow
-3. Click "Run workflow" and choose one of the following options:
+1. Ensure there is a Jira _Release_ ticket in the [`MCP` project](https://jira.mongodb.org/projects/MCP) for the new release and move it to _In Progress_.
```
nirinchev, line 152: Not opposed to this but curious to hear what's the benefit of having a release ticket as opposed to just looking at the tickets linked to the vNext version?

nirinchev (commented):
I think we need to mention that we need to rename the vNext release in Jira, release it, then create a new vNext with start date today.

CONTRIBUTING.md
gagik: biggest benefit is being able to track the release status & who did the release through Jira. Can be useful i.e. for external stakeholders to find the right person to contact or for them to block their release against ours.

nirinchev: Fair, but in that case, I'd suggest we need to have some automation that would generate the ticket. It sounds a bit tedious to link a bunch of tickets together.

gagik: I agree about automation
> It sounds a bit tedious to link a bunch of tickets together.

To clarify, this release ticket is just a regular ticket tied to the release version so no extra linking needed.

nirinchev: Ah, I read the "correctly mapped" section from point 2 as we'd need to link them to the release ticket, not that we link the version from the release ticket and then look at the tickets for that version.

PR #706, opened 2025-10-29
chore: include tool args for perf advisor telemetry MCP-277

nirinchev:
## Proposed changes

Reworks telemetry events a bit - now we can provide arbitrary kvp data from `resolveTelemetryMetadata` in tools. This allows us to include the operations as additional field in properties. Additionally, renamed the fixed fields in `TelemetryToolMetadata` to match the names we're sending so we can just spread the object instead of the custom handling we had previously.


tests/accuracy/sdk/accuracyTestingClient.ts
```diff
@@ -91,11 +91,15 @@ export class AccuracyTestingClient {
             return [`--${key}`, value];
         });
 
-        const args = [MCP_SERVER_CLI_SCRIPT, "--connectionString", mdbConnectionString, ...additionalArgs];
+        const args = [MCP_SERVER_CLI_SCRIPT, mdbConnectionString, ...additionalArgs];
```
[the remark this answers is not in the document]
blva: is this a leftover from the other PR?

src/tools/tool.ts
```diff
@@ -303,21 +303,10 @@ export abstract class ToolBase {
                 component: "tool",
                 duration_ms: duration,
                 result: result.isError ? "failure" : "success",
+                ...metadata,
```
blva, line 296: nice thanks!

tests/accuracy/sdk/accuracyTestingClient.ts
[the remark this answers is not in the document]
nirinchev: Ugh.. yeah - must have based on the wrong branch 🤦‍♂️

blva:
note: let's make sure to trigger the perf advisor tests here via GH actions

src/tools/atlas/read/getPerformanceAdvisor.ts
```diff
@@ -130,4 +131,14 @@ export class GetPerformanceAdvisorTool extends AtlasToolBase {
...
         }
     }
+
+    protected override resolveTelemetryMetadata(
+        result: CallToolResult,
+        args: ToolArgs<typeof this.argsShape>,
+        extra: RequestHandlerExtra<ServerRequest, ServerNotification>
+    ): TelemetryToolMetadata {
+        const baseMetadata = super.resolveTelemetryMetadata(result, args, extra);
+        baseMetadata.operations = args.operations;
```
blva: [q] why not capture this new field in `TelemetryToolMetadata`? keeping track of the data we populate there is handy - makes it easy to get a list of properties in the long term

nirinchev: That's fair - I think it's a bit annoying if it becomes unwieldy as we add more and more stuff. But you're right that by being overly permissive, we lose oversight into what's being sent to telemetry. I'll try to find a better solution.

kylelai1 (approved):
LGTM!! No specific qualms on the perf advisor tool or testing from me.

PR #707, opened 2025-10-29
chore: skip atlas tests in mongodb forks

blva:
## Proposed changes

- dependabot prs were failing because we were not properly skipping local tests 

## Checklist


gagik:
Come to think of it: should we skip these tests or just require manual approval for them? I imagine some dep updates might be entirely about Atlas-related stuff

blva:
> Come to think of it: should we skip these tests or just require manual approval for them? I imagine some dep updates might be entirely about Atlas-related stuff

I can check if that's not already supported and add as a follow-up, want to clean up our dependabot PRs first

Issue #705, opened 2025-10-28
[Bug]: Unable to connect using connection string


Ateeb626:
### Version

1.2.0

### App

- [ ] Cursor
- [ ] Windsurf
- [ ] VSCode
- [ ] VSCode Insiders
- [ ] Claude Desktop
- [x] Other

### Affected Models (if applicable)

- [ ] Claude 3.5 Sonnet
- [ ] Claude 3.7 Sonnet
- [ ] GPT-4a
- [ ] o4-mini
- [x] Other

### Bug Description

I am unable to connect to the mcp server using the connection string,  it gives an error related to docker even though I am not using docker to connect 

<img width="905" height="141" alt="Image" src="https://github.com/user-attachments/assets/6613001f-6e3a-4dad-8f29-8482aebdceac" />

jeroenvervaeke:
Hi thanks for opening this issue.

Could you confirm the following for me: after you see the error message in the console, does the process exit? (Or does the process keep running?)

Could you also share:
- which tool you're using to interact with our MCP server
- tell us how you are running MongoDB locally

Thanks!

Ateeb626:
> Hi thanks for opening this issue.
> 
> Could you confirm the following for me: after you see the error message in the console, does the process exit? (Or does the process keep running?)
> 
> Could you also share:
> 
> * which tool you're using to interact with our MCP server
> * tell us how you are running MongoDB locally
> 
> Thanks!

Hey, it keeps running.
 I am just running it in the terminal. When I use cursor to interact with the mcp server, it does work but not like this in the terminal, so I don't think its an issue with my MongoDb setup
I haven't setup our MongoDB, I actually just ssh into it to connect to locally.

kmruiz:
Hi @Ateeb626 

can you clarify what you mean with this statement?

> When I use cursor to interact with the mcp server, it does work but not like this in the terminal, so I don't think its an issue with my MongoDb setup

What do you mean that it works? To clarify, the MCP Server is not meant to be used from the terminal. You start the server and connect with Cursor/Copilot/Windsurf, and if you are able to query your MongoDB cluster in your IDE, it means it works fine.

If your Cursor is in your machine, and the MCP Server is in another machine, and Cursor can't connect to the MCP Server, you will need to ensure that:

1. The MCP Server allows remote connections (in the README.md and our docs we documented best practices for this)
2. There is no firewall blocking the connection between the MCP Server and Cursor.

[...]

PR #708, opened 2025-10-29
fix(atlas-local): improve error message when Docker is unavailable

jeroenvervaeke:
## Proposed changes
Replaced scary error messages with user-friendly messages that don't expose technical error details. When Docker is unavailable, show a simple message indicating Atlas Local tools are disabled but other tools continue to work normally.

Example output
```
[INFO] 1006001 - streamableHttpTransport: Server started on http://127.0.0.1:3000 (90571)
Cannot connect to Docker. Atlas Local tools are disabled. All other tools continue to work normally.
```

Jira: [\[MCP-276\] \[mongodb-mcp-server\] Make "docker unavailable" less scary.](https://jira.mongodb.org/browse/MCP-276)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -147,31 +147,28 @@ When adding new tools to the MCP server:
 
 ## Release Process
 
-Our release process is automated using GitHub Actions workflows:
+To release a new version of the MCP server, follow these steps:
 
-### Version Bumping
-
-1. To create a new version, go to the GitHub repository Actions tab
-2. Select the "Version Bump" workflow
-3. Click "Run workflow" and choose one of the following options:
+1. Ensure there is a Jira _Release_ ticket in the [`MCP` project](https://jira.mongodb.org/projects/MCP) for the new release and move it to _In Progress_.
+2. Verify that the Jira tickets you expect to be released are correctly mapped to the expected Release version. Add any additional required documentation to the release ticket.
+3. To create a new version, go to the GitHub repository Actions tab and run the "Prepare Release" workflow with one of the following options:
    - `patch` (e.g., 1.0.0 → 1.0.1) for backward-compatible bug fixes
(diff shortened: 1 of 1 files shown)


src/common/atlasLocal.ts
```diff
@@ -12,18 +12,14 @@ export const defaultCreateAtlasLocalClient: AtlasLocalClientFactoryFn = async ()
...
+                "Cannot connect to Docker. Atlas Local tools are disabled. All other tools continue to work normally."
             );
         }
-    } catch (importError) {
+    } catch {
         console.warn(
-            "Failed to import Atlas Local client (platform not supported), atlas-local tools will be disabled (error: ",
-            importError,
-            ")"
+            "Atlas Local is not supported on this platform. Atlas Local tools are disabled. All other tools continue to work normally."
```
gagik, line 22: should we not include the error message? seems useful. we're only assuming it's not supported. Maybe check the type of error and based on that show this or a different error message?

gagik: although I understand it from perspective of "they likely don't care about it".

jeroenvervaeke: I removed it because end users get confused by the error message. (And they likely don't care about it)

The error message will always say the platform is not supported; that's the only reason this import would fail, so it's essentially duplicate information.

nirinchev: Did we get any reports of users getting confused by it? I wonder what platform were they on that we don't currently support.

jeroenvervaeke: Yes, this is the ticket/reason why I opened this PR: https://github.com/mongodb-js/mongodb-mcp-server/issues/705

They did not get the error due to the unsupported platform, but in the other branch, a few lines higher, when Docker is not running.

kmruiz (approved):

PR #709, opened 2025-10-31
chore: automatic README generation from user config schema

gagik:
This is another step towards making the user schema the one source of truth about our configuration.

### DIFF
--- a/eslint-rules/enforce-zod-v4.test.js
+++ b/eslint-rules/enforce-zod-v4.test.js
@@ -1,0 +1,153 @@
+import path from "path";
+import { RuleTester } from "eslint";
+import { describe, it } from "vitest";
+import tsParser from "@typescript-eslint/parser";
+import rule from "./enforce-zod-v4.js";
+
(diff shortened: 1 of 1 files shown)


eslint-rules/enforce-zod-v4.js
```diff
@@ -0,0 +1,54 @@
...
+
+// The file that is allowed to import from zod/v4
+const configFilePath = path.resolve(import.meta.dirname, "../src/common/config.ts");
+
+// Ref: https://eslint.org/docs/latest/extend/custom-rules
+export default {
+    meta: {
+        type: "problem",
+        docs: {
+            description: "Only allow importing 'zod/v4' in config.ts, all other imports are allowed elsewhere.",
```
gagik: zod v4 has some nice features (and actually exists already as a separate import in our current v3 version) but we want to make sure we don't use this for tool schemas until https://github.com/modelcontextprotocol/typescript-sdk/issues/555 is resolved

addaleax (approved):

PR #710, opened 2025-10-31
chore: get default user config from schema

gagik:
This removes a lot of boilerplate validation code and makes the UserConfigSchema the main source of truth for our parsing. The only part which is still "unsafe" is arguments we haven't defined in the schema which are part of arg-parser options. But those are also the args we don't use directly.

There's actually a couple more follow-ups I want to do with i.e. storing secrets but don't want to make this too big.

### DIFF
--- a/src/tools/mongodb/create/createIndex.ts
+++ b/src/tools/mongodb/create/createIndex.ts
@@ -4,7 +4,7 @@ import { DbOperationArgs, MongoDBToolBase } from "../mongodbTool.js";
 import { type ToolArgs, type OperationType } from "../../tool.js";
 import type { IndexDirection } from "mongodb";
 import { quantizationEnum } from "../../../common/search/vectorSearchEmbeddingsManager.js";
-import { similarityValues, type Similarity } from "../../../common/schemas.js";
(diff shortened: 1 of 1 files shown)


src/common/schemas.ts
```diff
@@ -0,0 +1,7 @@
+import z4 from "zod/v4";
+import z3 from "zod/v3";
+
+const similarityValues = ["cosine", "euclidean", "dotProduct"] as const;
+
+export const similarityEnumV4 = z4.enum(similarityValues);
+export const similarityEnum = z3.enum(similarityValues);
```
gagik: this is one con of this transitional period but it'd be easy to switch over afterwards

tsconfig.cjs.json
```diff
@@ -1,6 +1,7 @@
 {
   "extends": "./tsconfig.build.json",
   "compilerOptions": {
+    "noCheck": true,
```
gagik: the CJS compiler gets confused by some Zod type things for some tools. We already check in ESM anyways so it seems worth skipping type checks here anyhow. What we're interested in is the build succeeding.

kmruiz (approved):

PR #714, opened 2025-11-04
chore: reconfigure permission model for Github actions MCP-279

himanshusinghs:
## Proposed changes
Modifies the Github workflows (Code Health and Code Health from fork) to work with least privileges and fixes the accidental problem where dependabot created PRs and forked PRs were not being tested for the modified dependencies and PR contents.

Additionally, we don't persist Git credentials when not required.

## Checklist

### DIFF
--- a/.github/workflows/cleanup-atlas-env.yml
+++ b/.github/workflows/cleanup-atlas-env.yml
@@ -13,6 +13,8 @@ jobs:
     steps:
       - uses: GitHubSecurityLab/actions-permissions/monitor@v1
       - uses: actions/checkout@v5
+        with:
+          persist-credentials: false
       - uses: actions/setup-node@v6
         with:
           node-version-file: package.json
--- a/.github/workflows/accuracy-tests.yml
+++ b/.github/workflows/accuracy-tests.yml
@@ -29,6 +29,8 @@ jobs:
     steps:
       - uses: GitHubSecurityLab/actions-permissions/monitor@v1
       - uses: actions/checkout@v5
+        with:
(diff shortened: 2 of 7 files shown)


.github/workflows/code-health-fork.yml
```diff
@@ -1,18 +1,17 @@
 ---
 name: Code Health (fork)
 on:
-  pull_request_target:
```
fmenezes, line 4: if we are changing this, do we need separated workflows for fork and non fork?

himanshusinghs: We already have separated workflows for both forks (code-health-fork.yml) and non-forks (code-health.yml). The problem earlier was that code-health-fork.yml was never testing the actual changes of the pull request because the workflow was triggered by `pull_request_target` trigger which was acting in the context of `main` branch and never checking out the pull request changes.

.github/workflows/cleanup-atlas-env.yml
```diff
@@ -13,6 +13,8 @@ jobs:
     steps:
       - uses: GitHubSecurityLab/actions-permissions/monitor@v1
       - uses: actions/checkout@v5
```
nirinchev, line 15: As an alternative, consider switching to v6-beta as that has a mitigation for the persist-credentials issue.

himanshusinghs: The release is only 3 days. I am thinking of letting it marinate a bit before pulling it in. Will keep on eye on it though. Thanks for the heads up.

PR #716, opened 2025-11-06
chore: add connection metadata to telemetry

blva:
## Proposed changes
- rename atlas tool metadata to just AtlasMetadata
- added connection auth type to all mongodb commands and connect commands

## Checklist

### DIFF
--- a/src/tools/atlas/connect/connectCluster.ts
+++ b/src/tools/atlas/connect/connectCluster.ts
@@ -308,11 +308,11 @@ export class ConnectClusterTool extends AtlasToolBase {
 
     protected override resolveTelemetryMetadata(
         result: CallToolResult,
-        args: Parameters<ToolCallback<typeof this.argsShape>>
+        ...args: Parameters<ToolCallback<typeof this.argsShape>>
     ): ConnectionMetadata {
         const parentMetadata = super.resolveTelemetryMetadata(result, ...args);
         const connectionMetadata = this.getConnectionInfoMetadata();
-        if (connectionMetadata.project_id !== undefined) {
+        if (connectionMetadata && connectionMetadata.project_id !== undefined) {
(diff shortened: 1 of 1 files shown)


src/telemetry/types.ts
```diff
@@ -141,17 +141,22 @@ export type CommonProperties = {
...
 };
 
-export type PerfAdvisorToolMetadata = AtlasToolMetadata & {
+export type PerfAdvisorToolMetadata = AtlasMetadata & {
     operations: string[];
 };
+
+export type ConnectionMetadata = AtlasMetadata &
+    AtlasLocalToolMetadata & {
+        connection_auth_type?: string;
```
blva, line 157: planning to add host info here - starting with connection auth type as an example of something we can capture

src/tools/atlas/connect/connectCluster.ts
```diff
@@ -303,4 +305,24 @@ export class ConnectClusterTool extends AtlasToolBase {
...
+        const { project_id, org_id, ...restConnectionMetadata } = connectionMetadata;
+        const finalProjectId = parentMetadata.project_id ?? project_id;
+        const finalOrgId = parentMetadata.org_id ?? org_id;
+        return {
+            ...parentMetadata,
+            ...restConnectionMetadata,
+            // Only include project_id and org_id if they are defined
+            ...(finalProjectId !== undefined && { project_id: finalProjectId }),
+            ...(finalOrgId !== undefined && { org_id: finalOrgId }),
+        };
```
himanshusinghs: This bit is duplicated, maybe we can extract this to a small function and re-use?

blva: this was actually such a code smell! i've fixed it with only the relevant items

PR #717, opened 2025-11-07
chore: adds field embeddings validation for quantization "none" and warn when vectorSearch is not configured correctly

himanshusinghs:
## Proposed changes
Proposes the following:
1. Modifies the description of createIndex tool's index definition object. We will mention `vectorSearch` only when the feature is enabled. Additionally adds additional tests for the metadata - [198b36e](https://github.com/mongodb-js/mongodb-mcp-server/pull/717/commits/198b36e7b26fd73dd87c935e78630001ba55582b)
2. Adds validation for document field embeddings when the index quantization is set to `none` - [5cd2e51](https://github.com/mongodb-js/mongodb-mcp-server/pull/717/commits/5cd2e51f0e78ec272b4d2524a2a1e5c8dd5001f3)
3. Adds a warning when vector search is not correctly configured - [fbb3d0f](https://github.com/mongodb-js/mongodb-mcp-server/pull/717/commits/fbb3d0f35f4b7ca8e04210b84103acef37256850)

## Checklist

### DIFF
--- a/src/common/search/vectorSearchEmbeddingsManager.ts
+++ b/src/common/search/vectorSearchEmbeddingsManager.ts
@@ -190,11 +190,13 @@ export class VectorSearchEmbeddingsManager {
                 try {
                     return fieldRef.toFloat32Array();
                 } catch {
-                    try {
-                        return fieldRef.toBits();
-                    } catch {
-                        return undefined;
-                    }
+                    // nothing to do here
+                }
+
+                try {
+                    return fieldRef.toBits();
+                } catch {
+                    // nothing to do here
                 }
             }
 


src/common/search/vectorSearchEmbeddingsManager.ts
```diff
@@ -198,9 +198,33 @@ export class VectorSearchEmbeddingsManager {
...
+                        error: "dimension-mismatch",
+                    });
+                }
+
+                if (!fieldRef.every((e) => this.isANumber(e))) {
+                    return constructError({
+                        actualNumDimensions: fieldRef.length,
+                        actualQuantization: "none",
+                        error: "not-numeric",
+                    });
```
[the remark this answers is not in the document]
himanshusinghs: Good point - addressed in my recent commit.

src/common/config.ts
```diff
@@ -372,6 +372,24 @@ export function registerKnownSecretsInRootKeychain(userConfig: Partial<UserConfi
     maybeRegister(userConfig.username, "user");
 }
 
+function warnIfVectorSearchNotEnabledCorrectly(config: UserConfig): void {
+    const vectorSearchEnabled = config.previewFeatures.includes("vectorSearch");
+    const embeddingsProviderConfigured = !!config.voyageApiKey;
+    if (vectorSearchEnabled && !embeddingsProviderConfigured) {
+        console.warn(`\
+Warning: Vector search is enabled but no embeddings provider is configured.
+- Set the 'voyageApiKey' configuration option to enable auto-embeddings during document insertion and text-based queries with $vectorSearch.\
```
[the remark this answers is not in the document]
himanshusinghs: Addressed

src/tools/mongodb/read/aggregate.ts
```diff
@@ -32,20 +31,26 @@ If the user requests additional filtering, include filters in \`$vectorSearch.fi
...
-    ),
-    responseBytesLimit: z.number().optional().default(ONE_MB).describe(`\
+- If no requested filters are valid prefilters, omit the filter key from $vectorSearch.\
+`;
+
+export const getAggregateArgs = (vectorSearchEnabled: boolean) =>
+    ({
+        pipeline: z
+            .array(vectorSearchEnabled ? z.union([AnyAggregateStage, VectorSearchStage]) : AnyAggregateStage)
+            .describe(pipelineDescription),
```
nirinchev: The pipeline description here is static and includes references to vector search - should we modify it so that we don't include them if the feature is disabled?

himanshusinghs: That's certainly an option but I think that the description applies in a generic sense to the pipeline and I don't think it would be confusing for the LLMs. Or do you feel otherwise?

nirinchev: Well, most of the description is usage rules for the `$vectorSearch` and when vector search is disabled, we take in any ejson for the stages - I don't know if it'll be confusing to the LLM but generally, the way the description is worded, it does imply one could run a $vectorSearch query (which admittedly, they can, as long as they provide the embeddings).

src/common/config.ts
```diff
@@ -372,6 +372,24 @@ export function registerKnownSecretsInRootKeychain(userConfig: Partial<UserConfi
     maybeRegister(userConfig.username, "user");
 }
 
+function warnIfVectorSearchNotEnabledCorrectly(config: UserConfig): void {
+    const vectorSearchEnabled = config.previewFeatures.includes("vectorSearch");
+    const embeddingsProviderConfigured = !!config.voyageApiKey;
+    if (vectorSearchEnabled && !embeddingsProviderConfigured) {
+        console.warn(`\
```
gagik: You use warn in other places but `console.warn` here, is it intentional?

gagik (approved):

src/common/config.ts
himanshusinghs: Ahh - that's because the place where we use `warn` is where the console.warn is being provided through the function arguments. Likely for some tests. I had the tests covered differently so it should be fine.

himanshusinghs: Actually no, the new warning is not covered. I will write a test quickly.

himanshusinghs: added tests in [2e10e74](https://github.com/mongodb-js/mongodb-mcp-server/pull/717/commits/2e10e748e4f8b69542b2286cd905f9ef47f6eb3c)

src/tools/mongodb/read/aggregate.ts
himanshusinghs: Updated description in [c6b6331](https://github.com/mongodb-js/mongodb-mcp-server/pull/717/commits/c6b6331bd583e4f43b6cbb8af44d13d04254bde1)

src/common/config.ts
```diff
@@ -319,7 +319,7 @@ export function warnAboutDeprecatedOrUnknownCliArgs(
     if (knownArgs.connectionString) {
         usedDeprecatedArgument = true;
         warn(
-            "The --connectionString argument is deprecated. Prefer using the MDB_MCP_CONNECTION_STRING environment variable or the first positional argument for the connection string."
+            "Warning: The --connectionString argument is deprecated. Prefer using the MDB_MCP_CONNECTION_STRING environment variable or the first positional argument for the connection string."
```
gagik, line 322: shouldn't `warn` already prefix warning? it's weird to hard code this every time.

himanshusinghs: Makes sense. I will adapt this.

tests/integration/tools/mongodb/create/insertMany.test.ts
```diff
@@ -169,14 +187,26 @@ describeWithMongoDB(
                     database: database,
                     collection: "test",
                     documents: [{ embedding: "oopsie" }],
+                    // Note: We are intentionally commenting out the
```
gagik, line 190: nit: i feel like this is already documented by the text expectations. I feel like the need to specify this is intentional makes me wonder if we should focus on what's missing

src/common/search/vectorSearchEmbeddingsManager.ts
```diff
@@ -197,70 +193,55 @@ export class VectorSearchEmbeddingsManager {
...
-                        }
-                    }
-                } else {
-                    if (!Array.isArray(fieldRef)) {
-                        return constructError({
-                            error: "not-a-vector",
-                        });
-                    }
+        if (fieldRef instanceof BSON.Binary) {
+            try {
```
gagik: can we separate this try-s into helper functions which returns undefined if it is unable to convert? This way we can have prevent all this nesting.
so something like 
```ts
const extractedElements = extractElements(fieldRef);
if (extractedElements == undefined) {
 return { error: "not-a-vector" };
}
...
extractElements = (fieldRef) {
     try {
           return fieldRef.toFloat32Array();
     } catch { }
     try {
           return fieldRef.toBits();
     } catch { }
     if (!Array.isArray(fieldRef)) {
           return undefined;
     }
     return fieldRef;
}
```
the numeric check could be included there too

src/common/config.ts
himanshusinghs: Just now noticed that I don't always prefix with warning, at times it is with a small indent. Additionally since the `warn` method argument is particularly to ease the testing, I think its best to leave the text as is to assert what needs to be logged.

tests/integration/tools/mongodb/create/insertMany.test.ts
himanshusinghs: Ahh unfortunately for me the test expectation alone were not clear enough. I instead started worrying about the correctness of this test because we were not even providing embeddingsParameter when that is to be expected to validate any embeddings.

src/common/search/vectorSearchEmbeddingsManager.ts
himanshusinghs: Thanks for the suggestion. It indeed made following through the logic much simpler, here it is - [1fcc7c7](https://github.com/mongodb-js/mongodb-mcp-server/pull/717/commits/1fcc7c73bdb596e869f904e174ee56d008dffb2f)

src/common/search/vectorSearchEmbeddingsManager.ts
```diff
@@ -185,6 +185,26 @@ export class VectorSearchEmbeddingsManager {
...
 
+        const extractUnderlyingVector = (fieldRef: unknown): ArrayLike<unknown> | undefined => {
+            if (fieldRef instanceof BSON.Binary) {
+                try {
+                    return fieldRef.toFloat32Array();
+                } catch {
+                    try {
+                        return fieldRef.toBits();
+                    } catch {
+                        return undefined;
```
gagik: this way the array case will never happen, I think you should have 2 empty catch statements instead of nesting

tests/integration/tools/mongodb/create/insertMany.test.ts
gagik: no strong feelings but imo it's better to add comments to the expectation (and add extra checks if necessary). I find this is a bit weird to have a half-commented "ideal" object but I also get your point

src/common/config.ts
gagik: > I don't always prefix with warning, at times it is with a small indent

I'd expect us to just always do this right? Is it because it's multi-line? Anyway, I'm fine keeping as is.

himanshusinghs: Yeap, there are parts which are optionally multiline so the prefix won't always apply.

src/common/search/vectorSearchEmbeddingsManager.ts
himanshusinghs: The array case isn't supposed to happen when the field is a `BSON.Binary`. In the original code, the array case was part of the else branch. Now, because we are always returning from the if branch, I thought of the else as unnecessary. Happy to bring it back if it is causing confusion.

himanshusinghs: Looks like I was mistaken. We still need to extract the underlying field so certainly need to follow through to the array case.

himanshusinghs: addressed [84a6105](https://github.com/mongodb-js/mongodb-mcp-server/pull/717/commits/84a61050552f01166f8ea086a578a1b946b70a6f)

PR #720, opened 2025-11-10
fix: operation ids in the openapi spec

fmenezes:
## Proposed changes
fixes the case where the openapi spec was not updating given the operation IDs changed

## Checklist

### DIFF
--- a/tests/integration/tools/atlas/performanceAdvisor.test.ts
+++ b/tests/integration/tools/atlas/performanceAdvisor.test.ts
@@ -49,6 +49,7 @@ describeWithAtlas("performanceAdvisor", (integration) => {
                         rootCertType: "ISRGROOTX1",
                         terminationProtectionEnabled: false,
                         versionReleaseSystem: "LTS",
+                        useAwsTimeBasedSnapshotCopyForFastInitialSync: false,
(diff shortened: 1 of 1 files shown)


src/common/atlas/apiClient.ts
```diff
@@ -351,213 +351,7 @@ export class ApiClient {
...
-    async listProjectIpAccessLists(options: FetchOptions<operations["listProjectIpAccessLists"]>) {
-        const { data, error, response } = await this.client.GET("/api/atlas/v2/groups/{groupId}/accessList", options);
-        if (error) {
-            throw ApiClientError.fromError(response, error);
-        }
-        return data;
-    }
-
-    // eslint-disable-next-line @typescript-eslint/explicit-function-return-type
-    async createProjectIpAccessList(options: FetchOptions<operations["createProjectIpAccessList"]>) {
```
jeroenvervaeke, line 363: [q] what happened to all these delete ones, are they unused?

fmenezes: I'm checking

fmenezes: this is to do with `x-xgen-operation-id-override`

PR #721, opened 2025-11-10
fix: creating multiple users on atlas connect [MCP-280]

fmenezes:
## Proposed changes

sometimes atlas connect tool would create multiple users instead of one, this fixes it.

## Checklist

### DIFF
--- a/tests/integration/tools/atlas/clusters.test.ts
+++ b/tests/integration/tools/atlas/clusters.test.ts
@@ -218,7 +218,7 @@ describeWithAtlas("clusters", (integration) => {
                             }
                         });
 
-                        it("disconnects before connecting to another cluster", async () => {
+                        it("disconnects and deletes the database user before connecting to another cluster", async () => {
                             const deleteDatabaseUserSpy = vitest.spyOn(
                                 integration.mcpServer().session.apiClient,
                                 "deleteDatabaseUser"


src/tools/atlas/connect/connectCluster.ts
```diff
@@ -218,6 +218,39 @@ export class ConnectClusterTool extends AtlasToolBase {
...
+            case "disconnected": {
+                await this.session.disconnect();
+
+                const { connectionString, atlas } = await this.prepareClusterConnection(
+                    projectId,
+                    clusterName,
+                    connectionType
+                );
+
+                createdUser = true;
```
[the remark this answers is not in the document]
fmenezes: prepareClusterConnection will always create a user

src/tools/atlas/connect/connectCluster.ts
```diff
@@ -218,6 +218,39 @@ export class ConnectClusterTool extends AtlasToolBase {
         const ipAccessListUpdated = await ensureCurrentIpInAccessList(this.session.apiClient, projectId);
         let createdUser = false;
 
+        const state = this.queryConnection(projectId, clusterName);
```
blva, line 221: could we also add integration test coverage to avoid regressions in the future? 

We could improve coverage with:
 1. validate that after we connect, only on DB user is created 
 2. have a scenario tun where we connect to a cluster then disconnect and validate that the DB user is deleted

fmenezes: done!

tests/integration/tools/atlas/clusters.test.ts
```diff
@@ -172,6 +187,58 @@ describeWithAtlas("clusters", (integration) => {
...
             });
 
+            describe("when connected", () => {
+                withCluster(
+                    integration,
+                    ({ getProjectId: getSecondaryProjectId, getClusterName: getSecondaryClusterName }) => {
+                        beforeAll(async () => {
+                            let connected = false;
+                            for (let i = 0; i < 10; i++) {
+                                const response = await integration.mcpClient().callTool({
```
blva, line 197: [q] why do we need to call connect several times, shouldn't we connect and wait a considerable amount of time and that should be enough?

tests/integration/tools/atlas/clusters.test.ts
```diff
@@ -172,6 +187,58 @@ describeWithAtlas("clusters", (integration) => {
...
+
+                                await sleep(500);
+                            }
+
+                            if (!connected) {
+                                throw new Error("Could not connect to cluster before tests");
+                            }
+                        });
+
+                        it("disconnects before connecting to another cluster", async () => {
```
blva:
```suggestion
                        it("disconnects and deletes the database user before connecting to another cluster", async () => {
```

blva (approved):
non-blocking comments, thanks for the fix and coverage!

tests/integration/tools/atlas/clusters.test.ts
fmenezes: this is the edge case where atlas can take a long time to provision the DB user to data plane, in which case we need to keep waiting

PR #727, opened 2025-11-12
chore: add pre-commit hooks [MCP-284]

fmenezes:
## Proposed changes

Add pre-commit hook to run `husky`, which runs `git-secrets`.

## Checklist

### DIFF
--- a/.github/workflows/code-health.yml
+++ b/.github/workflows/code-health.yml
@@ -153,6 +153,5 @@ jobs:
       - name: Install dependencies
         run: |
           sudo apt-get install -y git-secrets
-          npm ci
       - name: Run git-secrets
         run: git-secrets --register-aws && git-secrets --scan


.husky/pre-commit
```diff
@@ -0,0 +1,13 @@
...
+then
+   echo "git-secrets is not installed. Please run 'brew install git-secrets' or visit https://github.com/awslabs/git-secrets#installing-git-secrets"
+   exit 1
+fi
+
+git-secrets --register-aws > /dev/null
+git-secrets --scan
+
+npm run check
+npm test -- --run tests/unit
```
nirinchev: Can we not do this - there's no reason to have this run as part of pre-commit.

fmenezes: this is not the entire suite only the fast unit tests, I can remove but commits will still be red, I also moved it to `pre-push`, instead

.github/workflows/code-health.yml
```diff
@@ -136,3 +136,23 @@ jobs:
...
+          persist-credentials: false
+      - uses: actions/setup-node@v6
+        with:
+          node-version-file: package.json
+          cache: "npm"
+      - name: Install dependencies
+        run: |
+          sudo apt-get install -y git-secrets
+          npm ci
+      - name: Run pre-commit and pre-push hooks
```
gagik: I understand the idea behind having those aligned but beyond git-secret, isn't the rest redundant and handled by other runs? We can turn this into a git-secrets check instead

fmenezes: sure

fmenezes: done!

.husky/pre-commit
fmenezes: as discussed offline I've removed tests and left only `git-secrets` we can revisit it later

tests/integration/fixtures/httpsServerProxyTest.ts
```diff
@@ -43,6 +43,8 @@ export class HTTPServerProxyTestSetup {
     }
 
     requests: IncomingMessage[];
+
+    // note: these are self-signed certs for testing purposes, DO NOT use in production
```
fmenezes, line 47: @blva is this enough?

.github/workflows/code-health.yml
```diff
@@ -136,3 +136,23 @@ jobs:
...
+        with:
+          persist-credentials: false
+      - uses: actions/setup-node@v6
+        with:
+          node-version-file: package.json
+          cache: "npm"
+      - name: Install dependencies
+        run: |
+          sudo apt-get install -y git-secrets
+          npm ci
```
nirinchev: Do we need to install npm packages? I don't expect we want to scan for secrets in node_modules anyway?

fmenezes: will remove

PR #729, opened 2025-11-13
chore: refactor config initialisation for CLI to allow easy extension from other config sources as well MCP-288

himanshusinghs:
## Proposed changes
This PR proposes the following:
1. Removal of custom code for ENV variable parsing in favor of using yargs-parser.
2. Removal of custom code for validating CLI arguments in favor of using yargs-parser.
3. Decouple configured connection parameters (connectionString and driverOptions derived from it) from MCPConnectionManager by removing `driverOptions` from the MCPConnectionManager constructor and moving it instead of `MCPConnectionManager.connect`
4. Remove accidental leak of yargs-parser from exported library path

Furthermore this PR fixes two accidental bugs:
1. Connection String at positional argument was only considered when `--connectionString` flag was also provided which defeats the purpose of warning against using `--connectionString`. The final effect of this was that the value of `--connectionString` would have taken hold and MCP server will try connecting to that instead of the one specified in positional argument.
> The exact reason for this were the [following lines](https://github.com/mongodb-js/mongodb-mcp-server/pull/729/files#diff-9d5ba1a74fa3c1ca2e89c9b7a2c356807802c8b5be7473377ffd4d9ab348aef1L399-L402). We fix that by narrowing down [what classifies as connectionSpecifier](https://github.com/mongodb-js/mongodb-mcp-server/pull/729/files#diff-5e24f9727151a086f1fa56b721f5264e94dc483588b7e3e4fd0dcffea1c523f2R45-R48) and explicit [handling of connectionSpecifier over anything else](https://github.com/mongodb-js/mongodb-mcp-server/pull/729/files#diff-1f749d2d65647d0d43628092cf1e8662a543ce045d8ef39c9623c423fdb47dd8R112-R120). Additional tests were added to verify the new behaviour.
2. ConnectionManager was always using the `DriverOptions` of the configured connection in the user config due to which it could never be guaranteed for `switch-connection` tool to connect successfully even when the connection string would've been correct. The problem was apparent when the driver options used by configured connection was not compatible with the connection to be established by switch-connection tool. For example - consider configuring an OIDC connection, this would've made MCPConnectionManager bound to use configured `DriverOption` for any further connection and there is no guarantee that the next connection attempted by the MCPConnectionManager, lets say via switch-connection tool, would've been an OIDC connection with exact same parameters.
> To fix this, we have decoupled MCPConnectionManager from DriverOptions by removing DriverOptions from its interface and [instead accepting a ConnectionInfo like object in the `connect` method](https://github.com/mongodb-js/mongodb-mcp-server/pull/729/files#diff-f2868ae7aecc32593eab3b565097c879aee92af8503649750b334c79df072d14R174-R177). Now its the server [which generates the correct ConnectionInfo when it is trying to connect to the configured connection](https://github.com/mongodb-js/mongodb-mcp-server/pull/729/files#diff-8a8ae07582c9d433ec8c2e5c4310ff8901e604f4965c5b90a49117ad46c47595R331-R335).

This opens path for simply using yargs-parser to also parse a configuration file in the correct order or priority.

### Notes for reviewer:
1. Start with changes in `config.ts`. The overall idea here was to:
  1.1 Remove unnecessary code - validation logic, env and cli variable parsing logic. So you will see them removed and most of the work being done instead with yargs-parser.
  1.2 Separate different entities in different readable modules so you will see that UserConfigSchema is moved to its own file and the methods creating UserConfig are moved to separate files. More than just aesthetic, this also prevents yargs-parser accidentally leaking on the lib.
2. Check changes in ConnectionManager on how we've removed DriverOptions and what we're doing instead to connect to configured connection and later to different connections.
3. The rest of the changes are sideeffects of moving files and exports around.

### Notable user facing changes:
1. Earlier we were expecting nested ENV variables to be provided using `.` notation now we use `__` instead (default from yargs-parser). Rationale behind this is that `.` notation in env variables do not work for normal usage such as passing env variable inline to command or setting them using `export`. So we're essentially fixing another odd behaviour. Good thing is that we did not have any nested config key that needed parsing of nested ENV variable.

## Checklist

### DIFF
--- a/src/tools/mongodb/mongodbTool.ts
+++ b/src/tools/mongodb/mongodbTool.ts
@@ -8,7 +8,6 @@ import { LogId } from "../../common/logger.js";
 import type { Server } from "../../server.js";
 import type { ConnectionMetadata } from "../../telemetry/types.js";
 import type { ToolCallback } from "@modelcontextprotocol/sdk/server/mcp.js";
-import { generateConnectionInfoFromCliArgs } from "@mongosh/arg-parser";
 
 export const DbOperationArgs = {
     database: z.string().describe("Database name"),
@@ -30,11 +29,7 @@ export abstract class MongoDBToolBase extends ToolBase {
 
             if (this.config.connectionString) {
                 try {
-                    const connectionInfo = generateConnectionInfoFromCliArgs({
-                        ...this.config,
-                        connectionSpecifier: this.config.connectionString,
-                    });
-                    await this.session.connectToMongoDB(connectionInfo);
+                    await this.session.connectToConfiguredConnection();
                 } catch (error) {
                     this.session.logger.error({
                         id: LogId.mongodbConnectFailure,
--- a/src/common/session.ts
+++ b/src/common/session.ts
@@ -19,11 +19,11 @@ import type { ExportsManager } from "./exportsManager.js";
 import type { Client } from "@mongodb-js/atlas-local";
 import type { Keychain } from "./keychain.js";
 import type { VectorSearchEmbeddingsManager } from "./search/vectorSearchEmbeddingsManager.js";
+import { generateConnectionInfoFromCliArgs } from "@mongosh/arg-parser";
+import { type UserConfig } from "../common/config/userConfig.js";
 
 export interface SessionOptions {
-    apiBaseUrl: string;
-    apiClientId?: string;
-    apiClientSecret?: string;
+    userConfig: UserConfig;
     logger: CompositeLogger;
     exportsManager: ExportsManager;
     connectionManager: ConnectionManager;
@@ -40,6 +40,7 @@ export type SessionEvents = {
 };
 
 export class Session extends EventEmitter<SessionEvents> {
+    private readonly userConfig: UserConfig;
     readonly sessionId: string = new ObjectId().toString();
     readonly exportsManager: ExportsManager;
     readonly connectionManager: ConnectionManager;
@@ -57,9 +58,7 @@ export class Session extends EventEmitter<SessionEvents> {
     public logger: CompositeLogger;
 
     constructor({
-        apiBaseUrl,
-        apiClientId,
-        apiClientSecret,
+        userConfig,
         logger,
         connectionManager,
         exportsManager,
@@ -69,17 +68,18 @@ export class Session extends EventEmitter<SessionEvents> {
     }: SessionOptions) {
         super();
 
+        this.userConfig = userConfig;
         this.keychain = keychain;
         this.logger = logger;
         const credentials: ApiClientCredentials | undefined =
-            apiClientId && apiClientSecret
+            userConfig.apiClientId && userConfig.apiClientSecret
                 ? {
-                      clientId: apiClientId,
-                      clientSecret: apiClientSecret,
+                      clientId: userConfig.apiClientId,
+                      clientSecret: userConfig.apiClientSecret,
                   }
                 : undefined;
 
-        this.apiClient = new ApiClient({ baseUrl: apiBaseUrl, credentials }, logger);
+        this.apiClient = new ApiClient({ baseUrl: userConfig.apiBaseUrl, credentials }, logger);
         this.atlasLocalClient = atlasLocalClient;
         this.exportsManager = exportsManager;
         this.connectionManager = connectionManager;
@@ -142,6 +142,14 @@ export class Session extends EventEmitter<SessionEvents> {
         await this.apiClient.close();
         await this.exportsManager.close();
         this.emit("close");
+    }
+
+    async connectToConfiguredConnection(): Promise<void> {
+        const connectionInfo = generateConnectionInfoFromCliArgs({
+            ...this.userConfig,
+            connectionSpecifier: this.userConfig.connectionString,
+        });
+        await this.connectToMongoDB(connectionInfo);
     }
 
     async connectToMongoDB(settings: ConnectionSettings): Promise<void> {
--- a/src/server.ts
+++ b/src/server.ts
@@ -1,4 +1,3 @@
-import { generateConnectionInfoFromCliArgs } from "@mongosh/arg-parser";
 import type { McpServer } from "@modelcontextprotocol/sdk/server/mcp.js";
 import type { Session } from "./common/session.js";
 import type { Transport } from "@modelcontextprotocol/sdk/shared/transport.js";
@@ -328,11 +327,7 @@ export class Server {
                     context: "server",
                     message: `Detected a MongoDB connection string in the configuration, trying to connect...`,
                 });
-                const connectionInfo = generateConnectionInfoFromCliArgs({
-                    ...this.userConfig,
-                    connectionSpecifier: this.userConfig.connectionString,
-                });
-                await this.session.connectToMongoDB(connectionInfo);
+                await this.session.connectToConfiguredConnection();
             } catch (error) {
                 // We don't throw an error here because we want to allow the server to start even if the connection string is invalid.
                 this.session.logger.error({
--- a/src/transports/base.ts
+++ b/src/transports/base.ts
@@ -95,9 +95,7 @@ export abstract class TransportRunnerBase {
         });
 
         const session = new Session({
-            apiBaseUrl: this.userConfig.apiBaseUrl,
-            apiClientId: this.userConfig.apiClientId,
-            apiClientSecret: this.userConfig.apiClientSecret,
+            userConfig: this.userConfig,
             atlasLocalClient: await this.atlasLocalClient,
             logger,
             exportsManager,
--- a/tests/integration/helpers.ts
+++ b/tests/integration/helpers.ts
@@ -96,9 +96,7 @@ export function setupIntegrationTest(
         const connectionManager = new MCPConnectionManager(userConfig, logger, deviceId);
 
         const session = new Session({
-            apiBaseUrl: userConfig.apiBaseUrl,
-            apiClientId: userConfig.apiClientId,
-            apiClientSecret: userConfig.apiClientSecret,
+            userConfig,
             logger,
             exportsManager,
             connectionManager,
--- a/tests/integration/telemetry.test.ts
+++ b/tests/integration/telemetry.test.ts
@@ -21,7 +21,7 @@ describe("Telemetry", () => {
 
         const telemetry = Telemetry.create(
             new Session({
-                apiBaseUrl: "",
+                userConfig: defaultTestConfig,
                 logger,
                 exportsManager: ExportsManager.init(config, logger),
                 connectionManager: connectionManager,
--- a/tests/integration/tools/mongodb/mongodbTool.test.ts
+++ b/tests/integration/tools/mongodb/mongodbTool.test.ts
@@ -99,9 +99,7 @@ describe("MongoDBTool implementations", () => {
         deviceId = DeviceId.create(logger);
         const connectionManager = new MCPConnectionManager(userConfig, logger, deviceId);
         const session = new Session({
-            apiBaseUrl: userConfig.apiBaseUrl,
-            apiClientId: userConfig.apiClientId,
-            apiClientSecret: userConfig.apiClientSecret,
+            userConfig,
             logger,
             exportsManager,
             connectionManager,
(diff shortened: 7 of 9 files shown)


tests/integration/tools/mongodb/connect/connect.test.ts
```diff
@@ -88,6 +89,72 @@ describeWithMongoDB(
...
+        it("should be able to connect to next connection and not use the connect options of the connection setup during server boot", async () => {
+            const newConnectionString = `${integration.connectionString()}`;
+            // Note: The connect function is called with OIDC options for the
+            // configured string
+            expect(connectFnSpy).toHaveBeenNthCalledWith(
+                1,
+                expect.stringContaining(`${integration.connectionString()}/?directConnection=true`),
+                expect.objectContaining({
+                    // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+                    oidc: expect.objectContaining({ openBrowser: { command: "not-a-browser" } }),
```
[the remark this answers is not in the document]
himanshusinghs: not helpful

src/common/config/createUserConfig.ts
```diff
@@ -0,0 +1,197 @@
...
+    const [maybeConnectionSpecifier, ...unknownArguments] = positionalAndUnknownArguments;
+    // If the extracted connection specifier is not a connection specifier
+    // indeed, then we push it back to the unknown arguments list. This might
+    // happen for example when an unknown argument is provided without ever
+    // specifying a positional argument.
+    if (typeof maybeConnectionSpecifier !== "string" || !isConnectionSpecifier(maybeConnectionSpecifier)) {
+        if (maybeConnectionSpecifier) {
+            unknownArguments.unshift(maybeConnectionSpecifier);
+        }
+    }
```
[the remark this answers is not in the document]
himanshusinghs: addressed

tests/unit/common/config.test.ts
```diff
@@ -448,430 +490,468 @@ describe("config", () => {
...
+        // Note: These arguments were part of original test cases before
+        // refactor of Config but because now we use yargs-parser to strictly
+        // parse the config and do not allow unknown arguments to creep into the
+        // final results, these arguments never end up in the config. It is
+        // because we have the mongosh OPTIONS copied over from the repo and the
+        // copied object does not contain these are parse targets.
+
+        //  TODO: Whenever we finish importing OPTIONS from mongosh these test
+        // cases should be good to be enabled again.
+
```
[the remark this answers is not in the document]
himanshusinghs: addressed

src/common/config/configUtils.ts
```diff
@@ -4,63 +4,51 @@ import { ALL_CONFIG_KEYS } from "./argsParserOptions.js";
...
 export function isConnectionSpecifier(arg: string | undefined): boolean {
     return (
         arg !== undefined &&
         (arg.startsWith("mongodb://") ||
             arg.startsWith("mongodb+srv://") ||
-            !(arg.endsWith(".js") || arg.endsWith(".mongodb")))
+            // Strings starting with a hyphen `-` are generally a sign of CLI
+            // flag so we exclude them from the possibility of being a
+            // connection specifier.
+            !(arg.endsWith(".js") || arg.endsWith(".mongodb") || arg.startsWith("-")))
```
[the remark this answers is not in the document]
himanshusinghs: addressed

src/common/config.ts
```diff
@@ -1,430 +0,0 @@
...
-    previewFeatures: z4
-        .preprocess(
-            (val: string | string[] | undefined) => commaSeparatedToArray(val),
-            z4.array(z4.enum(previewFeatureValues))
-        )
-        .default([])
-        .describe("An array of preview features that are enabled."),
-});
-
-export type UserConfig = z4.infer<typeof UserConfigSchema> & CliOptions;
```
himanshusinghs, line 179: Moved to userConfig.ts as it is

src/common/config.ts
```diff
@@ -1,430 +0,0 @@
...
-
-export const config = setupUserConfig({
-    cli: process.argv,
-    env: process.env,
-});
```
himanshusinghs, line 184: The pre-constructed config object has now been removed in favour of explicitly creating one using createUserConfig and injecting in MCP server. This helps avoiding unexpected side-effects on first import such as warnings and possibly process closure.

src/common/config.ts
```diff
@@ -1,430 +0,0 @@
...
-        level: "local",
-    },
-    readPreference: "secondaryPreferred",
-    writeConcern: {
-        w: "majority",
-    },
-    timeoutMS: 30_000,
-    proxy: { useEnvironmentVariableProxies: true },
-    applyProxyToOIDC: true,
-};
```
himanshusinghs, line 198: This has been moved to ConnectionManager.connect and is only used when a connection attempt is requested without ever providing any DriverOptions.

src/common/config.ts
```diff
@@ -1,430 +0,0 @@
...
-        const fieldPath = key
-            .replace("MDB_MCP_", "")
-            .split(".")
-            .map((part) => SNAKE_CASE_toCamelCase(part));
-
-        setValue(result, fieldPath, value);
-    }
-
-    return result;
-}
```
himanshusinghs, line 264: The entire function has been removed. Please note that a key change is also that earlier we were expecting nested keys to be provided separated by `.` but that is not always possible in a shell environment and certainly not for common pattern of providing env variables. This has been noted in the user facing changes #1.

We rely on yargs-parser for all this now

src/common/config.ts
```diff
@@ -1,430 +0,0 @@
...
-    // if we have a positional argument that matches a connection string
-    // store it as the connection specifier and remove it from the argument
-    // list, so it doesn't get misunderstood by the mongosh args-parser
-    if (!parsed.nodb && isConnectionSpecifier(positionalArguments[0])) {
-        parsed.connectionSpecifier = positionalArguments.shift();
-    }
-
-    delete parsed._;
-    return parsed;
-}
```
himanshusinghs, line 307: The entire function is removed. It used yargs-parser but was having additional validation for config keys which is now entirely handled by yargs-parser.

src/common/config.ts
```diff
@@ -1,430 +0,0 @@
...
-    }
-
-    if (usedInvalidArgument || usedDeprecatedArgument) {
-        warn("- Refer to https://www.mongodb.com/docs/mcp-server/get-started/ for setting up the MCP Server.");
-    }
-
-    if (usedInvalidArgument) {
-        exit(1);
-    }
-}
```
himanshusinghs, line 350: Removed and being handled differently now. All the warnings are still maintained though.

src/common/config.ts
```diff
@@ -1,430 +0,0 @@
...
-    maybeRegister(userConfig.awsIamSessionToken, "password");
-    maybeRegister(userConfig.awsSecretAccessKey, "password");
-    maybeRegister(userConfig.awsSessionToken, "password");
-    maybeRegister(userConfig.password, "password");
-    maybeRegister(userConfig.tlsCAFile, "url");
-    maybeRegister(userConfig.tlsCRLFile, "url");
-    maybeRegister(userConfig.tlsCertificateKeyFile, "url");
-    maybeRegister(userConfig.tlsCertificateKeyFilePassword, "password");
-    maybeRegister(userConfig.username, "user");
-}
```
himanshusinghs, line 373: This has been maintained in createUserConfig.ts

src/common/config.ts
```diff
@@ -1,430 +0,0 @@
...
-`);
-    }
-
-    if (!vectorSearchEnabled && embeddingsProviderConfigured) {
-        warn(`\
-Warning: An embeddings provider is configured but the 'vectorSearch' preview feature is not enabled.
-- Enable vector search by adding 'vectorSearch' to the 'previewFeatures' configuration option, or remove the embeddings provider configuration if not needed.\
-`);
-    }
-}
```
himanshusinghs, line 391: This has been maintained in createUserConfig.ts

src/common/config.ts
```diff
@@ -1,430 +0,0 @@
...
-}: {
-    config: UserConfig;
-    defaults: Partial<DriverOptions>;
-}): DriverOptions {
-    const { driverOptions } = generateConnectionInfoFromCliArgs(config);
-    return {
-        ...defaults,
-        ...driverOptions,
-    };
-}
```
himanshusinghs, line 430: Removed as these were not needed anymore

src/common/config/createUserConfig.ts
```diff
@@ -0,0 +1,196 @@
...
+import { Keychain } from "../keychain.js";
+import type { Secret } from "../keychain.js";
+import { isConnectionSpecifier, matchingConfigKey } from "./configUtils.js";
+import { OPTIONS } from "./argsParserOptions.js";
+import { UserConfigSchema, type UserConfig } from "./userConfig.js";
+
+export type CreateUserConfigHelpers = {
+    onWarning: (message: string) => void;
+    onError: (message: string) => void;
+    closeProcess: (exitCode: number) => never;
```
gagik, line 12:
```suggestion
    onClose: (exitCode: number) => never;
```
(I see value in either naming, so up to you)

src/common/config/createUserConfig.ts
```diff
@@ -0,0 +1,196 @@
...
+            // To avoid populating `_` with end-of-flag arguments we explicitly
+            // populate `--` variable and altogether ignore them later.
+            "populate--": true,
+        },
+    });
+
+    // We don't make use of end of flag arguments but also don't want them to
+    // end up alongside unknown arguments so we are extracting them and having a
+    // no-op statement so ESLint does not complain.
+    void endOfFlagArguments;
```
gagik: can you name it with `_`? or add an eslint-ignore with a comment? This is preferrable to me over a no-op promise call

src/common/config/createUserConfig.ts
```diff
@@ -0,0 +1,196 @@
...
+        deprecatedCliArgumentWarnings: cliArguments
+            .filter((argument) => argument.startsWith("--connectionString"))
+            .map((argument) => {
+                if (argument.startsWith("--connectionString")) {
+                    return "Warning: The --connectionString argument is deprecated. Prefer using the MDB_MCP_CONNECTION_STRING environment variable or the first positional argument for the connection string.";
+                }
+
+                return undefined;
+            })
+            .filter((argument) => typeof argument === "string"),
```
gagik: is this same as
```ts
cliArguments.some((argument) => argument.startsWith("--connectionString")) ? 
"Warning: The --connectionString argument is deprecated. Prefer using the MDB_MCP_CONNECTION_STRING environment variable or the first positional argument for the connection string." 
: undefined
```
?

src/tools/mongodb/mongodbTool.ts
```diff
@@ -29,7 +30,11 @@ export abstract class MongoDBToolBase extends ToolBase {
 
             if (this.config.connectionString) {
                 try {
-                    await this.connectToMongoDB(this.config.connectionString);
+                    const connectionInfo = generateConnectionInfoFromCliArgs({
+                        ...this.config,
+                        connectionSpecifier: this.config.connectionString,
```
gagik: should we have a version of this function like `this.session.connectWithConfig` that does this mapping generation + mapping for us?

src/common/config/createUserConfig.ts
himanshusinghs: Yea certainly the same. I wrote the original to allow for easy extension later in case there were more deprecated arguments. That's not in near future so I will update as suggested.

src/tools/mongodb/mongodbTool.ts
himanshusinghs: That's a good call. I added this one very recently and now there are two instances of it, perhaps the right time to pull them together.

src/common/config/createUserConfig.ts
himanshusinghs: addressed in here - [dd0d98b](https://github.com/mongodb-js/mongodb-mcp-server/pull/729/commits/dd0d98b6f0172bcaa94bf431c8bed008fab0ea85)

src/common/config/createUserConfig.ts
himanshusinghs: addressed in here - [dd0d98b](https://github.com/mongodb-js/mongodb-mcp-server/pull/729/commits/dd0d98b6f0172bcaa94bf431c8bed008fab0ea85)

src/tools/mongodb/mongodbTool.ts
himanshusinghs: addressed in here - [8e32474](https://github.com/mongodb-js/mongodb-mcp-server/pull/729/commits/8e32474a430d4647fadc1f1e4a5b1e4abfa46039)

PR #740, opened 2025-11-19
chore: add createSessionConfig hook MCP-294

gagik:
Add a new async hook for library consumers to provide a configuration right before establishing an MCP session.

### DIFF
--- a/src/transports/base.ts
+++ b/src/transports/base.ts
@@ -22,7 +22,7 @@ import type { Client } from "@mongodb-js/atlas-local";
 import { VectorSearchEmbeddingsManager } from "../common/search/vectorSearchEmbeddingsManager.js";
 import type { ToolBase, ToolConstructorParams } from "../tools/tool.js";
 
-type CreateSessionConfigFn = (userConfig: UserConfig) => Promise<UserConfig>;
+type CreateSessionConfigFn = (userConfig: UserConfig) => Promise<UserConfig> | UserConfig;
 
 export type TransportRunnerConfig = {
     userConfig: UserConfig;


src/transports/base.ts
```diff
@@ -29,7 +30,11 @@ export type TransportRunnerConfig = {
     createAtlasLocalClient?: AtlasLocalClientFactoryFn;
     additionalLoggers?: LoggerBase[];
     telemetryProperties?: Partial<CommonProperties>;
-    tools?: (new (params: ToolConstructorParams) => ToolBase)[];
+    /**
+     * Hook which allows library consumers to fetch configuration from external sources (e.g., secrets managers, APIs)
+     * or modify the existing configuration before the session is created.
+     */
+    createSessionConfig?: CreateSessionConfigFn;
```
gagik, line 39: I am not too sure if I understood the usecase for this correctly so let me know if this should be used differently.

src/transports/base.ts
```diff
@@ -22,6 +22,8 @@ import type { Client } from "@mongodb-js/atlas-local";
 import { VectorSearchEmbeddingsManager } from "../common/search/vectorSearchEmbeddingsManager.js";
 import type { ToolBase, ToolConstructorParams } from "../tools/tool.js";
 
+type CreateSessionConfigFn = (userConfig: UserConfig) => Promise<UserConfig>;
```
addaleax: Actually, maybe this could be

```suggestion
type CreateSessionConfigFn = (userConfig: UserConfig) => Promise<UserConfig> | UserConfig;
```

(slightly nicer DX)

himanshusinghs (approved):
Yea that's precisely it. Thanks for getting it it 🚀

PR #743, opened 2025-11-20
feat: allow reading config from a file and support toolMetadataOverrides MCP-292 MCP-300

himanshusinghs:
## Proposed changes

This PR introduces support for reading configuration from a file specified either by command line argument `--config` or by environment variable `MDB_MCP_CONFIG`. The configuration from the file is used as the base on top of which configuration from environment variables are applied and then config from CLI arguments. The precedence order is provided and maintained by yargs-parser itself.

Furthermore, we've added a new config property called `toolMetadataOverrides` which allows specifying name and description overrides for tool. It is basically a map of internal tool name to the override object.

```
{
  "toolMetadataOverrides": {
    "insert-many": {
      "name": "insertMany",
      "description": "..."
    },
    "delete-many": {
      "name: "deleteMany"
      // either property (name, description) is optional
    }
  }
}
```

We expect users to be using config file for specifying this configuration and ensuring that access to the config is locked down to the user running the MongoDB MCP Server. Doing the same through command line arguments is gruesome but possible and not encouraged at all. An example:

CLI Arguments (nested keys separated by dot)
```
npx mongodb-mcp-server \
--toolMetadataOverrides.insert-many.name="insertMany" \
--toolMetadataOverrides.insert-many.description="..."
```

## Checklist

### DIFF
--- a/CONFIGURATION.md
+++ b/CONFIGURATION.md
@@ -84,6 +84,8 @@ When using the `disk` logger, log files are stored in:
 
 You can override the log directory with the `logPath` option.
 
+> **🔒 Security Note:** When configuring a custom `logPath`, ensure the directory is owned and writable only by the user running the MongoDB MCP server process. On Linux/macOS, use `chmod 700` and verify ownership with `chown`. On Windows, restrict write permissions to the service account only.
+
 ### Disabled Tools
 
 You can disable specific tools or categories of tools by using the `disabledTools` option. This option accepts an array of strings,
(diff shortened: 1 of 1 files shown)


fmenezes:
please update README.md

fmenezes:
please add validation to guard rail for name collision

src/tools/tool.ts
```diff
@@ -168,14 +196,28 @@ export abstract class ToolBase {
...
+        /**
+         * This is very similar to RegisteredTool.update, but without the bugs
+         * around the name. In the upstream update method, the name is captured
+         * in the closure and not updated when the tool name changes. This means
+         * that you only get one name update before things end up in a broken
+         * state. See
+         * https://github.com/modelcontextprotocol/typescript-sdk/issues/414 for
+         * more details.
+         */
+        this.update = (updates: {
```
gagik, line 216: do we really want to make it possible to rename tool mid-way? 
This really makes our state less mutable and more messy. If someone wants to do this, they should look into using our tool-related hooks and defining custom tools as they likely need custom behavior at this point.
I'd prefer an approach where we only set the overrides on initalization, in which case we don't have getters and separate "internal"/"default" fields.

himanshusinghs: We are already renaming tools midway. What the change is suggesting is that the rename should respect the tool metadata overrides.

scripts/generateArguments.ts
```diff
@@ -3,7 +3,7 @@
 /**
  * This script generates argument definitions and updates:
  * - server.json arrays
- * - README.md configuration table
+ * - CONFIGURATION.md configuration table
```
gagik, line 6: subjective but i like this is in README, I really hate having to click extra things to find more crucial information, it's also worse for the LLM agents that might be reading this. I think we can make the table more compact by combining the ENV and -- variable options into 1 column

himanshusinghs:
Tabling this PR for now until we have more clarity on how we want to proceed with this.

PR #744, opened 2025-11-20
chore: bump mcp sdk

nirinchev:
## Proposed changes

This bumps the MCP SDK npm package and fixes some breaking changes.

### DIFF
--- a/tests/integration/helpers.ts
+++ b/tests/integration/helpers.ts
@@ -7,7 +7,7 @@ import { McpServer } from "@modelcontextprotocol/sdk/server/mcp.js";
 import { Client } from "@modelcontextprotocol/sdk/client/index.js";
 import { InMemoryTransport } from "./inMemoryTransport.js";
 import { type UserConfig } from "../../src/common/config/userConfig.js";
-import { McpError, ResourceUpdatedNotificationSchema } from "@modelcontextprotocol/sdk/types.js";
+import { ResourceUpdatedNotificationSchema } from "@modelcontextprotocol/sdk/types.js";
 import { afterAll, afterEach, beforeAll, describe, expect, it, vi } from "vitest";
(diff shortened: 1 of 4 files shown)


src/tools/tool.ts
```diff
@@ -60,7 +60,6 @@ export abstract class ToolBase {
     protected get annotations(): ToolAnnotations {
         const annotations: ToolAnnotations = {
             title: this.name,
-            description: this.description,
```
nirinchev, line 63: AFAICT the description should never have been part of the annotations and now is rejected as it's an unknown key.

src/server.ts
```diff
@@ -100,7 +100,6 @@ export class Server {
         this.mcpServer.server.registerCapabilities({
             logging: {},
             resources: { listChanged: true, subscribe: true },
-            instructions: this.getInstructions(),
```
nirinchev, line 103: This was similarly never part of capabilities and was previously ignored. Now it's correctly rejected. I moved it to base.ts where it should live.

tests/integration/helpers.ts
```diff
@@ -325,12 +322,11 @@ export function expectDefined<T>(arg: T): asserts arg is Exclude<T, undefined |
     expect(arg).not.toBeNull();
 }
 
-function validateToolAnnotations(tool: ToolInfo, name: string, description: string): void {
+function validateToolAnnotations(tool: ToolInfo, name: string, operationType: OperationType): void {
     expectDefined(tool.annotations);
     expect(tool.annotations.title).toBe(name);
-    expect(tool.annotations.description).toBe(description);
 
-    switch (tool.operationType) {
```
nirinchev, line 333: This never worked - `operationType` would always be `undefined`, resulting in this switch getting largely ignored. This now fails correctly, which forced me to propagate the expectation across multiple tests, which is the reason for most of the changes here.

himanshusinghs (approved):

PR #747, opened 2025-11-24
feat: introduces dry mode for CLI MCP-297

himanshusinghs:
## Proposed changes

Introduces dry mode for dumping used configuration and enabled tools.

## Checklist

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -168,6 +168,7 @@ export async function handleDryRunRequest(config: UserConfig): Promise<never> {
             },
         });
         await runner.start();
+        await runner.close();
         process.exit(0);
     } catch (error) {
         console.error(`Fatal error running server in dry run mode: ${error as string}`);


src/common/config/argsParserOptions.ts
```diff
@@ -60,6 +60,7 @@ export const OPTIONS = {
         "apiDeprecationErrors",
         "apiStrict",
         "disableEmbeddingsValidation",
+        "dry",
```
gagik: I don't think this should live here

src/transports/dryModeRunner.ts
```diff
@@ -0,0 +1,79 @@
...
+    },
+    error(message) {
+        console.error(message);
+    },
+};
+
+export class DryModeRunner extends TransportRunnerBase {
+    private server: Server | undefined;
+    private exitProcess: DryModeTestHelpers["exit"];
+    private consoleLogger: DryModeTestHelpers["logger"];
```
gagik:
```suggestion
    private logger: DryModeTestHelpers["logger"];
```
(nit) this also isn't dependent on a console

src/transports/dryModeRunner.ts
```diff
@@ -0,0 +1,79 @@
...
+        const tools = this.server?.tools.map((tool) => ({
+            name: tool.name,
+            description: tool.description,
+            category: tool.category,
+        }));
+        this.consoleLogger.log("Enabled tools:");
+        this.consoleLogger.log(JSON.stringify(tools, null, 2));
+    }
+
+    static async assertDryMode(
```
gagik: this isn't really an assertion, is it? 
I think we should move this logic into `index.ts`

```ts
new this(...)
```
I never knew you could do that 😃 JS is cool and terrible

src/transports/dryModeRunner.ts
```diff
@@ -0,0 +1,79 @@
...
+    private exitProcess: DryModeTestHelpers["exit"];
+    private consoleLogger: DryModeTestHelpers["logger"];
+
+    constructor({ exit, logger, ...transportRunnerConfig }: DryModeRunnerConfig) {
+        super(transportRunnerConfig);
+        this.exitProcess = exit;
+        this.consoleLogger = logger;
+    }
+
+    async start(): Promise<void> {
```
gagik: why not just have this do everything? i.e. combine all the dump functions into it.

src/common/config/argsParserOptions.ts
gagik: TIL we have been doing this for all our variables... that's not great :(

nirinchev: Should this be `dryRun` instead? `dry` sounds a bit awkward.

himanshusinghs: > I don't think this should live here
yargs-parser parses the arguments considering these `OPTIONS` provided to it and gathers the rest (ones that are not part of `OPTIONS`) as unknown arguments which we use further to post warnings about unknown arguments passed.

The alternative is to not pass yargs-parser these `OPTIONS` and do the verification manually which I don't think bring much value.

src/transports/dryModeRunner.ts
himanshusinghs: but it _is_ dependent on `console` :) `logger` is a base class property which I did not want to override because otherwise I would have to conform to the interface.

And because in dryMode we plan to exit as soon as we dump, there is not much sense in logging via Session loggers. This should ideally work the same way as our warnings do and they work through `console` logging.

src/common/config/argsParserOptions.ts
himanshusinghs: dry sounds a bit awkward.

> more like dry sounds a bit dry? 😄 I have updated the name to `dryRun` instead

src/transports/dryModeRunner.ts
himanshusinghs: > this isn't really an assertion, is it?

thanks for pointing this. I have renamed the helpers to be more aligned with what they do.

src/transports/dryModeRunner.ts
himanshusinghs: I followed on the suggestion and have combined dump calls within the start method itself.

tests/integration/helpers.ts
```diff
@@ -5,7 +5,7 @@ import { Server, type ServerOptions } from "../../src/server.js";
 import { Telemetry } from "../../src/telemetry/telemetry.js";
 import { McpServer } from "@modelcontextprotocol/sdk/server/mcp.js";
 import { Client } from "@modelcontextprotocol/sdk/client/index.js";
-import { InMemoryTransport } from "./inMemoryTransport.js";
+import { InMemoryTransport } from "../../src/transports/inMemoryTransport.js";
```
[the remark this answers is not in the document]
himanshusinghs: The file is in the diff of the PR, it was just in earlier commits.

src/transports/dryModeRunner.ts
```diff
@@ -0,0 +1,49 @@
...
+    private dumpConfig(): void {
+        this.consoleLogger.log("Configuration:");
+        this.consoleLogger.log(JSON.stringify(this.userConfig, null, 2));
+    }
+
+    private dumpTools(): void {
+        const tools = this.server?.tools.map((tool) => ({
+            name: tool.name,
+            category: tool.category,
+        }));
```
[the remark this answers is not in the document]
himanshusinghs: Makes sense, will apply this.

tests/e2e/cli.test.ts
```diff
@@ -0,0 +1,29 @@
...
+import { promisify } from "util";
+import { describe, expect, it } from "vitest";
+import packageJson from "../../package.json" with { type: "json" };
+
+const execAsync = promisify(exec);
+const CLI_PATH = path.join(import.meta.dirname, "..", "..", "dist", "index.js");
+
+describe("CLI entrypoint", () => {
+    it("should handle version request", async () => {
+        const { stdout, stderr } = await execAsync(`${process.execPath} ${CLI_PATH} --version`);
```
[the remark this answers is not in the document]
himanshusinghs: Good call. I have addressed this in my last commit.

tests/e2e/cli.test.ts
```diff
@@ -0,0 +1,29 @@
...
+        expect(stdout).toContain(packageJson.version);
+        expect(stderr).toEqual("");
+    });
+
+    it("should handle help request", async () => {
+        const { stdout, stderr } = await execAsync(`${process.execPath} ${CLI_PATH} --help`);
```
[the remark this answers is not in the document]
himanshusinghs: Good call. I have addressed this in my last commit.

tests/e2e/cli.test.ts
```diff
@@ -0,0 +1,29 @@
...
+        expect(stdout).toContain("For usage information refer to the README.md");
+        expect(stderr).toEqual("");
+    });
+
+    it("should handle dry run request", async () => {
+        const { stdout, stderr } = await execAsync(`${process.execPath} ${CLI_PATH} --dryRun`);
```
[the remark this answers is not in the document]
himanshusinghs: Good call. I have addressed this in my last commit.

src/index.ts
```diff
@@ -133,17 +143,34 @@ main().catch((error: unknown) => {
...
 
-function assertVersionMode(config: UserConfig): void | never {
-    if (config.version) {
-        console.log(packageInfo.version);
+function handleVersionRequest(): never {
+    console.log(packageInfo.version);
+    process.exit(0);
+}
+
+export async function handleDryRunRequest(config: UserConfig): Promise<never> {
```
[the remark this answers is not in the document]
himanshusinghs: e2e for this function covers the same grounds if not more. I will update the guidelines to include e2e tests as well.

src/index.ts
```diff
@@ -133,17 +143,34 @@ main().catch((error: unknown) => {
...
+            logger: {
+                log(message): void {
+                    console.log(message);
+                },
+                error(message): void {
+                    console.error(message);
+                },
+            },
+        });
+        await runner.start();
```
nirinchev, line 170: Should we also stop the runner? Probably inconsequential, but seems like good hygiene.

himanshusinghs: addressed this

src/index.ts
```diff
@@ -133,17 +143,35 @@ main().catch((error: unknown) => {
...
 
-function assertVersionMode(config: UserConfig): void | never {
-    if (config.version) {
-        console.log(packageInfo.version);
+function handleVersionRequest(): never {
+    console.log(packageInfo.version);
+    process.exit(0);
+}
+
+export async function handleDryRunRequest(config: UserConfig): Promise<never> {
```
gagik, line 157: I think this function might as well be moved to [dryModeRunner.ts](https://github.com/mongodb-js/mongodb-mcp-server/pull/747/files#diff-d4334c4c982b73fe62b57d09c5d1c16a069f90f42598557f8c1af0f7270c25ec)

gagik: although I also am fine with keeping as is because of usage of `process.exit`

gagik: maybe these functions shouldn't process.exit themselves and instead return and then a larger function would process.exit?

gagik (approved):
LGTM, one note about process.exit

src/index.ts
himanshusinghs: > I think this function might as well be moved to [dryModeRunner.ts]

It was originally there but I moved it to index file based on the suggestion [here](https://github.com/mongodb-js/mongodb-mcp-server/pull/747#discussion_r2556901561). For me, either is fine but I still moved it to index only to have all the CLI args handling in one place.

I think its fine for them to take control of exiting - that's a pattern I have seen commander where different args have different action paths and they all independently terminate the process.

PR #748, opened 2025-11-24
feat: add ability to override parameters using HTTP headers MCP-293

gagik:
This adds header and query param-based overrides.

### DIFF
--- a/tests/unit/common/config.test.ts
+++ b/tests/unit/common/config.test.ts
@@ -65,6 +65,7 @@ const expectedDefaults = {
     embeddingsValidation: true,
     previewFeatures: [],
     dryRun: false,
+    allowRequestOverrides: false,
 };
 
 describe("config", () => {


src/transports/base.ts
```diff
@@ -90,20 +99,20 @@ export abstract class TransportRunnerBase {
         this.deviceId = DeviceId.create(this.logger);
     }
 
-    protected async setupServer(): Promise<Server> {
+    protected async setupServer(request?: RequestContext): Promise<Server> {
+        // Apply config overrides from request context (headers and query parameters)
+        let userConfig = applyConfigOverrides({ baseConfig: this.userConfig, request });
```
gagik: should we make this config override opt-in?
Can it be problematic that people who update the MCP server don't release that others can override some other settings now?

src/common/config/configOverrides.ts
```diff
@@ -0,0 +1,159 @@
...
+    return result;
+}
+
+/**
+ * Extracts config overrides from HTTP headers or query parameters.
+ */
+function extractConfigOverrides(
+    mode: "header" | "query",
+    source: Record<string, string | string[] | undefined> | undefined
+): Partial<Record<keyof typeof UserConfigSchema.shape, unknown>> {
```
addaleax, line 76: Do you happen to know how this is different from

```suggestion
): Partial<Record<z.infer<typeof UserConfigSchema>, unknown>> {
```

? I've mostly seen us use this form rather than accessing properties directly

src/transports/base.ts
himanshusinghs: Yeah the override should be opt-in using the `--allowRequestOverrides` flag. Just realised that I did not write this info in the ticket but its further explained in Goal#3 in the scope.

src/transports/base.ts
```diff
@@ -90,20 +99,20 @@ export abstract class TransportRunnerBase {
...
-                version: packageInfo.version,
-            },
-            {
-                instructions: TransportRunnerBase.getInstructions(userConfig),
-            }
-        );
+        // fetch or modify configuration after applying request context overrides
+        if (this.createSessionConfig) {
+            userConfig = await this.createSessionConfig({ userConfig, request });
+        }
```
himanshusinghs: Since you already provide `request` to createSessionConfig hook, at this point, it would be nice to introduce a default createSessionConfig hook which applies the same override within the hook itself. The idea being that users of the hook will have a chance to actually dictate how they wish to apply the overrides if they ever decide to not use our default override behaviour.

gagik: yeah it's a worthy consideration.
I decided to go with this being a post-override step because manually applying the overrides is cumbersome and I imagine most won't bother. so it'd be better for this to be handled by default and if there is a case where they want custom overrides, then we can look into that.

src/common/config/configOverrides.ts
gagik: `z.infer<typeof UserConfigSchema>` gives the TypeScript type of the parsed/validated data i.e. what this schema is going to parse into, `typeof Schema.shape` gives the types of the schema definition itself

gagik: in some cases the latter is what I'm looking for but it's true there's cases i.e. when looking at their keys where both would work.

src/common/config/configOverrides.ts
```diff
@@ -0,0 +1,164 @@
...
+}: {
+    baseConfig: UserConfig;
+    request?: RequestContext;
+}): UserConfig {
+    if (!request) {
+        return baseConfig;
+    }
+
+    // Only apply overrides if allowRequestOverrides is enabled
+    if (!baseConfig.allowRequestOverrides) {
```
nirinchev: Do we want to error out if the client attempts to request an override instead of silently ignoring it?

src/common/config/configOverrides.ts
```diff
@@ -0,0 +1,164 @@
...
+    behavior: OverrideBehavior
+): unknown {
+    if (typeof behavior === "function") {
+        // Custom logic function returns the value to use (potentially transformed)
+        // or throws an error if the override cannot be applied
+        try {
+            return behavior(baseValue, overrideValue);
+        } catch (error) {
+            throw new Error(
+                `Cannot apply override for ${key} from ${JSON.stringify(baseValue)} to ${JSON.stringify(overrideValue)}: ${error instanceof Error ? error.message : String(error)}`
```
nirinchev: This has the potential to leak the base value to the client - do we want that?

src/common/config/configOverrides.ts
```diff
@@ -0,0 +1,164 @@
...
+    }
+    switch (behavior) {
+        case "override":
+            return overrideValue;
+
+        case "merge":
+            if (Array.isArray(baseValue) && Array.isArray(overrideValue)) {
+                return [...(baseValue as unknown[]), ...(overrideValue as unknown[])];
+            }
+            throw new Error("Cannot merge non-array values, did you mean to use the 'override' behavior?");
```
nirinchev: Can we include the key here? Otherwise it's kind of hard to know what you did wrong.

src/common/config/configUtils.ts
```diff
@@ -82,3 +106,33 @@ export function commaSeparatedToArray<T extends string[]>(str: string | string[]
...
+
+/**
+ * Preprocessor for boolean values that handles string "true"/"false" correctly.
+ * Zod's coerce.boolean() treats any non-empty string as true, which is not what we want.
+ */
+export function parseBoolean(val: unknown): unknown {
+    if (typeof val === "string") {
+        const lower = val.toLowerCase().trim();
+        if (lower === "false" || lower === "0") return false;
+        if (lower === "true" || lower === "1") return true;
```
nirinchev:
```suggestion
        if (lower === "false" || lower === "0") {
          return false;
        }
        if (lower === "true" || lower === "1") {
          return true;
        }
```

src/common/config/configUtils.ts
```diff
@@ -82,3 +106,33 @@ export function commaSeparatedToArray<T extends string[]>(str: string | string[]
 
     return str as T;
 }
+
+/**
+ * Preprocessor for boolean values that handles string "true"/"false" correctly.
+ * Zod's coerce.boolean() treats any non-empty string as true, which is not what we want.
+ */
+export function parseBoolean(val: unknown): unknown {
```
nirinchev, line 114: This function appears to treat any non-empty string as true too (except `false` and `0`) - is this intentional?

src/common/config/userConfig.ts
```diff
@@ -44,16 +54,18 @@ export const UserConfigSchema = z4.object({
         .describe("An array of logger types.")
         .register(configRegistry, {
             defaultValueDescription: '`"disk,mcp"` see below*',
+            overrideBehavior: "merge",
```
nirinchev: This doesn't seem right - should a client be able to add loggers?

src/common/config/userConfig.ts
```diff
@@ -66,105 +78,140 @@ export const UserConfigSchema = z4.object({
...
         .default(getExportsPath())
         .describe("Folder to store exported data files.")
-        .register(configRegistry, { defaultValueDescription: "see below*" }),
+        .register(configRegistry, { defaultValueDescription: "see below*", overrideBehavior: "not-allowed" }),
     exportTimeoutMs: z4.coerce
         .number()
         .default(300_000)
-        .describe("Time in milliseconds after which an export is considered expired and eligible for cleanup."),
+        .describe("Time in milliseconds after which an export is considered expired and eligible for cleanup.")
+        .register(configRegistry, { overrideBehavior: "override" }),
```
nirinchev: This sounds sketchy - can I setup my exports to timeout after a year, then flood the server storage with exports?

src/common/config/userConfig.ts
```diff
@@ -66,105 +78,140 @@ export const UserConfigSchema = z4.object({
...
+        )
+        .register(configRegistry, { overrideBehavior: "override" }),
     voyageApiKey: z4
         .string()
         .default("")
         .describe(
             "API key for Voyage AI embeddings service (required for vector search operations with text-to-embedding conversion)."
         )
-        .register(configRegistry, { isSecret: true }),
+        .register(configRegistry, { isSecret: true, overrideBehavior: "not-allowed" }),
```
nirinchev: I wonder why this is not allowed? E.g. if I have my own voyage api key, shouldn't I be able to use it?

src/common/config/userConfig.ts
```diff
@@ -66,105 +78,140 @@ export const UserConfigSchema = z4.object({
...
     disableEmbeddingsValidation: z4
-        .boolean()
+        .preprocess(parseBoolean, z4.boolean())
         .default(false)
-        .describe("When set to true, disables validation of embeddings dimensions."),
+        .describe("When set to true, disables validation of embeddings dimensions.")
+        .register(configRegistry, {
+            overrideBehavior: oneWayOverride(false),
```
nirinchev: Wonder why this is oneWayOverride?

src/common/config/userConfig.ts
```diff
@@ -66,105 +78,140 @@ export const UserConfigSchema = z4.object({
...
+        .register(configRegistry, { overrideBehavior: "override" }),
     previewFeatures: z4
         .preprocess(
             (val: string | string[] | undefined) => commaSeparatedToArray(val),
             z4.array(z4.enum(previewFeatureValues))
         )
         .default([])
-        .describe("An array of preview features that are enabled."),
+        .describe("An array of preview features that are enabled.")
+        .register(configRegistry, { overrideBehavior: "merge" }),
```
nirinchev: Why are we merging these arrays?

src/transports/base.ts
nirinchev: I don't think we should be applying the header overrides before we pass in the config to the hook. That way we'd prevent users from customizing the override behavior. We can export the `applyConfigOverrides` function or we could construct an "overrides" object they can just apply like `userConfig = { ...userConfig, overrides }`, but I still believe we should give them the "clean" config and let the consumer decide what to do with it.

nirinchev (commented):
Haven't looked at the tests yet, but I figured it's better to leave the comments I have so far so you don't wait for the full review.

src/common/config/userConfig.ts
gagik: This is just following the scope document. Agree some can be adjusted cc: @himanshusinghs

src/transports/base.ts
gagik: agree generally but now that overrides are an opt-in feature so it'd be weird for one to enable overrides while also expecting a clean slate. I can adjust `applyConfigOverrides` to allow this even when they're disabled
I agree re: exporting `applyConfigOverrides` though

src/common/config/configUtils.ts
gagik: yeah, this was a good balance between making sure old behaviors will work similar way (so if someone has `--readonly=yes` before, it'd still be true and not suddenly become false. while also making sure intentional setting of the value works too.
This helps us have the same logic for CLI and request parsing/validation. Though open to adjusting if there's concerns here.

src/common/config/userConfig.ts
gagik: if validation is enabled (i.e. disableValidation is false) then we would not want a client to disable validation and let the MCP server insert arbitrary data in a vector search field (which apparently breaks vector search)

By the way, I find the double negative to be confusing so it'd be great to rename this in general

src/transports/base.ts
gagik: decided to go with the clean hook after all, assumption being that if you're `createSessionConfig`  you're open to very manual things and documentation.
will need to have docs in a different PR for library eventually

src/common/config/configOverrides.ts
himanshusinghs: I think its best to error out instead of silently ignoring because otherwise we might leave clients in the dark about why their overrides are not being respected.

src/common/config/userConfig.ts
himanshusinghs: I think the client should be able to request `mcp` logger at-least. This was the reason I added the logger as mergeable config so that client can configure the MCP logger.

src/common/config/userConfig.ts
himanshusinghs: Good point. But what if the client wanted a smaller timeout of their export? We can cap the overridable max timeout to the server configured timeout? But yea if it becomes too complex, I am fine leaving it as well.

src/common/config/userConfig.ts
himanshusinghs: This is based on understanding that individual clients might want to enable specific preview features for their session.

gagik: I think this should either be override (so you can remove features if needed) or not-allowed (so no experimental features can be enabled)

src/common/config/configUtils.ts
nirinchev: I wonder if we should just error out if the string value is different from true/false/1/0? Because based on the same logic, if we have `--readonly=no`, that would now make the server readonly.

src/common/config/userConfig.ts
nirinchev: We decided to blanket disallow this at the end, right?

src/common/config/configUtils.ts
gagik: I'm open to this but it does mean this is a breaking change. 
in latest v1.2.0 `--readOnly no` would make the server readOnly
`--readOnly false` would make the server write-able

nirinchev: Right, the question is if this is intentional - I would say this is more of a bugfix than a breaking change. Additionally, if we throw, it won't be a silent breaking change - we'll give users feedback about what to fix, which I think is a reasonable compromise.

gagik: okay apparently this preprocessing only really applies to these headers.
everything else is configured by yargs-parser... so it's about changing settings there

src/common/config/userConfig.ts
gagik: Yes

src/common/config/userConfig.ts
gagik: setting this to onlyLowerThanBaseValueOverride

package.json
```diff
@@ -67,7 +67,7 @@
     "generate:api": "./scripts/generate.sh",
     "generate:arguments": "tsx scripts/generateArguments.ts",
     "pretest": "pnpm run build",
-    "test": "vitest --project eslint-rules --project unit-and-integration --coverage",
+    "test": "vitest --project eslint-rules --project unit-and-integration --coverage --run",
```
gagik, line 70: drive-by: makes it easier to for LLMs to run this... we can have `test:watch` if needed

src/common/config/configUtils.ts
gagik: changed logic to be:
- if "true" or "false" -> parse into boolean
- any other value error with invalid boolean

note: this doesn't affect our CLI passing, this only affects query params. CLI stuff is handled by yargs-parser

src/common/config/userConfig.ts
gagik: changed this to be only subset of base value (so you can disable preview features but not add any new ones)

src/common/config/configOverrides.ts
```diff
@@ -0,0 +1,178 @@
...
+/**
+ * Parses a string value to the appropriate type using the Zod schema.
+ */
+function parseConfigValue(key: keyof typeof UserConfigSchema.shape, value: unknown): unknown {
+    const fieldSchema = UserConfigSchema.shape[key];
+    if (!fieldSchema) {
+        throw new Error(`Invalid config key: ${key}`);
+    }
+
+    return fieldSchema.safeParse(value).data;
```
himanshusinghs, line 121: @gagik the safeparse would silently insert `undefined` when value is incorrect here

gagik: that seems fine right?

gagik: ah, right we could error in way which is more useful

himanshusinghs: yea I think we should error out instead because otherwise we might be overriding with potentially wrong values.

PR #766, opened 2025-11-26
chore: disallow overriding of internal tool implementations

himanshusinghs:
## Proposed changes

This PR follows up on recent conversation regarding tool implementations and the idea that we should not allow library consumers to remove internal tool implementations and if they wish to disable internal tools, they can do so using the session config hook.

Following the outcome of the same conversation, this implements a name collision check to ensure we don't register same named tools more than once.

Additionally, the last commit ([9cc5526](https://github.com/mongodb-js/mongodb-mcp-server/pull/766/commits/9cc5526f2592b1054eff793a357ccd0af084c9a9)), moves the `name`, `operationType` and `category` to static properties of ToolClass so that library consumers have a chance of disabling multiple tools conditionally without statically typing the tool names.

I would advice reviewing the PR in two different steps:
1. Review the change that disallow overriding of internal tool implementation and implements the tool name collision check, basically the first two commits - [Link here](https://github.com/mongodb-js/mongodb-mcp-server/compare/main...776c8e3).
2. Review the change that moves the `name`, `operationType` and `category` to static properties of ToolClass to make it easy for library consumers, the last commit - [Link here](https://github.com/mongodb-js/mongodb-mcp-server/pull/766/commits/9cc5526f2592b1054eff793a357ccd0af084c9a9)

## Checklist

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -60,7 +60,6 @@ export class Server {
     private readonly telemetry: Telemetry;
     public readonly userConfig: UserConfig;
     public readonly elicitation: Elicitation;
-    private readonly internalToolImplementations: ToolClass[] = AllTools;
     private readonly additionalToolImplementations: ToolClass[];
     public readonly tools: ToolBase[] = [];
     public readonly connectionErrorHandler: ConnectionErrorHandler;
@@ -242,7 +241,7 @@ export class Server {
 
     private registerTools(): void {
         const toolImplementations = [
-            ...this.internalToolImplementations.map((toolConstructor) => ({
+            ...AllTools.map((toolConstructor) => ({
                 toolConstructor,
                 source: "internal",
             })),
--- a/src/tools/atlas/tools.ts
+++ b/src/tools/atlas/tools.ts
@@ -11,8 +11,9 @@ import { ListOrganizationsTool } from "./read/listOrgs.js";
 import { ConnectClusterTool } from "./connect/connectCluster.js";
 import { ListAlertsTool } from "./read/listAlerts.js";
 import { GetPerformanceAdvisorTool } from "./read/getPerformanceAdvisor.js";
+import type { ToolClass } from "../tool.js";
 
-export const AtlasTools = [
+export const AtlasTools: ToolClass[] = [
     ListClustersTool,
     ListProjectsTool,
     InspectClusterTool,
--- a/src/tools/atlasLocal/tools.ts
+++ b/src/tools/atlasLocal/tools.ts
@@ -2,5 +2,11 @@ import { DeleteDeploymentTool } from "./delete/deleteDeployment.js";
 import { ListDeploymentsTool } from "./read/listDeployments.js";
 import { CreateDeploymentTool } from "./create/createDeployment.js";
 import { ConnectDeploymentTool } from "./connect/connectDeployment.js";
+import type { ToolClass } from "../tool.js";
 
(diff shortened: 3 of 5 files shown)


tests/integration/tools/mongodb/create/createIndex.test.ts
```diff
@@ -510,7 +510,7 @@ describeWithMongoDB(
                 expect(indexes).toHaveLength(1);
                 expect(indexes[0]?.name).toEqual("vector_1_vector");
                 expect(indexes[0]?.type).toEqual("vectorSearch");
-                expect(indexes[0]?.status).toEqual("PENDING");
+                expect(indexes[0]?.status).toEqual(expect.stringMatching(/PENDING|BUILDING/));
```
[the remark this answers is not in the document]
himanshusinghs: either state - pending / building is a good indicator for index building but I haven't seen READY state so far so I am gonna keep my change for now.

gagik:
Lacking context about the conversation, what is the concern here? If they want to override the default tools, I don't see why they shouldn't be able to. 
They might want to replace our tools with their modified versions and it'd be a lot more annoying to deal with the existence of both and go through hacks to disable the rest.

himanshusinghs:
For context - we discussed that we already have means to replace the internal tools (using disabledTools and adding new tools) and having another method does not make sense. 

The usecase of overriding our internal tools is still supported, users just need to disable the internal tool and provide their own implementation but under a different name of their choosing. I don't think disabling tools is a hack, its a feature we have supported all the way so far.

src/tools/atlas/connect/connectCluster.ts
```diff
@@ -30,9 +30,9 @@ export const ConnectClusterArgs = {
 };
 
 export class ConnectClusterTool extends AtlasToolBase {
-    public name = "atlas-connect-cluster";
+    static toolName = "atlas-connect-cluster";
```
gagik, line 33: `ConnectClusterTool.toolName`? it's better to avoid stutter

himanshusinghs: The static property `name` is unfortunately reserved for `Function.name` and TS forbids overriding that.

src/server.ts
```diff
@@ -60,7 +60,8 @@ export class Server {
     private readonly telemetry: Telemetry;
     public readonly userConfig: UserConfig;
     public readonly elicitation: Elicitation;
-    private readonly toolConstructors: (new (params: ToolConstructorParams) => ToolBase)[];
+    private readonly internalToolImplementations: ToolClass[] = AllTools;
```
gagik: not much benefit in having this alias I think

src/tools/atlas/connect/connectCluster.ts
gagik: TIL

src/server.ts
himanshusinghs: Ahh! indeed right. This was earlier used for testing, we don't need that anymore.

himanshusinghs: addressed in [680d072](https://github.com/mongodb-js/mongodb-mcp-server/pull/766/commits/680d0723ff976dbffad955d44ac1ee9a30c7dc0b)

himanshusinghs:
Following an internal conversation, we will continue to support the existing interface of allowing library users to specify tool implementations while ensuring that tools have unique names during tool registration.

This PR will be closed and the change for implementing uniqueness will be added to #767

PR #767, opened 2025-11-26
chore: add developer guide and code examples to extend MCP server via library exports MCP-299

himanshusinghs:
## Proposed changes

This PR adds a guide targeting developers, explaining how to extend the MongoDB MCP server using the exports available from the library. Includes working code examples for the following use-cases:
1. Override Server Configuration
2. Per-Session Configuration
3. Adding Custom Tools
4. Selective Tool Registration

Most of the document was written by CoPilot, so it'd be nice to actually ready through it once. I have ensured correctness of the material particularly API reference and examples.

Please note of the following changes:
- I have moved `operationType` as static property to the tool classes so that its easier for library consumers to address a few use-cases, also mentioned in the guidelines.
- The library exports were also modified a bit to support the examples mentioned in the guide.
- Includes a fix for `InsertManyTool` using unexpected Zod schema

## Checklist

### DIFF
--- a/src/common/config/userConfig.ts
+++ b/src/common/config/userConfig.ts
@@ -113,10 +113,10 @@ export const UserConfigSchema = z4.object({
     httpPort: z4.coerce
         .number()
         .int()
-        .min(1, "Invalid httpPort: must be at least 1")
+        .min(0, "Invalid httpPort: must be at least 0")
         .max(65535, "Invalid httpPort: must be at most 65535")
         .default(3000)
-        .describe("Port number for the HTTP server (only used when transport is 'http').")
+        .describe("Port number for the HTTP server (only used when transport is 'http'). Use 0 for a random port.")
         .register(configRegistry, { overrideBehavior: "not-allowed" }),
     httpHost: z4
         .string()
@@ -125,7 +125,7 @@ export const UserConfigSchema = z4.object({
         .register(configRegistry, { overrideBehavior: "not-allowed" }),
     httpHeaders: z4
         .object({})
-        .passthrough()
+        .loose()
         .default({})
         .describe(
             "Header that the HTTP server will validate when making requests (only used when transport is 'http')."
--- a/tests/unit/common/config.test.ts
+++ b/tests/unit/common/config.test.ts
@@ -690,17 +690,17 @@ describe("config", () => {
         });
 
         describe("httpPort", () => {
-            it("must be above 1", () => {
+            it("must be above 0", () => {
                 const onErrorFn = vi.fn();
                 const onExitFn = vi.fn<CreateUserConfigHelpers["closeProcess"]>();
                 createUserConfig({
                     onError: onErrorFn,
                     closeProcess: onExitFn,
-                    cliArguments: ["--httpPort", "0"],
+                    cliArguments: ["--httpPort", "-1"],
                 });
                 expect(onErrorFn).toBeCalledWith(
                     expect.stringContaining(
-                        "Invalid configuration for the following fields:\nhttpPort - Invalid httpPort: must be at least 1"
+                        "Invalid configuration for the following fields:\nhttpPort - Invalid httpPort: must be at least 0"
                     )
                 );
                 expect(onExitFn).toBeCalledWith(1);


MCP_SERVER_LIBRARY.md
```diff
@@ -0,0 +1,973 @@
...
+- [API Reference](#api-reference)
+- [Advanced Topics](#advanced-topics)
+- [Examples](#examples)
+
+## Overview
+
+The MongoDB MCP Server can be embedded in your own Node.js applications and customized to meet specific requirements. The library exports provide full control over:
+
+- Server configuration and initialization
+- Per-session(MCP Client session) configuration hooks
```
gagik:
```suggestion
- Per-session (MCP Client session) configuration hooks
```

MCP_SERVER_LIBRARY.md
```diff
@@ -0,0 +1,973 @@
...
+  createSessionConfig?: (context: {
+    userConfig: UserConfig;
+    request?: RequestContext;
+  }) => Promise<UserConfig> | UserConfig;
+}
+```
+
+### ToolBase
+
+Base class for implementing custom tools.
```
gagik: This seems to be listing code implementations of classes for the most part. Not sure how useful that is.
Text here lives/should live in JSDoc. these are just different classes exported by the library; I don't see any value in barebones listing them in markdown.

It'd be better to see them as they're being imported / used.

himanshusinghs: Yea and then there's the headache of always keeping them up to date. I guess I will move the JSDoc to the actual classes itself and link to the class here for the API reference.

himanshusinghs: I have moved the doc to TSDoc and the examples covered in the use-case section shows how to use them practically.

gagik: that's a good solution! we can always generate a website/docs from TSDoc if we'd like

src/tools/mongodb/create/insertMany.ts
```diff
@@ -15,27 +15,29 @@ const zSupportedEmbeddingParametersWithInput = zSupportedEmbeddingParameters.ext
...
+    protected argsShape = this.isFeatureEnabled("search")
+        ? {
+              ...commonArgs,
+              embeddingParameters: zSupportedEmbeddingParametersWithInput
+                  .optional()
+                  .describe(
+                      "The embedding model and its parameters to use to generate embeddings for fields with vector search indexes. Note to LLM: If unsure which embedding model to use, ask the user before providing one."
+                  ),
+          }
+        : commonArgs;
```
himanshusinghs, line 39: This fixes an unexpected ZodSchema because of earlier optional empty object getting injected when feature flag was not enabled. This was caught by one of the examples in the guide where a tool call was getting rejected right away.

src/tools/index.ts
```diff
@@ -1,13 +1,20 @@
...
+import * as AtlasTools from "./atlas/tools.js";
+import * as AtlasLocalTools from "./atlasLocal/tools.js";
+import * as MongoDbTools from "./mongodb/tools.js";
 
-const AllTools = [...MongoDbTools, ...AtlasTools, ...AtlasLocalTools];
+const AllTools = {
+    ...MongoDbTools,
+    ...AtlasTools,
+    ...AtlasLocalTools,
+} as const;
```
himanshusinghs: Please note that we're now exporting record like exports instead of array like exports. The changes allows nitpicking of individual tools.

knip.json
```diff
@@ -6,7 +6,14 @@
     "scripts/**/*.ts",
     "eslint-rules/*.js"
   ],
-  "ignore": ["tests/integration/fixtures/curl.mjs", "tests/vitest.d.ts"],
+  "ignore": [
+    // Knip, for some reason, is not able to link the exported tools to the
+    // final exports in `tools/index.ts` and complains about unused exports. For
+    // that reason we're ignoring the tool definitions.
```
himanshusinghs: @gagik fyi
If there's a way to avoid this, I'd be happy to follow that.

src/tools/index.ts
gagik: I think we should export this as an array and also do:
```ts
export * from from "./mongodb/tools.js";
...
```
This way the IDE can suggest these imports and no real nitpicking is required

MCP_SERVER_LIBRARY.md
```diff
@@ -0,0 +1,1119 @@
...
+
+#### AllTools
+
+An object containing all internal tool classes (MongoDB, Atlas, and Atlas Local tools combined).
+
+```typescript
+import { AllTools, MongoDbTools, AtlasTools } from "mongodb-mcp-server/tools";
+
+// Pick a specific tool
+const MyTool = AllTools.AggregateTool;
```
gagik: with the change I suggested for exmaple, one could simply
`import { AllTools, MongoDbTools, AtlasTools, AggregateTool } from "mongodb-mcp-server/tools";`

knip.json
gagik: knip has a concept of "entry points" which is files that are "production" (i.e. we're fine with any export since this is being used by users)
`tools/index.ts` is a new entry point so it should be added to `entry` in `knip.json` instead

knip.json
```diff
@@ -6,7 +6,14 @@
...
+    "src/tools/**/*.ts",
```
gagik: `    "src/tools/**/*.ts",` is already in `entry` I don't think there should be a need to ignore this

MCP_SERVER_LIBRARY.md
```diff
@@ -0,0 +1,1119 @@
...
+  userConfig: UserConfigSchema.parse({
+    transport: "stdio",
+    // Don't provide a default connection string
+    connectionString: undefined,
+  }),
+  // Register all internal tools except the default connect tools, plus our custom tools
+  tools: [
+    ...Object.values(AllTools).filter(
+      (tool) => tool.operationType !== "connect"
+    ),
```
gagik: can we have a test to enforce we don't break this behavior in the future?

himanshusinghs: We do have TS checks in place, the project won't pass the TSBuild if `operationType` is not provided as static property.

src/tools/index.ts
himanshusinghs: But it is the same?

`export * from "../mongodb/tools.js"`

needs to be imported as:
`import { ConnectTool } from "mongodb-mcp-server/tools"`

and currently you would do the same either by:
```
import { AllTools } from "mongodb-mcp-server/tools"
AllTools.ConnectTool
```
or
```
import { MongoDbTools } from "mongodb-mcp-server/tools"
MongoDbTools.ConnectTool
```

Actually with the current change, we do get per category division already which we would miss if we do:
`export * from "../mongodb/tools.js"`

unless we do
`export * as MongoDbTools from "../mongodb/tools.js"`

which then essentially is the same.

knip.json
himanshusinghs: Perfect that's what I need, thank you 🙌

MCP_SERVER_LIBRARY.md
himanshusinghs: Besides the TSCheck I added a test to ensure the ToolClass conformance - [c786c6b](https://github.com/mongodb-js/mongodb-mcp-server/pull/767/commits/c786c6b2a3500a774a8687d9de299636a1c97a23)

src/tools/index.ts
himanshusinghs: Based on our discussion, I have updated the export to have individual exports alongside the list of all the tools for convenience.

gagik (approved):

PR #771, opened 2025-12-01
chore: export UserConfigSchema for consumers

nirinchev:
## Proposed changes

This will make it easier for library consumers (i.e. vscode) to construct a valid user config with all defaults.


himanshusinghs:
Just to mention - #767 covers this so if its not urgent shall we wait for that to land. Happy to have this merged as well.

himanshusinghs (approved):

PR #777, opened 2025-12-02
chore: adopt strict validation for all command line arguments MCP-298

gagik:
This will not work until https://github.com/mongodb-js/mongosh/pull/2589 is released but making it available for early reviews.

This gets rid of our hardcoded dependency on `OPTIONS` for both internal and mongosh-related argument parsing and instead relies completely on our (and mongosh's) Zod schema. 

This provides the following advantages:
1. We no longer need to manually append the `OPTIONS` array when adding new config fields which is easy to forget. This actually revealed that we had forgotten to add i.e. `vectorSearchDimensions` to this array before because of the redundancy involved.
2. We can reliably validate both our and mongosh argument types.
3. We can ensure we're keep up with changes to mongosh's connection string cases.
4. We get "out-of-the-box" support for more complex types like objects with `--object.field=3` which would otherwise have require us to manually specify different scenarios in the `OPTIONS` array. 

As a consequence of some of these advantages, this PR also...:
1. Adds support to set `httpHeaders` using the CLI (or env), either as `--httpHeaders='{ "test": "1" }'` or `--httpHeaders.test='1'`!
2. Better alignment with some of the mongosh logic - for example `ssl`-related fields actually get replaced with `tls` equivalents; and some fields are also unsupported. We would correctly throw an error in this case (which we weren't doing before).
3. We likely got some new type checking options now.

Some considerations:
1. For the future, we might want to consider adopting a subset of mongosh's arguments; in that case it's important to make sure that this subset can cover all usecases.
4. There's a general risk of clashing names. For example readonly exists in both. By default our configuration takes precedence; just something to keep in mind

### DIFF
--- a/tests/integration/elicitation.test.ts
+++ b/tests/integration/elicitation.test.ts
@@ -38,6 +38,7 @@ describe("Elicitation Integration Tests", () => {
                     expect(mockElicitInput.mock).toHaveBeenCalledWith({
                         message: expect.stringContaining("You are about to drop the `test-db` database"),
                         requestedSchema: Elicitation.CONFIRMATION_SCHEMA,
+                        mode: "form",
                     });
 
                     // Should attempt to execute (will fail due to no connection, but confirms flow worked)
@@ -82,6 +83,7 @@ describe("Elicitation Integration Tests", () => {
                     expect(mockElicitInput.mock).toHaveBeenCalledWith({
                         message: expect.stringContaining("You are about to drop the `test-collection` collection"),
                         requestedSchema: expect.objectContaining(Elicitation.CONFIRMATION_SCHEMA),
+                        mode: "form",
                     });
                 });
 
@@ -101,6 +103,7 @@ describe("Elicitation Integration Tests", () => {
                     expect(mockElicitInput.mock).toHaveBeenCalledWith({
                         message: expect.stringContaining("You are about to delete documents"),
                         requestedSchema: expect.objectContaining(Elicitation.CONFIRMATION_SCHEMA),
+                        mode: "form",
                     });
                 });
 
@@ -120,6 +123,7 @@ describe("Elicitation Integration Tests", () => {
                     expect(mockElicitInput.mock).toHaveBeenCalledWith({
                         message: expect.stringContaining("You are about to create a database user"),
                         requestedSchema: expect.objectContaining(Elicitation.CONFIRMATION_SCHEMA),
+                        mode: "form",
                     });
                 });
 
@@ -140,6 +144,7 @@ describe("Elicitation Integration Tests", () => {
                             "You are about to add the following entries to the access list"
                         ),
                         requestedSchema: expect.objectContaining(Elicitation.CONFIRMATION_SCHEMA),
+                        mode: "form",
                     });
                 });
             });
@@ -222,6 +227,7 @@ describe("Elicitation Integration Tests", () => {
                         /You are about to execute the `list-databases` tool which requires additional confirmation. Would you like to proceed\?/
                     ),
                     requestedSchema: expect.objectContaining(Elicitation.CONFIRMATION_SCHEMA),
+                    mode: "form",
                 });
             });
 
@@ -265,6 +271,7 @@ describe("Elicitation Integration Tests", () => {
                 expect(mockElicitInput.mock).toHaveBeenCalledWith({
                     message: expect.stringMatching(/project.*507f1f77bcf86cd799439011/),
                     requestedSchema: expect.objectContaining(Elicitation.CONFIRMATION_SCHEMA),
+                    mode: "form",
                 });
             });
 
@@ -283,6 +290,7 @@ describe("Elicitation Integration Tests", () => {
                 expect(mockElicitInput.mock).toHaveBeenCalledWith({
                     message: expect.stringMatching(/mydb.*database/),
                     requestedSchema: expect.objectContaining(Elicitation.CONFIRMATION_SCHEMA),
+                    mode: "form",
                 });
             });
         },
--- a/tests/integration/tools/mongodb/delete/dropIndex.test.ts
+++ b/tests/integration/tools/mongodb/delete/dropIndex.test.ts
@@ -476,6 +476,7 @@ describe.each([{ vectorSearchEnabled: false }, { vectorSearchEnabled: true }])(
                                 message: expect.stringContaining(
                                     "You are about to drop the search index named `searchIdx` from the `mflix.movies` namespace"
                                 ),
+                                mode: "form",
                                 requestedSchema: Elicitation.CONFIRMATION_SCHEMA,
                             });
 
@@ -503,6 +504,7 @@ describe.each([{ vectorSearchEnabled: false }, { vectorSearchEnabled: true }])(
                                 message: expect.stringContaining(
                                     "You are about to drop the search index named `searchIdx` from the `mflix.movies` namespace"
                                 ),
+                                mode: "form",
                                 requestedSchema: Elicitation.CONFIRMATION_SCHEMA,
                             });
                             expect(dropSearchIndexSpy).not.toHaveBeenCalled();
--- a/tests/unit/elicitation.test.ts
+++ b/tests/unit/elicitation.test.ts
@@ -82,6 +82,7 @@ describe("Elicitation", () => {
             expect(mockElicitInput.mock).toHaveBeenCalledWith({
                 message: testMessage,
                 requestedSchema: Elicitation.CONFIRMATION_SCHEMA,
+                mode: "form",
             });
         });
 


.gitignore
```diff
@@ -14,4 +14,8 @@ coverage
 # Generated assets by accuracy runs
 .accuracy
 
-.DS_Store
\ No newline at end of file
+.DS_Store
+
+# Development tool files
+.yalc
```
gagik, line 20: I like [yalc](https://github.com/wclr/yalc) so would be great if this is gitgnored

tests/unit/common/config.test.ts
```diff
@@ -577,362 +592,170 @@ describe("config", () => {
...
 
             it("allows stderr logger", () => {
-                const actual = createUserConfig({ cliArguments: ["--loggers", "stderr"] });
-                expect(actual.loggers).toEqual(["stderr"]);
-            });
-        });
-    });
-});
-
-describe("Warning and Error messages", () => {
```
gagik, line 798: moved to e2e tests instead

src/common/config/createUserConfig.ts
```diff
@@ -1,148 +1,120 @@
...
-    const { unknownCliArgumentErrors, deprecatedCliArgumentWarning, userAndArgsParserConfig, connectionSpecifier } =
-        parseUserConfigSources(cliArguments);
-
-    if (unknownCliArgumentErrors.length) {
-        const errorMessage = `
-${unknownCliArgumentErrors.join("\n")}
-- Refer to https://www.mongodb.com/docs/mcp-server/get-started/ for setting up the MCP Server.
-`;
-        onError(errorMessage);
-        return closeProcess(1);
```
gagik, line 44: changed this callback-based logic to instead be returning information to be used at a higher level

tests/unit/common/config.test.ts
```diff
@@ -279,27 +260,27 @@ describe("config", () => {
...
                 },
                 {
                     cli: ["--sslCRLFile", "/var/file"],
-                    expected: { sslCRLFile: "/var/file" },
+                    expected: { tlsCRLFile: "/var/file" },
                 },
                 {
                     cli: ["--sslCertificateSelector", "pem=pom"],
-                    expected: { sslCertificateSelector: "pem=pom" },
+                    expected: { tlsCertificateSelector: "pem=pom" },
```
gagik, line 272: this is what the actually happens in mongosh since ssl arguments are deprecated

server.json
```diff
@@ -111,14 +118,14 @@
...
-          "format": "string",
+          "format": "number",
           "isSecret": false
         },
         {
           "name": "MDB_MCP_IDLE_TIMEOUT_MS",
           "description": "Idle timeout for a client to disconnect (only applies to http transport).",
           "isRequired": false,
-          "format": "string",
+          "format": "number",
```
gagik, line 128: seems like there were some issues with the `server.json` generation before 🤷

src/common/config/userConfig.ts
```diff
@@ -222,3 +217,14 @@ export const UserConfigSchema = z4.object({
         )
         .register(configRegistry, { overrideBehavior: "not-allowed" }),
 });
+
+export const UserConfigSchemaWithCliOptions = z4.object({
+    ...UserConfigSchema.shape,
+    ...CliOptionsSchema.shape,
+});
```
himanshusinghs: We need to at-least deprecate `nodb` here otherwise that will hinder the rightful parsing of a connection specifier.

src/common/config/userConfig.ts
```diff
@@ -11,11 +10,7 @@ import {
     parseBoolean,
 } from "./configUtils.js";
 import { previewFeatureValues, similarityValues } from "../schemas.js";
-
-// TODO: UserConfig should only be UserConfigSchema and not an intersection with
-// CliOptions. When we pull apart these two interfaces, we should fix this type
-// as well.
-export type UserConfig = z4.infer<typeof UserConfigSchema> & CliOptions;
```
himanshusinghs, line 18: This is going backwards from the idea of separating `UserConfig` and Mongosh related config apart.

I don't think we should continue down this path of keeping the two together.

himanshusinghs: Ok I see that you left a similar todo in createUserConfig intact. If the idea to separate them still exist and that todo is meant for that then I am good.

gagik: yeah

scripts/generateArguments.ts
```diff
@@ -5,15 +5,15 @@
...
  *
- * It uses the Zod schema and OPTIONS defined in src/common/config.ts
+ * It uses the UserConfig Zod Schema.
  */
 
 import { readFileSync, writeFileSync } from "fs";
 import { join, dirname } from "path";
 import { fileURLToPath } from "url";
-import { UserConfigSchema, configRegistry } from "../src/common/config/userConfig.js";
+import { UserConfigSchemaWithCliOptions, configRegistry } from "../src/common/config/userConfig.js";
```
himanshusinghs: We're using the UserConfig schema together with CliOptions schema, yet, in the final server.json I don't see the entries from CliOptions schema? We should expect those right?

src/index.ts
```diff
@@ -49,7 +49,29 @@ import { DryRunModeRunner } from "./transports/dryModeRunner.js";
...
-    const config = createUserConfig();
+    const {
+        error,
+        warnings,
+        parsed: config,
+    } = createUserConfig({
+        args: process.argv.slice(2),
+    });
+
+    if (config == null || (error && error.length)) {
```
himanshusinghs:
```suggestion
    if (!config || (error && error.length)) {
```
If we wish to capture both null and undefined without annoying eslint 😄

tsconfig.cjs.json
```diff
@@ -1,8 +1,7 @@
 {
   "extends": "./tsconfig.build.json",
   "compilerOptions": {
-    "module": "commonjs",
-    "moduleResolution": "node",
+    "module": "node16",
```
himanshusinghs: I fear this will break the import / require in vscode extension. Was this necessary?

gagik: confusingly, "module": "commonjs" is basically a version so old it only is commonjs.
Node16 can be both CommonJS and ESM, it just supports more APIs.
we might need to bump VSCode tsconfig to use node16 too but this is needed in general.

src/index.ts
gagik: let me blow your mind
```ts
undefined == null // true
undefined === null // false
```

gagik: that's why I usually try to do `== null`. `== undefined` would also work

scripts/generateArguments.ts
gagik: hmmmm..... I guess in `server.json` we should; should we have any fields in the README?

gagik: let's follow-up on this. otherwise we're going to dump a bunch of unrelated fields into it

tsconfig.cjs.json
gagik: I'm investigating this; might need some more adjustments

gagik: okay you were totally right; confusingly node16 _supports_ CJS but doesn't output it even if you try to force it to.
instead needed to manually force the path... oh well

src/index.ts
gagik: ahh I see lol eslint doesn't like this. fair enough

scripts/generateArguments.ts
himanshusinghs: Yea - I am also unsure if they deserve a place in server.json and README but worth noting that somehow with this script they don't end up there. Happy to address this as a follow up.

src/common/config/userConfig.ts
himanshusinghs: @gagik did you find a way to deprecate this? I notice that we now have a warning message but the argument still end up deflecting the positional connection string.

src/index.ts
```diff
@@ -49,7 +49,29 @@ import { DryRunModeRunner } from "./transports/dryModeRunner.js";
...
+        parsed: config,
+    } = createUserConfig({
+        args: process.argv.slice(2),
+    });
+
+    if (!config || (error && error.length)) {
+        console.error(`
+            ${error}
+            - Refer to https://www.mongodb.com/docs/mcp-server/get-started/ for setting up the MCP Server.
+            `);
```
himanshusinghs:
```suggestion
        console.error(`
${error}
- Refer to https://www.mongodb.com/docs/mcp-server/get-started/ for setting up the MCP Server.
`);
```

I am assuming that the leading whitespaces were not intentional.

src/index.ts
himanshusinghs: yea unfortunately otherwise I do like this neat JS trick 💯

src/common/config/userConfig.ts
gagik: deprecation to me means the behavior continues to work. We warn them to remove this so we can eventually throw an error. Otherwise we're breaking behavior

src/common/config/createUserConfig.ts
```diff
@@ -22,6 +22,14 @@ export function createUserConfig({ args }: { args: string[] }): {
         return { error: parseError, warnings, parsed: undefined };
     }
 
+    if (parsed.nodb) {
+        return {
```
gagik, line 26: This ended up being more manual because of some mongosh type integrity stuff. We will need to follow-up to make this and more arguments use a subset of functionality

scripts/generateArguments.ts
gagik: this is existing behavior and I think the code is self-explanatory, let's follow up on this later

PR #779, opened 2025-12-03
chore: bump mcp SDK, refactor tool arguments

gagik:
This bumps the SDK to the latest versions which patches some security issues. This also simplifies our typing and structure of our arguments so we can have a more consistant and simpler to reason about API across tools.

### DIFF
--- a/src/tools/tool.ts
+++ b/src/tools/tool.ts
@@ -470,18 +470,29 @@ export abstract class ToolBase {
             }
         };
 
-        this.registeredTool = (
-            server.mcpServer.tool as (
-                // Note: We use explicit type casting here to avoid  "excessively deep and possibly infinite" errors
-                // that occur when TypeScript tries to infer the complex generic types from `typeof this.argsShape`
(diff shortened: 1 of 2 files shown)


src/tools/tool.ts
```diff
@@ -456,19 +464,24 @@ export abstract class ToolBase {
...
 
-        this.registeredTool = server.mcpServer.tool(
-            this.name,
-            this.description,
-            this.argsShape,
-            this.annotations,
-            callback
-        );
+        this.registeredTool = (
+            server.mcpServer.tool as (
```
nirinchev: `.tool` is deprecated - should we swap this out for `registerTool`?

gagik: yeah I tried and then [reverted](github.com/mongodb-js/mongodb-mcp-server/pull/779/commits/f881902d800103d4a7eb99ce7079cf3a46dece95), I thought there was something wrong with the SDK because I had incorrectly documented experiences from it before but it was actually our mock. Re-added

PR #780, opened 2025-12-03
feat: add structured content

nirinchev:
## Proposed changes

This is mostly a POC for how adding structured content to tools could look like. I'm still torn between adding it as is and adding it also to the `formatUntrustedData` helper - I've decided to keep them separate because I'm hoping we will eventually be able to remove the helper, though I'm very open to being convinced either way.

### DIFF
--- a/src/tools/mongodb/metadata/listDatabases.ts
+++ b/src/tools/mongodb/metadata/listDatabases.ts
@@ -1,24 +1,31 @@
-import type { CallToolResult } from "@modelcontextprotocol/sdk/types.js";
 import { MongoDBToolBase } from "../mongodbTool.js";
 import type * as bson from "bson";
-import type { OperationType } from "../../tool.js";
+import type { OperationType, ToolResult } from "../../tool.js";
 import { formatUntrustedData } from "../../tool.js";
+import z, { type ZodNever } from "zod";
+
+export const ListDatabasesToolOutputShape = {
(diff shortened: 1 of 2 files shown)


src/tools/mongodb/create/insertMany.ts
```diff
@@ -70,8 +77,14 @@ export class InsertManyTool extends MongoDBToolBase {
             `Inserted \`${result.insertedCount}\` document(s) into ${database}.${collection}.`,
             `Inserted IDs: ${Object.values(result.insertedIds).join(", ")}`
         );
+
         return {
             content,
+            structuredContent: {
```
fmenezes, line 83: FYI the spec says we should return the json serialised instead of formatted text at https://modelcontextprotocol.io/specification/2025-06-18/server/tools#structured-content, I don't know if you would be thinking about automating it as well

nirinchev: I think we touched on it at some point in Slack, but in my opinion the target of the "SHOULD" here is the backwards compatibility part, not the exact shape of the response. That is, we want to make sure clients that don't support structured content still work as expected, but I don't believe those matching precisely is critical.

I've decided to keep the prompt injection mitigation in place as, if a client does not support structured content, returning json with potentially malicious content is risky. Technically, automating it further is feasible - we can have `formatUntrustedData` both construct the content with the injection mitigation and returned the structured content as is, but it seemed like we're not winning a lot by doing that (the code would get more complex with a bunch of generics to enforce the output shape).

PR #782, opened 2025-12-04
chore: add support for loading config from file MCP-327

himanshusinghs:
## Proposed changes

Adds support for loading configuration from a file. As a drive by:
- have added coverage for Server.tools interface as well
- disabled overriding `voyageApiKey` from request headers

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -565,7 +565,7 @@ For a full list of roles and their privileges, see the [Atlas User Roles documen
 
 #### Configuration File
 
-Store configuration in a JSON file and load it using the `MDB_MCP_CONFIG` environment variable (recommended) or `--config` command-line argument.
+Store configuration in a JSON file and load it using the `MDB_MCP_CONFIG` environment variable.
 
 > **🔒 Security Best Practice:** Prefer using the `MDB_MCP_CONFIG` environment variable for sensitive fields over the configuration file or `--config` CLI argument. Command-line arguments are visible in process listings.
 


src/common/config/userConfig.ts
```diff
@@ -178,7 +178,7 @@ const ServerConfigSchema = z4.object({
         .describe(
             "API key for Voyage AI embeddings service (required for vector search operations with text-to-embedding conversion)."
         )
-        .register(configRegistry, { isSecret: true, overrideBehavior: "override" }),
+        .register(configRegistry, { isSecret: true, overrideBehavior: "not-allowed" }),
```
gagik, line 181: was this where we landed on? I don't remember this being an issue

README.md
```diff
@@ -562,6 +563,63 @@ For a full list of roles and their privileges, see the [Atlas User Roles documen
 
 ### Configuration Methods
 
+#### Configuration File
+
+Store configuration in a JSON file and load it using the `MDB_MCP_CONFIG` environment variable (recommended) or `--config` command-line argument.
+
+> **🔒 Security Best Practice:** Prefer using the `MDB_MCP_CONFIG` environment variable over the `--config` CLI argument. Command-line arguments are visible in process listings.
```
gagik:
```suggestion
> **🔒 Security Best Practice:** Prefer using the `MDB_MCP_CONFIG` environment variable for sensitive fields over the configuration file or `--config` CLI argument. Command-line arguments are visible in process listings.
```

README.md
```diff
@@ -562,6 +563,63 @@ For a full list of roles and their privileges, see the [Atlas User Roles documen
...
+```
+
+**Windows PowerShell:**
+
+```powershell
+$env:MDB_MCP_CONFIG="C:\path\to\config.json"
+npx -y mongodb-mcp-server@latest
+```
+
+Alternatively, use `--config` argument (less secure):
```
gagik: I think this is too extreme. Knowing where a config file is is very trivial, let's just not mention `--config` if it seems problematic

README.md
```diff
@@ -562,6 +563,63 @@ For a full list of roles and their privileges, see the [Atlas User Roles documen
 
 ### Configuration Methods
 
+#### Configuration File
+
+Store configuration in a JSON file and load it using the `MDB_MCP_CONFIG` environment variable (recommended) or `--config` command-line argument.
+
+> **🔒 Security Best Practice:** Prefer using the `MDB_MCP_CONFIG` environment variable over the `--config` CLI argument. Command-line arguments are visible in process listings.
+
+> **🔒 File Security:** Ensure your configuration file has proper ownership and permissions, limited to the user running the MongoDB MCP server:
```
gagik, line 572: I'm not super sold that this is a super meaningful security tip but we can leave it. Is the idea that a malicious process wouldn't be my user? Because I don't see which risks this mitigates and the idea of a malicious process that has permissions of a different user is quite niche.

 I mean we don't even say this for mongosh: https://www.mongodb.com/docs/manual/reference/configuration-options/#configuration-file

src/common/config/userConfig.ts
himanshusinghs: yea this the decision taken in the call for both connectionString and voyageApiKey (secrets)

README.md
himanshusinghs: Ensuring file permissions is a common security practice and I don't think there is anything problematic in advising this regardless of whether we do it for another product or not.

README.md
himanshusinghs: Yea knowing where a config file is trivial particularly when we don't take active measures. Here the suggested active measures are:
1. Don't provide file location in CLI arguments
2. Limit file access to a user so that even if you provide file location in arguments, its not accessible to other users.

README.md
gagik: yeah it doesn't harm, we can leave that

README.md
gagik: Maybe let's just delete `Alternatively, use `--config` argument (less secure):`? If we don't want to encourage it let's not actively mention it

gagik:
Let's delete the `--config` instructions though. If we give instructions then one will do it.

README.md
himanshusinghs: Ahh yea that makes sense. Adding it now.

himanshusinghs: Removed it in [d5d4ebe](https://github.com/mongodb-js/mongodb-mcp-server/pull/782/commits/d5d4ebe92f08250e66f6abf828bb871c0771bb71)

PR #783, opened 2025-12-04
feat: implement MCP-UI support

tsck:
## Summary
- Adds support of the [MCP-UI](https://mcpui.dev/)
- Adds build tooling for React components so that they can be rendered via an iframe client
- Implements UI for `list-database` tool calls
- Adds the ability to add additional UI's on server instantiation

## Testing
I've been using [MCPJam Inspector](https://github.com/MCPJam/inspector) which has MCP-UI implemented in the client.

## Screenshot
Screenshot 2025-12-04 at 10 11 20 AM

### DIFF
--- a/src/tools/tool.ts
+++ b/src/tools/tool.ts
@@ -22,7 +22,7 @@ export type ToolExecutionContext = {
 };
 
 /**
- * The ~`type of operation the tool performs. This is used when evaluating if a tool is allowed to run based on
+ * The type of operation the tool performs. This is used when evaluating if a tool is allowed to run based on
  * the config's `disabledTools` and `readOnly` settings.
  * - `metadata` is used for tools that read but do not access potentially user-generated
  *   data, such as listing databases, collections, or indexes, or inferring collection schema.


src/tools/tool.ts
```diff
@@ -675,6 +722,64 @@ export abstract class ToolBase {
...
+                    id: LogId.toolExecute,
+                    context: `tool - ${this.name}`,
+                    message: `structuredContent failed validation against outputSchema, skipping UI resource: ${validation.error.message}`,
+                });
+                return result;
+            }
+        }
+
+        const uiResource = createUIResource({
+            uri: `ui://${this.name}/${Date.now()}`,
```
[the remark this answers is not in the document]
tsck: I'm not sure we'd want this deterministic. If the same tool is called multiple times, each invocation generates new data, so unique URIs ensure each result is treated as a distinct resource. Open to other ideas here if my thinking is incorrect!

package.json
TheSonOfThomp: looks like this was renamed 
```suggestion
    "start": "node dist/index.js --transport http --loggers stderr mcp --previewFeatures search",
```

vite.ui.config.ts
```diff
@@ -0,0 +1,104 @@
...
+// Use node_modules/.cache for generated HTML entries - these are build artifacts, not source files
+const entriesDir = resolve(__dirname, "node_modules/.cache/mongodb-mcp-server/ui-entries");
+const templatePath = resolve(__dirname, "src/ui/build/template.html");
+const mountPath = resolve(__dirname, "src/ui/build/mount.tsx");
+
+/**
+ * Discover all component directories in src/ui/components/
+ * Each directory should have an index.ts that exports the component
+ */
+function discoverComponents(): string[] {
```
TheSonOfThomp: nit: why `discover`?

vite.ui.config.ts
```diff
@@ -0,0 +1,104 @@
...
+
+                const outputPath = join(entriesDir, `${componentName}.html`);
+                writeFileSync(outputPath, html);
+                console.log(`[generate-html-entries] Generated ${componentName}.html`);
+            }
+        },
+    };
+}
+
+const components = discoverComponents();
```
TheSonOfThomp: nit: we're running this search here, and then again in the `generateHtmlEntries` plugin. Could either use this pre-computed value in the plugin, or memoize it to speed up builds a bit

package.json
```diff
@@ -132,6 +152,8 @@
     "node-machine-id": "1.1.12",
     "oauth4webapi": "^3.8.0",
     "openapi-fetch": "^0.15.0",
+    "react": "^18.3.0",
+    "react-dom": "^18.3.0",
```
TheSonOfThomp: `dependencies` or `devDependencies`?

src/tools/tool.ts
```diff
@@ -675,6 +712,61 @@ export abstract class ToolBase {
...
+        return this.uiRegistry?.get(this.name);
+    }
+
+    /**
+     * Appends a UIResource to the tool result.
+     *
+     * @param result - The result from the tool's `execute()` method
+     * @returns The result with UIResource appended if conditions are met, otherwise unchanged
+     */
+    private appendUIResourceIfAvailable(result: CallToolResult): CallToolResult {
```
TheSonOfThomp: nit: not sure if this is a naming convention across the repo, but just `appendUIResource` with docs for what happens if it's not available feels more correct

src/tools/tool.ts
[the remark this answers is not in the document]
TheSonOfThomp: Not sure how the caching works exactly, but I think a static uri would be more performant? I don't think it caches the `uiMetadata`

src/tools/mongodb/metadata/listDatabases.ts
TheSonOfThomp, line 1: does this need to be `.tsx`?

vite.ui.config.ts
tsck: Removed this completely in the change to inline HTML strings

vite.ui.config.ts
tsck: Fixed this completely in the change to inline HTML strings

package.json
tsck: Oh yeah, good point! Will update to dev

src/tools/tool.ts
tsck: No argument on this from me. "if available" just seemed more technically correct based on what the method does, but I don't see an issue with shortening it.

src/tools/mongodb/metadata/listDatabases.ts
tsck: Good catch! I changed this at the very beginning so I could inline the jsx and validate things were working locally. Will change back

src/tools/mongodb/metadata/listDatabases.ts
```diff
@@ -3,22 +3,34 @@ import { MongoDBToolBase } from "../mongodbTool.js";
 import type * as bson from "bson";
 import type { OperationType } from "../../tool.js";
 import { formatUntrustedData } from "../../tool.js";
+import { ListDatabasesOutputSchema, type ListDatabasesOutput } from "../../../ui/components/ListDatabases/schema.js";
```
nirinchev: `listDatabases` should not depend/reference the UI components. Instead, these schemas should be exported here and imported in the UI component.

src/tools/tool.ts
```diff
@@ -675,6 +712,61 @@ export abstract class ToolBase {
...
         return metadata;
     }
+
+    /**
+     * Get the UI HTML string for this tool from the registry.
+     * Returns the registered UI HTML, or undefined if no UI exists for this tool.
+     */
+    protected getUI(): string | undefined {
+        return this.uiRegistry?.get(this.name);
+    }
```
nirinchev: Do we need this now that it's all abstracted away in the base tool? I don't suppose inheritors need to reference `getUI`, so we should probably just fetch the HTML when needed.

src/tools/tool.ts
```diff
@@ -675,6 +712,61 @@ export abstract class ToolBase {
...
+            const validation = schema.safeParse(result.structuredContent);
+            if (!validation.success) {
+                this.session.logger.warning({
+                    id: LogId.toolExecute,
+                    context: `tool - ${this.name}`,
+                    message: `structuredContent failed validation against outputSchema, skipping UI resource: ${validation.error.message}`,
+                });
+                return result;
+            }
+        }
```
nirinchev: I don't think we should be doing this validation here. Right now we're in a bit of a flux since we don't have fully-fleshed out structured content support, but eventually, we'd like to use the type system to ensure `structuredContent` matches the schema. For this PR, I'd suggest that we just trust that it's adhering to the schema, without doing any manual validations.

src/tools/tool.ts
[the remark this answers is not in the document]
nirinchev: I would also experiment without the Date component - since the html is static, it lends itself well to caching, as long as the client correctly provides the distinct `initial-render-data` for each tool call.

src/ui/registry/registry.ts
```diff
@@ -0,0 +1,47 @@
+import { uiHtml } from "../generated/uiHtml.js";
+
+/**
+ * UI Registry that manages bundled UI HTML strings for tools.
+ *
+ * The default UIs are embedded at build time via the generated uiHtml module.
+ * Custom UIs can be provided at runtime to override or extend the defaults.
+ */
+export class UIRegistry {
+    private customUIs: Map<string, string> = new Map();
```
nirinchev: We should probably make this lazy, otherwise it will load a bunch of long strings on startup. I'm fine with caching the values, but we should only load them in memory when needed.

src/ui/registry/registry.ts
```diff
@@ -0,0 +1,47 @@
+import { uiHtml } from "../generated/uiHtml.js";
```
nirinchev: This seems to be .gitignored - while I know there's different opinions about generated code, I think I am more in favor of committing it as it makes it easier to review what's going on without checking out the repo and building it locally. If you are worried about it polluting the diff view, you can add the linguist-generated git attribute.

src/ui/registry/uiMap.ts
```diff
@@ -0,0 +1,12 @@
...
+ * Mapping from tool names to their UI component names.
+ *
+ * Tool names use kebab-case (e.g., 'list-databases')
+ * Component names use PascalCase (e.g., 'ListDatabases')
+ *
+ * The component name corresponds to the folder in src/ui/components/
+ * This mapping is used at BUILD TIME to generate the uiHtml module.
+ */
+export const uiMap: Record<string, string> = {
+    "list-databases": "ListDatabases",
```
nirinchev: Should we just make it a convention instead of keeping a map like this? Then at build time, we can discover all components and flag cases where a component name doesn't have a corresponding tool name.

package.json
```diff
@@ -112,9 +128,15 @@
...
     "typescript-eslint": "^8.41.0",
     "uuid": "^13.0.0",
+    "vite": "^5.4.21",
+    "vite-plugin-node-polyfills": "^0.24.0",
+    "vite-plugin-singlefile": "^2.3.0",
     "vitest": "^3.2.4"
   },
   "dependencies": {
+    "@emotion/css": "^11.13.5",
+    "@leafygreen-ui/table": "^15.2.2",
```
nirinchev: [Noob question] Should these be dependencies or devDependencies?

src/tools/mongodb/metadata/listDatabases.ts
tsck: This unfortunately isn't possible, I don't think. The UI gets rendered off of the server and wouldn't have access to a server export in the bundle, so it creates a dependency issue on bundle. I understand that not all tools will have UI's but all tools will have schemas, but I couldn't see a way around this aside from duplicating the schema. Definitely open to ideas here though!

src/tools/tool.ts
tsck: Good call, its only ever used in the class now. Will remove!

src/tools/tool.ts
tsck: That works for me. This was sort of an "extra layer" I added since we now had `outputSchema` that would prevent a UI response from being appended at all if we knew the validation was broken. Without this the only downside is that a UI could be theoretically be sent and rendered in an error state since data is invalid. I'm ok with trusting it at this point though. Will remove.

src/ui/registry/uiMap.ts
tsck: Convention was actually the original design. I changed it to use a map in order to hedge against a possible scenario where a component has a different name then a tool but I guess there is always a 1:1 mapping so perhaps that will/should never exist. I was also thinking the explicit config/map prevented "magic" lol, since adding it to the map was explicit, whereas with convention once the component exists, its being bundled. This all said, I was on the fence about this already so definitely okay with making it a convention. It will make creating a UI one step easier. Will update the build and remove the mapping.

src/ui/registry/registry.ts
tsck: I'm okay with this. Yeah, I knew these bundles were quite large so I didn't know if you all would want them committed to the repo. Happy to go that way with it though

src/tools/mongodb/metadata/listDatabases.ts
tsck: @TheSonOfThomp Curious if you have any ideas here?

src/ui/registry/registry.ts
tsck: That makes sense, will update

src/tools/mongodb/metadata/listDatabases.ts
tsck: Actually another kind of related issue. 

In reality since this is feature flagged, there might need to be two separate schemas, for now. Obviously we want to support `outputSchema` for non-UI augmented responses. But if our `outputSchema` is generated expecting UI in the response, like in `list-databases`, and the feature is disabled, the client will deem the response invalid. So I think for now at least, UI schemas, and tool schemas, kind of need to be separate my necessity.

This doesn't change the fact that the tool is importing a UI schema, which I think you were trying to avoid.

I'm pushing up a suggested change where there is a separate `ToolBase.uiOutputSchema` prop that's overrides the `outputSchema` in the response if it there's UI and the feature is enabled. This will allow a text only schema as well.

src/tools/tool.ts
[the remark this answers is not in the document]
tsck: The examples in the spec always show an instance ID but removing date altogether doesn't _appear_ to break anything, so this seems to be ok 🚀

src/server.ts
```diff
@@ -61,6 +62,21 @@ export interface ServerOptions {
...
+     * import { readFileSync } from 'fs';
+     * const server = new Server({
+     *     // ... other options
+     *     customUIs: {
+     *         'list-databases': readFileSync('./my-custom-ui.html', 'utf-8'),
+     *     }
+     * });
+     * ```
+     */
+    customUIs?: Record<string, string>;
```
gagik: can this information live in the tool itself?

.gitignore
```diff
@@ -14,6 +15,9 @@ coverage
 # Generated assets by accuracy runs
 .accuracy
 
+# Generated UI module (rebuilt by `pnpm build:ui`)
+src/ui/generated
```
gagik: would it be more conventional for this to be in like `lib/`?

src/tools/mongodb/metadata/listDatabases.ts
gagik: > UI gets rendered off of the server

my impression was that all UI-related things are at built time and not the bundle, do I not understand that correctly?
very roughly what I imagining can we not make `tool.uiSchema` is imported by i.e. `component-generator` and then the generated `getComponent()` is imported by `tool`?

gagik: alternative thing that maybe matches what you're saying: 

why have UI <-> tool be so deeply tied at all? 
can't we just have strongly typed UI text generators function and let any tool import them as needed?
i.e. (again very rough idea) 
```ts
// in UI src
function getListDatabasesHtml({ databases: Database[]): string;
// ...

// in a tool class
import { getListDatabasesHtml } from 'mongodb-mcp-server/ui';

class SomeTool { getUI() => getListDatabasesHtml({ databases }); }}; // this returns a string
```
then these generated UI components are a lot more reusable

tsck: > my impression was that all UI-related things are at built time and not the bundle, do I not understand that correctly?

As in, they're JIT bundling as opposed to doing so at build? This was something we went back and forth on in our last sync, but we ended up landing on bundling at build. I forget fully why, but we're not bundling on request. Please let me know if I'm misunderstanding the question!

> can't we just have strongly typed UI text generators function and let any tool import them as needed?
i.e. (again very rough idea)

If I'm understanding correctly, this was the original design to some extent in https://github.com/mongodb-js/mongodb-mcp-server/pull/783/commits/0f9a81229ca340a305bd02866352afddb6c0e388 where the tool itself was getting the UI. The issue communicated was that this pushed the UI business logic up to the tool itself. I do think the current design of not having to do anything in the tool to get the UI is quite nice.

package.json
tsck: Nope that's a great question, they are definitely dev dependencies in this case. Updating!

src/server.ts
nirinchev: I would prefer that it lives outside so we have a clear separation between the business logic and the presentation.

src/tools/mongodb/metadata/listDatabases.ts
nirinchev: I'll need to dive deeper here, but a couple of questions:
1. Does having UI impact the shape of output schema? Why is that?
2. At build time, why can't we have `ListDatabases.tsx` import the schema from listDatabases.ts? Alternatively, we can just import the typescript type - we probably don't need the safe parsing there and just trust that the data we receive has the correct shape.

src/ui/registry/registry.ts
tsck: This change has been made. One thing I realized retrospectively is that because bundles are minified, it has nondeterministic variable names generated. This means that every build generates a unique file, which shows as a diff and needs to be committed. That obviously won't work.

The way around this is to use something like terser, which has deterministic variable naming support, but would add time to the build. The time would be negligible currently, but I think we should revisit this. With these bundles being so massive and minified, is there really value in having them committed to the repo? I'm just not sure someone will be debugging a file like this in Github. It leads me to feel that ignoring them is still the right move here.

vite.ui.config.ts
```diff
@@ -0,0 +1,157 @@
...
+                }
+                const html = readFileSync(htmlFile, "utf-8");
+
+                const toolModuleContent = `/**
+ * AUTO-GENERATED FILE - DO NOT EDIT MANUALLY
+ * Generated by: vite build --config vite.ui.config.ts
+ * Tool: ${toolName}
+ * Component: ${componentName}
+ */
+export default ${JSON.stringify(html)};
```
gagik: very nit but can this be a named const export like `${componentName}` maybe with some suffix? would make future refactoring easier

src/ui/registry/registry.ts
```diff
@@ -0,0 +1,38 @@
...
+            return customUI;
+        }
+
+        const cached = this.cache.get(toolName);
+        if (cached !== undefined) {
+            return cached;
+        }
+
+        try {
+            const module = (await import(`../generated/tools/${toolName}.js`)) as { default: string };
```
gagik: not 100% sure if this works in CJS but if it does sgtm

src/tools/mongodb/metadata/listDatabases.ts
gagik: If I understand the issue right, maybe what we're missing is, instead of defining a mapping tool names to their component string, we should also specify a transformer function which takes information from `structuredSchema` of the specified tool and produced the needed input for the component?

tsck: > Does having UI impact the shape of output schema? Why is that?

Ok I dug into this more and I think it's actually a bug in MCPJam. It has some validation code that runs when a schema is present that looks like this:

```ts
//  Validate raw content string (if it's a string)
if (typeof result.content[0].text === "string") {
  try {
    const parsedContent = JSON.parse(result.content[0].text);
    // ...
  } catch (e) {
  ```
In our case `result.content[0].text` is "Found 3 databases" - which isn't valid JSON. This seems incorrect though according to the MCP spec as far as I can tell. I don't know why it doesn't validate when the UI response is present. Perhaps it skips it when it knows it's going to render UI, but I didn't dig deep enough. 

Either way, I'm going revert this change because I think it is solving for a bug/quirk in MCPJam.

> At build time, why can't we have ListDatabases.tsx import the schema from listDatabases.ts?

This was because in the import chain, `listDatabases` imports a node only package. But I just thought of a very simple solution - why not just have a shard `schemas/` directory at the same level as `tools/` and `ui/`? I'll push up a change for this for now, but let me know if this is problematic?

File: .gitignore (hunk 14 above)
Comment (tsck) [reply .gitignore]: I think this makes sense. I'll update the script accordingly

File: vite.ui.config.ts (hunk 15 above)
Comment (tsck) [reply vite.ui.config.ts]: That works for me. I just went with `export const ${componentName}Html` but let me know if you had something else in mind

File: src/ui/registry/registry.ts (hunk 10 above)
Comment (tsck) [reply src/ui/registry/registry.ts]: Another argument for not ignoring this is that if you do update something, you have to remember to build before committing or the generated file will be out of date.

File: src/tools/mongodb/metadata/listDatabases.ts (hunk 6 above)
Comment (tsck) [reply src/tools/mongodb/metadata/listDatabases.ts]: Just remembering that we also have `src/common/` and `src/common/schemas.ts` specifically, but not sure if this actually fits here

Comment (nirinchev) [reply src/tools/mongodb/metadata/listDatabases.ts]: I like the colocation of the tool implementation and the output schema, but if we can't find an alternative, I think separating them is not the end of world.

> This was because in the import chain, listDatabases imports a node only package

Is this still an issue if we export the output schema type definition and do a type-only import? Assuming all the validations work as intended, I assume that in the react component we only care about the render data shape and don't need anything from the actual zod schema.

File: src/ui/registry/registry.ts (hunk 10 above)
Comment (nirinchev) [reply src/ui/registry/registry.ts]: Yeah, that's fair. The main challenge with not committing them that I see is that we'd need to still ensure they end up in the final package that we publish to npm. It also creates a risk that what is being tested is not guaranteed to be the same as what's being published, though that's probably not overly concerning.

File: src/tools/mongodb/metadata/listDatabases.ts (hunk 6 above)
Comment (tsck) [reply src/tools/mongodb/metadata/listDatabases.ts]: > Is this still an issue if we export the output schema type definition and do a type-only import? Assuming all the validations work as intended, I assume that in the react component we only care about the render data shape and don't need anything from the actual zod schema.

I will test this and let you know shortly

File: src/ui/registry/registry.ts (hunk 10 above)
Comment (tsck) [reply src/ui/registry/registry.ts]: > we'd need to still ensure they end up in the final package that we publish to npm

I guess I'm not seeing the scenario where we'd see this not occur given that its a part of the build step.

> risk that what is being tested is not guaranteed to be the same as what's being published, though that's probably not overly concerning.

I think perhaps this argument could be made for any frontend component that is ultimately bundled. You build the component in a dev environment but then its bundled into a production bundle when its ultimately published. So there is a bit of relying on the build tool to bundle as expected.

This all said, if you feel strongly about having them committed, I'm happy to explore integrating Terser and seeing if we can bundle in a deterministic way.

File: src/tools/mongodb/metadata/listDatabases.ts (hunk 6 above)
Comment (tsck) [reply src/tools/mongodb/metadata/listDatabases.ts]: Ok thinking about this further, I think it's a really good point. We had validation in the component originally because it was going to be served at a URL and not bundled on the server. Now that's not the case. So validating in the component is completely redundant. I'm going to remove the validation from the component and colocated the schema back to the tool. I'll resolve this thread once that's pushed.

File: src/ui/registry/registry.ts (hunk 10 above)
Comment (tsck) [reply src/ui/registry/registry.ts]: I have gone back to ignoring the generated files for now given the reasons stated above, so I'll resolve this thread. If there is strong disagreement please re-open

File: src/tools/tool.ts [hunk 17]
```diff
@@ -449,15 +486,17 @@ export abstract class ToolBase {
                 });
 
                 const result = await this.execute(args, { signal });
-                this.emitToolEvent(args, { startTime, result });
+                const finalResult = await this.appendUIResource(result);
```
```suggestion
                const toolCallResult = await this.execute(args, { signal });
                this.emitToolEvent(args, { startTime, result: toolCallResult });
                const result = await this.appendUIResource(toolCallResult);
```
nit: just to be clearer with the naming about what is what

src/tools/tool.ts
```diff
@@ -449,15 +486,17 @@ export abstract class ToolBase {
...
+
+                this.emitToolEvent(args, { startTime, result: finalResult });
```
gagik: is it intentional here to emit a tool event twice?

gagik (commented):
Haven't tried running the project itself just yet but changes LGTM! Would be nice to add some tests to validate this is working as intended.

tsck:
> Haven't tried running the project itself just yet but changes LGTM! Would be nice to add some tests to validate this is working as intended.

@gagik I have another PR open for that! It's probably a bit outdated now because its based on this PR which has changed since, but I kept it separate in order to not bloat this more. Happy to merge it in if that's better on your end https://github.com/mongodb-js/mongodb-mcp-server/pull/787

Also just a note about trying it - it won't work in the built-in inspector because that doesn't support mcp ui. To try the UI response you'll need to use another inspector, I've been using [MCPJam](https://github.com/MCPJam/inspector)

src/tools/tool.ts
tsck: I don't believe its getting emitted again after this

src/tools/tool.ts
```diff
@@ -404,14 +432,23 @@ export abstract class ToolBase {
      * or inputs during tool execution.
      */
     protected readonly elicitation: Elicitation;
-
-    constructor({ category, operationType, session, config, telemetry, elicitation }: ToolConstructorParams) {
+    protected readonly uiRegistry?: UIRegistry;
```
nirinchev: Seems like this is always provided + we can make it private as I don't believe inheritors need to know about it.

```suggestion
    private readonly uiRegistry: UIRegistry;
```

src/ui/registry/registry.ts
```diff
@@ -0,0 +1,51 @@
...
+
+        const cached = this.cache.get(toolName);
+        if (cached !== undefined) {
+            return cached;
+        }
+
+        try {
+            const module = (await import(`../lib/tools/${toolName}.js`)) as Record<string, string>;
+            const exportName = `${toPascalCase(toolName)}Html`;
+            const html = module[exportName];
```
nirinchev: If this is generated, might be a good idea to add a comment to explain it.

src/server.ts
nirinchev: This is still eagerly evaluating large htmls - would be better to make it a function instead.

src/tools/tool.ts
```diff
@@ -19,7 +22,7 @@ export type ToolExecutionContext = {
 };
 
 /**
- * The type of operation the tool performs. This is used when evaluating if a tool is allowed to run based on
+ * The ~`type of operation the tool performs. This is used when evaluating if a tool is allowed to run based on
```
nirinchev: Is this intentional?

tsck: Nope 😅 thanks for catching, will update before merge

PR #786, opened 2025-12-08
chore: fix publish workflow for MCP registry

himanshusinghs:
## Proposed changes
Our publish job which publishes the artifacts to MCP registry depends on the docker image being available already but its not when we publish. The docker images are published on a scheduled basis or on manual dispatch.

This PR changes the publish workflow such that we publish to docker as soon as we have the npm package listed in the registry and then proceed ahead with MCP registry publish.

An example failure - https://github.com/mongodb-js/mongodb-mcp-server/actions/runs/19944234451/job/57189691853

## Checklist

### DIFF
--- a/.github/workflows/mcp-publish.yml
+++ b/.github/workflows/mcp-publish.yml
@@ -1,0 +1,28 @@
+---
+name: Publish to MCP Registry
+on:
+  workflow_call:
+  workflow_dispatch:
+
+jobs:
+  mcp-publish:
+    runs-on: ubuntu-latest
+    environment: Production
+    permissions:
+      id-token: write
+      contents: read
+    steps:
+      - uses: GitHubSecurityLab/actions-permissions/monitor@v1
+      - uses: actions/checkout@v5
+        with:
+          persist-credentials: false
+
+      - name: Install MCP Publisher
+        run: |
+          curl -L "https://github.com/modelcontextprotocol/registry/releases/latest/download/mcp-publisher_$(uname -s | tr '[:upper:]' '[:lower:]')_$(uname -m | sed 's/x86_64/amd64/;s/aarch64/arm64/').tar.gz" | tar xz mcp-publisher
+
+      - name: Login to MCP Registry
+        run: ./mcp-publisher login github-oidc
+
+      - name: Publish to MCP Registry
+        run: ./mcp-publisher publish
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -105,50 +105,38 @@ jobs:
         run: |
           PACKAGE_NAME=$(jq -r '.name' < package.json)
           VERSION="${{ needs.check.outputs.VERSION }}"
+          # Strip the 'v' prefix for npm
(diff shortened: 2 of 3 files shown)


.github/workflows/docker-publish.yml
```diff
@@ -0,0 +1,50 @@
+---
+name: Build and Push Docker Image
```
gagik, line 2: q: should this be a workflow or an action? tbh not sure what pros and cons of each are

himanshusinghs: They both could be used for reusability. I used the workflow here only because they have better support for credentials masking. I don't have a strong feeling about it though as I only recently found about this and was curious to see this in action.

.github/workflows/publish.yml
```diff
@@ -101,12 +101,42 @@ jobs:
         run: |
           gh release create ${{ needs.check.outputs.VERSION }} --title "${{ needs.check.outputs.VERSION }}" --generate-notes --target ${{ github.sha }} ${{ (needs.check.outputs.RELEASE_CHANNEL != 'latest' && '--prerelease') || ''}}
 
-      - name: Install MCP Publisher
+      - name: Wait for package to be available on npm
         run: |
-          curl -L "https://github.com/modelcontextprotocol/registry/releases/latest/download/mcp-publisher_$(uname -s | tr '[:upper:]' '[:lower:]')_$(uname -m | sed 's/x86_64/amd64/;s/aarch64/arm64/').tar.gz" | tar xz mcp-publisher
+          PACKAGE_NAME=$(jq -r '.name' < package.json)
```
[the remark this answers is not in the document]
himanshusinghs: jq is available in runners and package.json will always be there

.github/workflows/publish.yml
```diff
@@ -101,12 +101,42 @@ jobs:
...
+          MAX_ATTEMPTS=30
+          SLEEP_SECONDS=10
 
-      - name: Login to MCP Registry
-        run: ./mcp-publisher login github-oidc
+          echo "Waiting for ${PACKAGE_NAME}@${NPM_VERSION} to be available on npm..."
 
-      - name: Publish to MCP Registry
-        run: ./mcp-publisher publish
+          for i in $(seq 1 $MAX_ATTEMPTS); do
```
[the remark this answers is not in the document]
himanshusinghs: we will do that when we across such requirement

.github/workflows/mcp-publish.yml
```diff
@@ -0,0 +1,28 @@
+---
+name: Publish to MCP Registry
+on:
+  workflow_call:
+  workflow_dispatch:
```
fmenezes, line 5: nit: I think you only need this on the other one

```suggestion
```

himanshusinghs: This in intentional because we still want to be able to publish to MCP registry manually in case the relevant parts in publish job fails.

Also relevant for the current v1.3.0 release which was never published in MCP registry so this is expected to help with that as well.

cveticm (approved):
lgtm

PR #796, opened 2025-12-11
chore: add telemetry properties for vector search and mcp auto-embed usage

himanshusinghs:
## Proposed changes
This PR adds the following telemetry properties to Server events:
1. `previewFeatures`: Answers what preview features are being enabled
2. `embeddingProviderConfigured`: Answers whether there is an embedding provider configured or not

and adds another property `embeddingsGeneratedBy` to Tool events to answer whether MCP server auto-embeddings feature is being used or not.

## Checklist

### DIFF
--- a/src/tools/mongodb/read/aggregate.ts
+++ b/src/tools/mongodb/read/aggregate.ts
@@ -352,8 +352,10 @@ Returning ${documents.length} documents${appliedLimitText ? ` ${appliedLimitText
     ): ConnectionMetadata | AutoEmbeddingsUsageMetadata {
         const [maybeVectorStage] = args.pipeline;
         if (
-            maybeVectorStage &&
-            (maybeVectorStage as z.infer<typeof VectorSearchStage>)?.["$vectorSearch"]?.embeddingParameters &&
+            maybeVectorStage !== null &&
+            maybeVectorStage instanceof Object &&
+            "$vectorSearch" in maybeVectorStage &&
+            "embeddingParameters" in maybeVectorStage["$vectorSearch"] &&
             this.config.voyageApiKey
         ) {
             return {


src/tools/mongodb/read/aggregate.ts
```diff
@@ -344,4 +345,23 @@ The aggregation resulted in ${aggResultsCount === undefined ? "indeterminable nu
...
+            this.config.voyageApiKey
+        ) {
+            return {
+                ...super.resolveTelemetryMetadata(args, { result }),
+                embeddingsGeneratedBy: "mcp",
+            };
+        } else {
+            return super.resolveTelemetryMetadata(args, { result });
+        }
+    }
```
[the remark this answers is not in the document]
himanshusinghs: the checks are isolated and expected to change as we add support for auto-embed index.

src/telemetry/types.ts
```diff
@@ -45,6 +45,8 @@ export type ServerEventProperties = {
     read_only_mode?: TelemetryBoolSet;
     disabled_tools?: string[];
     confirmation_required_tools?: string[];
+    previewFeatures?: string[];
```
fmenezes, line 48: note: while we can add arrays in telemetry, this really makes stuff like amplitude difficult as the values appear as one text with spaces normally.

ordering of strings is also a problem as you can't queries based on keys or position.

since this pattern is already stablished let's proceed, but ideally we should be removing arrays from segment events.

src/tools/mongodb/read/aggregate.ts
```diff
@@ -344,4 +345,23 @@ The aggregation resulted in ${aggResultsCount === undefined ? "indeterminable nu
 Returning ${documents.length} documents${appliedLimitText ? ` ${appliedLimitText}` : "."}\
 `;
     }
+
+    protected resolveTelemetryMetadata(
+        args: ToolArgs<typeof this.argsShape>,
+        { result }: { result: CallToolResult }
+    ): ConnectionMetadata | AutoEmbeddingsUsageMetadata {
+        const [maybeVectorStage] = args.pipeline;
```
gagik, line 353: is the vector stage always first?

himanshusinghs: Yea Vector stage, if included, should always be at the first position in the pipeline.

src/tools/mongodb/read/aggregate.ts
```diff
@@ -344,4 +345,23 @@ The aggregation resulted in ${aggResultsCount === undefined ? "indeterminable nu
...
+        if (
+            maybeVectorStage &&
+            (maybeVectorStage as z.infer<typeof VectorSearchStage>)?.["$vectorSearch"]?.embeddingParameters &&
```
gagik: this is hard to read; can it be
```ts
`maybeVectorStage instanceof Object && "$vectorSearch" in maybeVectorStage && "embeddingParameters" in maybeVectorStage["$vectorSearch"]
```

gagik (approved):

PR #798, opened 2025-12-12
docs: list all available tools with a generate script MCP-340

gagik:
Our README and consequently our documentation was missing certain tools. To make sure we keep track of this, this adds a new script for generating documentation for our tools based on our tool exports.

Will trigger a DOCS ticket after merge too so we can list all tools properly.


scripts/generate/generateArguments.ts
```diff
@@ -302,21 +300,21 @@ function generateReadmeConfigTable(argumentInfos: ArgumentInfo[]): string {
 
         const desc = argumentInfo.description.replace(/\|/g, "\\|"); // Escape pipes in description
         rows.push(
-            `| ${cliOption.padEnd(38)} | ${envVarName.padEnd(51)} | ${defaultValueString.padEnd(75)} | ${desc.padEnd(199)} |`
+            `| ${`${envVarName} or ${cliOption}`.padEnd(89)} | ${defaultValueString.padEnd(75)} | ${desc.padEnd(199)} |`
```
gagik: drive-by: it'd a lot more compact and easy to read the table if we combine the CLI argument and the env argument into one column

README.md
```diff
@@ -345,37 +350,37 @@ The MongoDB MCP Server can be configured using multiple methods, with the follow
...
-| `vectorSearchSimilarityFunction`       | `MDB_MCP_VECTOR_SEARCH_SIMILARITY_FUNCTION`         | `"euclidean"`                                                                                          | Default similarity function for vector search: 'euclidean', 'cosine', or 'dotProduct'.                                                                                                          |
-| `voyageApiKey`                         | `MDB_MCP_VOYAGE_API_KEY`                            | `""`                                                                                                   | API key for Voyage AI embeddings service (required for vector search operations with text-to-embedding conversion).                                                                             |
+| Environment Variable / CLI Option                                                              | Default                                                                                                | Description                                                                                                                                                                                     |
```
gagik, line 353: as a drive by I am also combining these columns to have a more compact viewing experience

scripts/generate/generateToolDocumentation.ts
```diff
@@ -0,0 +1,149 @@
...
+            telemetry: {
+                emitEvents: () => {},
+            } as never,
+            elicitation: {
+                requestConfirmation: () => Promise.resolve(false),
+            } as never,
+        };
+
+        try {
+            const instance = new ToolClass(dummyParams);
```
himanshusinghs, line 59: Looks like we could've benefitted from the parameters being static on the class 😄

himanshusinghs (approved):
Thanks for this. This is gonna remove much of guesswork now 🙌

scripts/generate/generateToolDocumentation.ts
gagik: yeah this is a bit of a hack.... but I figured for a script like this it'll do for now.
should we make these static I'd be happy to remove

PR #800, opened 2025-12-15
chore: support schema overrides and custom parserOptions MCP-338

gagik:
This allows for more flexibility when parsing, including specifying custom defaults, environment prefixes, etc..

### DIFF
--- a/MCP_SERVER_LIBRARY.md
+++ b/MCP_SERVER_LIBRARY.md
@@ -820,7 +820,11 @@ _Note: This is what MongoDB MCP server uses internally._
 **Example:**
 
 ```typescript
-import { parseUserConfig, StdioRunner, UserConfigSchema } from "mongodb-mcp-server";
+import {
+  parseUserConfig,
+  StdioRunner,
+  UserConfigSchema,
+} from "mongodb-mcp-server";
 
 // Parse config from process.argv and environment variables
 const config = parseUserConfig({
@@ -829,7 +833,7 @@ const config = parseUserConfig({
   // This can be used, for example, to set new defaults.
   overrides: {
     readOnly: UserConfigSchema.shape.readOnly.default(true),
-  }
+  },
 });
 
 const runner = new StdioRunner({ userConfig: config });


src/lib.ts
```diff
@@ -1,7 +1,7 @@
 export { Server, type ServerOptions } from "./server.js";
 export { Session, type SessionOptions } from "./common/session.js";
 export { type UserConfig, UserConfigSchema } from "./common/config/userConfig.js";
-export { createUserConfig as parseCliArgumentsAsUserConfig } from "./common/config/createUserConfig.js";
```
gagik, line 4: so I unfortunately missed that we're doing this; it's weird to have this discrepancy where our external export is named differently than our internal one. To me the idea that the name isn't straightforward was a code smell.

`parseCliArgumentsAsUserConfig` is inherently misleading though since by default it also parses environment variables. This can mislead users so it's best to rename it.

It is technically a library breaking change though.

MCP_SERVER_LIBRARY.md
```diff
@@ -811,26 +811,19 @@ const customConfig = UserConfigSchema.parse({
...
+### parseUserConfig
 
 Utility function to parse command-line arguments and environment variables into a UserConfig object, using the same parsing logic as the MongoDB MCP server CLI.
 
 _Note: This is what MongoDB MCP server uses internally._
 
-```typescript
-function parseCliArgumentsAsUserConfig(options?: {
-  args?: string[];
-  helpers?: CreateUserConfigHelpers;
```
gagik, line 823: the concept of CreateUserConfigHelpers was unused.

src/lib.ts
cveticm: Then we anticipate the release with this work will bump major version, correct?

MCP_SERVER_LIBRARY.md
```diff
@@ -811,26 +811,19 @@ const customConfig = UserConfigSchema.parse({
...
-
 **Example:**
 
 ```typescript
-import { parseCliArgumentsAsUserConfig, StdioRunner } from "mongodb-mcp-server";
+import { parseUserConfig, StdioRunner } from "mongodb-mcp-server";
 
 // Parse config from process.argv and environment variables
-const config = parseCliArgumentsAsUserConfig();
+const config = parseUserConfig();
```
cveticm: Q for my own learning: Why is it that we do not have to provide any inputs here?

From the implementation, `parseUserConfig()` requires `args:string[]` ([here](https://github.com/mongodb-js/mongodb-mcp-server/pull/800/files#:~:text=%7D%3A%20%7B-,args%3A%20string%5B%5D%3B,-overrides%3F%3A%20z4)) I don't see where in the implementation process.argv and environment variables are taken as args if no args is provided.

gagik: good find! will add this

src/common/config/parseUserConfig.ts
```diff
@@ -4,19 +4,51 @@ import type { Secret } from "../keychain.js";
...
+    args: string[];
+    overrides?: z4.ZodRawShape;
+    parserOptions?: ParserOptions;
+}): {
     warnings: string[];
     parsed: UserConfig | undefined;
     error: string | undefined;
 } {
-    const { error: parseError, warnings, parsed } = parseUserConfigSources(args);
+    const schema = overrides
```
cveticm, line 44: Non-blocking q: Which takes precedent: env vars or overrides?
I suspect env vars are applied to the UserConfiSchema before we get to this point and so overrides take precedent.

gagik: generally overrides are applied to the schema and in practice likely only going to be used to modify the default, i.e.
```ts
{
    readOnly: UserConfigSchema.shape.readOnly.default(true)
}
```
so any value (CLI, ENV) will override the default.

I will add some documentation to make it clearer

src/lib.ts
nirinchev: Technically, yes, though to avoid a major version bump, let's re-export with the old name and mark as deprecated?

PR #812, opened 2025-12-18
feat(mcp-ui): add hook for posting data back to MCP client

tsck:
## Proposed changes
- Add `useHostCommunication` hook for sending UI actions to the parent window via postMessage
- Updates `useRenderData` hook to capture origin, to be used in comms once available for better security
- Adds tests for both data hooks

### DIFF
--- a/tests/unit/ui/useHostCommunication.test.ts
+++ b/tests/unit/ui/useHostCommunication.test.ts
@@ -36,13 +36,14 @@ describe("useHostCommunication", () => {
         result.current.intent("create-task", { title: "Test Task" });
 
         expect(postMessageMock).toHaveBeenCalledWith(
-            {
+            expect.objectContaining({
                 type: "intent",
-                payload: {
+                // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+                payload: expect.objectContaining({
                     intent: "create-task",
                     params: { title: "Test Task" },
-                },
-            },
+                }),
+            }),
             "*"
         );
     });
@@ -53,13 +54,14 @@ describe("useHostCommunication", () => {
         result.current.intent("cancel", {});
 
         expect(postMessageMock).toHaveBeenCalledWith(
-            {
+            expect.objectContaining({
                 type: "intent",
-                payload: {
+                // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+                payload: expect.objectContaining({
                     intent: "cancel",
                     params: {},
-                },
-            },
+                }),
+            }),
             "*"
         );
     });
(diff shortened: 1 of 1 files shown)


mointerior36-lang (approved):

src/ui/hooks/useHostCommunication.ts
```diff
@@ -0,0 +1,140 @@
...
+        () => ({
+            intent,
+            notify,
+            prompt,
+            tool,
+            link,
+            reportSizeChange,
+        }),
+        [intent, notify, prompt, tool, link, reportSizeChange]
+    );
```
TheSonOfThomp: Not sure this `useMemo` is necessary, especially since each entry is wrapped in a `useCallback`

src/ui/hooks/useHostCommunication.ts
```diff
@@ -0,0 +1,140 @@
+import { useCallback, useMemo } from "react";
+
+interface SendMessageOptions {
+    targetOrigin?: string;
+}
+
+/** Return type for the useHostCommunication hook */
+interface UseHostCommunicationResult {
```
TheSonOfThomp, line 12: are any of these types available from `mcp-ui/server` ?

tests/unit/ui/useHostCommunication.test.ts
```diff
@@ -0,0 +1,235 @@
...
+
+interface HookOptions {
+    targetOrigin?: string;
+}
+
+/**
+ * Simple hook testing utility that renders a component using the hook
+ * and captures the result for assertions.
+ */
+function testHook(options?: HookOptions): UseHostCommunicationResult {
```
TheSonOfThomp: might be worth pulling in RTL

tests/unit/ui/useHostCommunication.test.ts
```diff
@@ -0,0 +1,235 @@
...
+        actions.intent("create-task", { title: "Test Task" });
+
+        expect(postMessageMock).toHaveBeenCalledWith(
+            {
+                type: "intent",
+                payload: {
+                    intent: "create-task",
+                    params: { title: "Test Task" },
+                },
+            },
```
TheSonOfThomp: nit: not necessary here, but might want to use `expect.objectContaining` for furute-proofing

src/ui/hooks/useHostCommunication.ts
tsck: Digging into this, the types aren't really there. However, I found an exported util `postUIActionResult` and helper functions for each post type. I'm going to use these and derive the types from them instead. Our implementation technically provided more functionality, like custom origins, but I think its better to stay as aligned with the sdk as possible

PR #817, opened 2025-12-23
feat: add body limit config option for streamable HTTP - MCP-348

blva:
## Proposed changes
Adds a new configuration option to control the maximum size of HTTP request bodies when using the HTTP transport. This addresses `PayloadTooLargeError` issues when clients send large JSON payloads.

- Default value: 100kb (Express.js default)
- Supports a given number of bytes with min being 100kb
- Only applies when transport is 'http'

## Checklist

### DIFF
--- a/src/common/config/userConfig.ts
+++ b/src/common/config/userConfig.ts
@@ -130,7 +130,10 @@ const ServerConfigSchema = z4.object({
     httpBodyLimit: z4.coerce
         .number()
         .int()
-        .min(TRANSPORT_PAYLOAD_LIMITS.http, `Invalid httpBodyLimit: must be at least ${TRANSPORT_PAYLOAD_LIMITS.http} bytes`)
+        .min(
+            TRANSPORT_PAYLOAD_LIMITS.http,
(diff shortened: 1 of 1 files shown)


src/common/config/userConfig.ts
```diff
@@ -126,6 +126,14 @@ const ServerConfigSchema = z4.object({
             "Header that the HTTP server will validate when making requests (only used when transport is 'http')."
         )
         .register(configRegistry, { overrideBehavior: "not-allowed" }),
+    httpBodyLimit: z4
+        .string()
+        .regex(/^\d+(?:kb|mb|gb)$/i, "Invalid httpBodyLimit: must be a string like '100kb', '1mb', or '1gb'")
+        .default("100kb")
+        .describe(
+            "Maximum size of the HTTP request body (only used when transport is 'http'). Supports formats like '100kb', '1mb', '50mb'. This is the Express.js json() middleware limit."
```
cveticm: nit as it took me some time to understand what was meant by this
```suggestion
            "Maximum size of the HTTP request body (only used when transport is 'http'). Supports formats like '100kb', '1mb', '50mb'. This value is passed as the optional limit parameter to the Express.js json() middleware"
```

PR #819, opened 2026-01-02
feat: add transport metadata to all tools - MCP-348

blva:
## Proposed changes
Adds the transport information to the tool metadata, hoping that it can support clients and LLMs with how they organize their request. 
- https://modelcontextprotocol.io/specification/2025-11-25/basic/index#_meta

## Checklist

### DIFF
--- a/src/tools/mongodb/create/insertMany.ts
+++ b/src/tools/mongodb/create/insertMany.ts
@@ -28,7 +28,7 @@ const commonArgs = {
 export class InsertManyTool extends MongoDBToolBase {
     public name = "insert-many";
     protected description =
-        "Insert an array of documents into a MongoDB collection. If the list of documents is above maxRequestPayloadBytes, consider inserting them in batches.";
+        "Insert an array of documents into a MongoDB collection. If the list of documents is above com.mongodb/maxRequestPayloadBytes, consider inserting them in batches.";
     protected argsShape = this.isFeatureEnabled("search")
         ? {
               ...commonArgs,
--- a/src/tools/mongodb/update/updateMany.ts
+++ b/src/tools/mongodb/update/updateMany.ts
@@ -8,7 +8,7 @@ import { zEJSON } from "../../args.js";
 export class UpdateManyTool extends MongoDBToolBase {
     public name = "update-many";
     protected description =
-        "Updates all documents that match the specified filter for a collection. If the list of documents is above maxRequestPayloadBytes, consider updating them in batches.";
+        "Updates all documents that match the specified filter for a collection. If the list of documents is above com.mongodb/maxRequestPayloadBytes, consider updating them in batches.";
     protected argsShape = {
         ...DbOperationArgs,
         filter: zEJSON()
--- a/tests/integration/tools/mongodb/create/insertMany.test.ts
+++ b/tests/integration/tools/mongodb/create/insertMany.test.ts
@@ -23,7 +23,7 @@ describeWithMongoDB("insertMany tool when search is disabled", (integration) => 
     validateToolMetadata(
         integration,
         "insert-many",
-        "Insert an array of documents into a MongoDB collection. If the list of documents is above maxRequestPayloadBytes, consider inserting them in batches.",
+        "Insert an array of documents into a MongoDB collection. If the list of documents is above com.mongodb/maxRequestPayloadBytes, consider inserting them in batches.",
         "create",
         [
             ...databaseCollectionParameters,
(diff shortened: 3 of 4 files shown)


src/tools/tool.ts
```diff
@@ -334,6 +360,42 @@ export abstract class ToolBase {
...
 
+    /**
+     * Returns tool-specific metadata that will be included in the tool's `_meta` field.
+     *
+     * This getter computes metadata based on the current configuration, including
+     * transport-specific constraints like request payload size limits.
+     *
+     * The metadata includes:
+     * - `mongodb.transport`: The transport protocol in use ("stdio" or "http")
+     * - `mongodb.maxRequestPayloadBytes`: Maximum request payload size for the current transport
```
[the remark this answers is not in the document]
jeroenvervaeke: I have the same question.

You mention `mongodb.*`, but the code uses `com.mongodb/*`

src/tools/tool.ts
```diff
@@ -17,6 +17,32 @@ export type ToolArgs<T extends ZodRawShape> = {
...
+ *
+ * @remarks
+ * These values are exposed in each tool's `_meta` field as `com.mongodb/maxRequestPayloadBytes`
+ * to inform LLMs about the size constraints when constructing tool call arguments.
+ */
+export const TRANSPORT_PAYLOAD_LIMITS: Record<TransportType, number> = {
+    /** Default limit for stdio transport (50MB - conservative system limit) */
+    stdio: 52_428_800,
+    /** Default limit for HTTP transport (100KB - Express.js default for JSON body) */
+    http: 102_400,
```
nirinchev: Should we extract those to consts in their respective runners and also apply them explicitly? Right now we're relying on this matching express's default, but there's nothing enforcing it.

nirinchev (approved):
One suggestion, otherwise looks good.

src/tools/tool.ts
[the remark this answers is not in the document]
blva: yes, that's because I applied the metadata best practices later and forgot to update the comments [apply _meta best practices](https://github.com/mongodb-js/mongodb-mcp-server/pull/819/commits/98a9a856530879fcda018f914a026a46372235c7)

we should always use the reverse DNS notation based on the MCP protocol `Implementations SHOULD use reverse DNS notation (e.g., com.example/ rather than example.com/).`

src/tools/tool.ts
blva: Yep, I have another PR to make it configurable, which I'll need to adjust if we agree to do that, but I can do this here proactively as well!

tests/unit/toolBase.test.ts
```diff
@@ -260,6 +260,36 @@ describe("ToolBase", () => {
             }
         });
     });
+
+    describe("toolMeta", () => {
```
nirinchev, line 265: Might be a good idea to add an actual integration test that ensures that the `_meta` in the response is indeed what `toolMeta` returns. Right now, if I delete line 597 in tool.ts, the tests will still pass.

src/tools/tool.ts
nirinchev: I see - I'm fine either way. The only thing to keep in mind is that in your other PR you're expressing the body limit as a string (100kb/1mb), whereas here you're translating it to raw bytes. I don't think it matters a whole lot to a model, but something to keep in mind.

blva: yeah, no worries I'll address it here and we can keep the other one smaller. I think for now I'll keep as a byte number

blva: oh! actually it seems that express also accepts a number, so we should be OK, I just had to triple check the way they calculate 100kb > in summary they use this library https://www.npmjs.com/package/bytes so we'll use:

bytes('1KB');
// output: 1024

tests/unit/toolBase.test.ts
blva: added an integration test

nirinchev (approved):

src/tools/mongodb/create/insertMany.ts
```diff
@@ -27,7 +27,8 @@ const commonArgs = {
 
 export class InsertManyTool extends MongoDBToolBase {
     public name = "insert-many";
-    protected description = "Insert an array of documents into a MongoDB collection";
+    protected description =
+        "Insert an array of documents into a MongoDB collection. If the list of documents is above maxRequestPayloadBytes, consider inserting them in batches.";
```
jeroenvervaeke: How will the LLM figure out how much `maxRequestPayloadBytes` is?

Should the description say `com.mongodb/maxRequestPayloadBytes` instead?
Or should it contain the exact number of bytes?

blva: it's a good question, my rationale is to avoid calling out the specific limit again here so it doesn't get outdated, so we could try to use the full metadata description instead, the format might also help hinting it's a metadata value

PR #820, opened 2026-01-02
chore: extend accuracy tests

blva:
## Proposed changes
- add accuracy test to check that large files are proactively split in batches for the `insertMany` command

## Checklist


nirinchev (commented):
Does this test work for you locally? Looking at the results on CI, there are a few issues:
1. The prompt breaks the html formatting (those long documents blow up the table displaying the results).
2. Only 1 `insert-many` is invoked and with that, with just a single document. Haven't investigated why that is - we might be blowing up the context window or something.
3. The accuracy tests are setup with stdio transport, so the 100kb limit shouldn't be applied here - unless you want to adjust this to use the http transport, we could instead use the new config option to lower the payload limit.

PR #822, opened 2026-01-05
chore: make fields public for library usage

gagik:
This makes the following fields public:
- different tool fields: description, argShape, invoke (our tool callback), operationType, category, outputSchema
- `registerTools` and `registerResources` - this can let one do this without having to call connect.

All of this is motivated by the needs identified in https://github.com/mongodb-js/compass/pull/7675 for Compass Assistant.

### DIFF
--- a/src/tools/tool.ts
+++ b/src/tools/tool.ts
@@ -413,25 +413,31 @@ export abstract class ToolBase {
         args: ToolArgs<typeof this.argsShape>,
         { signal }: ToolExecutionContext
     ): Promise<CallToolResult> {
-        if (!(await this.verifyConfirmed(args))) {
-            this.session.logger.debug({
-                id: LogId.toolExecute,
-                context: "tool",
-                message: `User did not confirm the execution of the \`${this.name}\` tool so the operation was not performed.`,
-                noRedaction: true,
-            });
-            return {
-                content: [
-                    {
-                        type: "text",
-                        text: `User did not confirm the execution of the \`${this.name}\` tool so the operation was not performed.`,
-                    },
-                ],
-            };
-        }
-        const startTime = Date.now();
+        let startTime: number = Date.now();
 
         try {
+            if (this.requiresConfirmation()) {
+                if (!(await this.verifyConfirmed(args))) {
+                    this.session.logger.debug({
+                        id: LogId.toolExecute,
+                        context: "tool",
+                        message: `User did not confirm the execution of the \`${this.name}\` tool so the operation was not performed.`,
+                        noRedaction: true,
+                    });
+                    return {
+                        content: [
+                            {
+                                type: "text",
+                                text: `User did not confirm the execution of the \`${this.name}\` tool so the operation was not performed.`,
+                            },
+                        ],
+                    };
+                }
+                // We do not want to include the elicitation time in the tool execution time
+                // so we reset the startTime to the current time. We may want to consider adding
+                // a separate field for elicitation time in the future.
+                startTime = Date.now();
+            }
             this.session.logger.debug({
                 id: LogId.toolExecute,
                 context: "tool",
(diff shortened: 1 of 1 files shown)


scripts/generate/generateToolDocumentation.ts
```diff
@@ -59,7 +59,7 @@ function extractToolInformation(): ToolInfo[] {
 
         try {
             const instance = new ToolClass(dummyParams);
-            // Access the protected description property
+            // Access the public description property
             const description =
                 (instance as unknown as { description?: string }).description || "No description available";
```
nirinchev: Is this casting necessary now?

src/tools/atlas/create/createAccessList.ts
```diff
@@ -18,13 +18,13 @@ export const CreateAccessListArgs = {
...
-    protected description = "Allow Ip/CIDR ranges to access your MongoDB Atlas clusters.";
+    public description = "Allow Ip/CIDR ranges to access your MongoDB Atlas clusters.";
     static operationType: OperationType = "create";
-    protected argsShape = {
+    public argsShape = {
         ...CreateAccessListArgs,
     };
 
-    protected async execute({
+    public async execute({
```
nirinchev: I'm not sure about this change. The callback that the base tool constructs does a bunch of things around `execute`, so I don't think we should be having two fairly different behaviors available to consumers. Instead, if we want to expose a way to directly invoke these tools, I would suggest keeping `execute` as protected (you can rename it to `executeImpl` or something like this) and then exposing a `invoke`/`execute` on the base tool that is basically extracting the `callback` we construct in tool.ts:501.

nirinchev (commented):
Looks good, but I have a suggestion about reverting a part of the publicizing.

src/tools/atlas/create/createAccessList.ts
gagik: good point. switched to a separate `invoke` function

gagik: I can imagine a world where we want to manage stuff like confirmation manually but it's hypothetically possible even with our wrapping by disabling it so I'm fine with that

src/tools/tool.ts
```diff
@@ -408,6 +408,60 @@ export abstract class ToolBase {
         { signal }: ToolExecutionContext
     ): Promise<CallToolResult>;
 
+    /** This is used internally by the server to invoke the tool. It can also be run manually to call the tool directly. */
+    public async invoke(
+        args: ToolArgs<typeof this.argsShape>,
+        { signal }: ToolExecutionContext
+    ): Promise<CallToolResult> {
+        const startTime = Date.now();
```
nirinchev: Should we construct the start time prior to the elicitation request? I see value in both, but considering this eventually gets reported as duration, we'd be measuring both the tool execution and users' response times, blurring the signal from that measurement.

src/tools/atlas/read/listDBUsers.ts
```diff
@@ -10,13 +10,13 @@ export const ListDBUsersArgs = {
 
 export class ListDBUsersTool extends AtlasToolBase {
     public name = "atlas-list-db-users";
-    protected description = "List MongoDB Atlas database users";
-    static operationType: OperationType = "read";
-    protected argsShape = {
+    public description = "List MongoDB Atlas database users";
+    public static operationType: OperationType = "read";
```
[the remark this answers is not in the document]
himanshusinghs: @gagik just making sure, are you handling static fields as static in Compass assistant?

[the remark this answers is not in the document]
himanshusinghs: but yea also the public keyword is likely not needed here?

src/tools/tool.ts
himanshusinghs: I am not sure if there is much value in knowing how long user took to approve/cancel but if we see one, lets keep both?

gagik: I think we'd want to track 2 times in this case. tool execution time should skip the elicitation IMO

gagik: would make sense do this in a follow up instead

PR #833, opened 2026-01-07
fix: incorrect aggregate pipelines construction MCP-351

nirinchev:
## Proposed changes

Fixes a few quirks with the `aggregate` tool:
* If we have $merge or $out stages, don't attempt to apply limits - this will not work as we can't have stages after `$merge` or `$out`
* Correctly check for `$search` and `$searchMeta` stages
* Tweak the message generation to avoid repeating the same thing when the total docs equals to the number of returned docs.

### DIFF
--- a/tests/integration/tools/mongodb/read/aggregate.test.ts
+++ b/tests/integration/tools/mongodb/read/aggregate.test.ts
@@ -197,11 +197,38 @@ describeWithMongoDB("aggregate tool", (integration) => {
             },
         });
         const content = getResponseContent(response);
-        expect(content).toEqual("The aggregation pipeline executed successfully");
+        expect(content).toEqual("The aggregation pipeline executed successfully.");
 
         const copiedDocs = await mongoClient.db(integration.randomDbName()).collection("outpeople").find().toArray();
         expect(copiedDocs).toHaveLength(3);
         expect(copiedDocs.map((doc) => doc.name as string)).toEqual(["Peter", "Laura", "Søren"]);
+    });
+
+    it("can run $merge stages in non-readonly mode", async () => {
+        const mongoClient = integration.mongoClient();
+        await mongoClient
+            .db(integration.randomDbName())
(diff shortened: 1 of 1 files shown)


tests/integration/tools/mongodb/read/aggregate.test.ts
```diff
@@ -153,6 +153,57 @@ describeWithMongoDB("aggregate tool", (integration) => {
...
+                database: integration.randomDbName(),
+                collection: "people",
+                pipeline: [{ $limit: 1 }],
+            },
+        });
+        const content = getResponseContent(response);
+        expect(content).toContain("The aggregation resulted in 1 documents");
+    });
+
+    it("can run $out stages in non-readonly mode", async () => {
```
blva, line 180: nit: should we also have a $merge test? or am i missing something?

nirinchev: Yeah, fair - I'll add one

src/tools/mongodb/read/aggregate.ts
```diff
@@ -108,6 +108,18 @@ export class AggregateTool extends MongoDBToolBase {
...
+                // This is a write pipeline, so special-case it and don't attempt to apply limits or caps
+                aggregationCursor = provider.aggregate(database, collection, pipeline);
+                const results = await aggregationCursor.toArray();
+                return {
+                    content: formatUntrustedData(
+                        "The aggregation pipeline executed successfully",
+                        ...(results.length > 0 ? [EJSON.stringify(results)] : [])
+                    ),
+                };
+            }
```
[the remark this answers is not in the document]
blva: could we decouple the two different ways of doing this into different methods where you get the result and then follow same logic to return the result? this also will help us understand this business logic in the long-term

himanshusinghs (approved):
Nice catch!

gagik (approved):

PR #834, opened 2026-01-08
chore: bring accuracy tests to the normal levels

himanshusinghs:
## Proposed changes
This PR fixes flakes and misses on the accuracy tests and brings them mostly to the normal levels. There are still a few of the tests to be fixed but I am leaving them out of scope intentionally as they don't concern / block my upcoming work.

As a drive-by I have fixed the incorrect usage of discriminated union.

## Checklist

### DIFF
--- a/src/tools/mongodb/create/createIndex.ts
+++ b/src/tools/mongodb/create/createIndex.ts
@@ -138,6 +138,12 @@ export class CreateIndexTool extends MongoDBToolBase {
     public argsShape = {
         ...DbOperationArgs,
         name: z.string().optional().describe("The name of the index"),
+        // Note: Although it is not required to wrap the discriminated union in
+        // an array here because we only expect exactly one definition to be
+        // provided here, we unfortunately cannot use the discriminatedUnion as
+        // is because Cursor is unable to construct payload for tool calls where
+        // the input schema contains a discriminated union without such
+        // wrapping. This is a workaround for enabling the tool calls on Cursor.
         definition: z
             .array(
                 z.discriminatedUnion("type", [
--- a/src/tools/mongodb/metadata/explain.ts
+++ b/src/tools/mongodb/metadata/explain.ts
@@ -15,6 +15,12 @@ export class ExplainTool extends MongoDBToolBase {
 
     public argsShape = {
         ...DbOperationArgs,
+        // Note: Although it is not required to wrap the discriminated union in
+        // an array here because we only expect exactly one method to be
+        // provided here, we unfortunately cannot use the discriminatedUnion as
+        // is because Cursor is unable to construct payload for tool calls where
+        // the input schema contains a discriminated union without such
+        // wrapping. This is a workaround for enabling the tool calls on Cursor.
         method: z
             .array(
                 z.discriminatedUnion("name", [
--- a/src/tools/mongodb/read/export.ts
+++ b/src/tools/mongodb/read/export.ts
@@ -14,6 +14,12 @@ export class ExportTool extends MongoDBToolBase {
     public argsShape = {
         ...DbOperationArgs,
         exportTitle: z.string().describe("A short description to uniquely identify the export."),
+        // Note: Although it is not required to wrap the discriminated union in
+        // an array here because we only expect exactly one exportTarget to be
+        // provided here, we unfortunately cannot use the discriminatedUnion as
+        // is because Cursor is unable to construct payload for tool calls where
+        // the input schema contains a discriminated union without such
+        // wrapping. This is a workaround for enabling the tool calls on Cursor.
         exportTarget: z
             .array(
                 z.discriminatedUnion("name", [


src/tools/mongodb/create/createIndex.ts
```diff
@@ -137,19 +137,17 @@ export class CreateIndexTool extends MongoDBToolBase {
...
+                z
+                    .object({
+                        type: z.literal("classic"),
+                        keys: z.object({}).catchall(z.custom<IndexDirection>()).describe("The index definition"),
+                    })
+                    .describe("Definition for a MongoDB index (e.g. ascending/descending/geospatial)."),
+                ...(this.isFeatureEnabled("search")
+                    ? [this.vectorSearchIndexDefinition, this.atlasSearchIndexDefinition]
+                    : []),
+            ])
```
himanshusinghs: Drive-by fix in createIndex, explain and export tool. We don't need to use an array to use discriminated unions. Its more natural and understandable for LLMs to pass an object with required properties.

tests/integration/tools/mongodb/read/export.test.ts
```diff
@@ -70,7 +70,7 @@ describeWithMongoDB(
                 },
                 {
                     name: "exportTarget",
-                    type: "array",
+                    anyOf: [{ type: "object" }, { type: "object" }],
```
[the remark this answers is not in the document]
himanshusinghs: We do expect two different object type because of union

tests/integration/tools/mongodb/metadata/explain.test.ts
```diff
@@ -19,7 +19,7 @@ describeWithMongoDB("explain tool", (integration) => {
             {
                 name: "method",
                 description: "The method and its arguments to run",
-                type: "array",
+                anyOf: [{ type: "object" }, { type: "object" }, { type: "object" }],
```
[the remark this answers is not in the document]
himanshusinghs: We do expect two different object type because of union

tests/integration/tools/mongodb/create/createIndex.test.ts
```diff
@@ -134,7 +128,7 @@ describeWithMongoDB(
             ...databaseCollectionParameters,
             {
                 name: "definition",
-                type: "array",
+                anyOf: [{ type: "object" }, { type: "object" }, { type: "object" }],
```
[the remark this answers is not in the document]
himanshusinghs: We do expect two different object type because of union

tests/accuracy/createDeployment.test.ts
```diff
@@ -1,51 +1,93 @@
...
+    });
+}
 
 describeAccuracyTests([
     {
         prompt: "Setup a local MongoDB cluster named 'local-cluster'",
+        mockedTools: {
+            // Mocking the create-deployment tool call here so that consecutive
+            // test runs remains unaffected
+            "atlas-local-create-deployment": mockCreateDeploymentResponse("local-cluster"),
```
blva, line 22: [q] do we mock calls in other accuracy tests?

tests/accuracy/createIndex.test.ts
```diff
@@ -29,14 +29,12 @@ describeAccuracyTests(
                         database: "mflix",
                         collection: "movies",
                         name: Matcher.anyOf(Matcher.undefined, Matcher.string()),
-                        definition: [
```
blva, line 32: [q] for my own understanding, looks like you're removing the array usage for the definitions in many of the tests, could you explain the reasoning?

tests/accuracy/createDeployment.test.ts
himanshusinghs: Ideally we don't need to mock calls in accuracy tests as the idea is to test end to end but in this case the test was leaving resources uncleared and there wasn't a clear way to remove them post test without extending the test suite so I chose instead to mock the create call.

In this PR, I have mocked all the places where create-deployment was leaving resources without clearing them up.

tests/accuracy/createIndex.test.ts
himanshusinghs: This is related to the refactor of usage of [discriminated unions](https://github.com/mongodb-js/mongodb-mcp-server/pull/834#discussion_r2672879404). To have a property that is supposed to be an object, guarded by a discriminated union, we don't _need_ to wrap it in an array.

In the refactor, I have moved away from the pattern of guarding an object using `z.array(z.discriminatedUnion)` where we were literally picking the first element of the array. The test case changes are just follow ups for that refactor.

tests/accuracy/createDeployment.test.ts
blva: got it, couldn't we call delete after the test runs?

himanshusinghs: The way we declare these accuracy tests is declarative which makes it easy and determinable. An extension is required to provide some programmatic control which I think we should avoid for as long as we have alternatives.

blva: got it, but will this not affect the deleteDeployment test since it expects a certain output? 

you could also consider merging the tests as deploymentsLifecycle then so you'd call deletion

src/tools/mongodb/create/createIndex.ts
nirinchev: Can you double check this works with Cursor, Claude, and VSCode? I remember the reason we have this weirdness with the array was because one of those tools had a lot of trouble with discriminated union json schemas. Could be that this is fixed now, but if it isn't, we may need to revert this.

tests/accuracy/createIndex.test.ts
blva: thank you for the explanation! makes sense!

src/tools/mongodb/create/createIndex.ts
himanshusinghs: That's a good context, thank you for that. I will check it.

tests/accuracy/createDeployment.test.ts
himanshusinghs: Yea this should not affect the deleteDeployment test since that one creates and deletes the deployment as well.

Merging the test as lifecycle is prone to order shuffling issue which might get hard to debug.

himanshusinghs:
I realised that the discriminated union refactor needed fixing in some accuracy tests as well. Will do that, verify the editor experience and ask for a re-review.

tests/accuracy/createDeployment.test.ts
blva: got it, ok, since the test accuracy seem healthy I'll go ahead and resolve this thread, thanks!

blva (approved):
LGTM given that we do the testing suggested by Nikola, thanks for the updates!

src/tools/mongodb/create/createIndex.ts
himanshusinghs: Unfortunately Cursor has trouble picking up the updated discriminatedUnion schema, VSCode and Claude works fine. I have reverted the changes and left a note about this so that I don't start biting this again in future 😄 Thank you again for flagging this. 🙌

src/tools/mongodb/read/aggregate.ts
```diff
@@ -58,6 +54,10 @@ export class AggregateTool extends MongoDBToolBase {
     public argsShape = {
         ...DbOperationArgs,
         ...getAggregateArgs(this.isFeatureEnabled("search")),
+        responseBytesLimit: z.number().optional().default(ONE_MB).describe(`\
+The maximum number of bytes to return in the response. This value is capped by the server's configured maxBytesPerQuery and cannot be exceeded. \
+Note to LLM: If the entire aggregation result is required, use the "export" tool instead of increasing this limit.\
```
nirinchev, line 59: Flagging this, though we don't need to fix in this PR - the export tool may not always be available (e.g. if the user disabled it), in which case this hint would be confusing.

himanshusinghs: This change is removing `responseBytesLimit` from the common args that are re-used in export and explain tool and inserting it explicitly in aggregate and find tool where it makes sense. I think it should be fine, unless I am missing something.

himanshusinghs: Ahh I see it now. The hint itself is referencing export tool. I will open a separate PR to fix this as this is an existing problem.

PR #836, opened 2026-01-08
feat: Add configurable api / auth client support

ciprian-tibulca:
## Proposed changes
this PR add support for configurable api client and (new) auth provider, needed by the remote mcp wrapper to inject the access token (that is received via the proxy) instead of using client creds
it is backward compatible and default behavior is unchanged when credentials are provided via UserConfig

- extract auth logic from apiClient into `AuthProvider` and `ClientCredentialsAuthProvider`
- unit tests for `ClientCredentialsAuthProvider` 
- `ApiClient` constructor now accepts an optional `authProvider` parameter
- If no `authProvider` is provided, it falls back to `AuthProviderFactory.create()` using credentials from options
- This enables dependency injection of custom authentication providers for testing and advanced use cases
- Updated `Session` class to use the new `createAtlasApiClient` factory function

### DIFF
--- a/src/common/atlas/apiClient.ts
+++ b/src/common/atlas/apiClient.ts
@@ -114,10 +114,7 @@ export class ApiClient {
     }
 
     public async validateAuthConfig(): Promise<void> {
-        const isValidAuthConfig = await this.authProvider?.validate();
-        if (!isValidAuthConfig) {
-            throw new Error("Invalid auth configuration");
-        }
+        await this.authProvider?.validate();
     }
 
     public async close(): Promise<void> {


tests/unit/common/auth/clientCredentials.test.ts
```diff
@@ -0,0 +1,317 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
```
wtrocki: :FYI we using Jest: https://github.com/10gen/mms/pull/152213

src/common/atlas/auth/authProvider.ts
```diff
@@ -0,0 +1,45 @@
...
+    access_token: string;
+    expires_at?: number;
+}
+
+export interface AuthProvider {
+    hasCredentials(): boolean;
+    getAccessToken(): Promise<string | undefined>;
+    validateAccessToken(): Promise<void>;
+    revokeAccessToken(): Promise<void>;
+    middleware(): Middleware;
```
wtrocki: Curious if this needs to be part of the interface?

src/common/atlas/auth/authProvider.ts
```diff
@@ -0,0 +1,45 @@
+import type { Middleware } from "openapi-fetch";
+import type { LoggerBase } from "../../logger.js";
+import { ClientCredentialsAuthProvider } from "./clientCredentials.js";
+
+export interface AccessToken {
+    access_token: string;
+    expires_at?: number;
+}
+
+export interface AuthProvider {
```
wtrocki, line 14: [nit] This interface looks more like OAuth client credentials provider.
Suggesting new non generic name

```suggestion
export interface OAuthClientCredentialsProvider {
```

src/common/atlas/auth/clientCredentials.ts
```diff
@@ -0,0 +1,172 @@
...
+    public async getAuthHeaders(): Promise<Record<string, string> | undefined> {
+        const accessToken = await this.getAccessToken();
+        return accessToken
+            ? {
+                  Authorization: `Bearer ${accessToken}`,
+              }
+            : undefined;
+    }
+
+    public hasCredentials(): boolean {
```
wtrocki: [question][nit] Do we need this method? I see that we set those 2 fields in constructor.

src/common/atlas/auth/authProvider.ts
ciprian-tibulca: the end goal is to let systems that embed the MCP Server as a package plug in their own auth provider implementation. This is not limited to client credentials, and might other flows as needed.

src/common/atlas/auth/clientCredentials.ts
ciprian-tibulca: hasCredentials is generic for any auth provider (not only for the current ClientCredentialsAuthProvider implementation). Please check https://github.com/mongodb-js/mongodb-mcp-server/pull/836#discussion_r2681948051

src/common/atlas/auth/authProvider.ts
ciprian-tibulca: yes, this is where the auth provider injects whatever auth data the outbound API request needs that is specific to the implemented auth flow. For client credentials that is `Authorization: Bearer <token>`, while other providers may need to add different headers or additional request metadata.

this is also related to https://github.com/mongodb-js/mongodb-mcp-server/pull/836#discussion_r2681948051

src/common/atlas/auth/clientCredentials.ts
```diff
@@ -0,0 +1,172 @@
...
+        }
+
+        if (!this.isAccessTokenValid()) {
+            this.accessToken = await this.getNewAccessToken();
+        }
+
+        return this.accessToken?.access_token;
+    }
+
+    public async validateAccessToken(): Promise<void> {
```
gagik: this is from older code but can we just get rid of this method? it's a confusing wrapper for `getAccessToken`

src/common/atlas/auth/clientCredentials.ts
```diff
@@ -0,0 +1,172 @@
...
+    userAgent: string;
+}
+
+export class ClientCredentialsAuthProvider implements AuthProvider {
+    private oauth2Client?: oauth.Client;
+    private oauth2Issuer?: oauth.AuthorizationServer;
+    private accessToken?: AccessToken;
+    private readonly options: ClientCredentialsAuthOptions;
+    private readonly logger: LoggerBase;
+    private static customFetch: typeof fetch = createFetch({
```
gagik: let's move this to the constructor, otherwise it breaks web compatibility, see https://github.com/mongodb-js/mongodb-mcp-server/commit/399c2e767c9c84cfb2eccd551f0b54ce9b98112a

src/common/atlas/auth/authProvider.ts
wtrocki: Ack. Althought methods cover oauth tokens and credentials based login (I assume hasCredentails method checks for that). Having ts-doc documentation on class/methods will be helpful to clarify purpose of the class.

src/common/atlas/auth/clientCredentials.ts
wtrocki: @gagik WDYT about future idea to have webcompatibility tests via github actions on PRs. 
It would be easy to miss other non trivial cases in future.

src/common/atlas/auth/clientCredentials.ts
ciprian-tibulca: leftover from the refactor, not needed anymore. Removed in the last commit, thanks.

src/common/atlas/auth/clientCredentials.ts
ciprian-tibulca: thanks, brought this code in sync with your changes, thanks

src/common/atlas/auth/authProvider.ts
```diff
@@ -0,0 +1,44 @@
...
+    hasCredentials(): boolean;
+    getAccessToken(): Promise<string | undefined>;
+    revokeAccessToken(): Promise<void>;
+    middleware(): Middleware;
+    getAuthHeaders(): Promise<Record<string, string> | undefined>;
```
gagik, line 18: maybe the function we should control/revoke is `getAuthHeaders` and `revokeAuthHeaders`?  in that case auth token ends up being an implementation detail of our client auth provider.

I feel like there's a small redundancy here otherwise as far as this generic provider is concerned. but not a blocker, just a thought

ciprian-tibulca: My initial intention was to extract the existing auth code into a separate AuthProvider (& client credentials impl) with minimal changes to the ApiClient and Session.

But from yours and @wtrocki comments, I agree more clarity is needed, so I did a bit of cleanup and made the interface more generic. It now exposes just three methods, validate(), getAuthHeaders(), and revoke(), so tokens stay as an implementation detail while still supporting other auth schemes and making future extensions easier.

PR #837, opened 2026-01-09
chore: minimize node-specific API usage

gagik:
These changes remove the need for some polyfilling on Compass end and make the server more flexible in terms of working in a browser environment.


gagik:
I'm going to add a e2e test to this as it's easier to reason about and mocking everything ruins the point of the test

PR #855, opened 2026-01-13
fix: fixes wrong path getting updated with generated embeddings MCP-357

himanshusinghs:
## Proposed changes
When auto-generation of embeddings are requested for nested field paths then the logic to update the document was setting a top level key with the provided field path and embeddings instead of setting the correct, nested field paths with the embeddings.

This PR fixes that by ensuring that we take care of nested field path by traversing the individual parts in documents and creating the intermediate paths as necessary.

## Checklist

### DIFF
--- a/tests/unit/helpers/manageNestedFieldPaths.test.ts
+++ b/tests/unit/helpers/manageNestedFieldPaths.test.ts
@@ -122,7 +122,7 @@ describe("manageNestedFieldPaths", () => {
             const doc: Record<string, unknown> = {};
             const deepPath = Array.from({ length: 101 }, (_, i) => `level${i}`).join(".");
             expect(() => setFieldPath(doc, deepPath, "value")).toThrow(
-                "Field path has too many nested levels (maximum 100 allowed)."
+                `Field path "${deepPath}" has too many nested levels (maximum 100 allowed).`
             );
         });
     });


src/helpers/manageNestedFieldPaths.ts
```diff
@@ -0,0 +1,37 @@
...
+            const traversedPath = parts.slice(0, i + 1).join(".");
+            throw new Error(
+                `Cannot set field at path '${fieldPath}': intermediate path '${traversedPath}' is not an object.`
+            );
+        }
+        current = current[part] as Record<string, unknown>;
+    }
+    const lastPart = parts[parts.length - 1] as string;
+    current[lastPart] = value;
+}
```
addaleax, line 42: Completely optional, but both of these functions would be a bit more cleaner/readable if written recursively instead of iteratively

tests/unit/helpers/manageNestedFieldPaths.test.ts
```diff
@@ -0,0 +1,126 @@
...
+
+    describe("setFieldPath", () => {
+        it("sets a top-level field", () => {
+            const doc: Record<string, unknown> = {};
+            setFieldPath(doc, "name", "test");
+            expect(doc).toEqual({ name: "test" });
+        });
+
+        it("sets a nested field on existing object", () => {
+            const doc: Record<string, unknown> = { info: { title: "Old Title" } };
```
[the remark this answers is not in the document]
himanshusinghs: updated the title to match the expectation

src/tools/mongodb/create/insertMany.ts
```diff
@@ -134,30 +135,13 @@ export class InsertManyTool extends MongoDBToolBase {
                 throw new MongoDBError(ErrorCodes.Unexpected, `Document at index ${documentIndex} does not exist.`);
             }
             // Ensure no nested fields are present in the field path.
-            this.deleteFieldPath(processedDocuments[documentIndex], fieldPath);
-            processedDocuments[documentIndex][fieldPath] = generatedEmbeddings[index];
+            deleteFieldPath(processedDocuments[documentIndex], fieldPath);
+            setFieldPath(processedDocuments[documentIndex], fieldPath, generatedEmbeddings[index]);
```
ciprian-tibulca: is the delete redundant as we're setting back the field in the next line?
unless I'm missing somethingm, the comment `Ensure no nested fields are present in the field path.` refers to the leaf field, which will be set to `generatedEmbeddings[index]` regardless if it has nested fields.

src/helpers/manageNestedFieldPaths.ts
```diff
@@ -0,0 +1,45 @@
...
+        throw new Error(`Cannot set field at provided path: intermediate path '${currentPath}' is not an object.`);
+    }
+
+    _setFieldPath(current[key] as Record<string, unknown>, rest, value, currentPath);
+}
+
+// The provided field path might include some internal Object properties such as
+// `__proto__`. For such paths, we need to ensure that we don't override the
+// derived properties on the object so we explicitly set these properties as
+// Object's own property.
```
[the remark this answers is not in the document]
himanshusinghs: The comment explains the same with different wordings. The suggested wordings are backwards.

src/tools/mongodb/create/insertMany.ts
himanshusinghs: Thanks for the catch 🙌 It was indeed redundant, I have now removed the deleteFieldPath call.

src/helpers/manageNestedFieldPaths.ts
himanshusinghs: Thanks for pointing it out. Simplified the implementation using a recursive approach.

src/helpers/manageNestedFieldPaths.ts
```diff
@@ -0,0 +1,37 @@
...
+    let current: Record<string, unknown> = document;
+    for (let i = 0; i < parts.length; i++) {
+        const part = parts[i];
+        const key = part as keyof typeof current;
+        if (!current[key]) {
+            return;
+        } else if (i === parts.length - 1) {
+            delete current[key];
+        } else {
+            current = current[key] as Record<string, unknown>;
```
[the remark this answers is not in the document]
himanshusinghs: method is removed now.

src/helpers/manageNestedFieldPaths.ts
```diff
@@ -0,0 +1,45 @@
...
+}
+
+function _setFieldPath(current: Record<string, unknown>, parts: string[], value: unknown, parentPath: string): void {
+    if (parts.length === 0) {
+        return;
+    }
+
+    const [key, ...rest] = parts;
+    if (!key) {
+        return;
```
ciprian-tibulca: this happens when fieldPath is not correct (e.g. "a.b.c.", "a...b.c")? If yes, I suggest validating the `fieldPath` param in `setFieldPath` rather than silently fail here

src/helpers/manageNestedFieldPaths.ts
```diff
@@ -0,0 +1,45 @@
+export function setFieldPath(document: Record<string, unknown>, fieldPath: string, value: unknown): void {
+    const parts = fieldPath.split(".");
+    _setFieldPath(document, parts, value, "");
```
ciprian-tibulca, line 14: [nit]
should we validate the depth of the path (e.g. MAX_DEPTH=64) to avoid blowing the stack (if fieldPath is extremely long)?

addaleax: What would be the goal of that? Would we throw an error in that case, and if so, why is that better than using the normal stack overflow exception?

If you do want this, I think there's a BSON nesting limit defined for the server that's 128 or something like that, you'd want do make sure we're over that limit at least.

ciprian-tibulca: main goal would be to fail fast with a clear error instead of a stack overflow, which is harder to debug and can crash the process if fieldPath is user controlled. 
agree on the limit, we can set it equal or above the BSON nesting limit which is 100 
https://www.mongodb.com/docs/manual/reference/limits/?atlas-provider=aws&atlas-class=general#mongodb-limit-Nested-Depth-for-BSON-Documents

addaleax: > which is harder to debug

It's maybe a little less clear to users, but it's not really easier or harder to debug for us, a stack overflow with a stack trace pointing at `_setFieldPath()` would be pretty unambiguous

> and can crash the process if fieldPath is user controlled.

No, it can't, JS engines won't let you do that.

src/helpers/manageNestedFieldPaths.ts
himanshusinghs: I missed handling this case, thanks for pointing it out. Covered this in my recent commit.

src/helpers/manageNestedFieldPaths.ts
```diff
@@ -0,0 +1,53 @@
+const MAX_DEPTH = 64;
```
addaleax:
```suggestion
const MAX_DEPTH = 1024;
```

Or at least something larger than 100, as discussed below – don't see any reason to keep the limit artificially low

src/helpers/manageNestedFieldPaths.ts
himanshusinghs: Based on the discussion here and the limits in the referenced doc, I have set the limit to 100.

src/helpers/manageNestedFieldPaths.ts
himanshusinghs: If the max allowed nesting for BSON docs is 100 then a limit of 100 should be fine, right?

src/helpers/manageNestedFieldPaths.ts
ciprian-tibulca: thanks for the change @himanshusinghs 

This validation adds clarity. Since this function runs in a loop, logging fieldPath will make it much easier to debug if we ever hit the limit:
`for (const [index, { fieldPath, documentIndex }] of flattenedEmbeddingsInput.entries())...`

Also, even if the process does not crash, in a multi-tenant setup a very deep fieldPath can still impact event loop & performance.

ciprian-tibulca (approved):

src/helpers/manageNestedFieldPaths.ts
```diff
@@ -0,0 +1,55 @@
...
+const MAX_DEPTH = 100;
+
+export function setFieldPath(document: Record<string, unknown>, fieldPath: string, value: unknown): void {
+    const parts = fieldPath.split(".");
+    if (parts.some((part) => !part.trim())) {
+        throw new Error(`Invalid field path: '${fieldPath}'`);
+    }
+
+    if (parts.length > MAX_DEPTH) {
+        throw new Error(`Field path has too many nested levels (maximum ${MAX_DEPTH} allowed).`);
```
ciprian-tibulca:
```suggestion
        throw new Error(`Field path "${fieldPath}" has too many nested levels (maximum ${MAX_DEPTH} allowed).`);
```

himanshusinghs: added this log

PR #856, opened 2026-01-13
chore: use unref conditionally to avoid electorn issues

gagik:
This is Node-specific can cause issues when running in electron otherwise. See https://github.com/electron/electron/issues/21162

### DIFF
--- a/src/telemetry/telemetry.ts
+++ b/src/telemetry/telemetry.ts
@@ -107,7 +107,10 @@ export class Telemetry {
                     });
                     resolve();
                 }, flushMaxWaitTime);
-                flushTimeout.unref();
+                // This is Node-specific and can cause issues when running in electron otherwise. See https://github.com/electron/electron/issues/21162
+                if (typeof flushTimeout.unref === "function") {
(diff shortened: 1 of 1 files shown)


himanshusinghs (approved):

src/telemetry/telemetry.ts
```diff
@@ -107,7 +107,10 @@ export class Telemetry {
                     });
                     resolve();
                 }, flushMaxWaitTime);
-                flushTimeout.unref();
+                // This is Node-specific and can cause issues when running in electron otherwise. See https://github.com/electron/electron/issues/21162
+                if (typeof flushTimeout.unref === "function") {
+                    flushTimeout.unref();
+                }
```
blva, line 113: what's the impact when you use it in web? is telemetry disabled?

gagik: in practice: nothing, unref just ensure setTimeout doesn't keep the Node process alive
https://docs.deno.com/api/node/process/~/Process.unref (this is Deno but one of the better docs on it)

PR #865, opened 2026-01-14
refactor(mcp-ui): update UI module generation to create .ts file for compilation

tsck:
## Proposed changes
**Issue:** The UI build was generating .js files for each UI and storing them in `ui/lib/tools/.js` as expected. These are what get found by the `UIRegistry`. `tsc` however, only compiles .ts files, so these generated modules never made it into `dist/`.

**Fix:** updated `vite.ui.config.ts` to generate .ts files as well so that tsc picks them up on build. 

Ticket: https://jira.mongodb.org/browse/MCP-363

## Test
- [x] Confirmed local build working
- [x] Confirmed tool UI properly being copied to `dist` on build

Screenshot 2026-01-14 at 4 01 51 PM

### DIFF
--- a/vitest.config.ts
+++ b/vitest.config.ts
@@ -1,5 +1,4 @@
 import { coverageConfigDefaults, defineConfig } from "vitest/config";
-import { resolve } from "path";
 
 // Shared exclusions for all projects
 // Ref: https://vitest.dev/config/#exclude
@@ -22,13 +21,6 @@ if (process.env.SKIP_ATLAS_LOCAL_TESTS === "true") {
 }
 
 export default defineConfig({
-    resolve: {
-        alias: {
-            // Map .js imports to .ts files in src/ui/lib for testing
-            // This allows dynamic imports with .js extension to work in tests where only .ts files exist
-            "./tools/list-databases.js": resolve(__dirname, "src/ui/lib/tools/list-databases.ts"),
-        },
-    },
     test: {
         environment: "node",
         testTimeout: 3600000,


vite.ui.config.ts
```diff
@@ -101,8 +101,8 @@ function generateUIModule(): Plugin {
                 }
                 const html = readFileSync(htmlFile, "utf-8");
 
-                // Generate as .js file so it works with dynamic imports in both
-                // production (from dist/) and development/test (from src/)
+                // Generate .ts file for tsc to compile to dist/
+                // Generate .js file for development/test (dynamic imports use .js extension)
```
gagik: can we not just access dist?

tsck: The problem is that in `registry` we need someway to access them. When built, `registry` also lives in `dist/` so there's no additional `dist/` to access. We solve this by using a relative path, accessing the .js files regardless of where `registry` is.

vite.ui.config.ts
```diff
@@ -111,13 +111,20 @@ function generateUIModule(): Plugin {
  */
 export const ${componentName}Html = ${JSON.stringify(html)};
 `;
+                writeFileSync(join(toolsDir, `${toolName}.ts`), toolModuleContent);
                 writeFileSync(join(toolsDir, `${toolName}.js`), toolModuleContent);
```
nirinchev: q: do we need to write the .js files now?

gagik:
I think we can reduce complexity further. We should consider just creating TypeScript files that are committed into our codebase. Instead of having this run on every build, we run it as a script similar to `generate:arguments` or `generate:api`. 

Then we can also generate a file like `index.ts` that imports all the tools without dynamic imports:
```ts
///src/ui/registry/tools/index.ts
import { ListDatbasesTool } from "./list-databases";
```
and the registry file would simply
```ts
import * as UITools from "./tools/index
```
Rough idea, let me know if it isn't clear

vite.ui.config.ts
tsck: Currently I believe we do given the dynamic import at runtime. Since Node doesn't have a built-in TypeScript parser/compiler, we wouldn't be able to dynamically import a .ts file.

tsck:
> I think we can reduce complexity further. We should consider just creating TypeScript files that are committed into our codebase. Instead of having this run on every build, we run it as a script similar to `generate:arguments` or `generate:api`.
> 
> Then we can also generate a file like `index.ts` that imports all the tools without dynamic imports:
> 
> ```ts
> ///src/ui/registry/tools/index.ts
> import { ListDatbasesTool } from "./list-databases";
> ```
> 
> and the registry file would simply
> 
> ```ts
> import * as UITools from "./tools/index
> ```
> 
> Rough idea, let me know if it isn't clear

This is an interesting idea. I will toy with this for a bit and see if I can get something working in this direction.

vite.ui.config.ts
tsck: Actually backtracking on this. Since even locally we're working off of `/dist` the only place `.js` is currently used is within tests. But perhaps there's a way around this. Will see if there's a way around this there.

tsck:
@gagik Ok I’ve had time to think this through. Really the core change being proposed is **committing the generated files rather than keeping them gitignored**. While the `index.ts` entry point is a nice cleanup, and removing `.js` files is something I'm looking into (currently they're just there for tests), those are orthogonal - we could add those to the current build process without committing files.

I actually discussed this strategy with @nirinchev in the [initial PR](https://github.com/mongodb-js/mongodb-mcp-server/pull/783#discussion_r2602066704) and decided against it. Here is why I believe we should keep these files ignored:

1. **Nondeterministic Builds:** Our current minification process is nondeterministic. If we commit these files, they will be different every time the file generation occurs, leading to constant, noisy diffs. Fixing this requires a slower build.  
2. **Repository Bloat:** These generated files are large (\~1MB even for this small UI). Committing them increases package size and clone times significantly as we scale.  
3. **Sync Issues:** Relying on a manual script (like generate:arguments) introduces the risk of the committed asset falling out of sync with the source React code if a step is missed.  
4. **Committed Files Are Not Human Readable:** For the most part these files are not human readable due to the fact that they’re minified. This makes the benefits a bit negligible to me.

### Note on the Proposed Import Strategy:
> ```ts
> ///src/ui/registry/tools/index.ts
> import { ListDatbasesTool } from "./list-databases";
> ```
> 
> and the registry file would simply
> 
> ```ts
> import * as UITools from "./tools/index
> ```

I believe this just switches us from lazy loading to static loading. Given the size of these UI files, I don't believe this would be ideal. It would cause us to load every UI into memory on startup, even if the UI is never needed.

### Quick Context on Bug:

It’s worth noting that the issue this PR fixes wasn't missed due to the build architecture, but just by my local environment having a stale `lib/` directory (an old `list-databases.ts` file was lingering). Otherwise, this would not have worked locally and would have been caught. To prevent this recurring, I’m adding a build step to clean the directory before regeneration. Will push this up shortly.

---

## **Proposed Middle Ground**

We can achieve the cleaner architecture you proposed (a unified entry point) *without* the downsides of committing large assets. We can generate the index file at build time, but keep it gitignored:

```ts
// src/ui/registry/tools/index.ts (generated at build time, gitignored)
export const UIToolLoaders: Record<string, () => Promise<string>> = {
    "list-databases": async () => (await import("./list-databases.js")).ListDatabasesHtml,
};
```

And the registry becomes cleaner still, similar to what you propose:

```ts
import { UIToolLoaders } from "./tools/index.js";

export class UIRegistry {
    private cache = new Map<string, string>();

    async get(toolName: string): Promise<string | null> {
        if (this.cache.has(toolName)) return this.cache.get(toolName)!;

        const loader = UIToolLoaders[toolName];
        if (!loader) return null;

        const html = await loader();
        this.cache.set(toolName, html);
        return html;
    }
}
```

Curious to hear your thoughts. Would this be a fair compromise here?

vite.ui.config.ts
tsck: Was able to figure this out with [import.meta.glob](https://vite.dev/guide/features#glob-import). TIL, I didn't know about this in in vitest 🙂. Change pushing shortly

gagik:
> **1. Nondeterministic Builds:** Our current minification process is nondeterministic. If we commit these files, they will be different every time the file generation occurs, leading to constant, noisy diffs. Fixing this requires a slower build.

Why is it nondeterministic? Even if so, this is why I want this to be a manually running script which gets run for example when bumping packages instead of being hidden from us during builds.

> **2. Repository Bloat:** These generated files are large (~1MB even for this small UI). Committing them increases package size and clone times significantly as we scale.

If these are 1MB when generated, isn't it more concerning that this is increasing **bundle size**? I personally don't really care about repository size. Adding the entirety of the MCP server to compass-web added +3mb to its bundle. If these files are so large (and presumably cannot be made smaller) this is a general concern for me.

> **3. Sync Issues:** Relying on a manual script (like generate:arguments) introduces the risk of the committed asset falling out of sync with the source React code if a step is missed.


The generate script is run when bumping packages so that'd be part of the process to ensure we don't miss it. We could even add a test to enforce this if strictly necessary. 
My larger issue is that there's an unnecessary generation being done every time we install or build when 90% of the time the work here isn't going to be modifying the UI components.

> **4. Committed Files Are Not Human Readable:** For the most part these files are not human readable due to the fact that they’re minified. This makes the benefits a bit negligible to me.

Yeah I agree with this bit is not the biggest benefit but not uncommon to 


Regarding the middle ground, 
- dynamic imports sound good to me (though I will need to verify if this can work in browsers #868). 
- I would also add that the UI registry itself should be conditionally imported based on the UI functionality being turned on (and it should be off my default in library context IMO).
- I still think we should consider commiting generated could be worthwhile. I am quite concerned by 1MB per file situation in general but conditional import of UI registry can alleviate this

tsck:
I feel like there are just a few suggestions and concerns here that are getting a bit conflated. I'm going to address each separately, if I'm missing something please let me know!

## ✅  Committing Generated FIles

> Why is it nondeterministic? Even if so, this is why I want this to be a manually running script which gets run for example when bumping packages instead of being hidden from us during builds.

My understanding was that the `esbuild` minification step generated nondeterministic variable names.

This said, I can see that you strongly about committing the generated files. I know @nirinchev felt this way as well. I do feel differently about this for the points mentioned, and though I'm not sold on the value, I don't think the downsides are majorly concerning either. 

I will concede to the majority here 🙂 - I'll get this working again and push a commit for it shortly

## ✅ Dynamic Imports

> dynamic imports sound good to me (though I will need to verify if this can work in browsers https://github.com/mongodb-js/mongodb-mcp-server/pull/868).

Dynamic imports are supported in the browser ([MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/import)). If you do run into issues though please let me know! 

## ❌ Conditional Registry

> I would also add that the UI registry itself should be conditionally imported based on the UI functionality being turned on (and it should be off my default in library context IMO).

I don't think I agree with this. The UI registry itself is quite small, just a few lines of code. If the benefit of this is that we can then statically import the UI modules in the registry since we're now dynamically importing the registry, this design would result in every UI module being loaded into memory / bundled any time the feature's turned on or needed. In comparison to the current design where a UI module is only loaded when that specific module is needed.

If the idea is that we'd dynamically load both the registry and the UI modules, I believe dynamically loading the registry code would be a very negligible gain.

Is there another benefit to this that I'm missing?

## ❕Bundle Size Concerns

>  If these are 1MB when generated, isn't it more concerning that this is increasing bundle size?

This is why I'm pushing back on static imports and encouraging us to be careful about our lazy loading strategy. Using static imports will bloat the bundle size that's sent over the wire to a client since every UI module will be included in the initial bundle. By loading them dynamically instead, they won't be included in that bundle, and instead be sent in separate bundles on an as needed basis. This is how we manage large dependencies in design system components as well.

I am wondering if there's a compression solution that's cross platform enough. If we compress the UI module, we could allow the registry to decompress the response. I think this would help in the browser. The registry would get sent in the first bundle, but then when it lazy loads the UI module, the UI module would be much smaller when it's sent over the wire.

tsck:
@gagik **TL;DR** - I've committed the generated files and moved the task out of the build as requested. I can't find a good way to deterministically generate the UI modules though unfortunately, so currently they all get regenerated every time the script is ran. Definitely open to any ideas you might have for solving this.

## Nondeterministic Details 
I'm struggling to get the `esbuild` to produce deterministic output. The minification generates different variable names between runs.

This happens because `esbuild` assigns short variable names based on the order modules are processed, and that order varies between runs. I did also oddly observe that the module count seems to vary between runs (1267-1271 modules).

I'm suspecting there might be a race condition in Vite/Rollup's module resolution, possibly in how import.meta.glob discovers components or how transitive dependencies are resolved in parallel. I'm really not sure though. The varying module count (1267-1271) suggests some modules are being discovered inconsistently, which then affects the order they're processed and thus the minified variable names.

I tried a few different paths with no success:

- Switched to `terser` (instead of `esbuild`): Still non-deterministic. This made it seem like the non-determinism happens before minification, in Rollup's module resolution/bundling phase, maybe? But I'm not sure.
- Clear Vite's cache before each build: Made it worse. 
- Sort inputs: This seemed to help, but didn't work.
- Run build twice (use second result): This was inconsistent. Even the second build varied, suggesting the instability isn't just cache-related
- Disable `optimizeDeps`: No improvement

Of course we can disable minification, but that would be an impractical solution given the already large size. So right now every module is just regenerated every time the script runs.

gagik:
> Dynamic imports are supported in the browser ([MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/import)). If you do run into issues though please let me know!

This might be true but dynamic imports have been problematic with us when bundling with webpack already.  compass-web currently needs to patch webpack config to get it to work otherwise (https://github.com/mongodb-js/compass/pull/7715, see next point). I don't mind leaving these dynamic imports _if_ we make the uiRegistry import conditional.

> If the idea is that we'd dynamically load both the registry and the UI modules, I believe dynamically loading the registry code would be a very negligible gain.

The big gain here is that it is much easier to use webpack to avoid bundling the registry rather than to try to target all possible dynamic imports. We have had issues with this, so it'd be best to avoid it, see: https://github.com/mongodb-js/compass/pull/7715 

We conditionally load the apiClient we use for Atlas tools for the same reason: https://github.com/mongodb-js/mongodb-mcp-server/pull/860

Hypothetically we can find a fix in compass-web but I don't see a great argument to bundle all the UI files if the UI features are disabled.

> Nondeterministic Details

I think non-deterministic output is fine because this would be run manually or on bumps. It's not ideal; thanks for the efforts in trying to find an alternative. To me this gives us better observability into what is going into our binary before release so it's not a negative per say.

This is manual approach and has its own risks but we could manually keep a hash of the `src/ui` folder and only re-run the script if this changes.

But otherwise, every time the script is run SGTM!

vitest.config.ts
```diff
@@ -21,15 +22,20 @@ if (process.env.SKIP_ATLAS_LOCAL_TESTS === "true") {
...
+    },
     test: {
         environment: "node",
         testTimeout: 3600000,
         hookTimeout: 3600000,
         setupFiles: ["./tests/setup.ts"],
         coverage: {
             exclude: [
-                // Required: import.meta.glob() in src/ui creates Vite virtual modules (\0 prefixed paths)
-                // that crash Istanbul reporters. See: https://github.com/vitest-dev/vitest/issues/5101
```
gagik, line 32: is this still worth noting?

gagik:
> By loading them dynamically instead, they won't be included in that bundle

In case of webpacked content, it would be. It won't be loaded into memory but it will be available in the bundle. At least that's my understand of how this works. When people i.e. download Compass, they download the whole thing always, there's no remote place to load the rest conditionally, so the bundle will always be there unless we manually adjust it with webpack. 

Conditional imports (like we do for API Client) would make it easier for us to exclude the bundling when we don't need the UI Registry.

vite.ui.config.ts
```diff
@@ -111,13 +105,54 @@ function generateUIModule(): Plugin {
...
+                    return `    "${toolName}": async () => {
+        const mod = await import("./tools/${toolName}.js");
+        return mod.${componentName}Html;
+    }`;
+                })
+                .join(",\n");
+
+            const loadersContent = `/**
+ * AUTO-GENERATED FILE - DO NOT EDIT MANUALLY
+ * Generated by: vite build --config vite.ui.config.ts
```
gagik: maybe mentioing the pnpm command `generate:ui` would be more flexible

gagik:
I think as a general test it'd be good to have https://github.com/mongodb-js/mongodb-mcp-server/pull/868 merged beforehand just to make sure any assumptions we have are correct. Browsers bundling is unpredictable.

tsck:
Above comments regarding conditionally loading registry were addressed in https://github.com/mongodb-js/mongodb-mcp-server/pull/871

vitest.config.ts
```diff
@@ -21,6 +22,13 @@ if (process.env.SKIP_ATLAS_LOCAL_TESTS === "true") {
 }
 
 export default defineConfig({
+    resolve: {
+        alias: {
+            // Map .js imports to .ts files in src/ui/lib for testing
+            // This allows dynamic imports with .js extension to work in tests where only .ts files exist
+            "./tools/list-databases.js": resolve(__dirname, "src/ui/lib/tools/list-databases.ts"),
```
gagik: this is odd, I would not expect this to be necessary though not sure what the solution is.

tsck: This ultimately was a result of [this thread](https://github.com/mongodb-js/mongodb-mcp-server/pull/865#discussion_r2693443972). The original approach generated both .js and .ts files so that something like this wasn't needed, but there was pushback on that. I'm open to other solutions here but I was stuck on getting this to work in ESM.

tsck: One solution could be to generate only .js files and update the tsconfig to copy over .js files as well. Would that be preferred? Might be weird in that the generated files will be .js and we're now committing them, but might be cleaner?

tsck: Spoke offline - this appears it may be unnecessary now. There was a failing test at some point that triggered this change, but we've changed a lot since then so we may have removed the need for this.

Looks good without it locally. Pushing up the change and resolving!

gagik (approved):

PR #868, opened 2026-01-15
chore: add browser tests and a web entry point

gagik:
Stacked on #871 

Browsers tests are useful in ensuring our future updates do not break its ability to be embedded in `compass-web`.

At this stage, the web entrypoint's main purpose is just to ensure that simply importing a subset of MCP functionality doesn't lead to a crash with a browser.

`tests/browser` is a pnpm workspace (but it will not be published as a separate package). This just allows us to have a better separation of dependencies.

We still have a large number of polyfills to get this running. All polyfills largely mirror that of Compass. Ideally we can look into being in sync with Compass here.

The polyfills are a feature rather than a bug. If we end up using certain node-specific features, we can make a decision on whether this should be polyfilled in web or if we should consider an environment-independent implementation instead.

### DIFF
--- a/tests/browser/polyfills/README.md
+++ b/tests/browser/polyfills/README.md
@@ -4,22 +4,24 @@ This directory contains polyfills for the browser environment. Most of the polyf
 
 We may also want to consider distributing a version of the library with those polyfills applied.
 
-
 #### Adding polyfills
 
 Ideally we want to avoid additional polyfills. There's a few approaches we can take:
 
 In `vite.config.ts`, if we want to simply error if one attempts to import the polyfilled package we can use the `throwErrorPolyfill` function.
+
 ```
     kerberos: throwErrorPolyfill(),
 ```
- 
+
 If we want the polyfill to be a no-op, we can use the `emptyPolyfill` function.
+
 ```
     express: emptyPolyfill(),
 ```
 
 For anything more custom, one should create a new file in this directory which matches the package name (for example `@mongodb-js/atlas-local` would be in `polyfills/@mongodb-js/atlas-local/index.ts`) and then use the `localPolyfill` function.
+
 ```
     "@mongodb-js/atlas-local": localPolyfill("@mongodb-js/atlas-local"),
 ```


src/ui/registry/registry.ts
```diff
@@ -33,17 +33,7 @@ export class UIRegistry {
...
-            if (html === undefined) {
-                return null;
-            }
-            this.cache.set(toolName, html);
-            return html;
-        } catch {
-            return null;
-        }
+        // TODO: Remove this once we have a proper build step for UI tools.
+        return null;
```
gagik: Just a patch until https://github.com/mongodb-js/mongodb-mcp-server/pull/865 is in.
All changes in `src/**` will be reverted

tests/browser/polyfills/README.md
```diff
@@ -0,0 +1,5 @@
+### Polyfills
+
+This directory contains polyfills for the browser environment. Most of the polyfills 1:1 match those in [compass-web](https://github.com/mongodb-js/compass/tree/main/packages/compass-web/polyfills) so they provide a way to ensure our integration can remain stable. Ideally, we would want to both reduce these polyfilsl and/or be more explicitly in sync with the polyfills used in compass-web and other target platforms.
```
cveticm: typo
```suggestion
This directory contains polyfills for the browser environment. Most of the polyfills 1:1 match those in [compass-web](https://github.com/mongodb-js/compass/tree/main/packages/compass-web/polyfills) so they provide a way to ensure our integration can remain stable. Ideally, we would want to both reduce these polyfills and/or be more explicitly in sync with the polyfills used in compass-web and other target platforms.
```

tests/browser/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+## Running Browser Tests
+
+```bash
+# From workspace root - install browser-tests and root
+pnpm install --workspace browser-tests
+
+# From this directory - run browser tests
+cd tests/browser-tests
+pnpm test --workspace browser-tests
```
cveticm: This snippet did not work for me. Below did:
```
pnpm install --workspace browser
cd tests/browser
pnpm test
```

Or 

```
pnpm install --workspace browser
pnpm --filter browser test
```

tests/browser/polyfills/README.md
```diff
@@ -0,0 +1,5 @@
+### Polyfills
```
cveticm, line 1: suggestion: Would it be useful to add documentation on how to add new polyfills if this is not an exhausted list?

gagik: great idea

tests/browser/README.md
gagik: thank you! these docs were quite all over the place

PR #869, opened 2026-01-15
fix: return error result with structured content

himanshusinghs:
## Proposed changes
Apparently when a tool is registered with `OutputSchema`, the error result should also include the defined structured content otherwise the OutputSchema validation fails for some client, particularly for an MCPClient created through TS modelcontextprotocol SDK.

## Checklist


himanshusinghs:
This was a misunderstanding! The OutputSchema validation only applies if the CallToolResult is not an error result.

PR #870, opened 2026-01-15
feat: allow session id to be passed externally MCP-369

nirinchev:
## Proposed changes

Allows for clients to send requests with externally-generated session ids. This means that session initialization is skipped which additionally requires that we switch to pure json responses instead of SSE. Sessions are now going to be created on demand with the externally-supplied id and will not be cleaned up until `idleTimeoutMS` fires.

### DIFF
--- a/src/common/config/userConfig.ts
+++ b/src/common/config/userConfig.ts
@@ -233,7 +233,7 @@ const ServerConfigSchema = z4.object({
         .boolean()
         .default(false)
         .describe(
-            "When true, the HTTP transport handling requests with a session ID supplied externally through the 'mcp-session-id' header. When an external ID is supplied, the initialization request is optional."
+            "When true, the HTTP transport allows requests with a session ID supplied externally through the 'mcp-session-id' header. When an external ID is supplied, the initialization request is optional."
         )
         .register(configRegistry, { overrideBehavior: "not-allowed" }),
     httpResponseType: z4
--- a/src/transports/streamableHttp.ts
+++ b/src/transports/streamableHttp.ts
@@ -142,6 +142,10 @@ export class StreamableHttpRunner extends TransportRunnerBase {
                 | { sessionId?: string; isImplicitInitialization?: false }
                 | { sessionId: string; isImplicitInitialization: true }
         ): Promise<void> => {
+            if (isImplicitInitialization && !sessionId) {
+                throw new Error("Implicit initialization requires externally-passed sessionId");
+            }
+
             const request: RequestContext = {
                 headers: req.headers as Record<string, string | string[] | undefined>,
                 query: req.query as Record<string, string | string[] | undefined>,
--- a/tests/integration/transports/streamableHttp.test.ts
+++ b/tests/integration/transports/streamableHttp.test.ts
@@ -33,7 +33,7 @@ describe("StreamableHttpRunner", () => {
         const requestHeaders: Record<string, string> = {
             ...additionalHeaders,
         };
-        if (shouldInitialize && sessionId) {
+        if (sessionId) {
             requestHeaders["mcp-session-id"] = sessionId;
         }
 
@@ -41,6 +41,9 @@ describe("StreamableHttpRunner", () => {
             requestInit: {
                 headers: requestHeaders,
             },
+            // If `sessionId` is set, the client will skip the initialize request.
+            // If we want to ensure the initialization request is sent, we set `sessionId` to undefined,
+            // even if we have an external session ID to use.
             sessionId: shouldInitialize ? undefined : sessionId,
         });
 
--- a/tests/unit/common/config/configOverrides.test.ts
+++ b/tests/unit/common/config/configOverrides.test.ts
@@ -234,6 +234,8 @@ describe("configOverrides", () => {
                     "voyageApiKey",
                     "allowRequestOverrides",
                     "dryRun",
+                    "externallyManagedSessions",
+                    "httpResponseType",
                 ]);
             });
 


src/transports/streamableHttp.ts
```diff
@@ -49,118 +49,141 @@ export class StreamableHttpRunner extends TransportRunnerBase {
...
-                res.status(404).json({
-                    jsonrpc: "2.0",
-                    error: {
-                        code: JSON_RPC_ERROR_CODE_SESSION_NOT_FOUND,
-                        message: "session not found",
-                    },
-                });
-                return;
+                if (this.userConfig.externallyManagedSessions) {
+                    console.log(`Session with ID ${sessionId} not found, initializing new session`);
```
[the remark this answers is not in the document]
nirinchev: /ar this is temporary and will be removed once testing is complete

src/common/config/userConfig.ts
```diff
@@ -229,6 +229,13 @@ const ServerConfigSchema = z4.object({
             "When true, runs the server in dry mode: dumps configuration and enabled tools, then exits without starting the server."
         )
         .register(configRegistry, { overrideBehavior: "not-allowed" }),
+    externallyManagedSessions: z4
+        .boolean()
+        .default(false)
+        .describe(
+            "When true, the HTTP transport will allow json responses without SSE. Use this in cases where sessions are managed externally, such as by Amazon Bedrock AgentCore."
```
ciprian-tibulca: this reads like two separate concerns bundled into one config flag:
* allow JSON responses without SSE
* sessions are managed externally
 
Would it make sense to split into 2 configs (transport mode vs session management mode)?

src/transports/streamableHttp.ts
```diff
@@ -49,118 +49,113 @@ export class StreamableHttpRunner extends TransportRunnerBase {
...
-                            message: `invalid request`,
-                        },
+            let transport: StreamableHTTPServerTransport;
+            if (this.userConfig.externallyManagedSessions && sessionId) {
+                transport = new StreamableHTTPServerTransport({
+                    enableJsonResponse: true,
+                });
+                server.session.logger.setAttribute("sessionId", sessionId);
+
+                this.sessionStore.setSession(sessionId, transport, server.session.logger);
```
ciprian-tibulca: just to confirm, we don't need to customize `onsessionclosed` for the scenario where sessions are externally managed?

src/transports/streamableHttp.ts
```diff
@@ -49,118 +49,113 @@ export class StreamableHttpRunner extends TransportRunnerBase {
...
-                if (!isInitializeRequest(req.body)) {
-                    res.status(400).json({
-                        jsonrpc: "2.0",
-                        error: {
-                            code: JSON_RPC_ERROR_CODE_INVALID_REQUEST,
-                            message: `invalid request`,
-                        },
+            let transport: StreamableHTTPServerTransport;
+            if (this.userConfig.externallyManagedSessions && sessionId) {
+                transport = new StreamableHTTPServerTransport({
```
ciprian-tibulca: [nit]
is it possible to unify the 2 transport init scenarios? something like this:

```
            transport = new StreamableHTTPServerTransport({
                enableJsonResponse: this.userConfig.externallyManagedSessions && sessionId ? true : false, // or separate configflag
                sessionIdGenerator: (): string => this.userConfig.externallyManagedSessions && sessionId
                    ? sessionId
                    : getRandomUUID(),
                onsessioninitialized: (sessionId): void => {
                    server.session.logger.setAttribute("sessionId", sessionId);

                    this.sessionStore.setSession(sessionId, transport, server.session.logger);
                },
                onsessionclosed: async (sessionId): Promise<void> => {
                    try {
                        await this.sessionStore.closeSession(sessionId, false);
                    } catch (error) {
                        this.logger.error({
                            id: LogId.streamableHttpTransportSessionCloseFailure,
                            context: "streamableHttpTransport",
                            message: `Error closing session ${sessionId}: ${error instanceof Error ? error.message : String(error)}`,
                        });
                    }
                },
            });
```

src/transports/streamableHttp.ts
nirinchev: When we enable json responses, we don't have true "sessions" as we don't have SSE. So the session close event will never fire: https://github.com/modelcontextprotocol/typescript-sdk/blob/aa81a66556fb4434d8a6d1b70f7ac9fc40b5d325/src/server/webStandardStreamableHttp.ts#L801

This is the code that handles the delete and `onSessionClosed` is invoked with undefined because the sessionId is never set.

ciprian-tibulca: I don't think sessions are tied to SSE, they are a correlating MCP messages across multiple HTTP requests (and we use them for holding the data plane connections). 
> This is the code that handles the delete and onSessionClosed is invoked with undefined because the sessionId is never set.

When sessions are externally managed, is the MCP client invoking the DELETE with undefined sessionId or not invoking it at all? And follow-up, what happens if the client sends DELETE with the externally managed session id?

I'm just trying to understand the new flow covered here, not pushing to implement onSessionClosed unless we find it is useful.

nirinchev: Right, sorry, that was imprecise formulation. With AgentCore, the client doesn't send initialization request or a DELETE for terminating the MCP server session as there terminating a session is equivalent to tearing down the process.

That being said, I agree that right now we're confounding json responses and session management unnecessarily. While this works for AgentCore, it's not a clean solution and I'll rework this to be more client-agnostic.

src/common/config/userConfig.ts
nirinchev: Yes, that's fair - I was thinking about it too narrowly - while we don't have a clear use case for independently setting those, there should be no reason to tie them together.

ciprian-tibulca (approved):

src/common/config/userConfig.ts
```diff
@@ -229,6 +229,20 @@ const ServerConfigSchema = z4.object({
             "When true, runs the server in dry mode: dumps configuration and enabled tools, then exits without starting the server."
         )
         .register(configRegistry, { overrideBehavior: "not-allowed" }),
+    externallyManagedSessions: z4
+        .boolean()
+        .default(false)
+        .describe(
+            "When true, the HTTP transport allows requests with a session ID supplied externally through the 'mcp-session-id' header. When an external ID is supplied, the initialization request is optional."
```
wtrocki, line 236: [nit] Do we have way to instruct end users how this external ID's can be supplied (header)

nirinchev: It already mentions that this is defined in the header, though I'm not sure if we should be more descriptive beyond that. We could be more descriptive in the docs page, though I doubt we'll be able to cover all the configuration methods.

PR #871, opened 2026-01-16
chore: import UI registry conditionally

gagik:
As it stands, UI registry needs to be imported conditionally because of the fact that it uses dynamic imports, which is problematic with compass-web webpack, see https://github.com/mongodb-js/compass/pull/7715

Some of it will be solved by #865 but it's still good to have the option to exclude custom UIs from the bundle. This will likely be necessary for `compass-web`.

This also refactors it so instead of configuring with `customUIs`, we directly pass a UIRegistry, this creates a better separation of concerns for our configuration options.

I will open further followups to our transport runner on top of this.


package.json
```diff
@@ -57,13 +57,12 @@
     "start": "node dist/index.js --transport http --loggers stderr mcp --previewFeatures search",
     "start:stdio": "node dist/index.js --transport stdio --loggers stderr mcp",
     "prepare": "husky && pnpm run build",
-    "build:clean": "rm -rf dist",
     "build:update-package-version": "tsx scripts/updatePackageVersion.ts",
     "build:esm": "tsc --project tsconfig.esm.json && chmod +x dist/esm/index.js",
     "build:cjs": "tsc --project tsconfig.cjs.json",
     "build:ui": "vite build --config vite.ui.config.ts",
     "build:universal-package": "tsx scripts/createUniversalPackage.ts",
-    "build": "pnpm run build:clean && pnpm run build:ui && concurrently \"pnpm run build:esm\" \"pnpm run build:cjs\" && pnpm run build:universal-package",
```
gagik, line 66: drive by, this is unnecessary

src/tools/tool.ts
```diff
@@ -786,26 +786,31 @@ export abstract class ToolBase {
...
-            },
-        });
+        const resultContent = result.content || [];
+        const content = uiResource ? [...resultContent, uiResource] : resultContent;
 
         return {
             ...result,
-            content: [...(result.content || []), uiResource],
+            content,
         };
```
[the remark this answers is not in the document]
gagik: eh, it's too confusing with all the `|| []`

tests/integration/transports/base.test.ts
```diff
@@ -0,0 +1,106 @@
...
+
+    afterEach(async () => {
+        if (server) {
+            await server.close();
+            server = undefined;
+        }
+    });
+
+    describe("UIRegistry conditional import", () => {
+        it("should not set UIRegistry when mcpUI preview feature is not enabled", async () => {
```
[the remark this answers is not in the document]
gagik: eh, not a fan of spys on an import

src/server.ts
```diff
@@ -111,7 +92,7 @@ export class Server {
         connectionErrorHandler,
         elicitation,
         tools,
-        customUIs,
+        uiRegistry,
```
gagik, line 95: since this is a preview feature in library, I'd assume it's fine to make a "breaking" change like this right?

himanshusinghs: Yea it should be fine. Unfortunately, we don't have a place in our changelog to call such changes out.

PR #875, opened 2026-01-19
CLOUDP-373012: Support Operational Metrics via EventEmitter interface

wtrocki:
General idea. Not ready for detailed review

## Problem Statement

The current implementation in server telemetry metrics by parsing log message strings using regex. This approach is brittle and could break silently if the upstream `mongodb-mcp-server` library changes its log message format.

##  Solution

This PR introduces a **dedicated monitoring system** that is completely separate from telemetry, providing type-safe event hooks for internal observability and metrics collection.

The MongoDB MCP Server will have two separate event systems:

1. **Telemetry** - User analytics sent to MongoDB backend (controlled by DO_NOT_TRACK)
      - Automatically configured and enabled to send data.
2. **Monitoring** - Internal observability for metrics collection 
      - Requires extension to send metrics to prometheus or other systems.

## Key Differences

| Aspect | Telemetry | Monitoring |
|--------|-----------|------------|
| **Purpose** | User analytics & product insights | Internal metrics & observability |
| **Destination** | MongoDB backend | Local metrics systems (e.g., Prometheus) |
| **Enabled** | Controlled by `DO_NOT_TRACK` env var | Always active |
| **Events** | Tool execution, Server lifecycle | Tool execution, Server lifecycle, **Connection events** |
| **Types** | `ToolEvent`, `ServerEvent` | `MonitoringToolEvent`, `MonitoringServerEvent`, `MonitoringConnectionEvent` |

## Important Notes

- **Connection events are ONLY emitted to monitoring**, never to telemetry
- Monitoring events are always emitted regardless of telemetry settings
- The two systems use separate event types to allow independent evolution

## Dedicated Monitoring System 

We created a **completely separate monitoring system** with its own types and event emitter:

#### Key Changes:

1. **Separate Monitoring Types** (`src/monitoring/types.ts`)
   - `MonitoringToolEvent` - Tool execution events
   - `MonitoringServerEvent` - Server lifecycle events
   - `MonitoringConnectionEvent` - Connection events (monitoring only!)
   - `MonitoringEventNames` - Constants for event names (no hardcoded strings)

2. **Dependency Injection Pattern** (`src/server.ts` & `src/transports/base.ts`)
   - Monitoring EventEmitter is created in `TransportRunnerBase` and injected into `Server`
   - Clean separation of concerns - no tight coupling
   - Server receives monitoring as a constructor parameter
   ```typescript
   // In TransportRunnerBase
   public readonly monitoring: EventEmitter<MonitoringEvents> = new EventEmitter();
   
   // Passed to Server during construction
   const server = new Server({
       // ... other params
       monitoring: this.monitoring,
   });
   ```

3. **Always Active** (Open to change?)
   - Monitoring events are always emitted regardless of `DO_NOT_TRACK` setting
   - Telemetry events only emitted when telemetry is enabled

4. **Privacy Compliant**
   - Connection events are ONLY emitted to monitoring, never to telemetry backend
   - Sensitive connection details stay local

5. **Type-Safe Event Names**
   - Use `MonitoringEventNames` constants instead of hardcoded strings
   - Prevents typos and provides autocomplete

## Usage Example

```typescript
import { 
    StdioRunner, 
    MonitoringEventNames,
    type MonitoringToolEvent,
    type MonitoringConnectionEvent
} from 'mongodb-mcp-server';
import { Histogram } from 'prom-client';

const runner = new StdioRunner(config);
await runner.start();

// Access monitoring from the runner (not from server)
runner.monitoring.on(MonitoringEventNames.TOOL_EXECUTED, (event: MonitoringToolEvent) => {
    mcpOperationDuration.observe({
        tool_name: event.tool_name,
        status: event.result
    }, event.duration_ms / 1000);
});

// Listen to connection events (NOT sent to telemetry!)
runner.monitoring.on(MonitoringEventNames.CONNECTION_LIFECYCLE, (event: MonitoringConnectionEvent) => {
    mcpConnectionDuration.observe({
        status: event.result
    }, event.duration_ms / 1000);
});
```

### Rejected Approach : Telemetry Event Hooks (first commit)

Initially, we enhanced the existing `TelemetryEvents` interface to pass actual event data as parameters:

```typescript
export interface TelemetryEvents {
    "events-emitted": [events: BaseEvent[]];
    "events-send-failed": [events: BaseEvent[], error: Error];
    "events-skipped": [events: BaseEvent[], reason: string];
}
```

However, this approach had limitations:
- Mixed user analytics (telemetry) with internal metrics (monitoring)
- Connection events would be sent to telemetry backend (privacy concern)
- Monitoring only worked when telemetry was enabled

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -106,7 +106,7 @@ export class Server {
     /**
      * Monitoring event emitter for internal observability and metrics collection.
      * Always active regardless of telemetry settings.
-     * Use this for Prometheus metrics, logging, and other internal monitoring.
+     * Use this for metrics, logging, and other internal monitoring.
      * Injected from the transport runner.
      */
     public readonly monitoring: EventEmitter<MonitoringEvents>;
@@ -119,6 +119,14 @@ export class Server {
 
     private readonly startTime: number;
     private readonly subscriptions = new Set<string>();
+
+    // Store event listener references for cleanup
+    private connectionMonitoringListeners?: {
+        onRequest: () => void;
+        onSuccess: (state: ConnectionStateConnected) => void;
+        onError: (state: ConnectionStateErrored) => void;
+        onClose: () => void;
+    };
 
     constructor({
         session,
@@ -240,6 +248,9 @@ export class Server {
     }
 
     async close(): Promise<void> {
+        // Clean up connection monitoring event listeners
+        this.cleanupConnectionMonitoring();
+
         await this.telemetry.close();
         await this.session.close();
         await this.mcpServer.close();
@@ -316,38 +327,64 @@ export class Server {
         // Track connection start time in closure scope for better encapsulation
         let connectionStartTime: number | undefined;
 
-        // Track connection request (start timing)
-        // Note: Multiple connection-request events can be emitted during a single connection attempt
-        // (e.g., for OIDC flows), so we only set the start time if it's not already set
-        this.session.connectionManager.events.on("connection-request", () => {
+        // Define event listeners and store references for cleanup
+        const onRequest = (): void => {
             if (connectionStartTime === undefined) {
                 connectionStartTime = Date.now();
             }
-        });
+        };
 
-        // Track successful connections
-        this.session.connectionManager.events.on("connection-success", (state: ConnectionStateConnected) => {
+        const onSuccess = (state: ConnectionStateConnected): void => {
             if (connectionStartTime !== undefined) {
                 const duration = Date.now() - connectionStartTime;
                 this.emitConnectionMonitoringEvent(MonitoringConnectionCommand.CONNECT, duration, "success", state);
                 connectionStartTime = undefined;
             }
-        });
+        };
 
-        // Track connection errors
-        this.session.connectionManager.events.on("connection-error", (state: ConnectionStateErrored) => {
+        const onError = (state: ConnectionStateErrored): void => {
             if (connectionStartTime !== undefined) {
                 const duration = Date.now() - connectionStartTime;
                 this.emitConnectionMonitoringEvent(MonitoringConnectionCommand.CONNECT, duration, "failure", state);
                 connectionStartTime = undefined;
             }
-        });
+        };
 
-        // Track disconnections
-        this.session.connectionManager.events.on("connection-close", () => {
+        const onClose = (): void => {
             // Duration is 0 for disconnect events as it's not meaningful
             this.emitConnectionMonitoringEvent(MonitoringConnectionCommand.DISCONNECT, 0, "success");
-        });
+        };
+
+        // Store listener references for cleanup
+        (this.connectionMonitoringListeners as typeof this.connectionMonitoringListeners) = {
+            onRequest,
+            onSuccess,
+            onError,
+            onClose,
+        };
+
+        // Register event listeners
+        // Note: Multiple connection-request events can be emitted during a single connection attempt
+        // (e.g., for OIDC flows), so we only set the start time if it's not already set
+        this.session.connectionManager.events.on("connection-request", onRequest);
+        this.session.connectionManager.events.on("connection-success", onSuccess);
+        this.session.connectionManager.events.on("connection-error", onError);
+        this.session.connectionManager.events.on("connection-close", onClose);
+    }
+
+    /**
+     * Clean up connection monitoring event listeners.
+     * Should be called when the server is being closed.
+     */
+    private cleanupConnectionMonitoring(): void {
+        if (!this.connectionMonitoringListeners) {
+            return;
+        }
+
+        this.session.connectionManager.events.off("connection-request", this.connectionMonitoringListeners.onRequest);
+        this.session.connectionManager.events.off("connection-success", this.connectionMonitoringListeners.onSuccess);
+        this.session.connectionManager.events.off("connection-error", this.connectionMonitoringListeners.onError);
+        this.session.connectionManager.events.off("connection-close", this.connectionMonitoringListeners.onClose);
     }
 
     private emitConnectionMonitoringEvent(
--- a/src/tools/tool.ts
+++ b/src/tools/tool.ts
@@ -547,7 +547,7 @@ export abstract class ToolBase {
      * Reference to the server instance. Set during tool registration.
      * Used for accessing monitoring events and other server-level functionality.
      */
-    protected server!: Server;
+    protected server?: Server;
 
     constructor({
         category,
@@ -568,10 +568,12 @@ export abstract class ToolBase {
     }
 
     public register(server: Server): boolean {
-        this.server = server;
         if (!this.verifyAllowed()) {
             return false;
         }
+
+        // Set server reference after verification passes
+        this.server = server;
 
         this.registeredTool =
             // Note: We use explicit type casting here to avoid  "excessively deep and possibly infinite" errors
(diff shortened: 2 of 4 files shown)


src/server.ts
```diff
@@ -275,6 +279,73 @@ export class Server {
         this.telemetry.emitEvents([event]);
     }
 
+    private connectionStartTime: number | undefined;
+
+    private setupConnectionTelemetry(): void {
+        // Track connection request (start timing)
+        this.session.connectionManager.events.on("connection-request", () => {
```
wtrocki: Do we need abstract this via telemetry or can we hook directly to connectionManager events

src/tools/tool.ts
```diff
@@ -542,6 +543,12 @@ export abstract class ToolBase {
      */
     protected readonly elicitation: Elicitation;
     private readonly uiRegistry: UIRegistry;
+    /**
+     * Reference to the server instance. Set during tool registration.
+     * Used for accessing monitoring events and other server-level functionality.
+     */
+    protected server!: Server;
```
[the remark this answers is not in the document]
wtrocki: Will do after initial review.

src/monitoring/types.ts
```diff
@@ -0,0 +1,124 @@
...
+/**
+ * Constants for monitoring event names.
+ */
+export const MonitoringEventNames = {
+    TOOL_EXECUTED: "tool-executed",
+    SERVER_LIFECYCLE: "server-lifecycle",
+    CONNECTION_LIFECYCLE: "connection-lifecycle",
+} as const;
+
+export type MonitoringEventName = (typeof MonitoringEventNames)[keyof typeof MonitoringEventNames];
```
gagik, line 19: No need for that const for some of these (not necessarily this, see below),
```ts
export type MonitoringEventName = "tool-executed" | "server-lifecycle" | "connection-lifecycle";
```
this might _feel_ like a downgrade from enum world but it has all the same benefits: autocompletition, enforcement, etc.
I think this is superior to having this separate variable, it goes out of hand quickly.

Also FYI: `export type` and `export const` can have the same name, so you can actually do `export type MonitoringEventNames` here I think.

src/monitoring/types.ts
```diff
@@ -0,0 +1,124 @@
...
+export type MonitoringEvent = MonitoringToolEvent | MonitoringServerEvent | MonitoringConnectionEvent;
+
+/**
+ * Monitoring event emitter interface
+ */
+export interface MonitoringEvents {
+    /**
+     * Emitted when a tool is executed
+     */
+    [MonitoringEventNames.TOOL_EXECUTED]: [event: MonitoringToolEvent];
```
gagik, line 117: okay I suppose for cases like this having this enum-like thing is okay.
I think it might also be possible to do `& Record<MonitoringEventNames, MonitoringToolEvent | MonitoringServerEvent | MonitoringConnectionEvent` to enforce this and then we can just use strings

src/tools/tool.ts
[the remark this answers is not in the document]
gagik: yeah agree with copilot here

src/tools/tool.ts
```diff
@@ -561,6 +568,7 @@ export abstract class ToolBase {
     }
 
     public register(server: Server): boolean {
+        this.server = server;
```
gagik: this should probably be after verifyAllowed?

src/tools/tool.ts
```diff
@@ -542,6 +543,12 @@ export abstract class ToolBase {
      */
     protected readonly elicitation: Elicitation;
     private readonly uiRegistry: UIRegistry;
+    /**
+     * Reference to the server instance. Set during tool registration.
+     * Used for accessing monitoring events and other server-level functionality.
```
gagik, line 548: if all we're doing right now is accessing monitoring object, it might be better to just have the monitoring object here instead of the entire server? to stick to minimum responsibility

src/transports/base.ts
```diff
@@ -211,6 +213,14 @@ export abstract class TransportRunnerBase {
...
     private readonly createSessionConfig?: CreateSessionConfigFn;
     private readonly createApiClient: ApiClientFactoryFn;
+    protected server?: Server;
+
+    /**
+     * Monitoring event emitter for internal observability and metrics collection.
+     * Always active regardless of telemetry settings.
+     * Use this for Prometheus metrics, logging, and other internal monitoring.
+     */
+    public readonly monitoring: EventEmitter<MonitoringEvents> = new EventEmitter();
```
gagik: should this be something that can be specified in the constructor?

src/transports/base.ts
```diff
@@ -211,6 +213,14 @@ export abstract class TransportRunnerBase {
     private readonly tools?: ToolClass[];
     private readonly createSessionConfig?: CreateSessionConfigFn;
     private readonly createApiClient: ApiClientFactoryFn;
+    protected server?: Server;
+
+    /**
+     * Monitoring event emitter for internal observability and metrics collection.
+     * Always active regardless of telemetry settings.
+     * Use this for Prometheus metrics, logging, and other internal monitoring.
```
gagik:
```suggestion
     * Use this for metrics, logging, and other internal monitoring.
```
this is user-library-facing so probably best to keep it general

src/tools/tool.ts
gagik: on the other hand, our runners do all end up exposing the whole server object, so actually this is fine as well...

wtrocki: Good find! I was going back and forth with the same thoughts but eventually ended up using server to be consistent.

src/transports/base.ts
wtrocki: To be clear: Do you mean creating EventEmitter() via constructor?

gagik: I was wondering if we should allow overriding `monitoring` by injecting it through the constructor.

src/server.ts
```diff
@@ -120,8 +138,12 @@ export class Server {
         this.userConfig = userConfig;
         this.elicitation = elicitation;
         this.connectionErrorHandler = connectionErrorHandler;
+        this.monitoring = monitoring;
         this.toolConstructors = tools ?? AllTools;
         this.uiRegistry = new UIRegistry({ customUIs });
+
+        // Track connection timing for monitoring
+        this.setupConnectionMonitoring();
```
gagik, line 154: so generally when the constructor has side effects (i'd consider event listeners as side effects) we have been making it private and move it to a `.create(...)` function. I think it's a nice pattern to stick to

src/server.ts
```diff
@@ -257,22 +279,104 @@ export class Server {
...
+        this.session.connectionManager.events.on("connection-success", (state: ConnectionStateConnected) => {
+            if (connectionStartTime !== undefined) {
+                const duration = Date.now() - connectionStartTime;
+                this.emitConnectionMonitoringEvent(MonitoringConnectionCommand.CONNECT, duration, "success", state);
+                connectionStartTime = undefined;
+            }
+        });
+
+        // Track connection errors
+        this.session.connectionManager.events.on("connection-error", (state: ConnectionStateErrored) => {
```
gagik: do we dispose these event listeners somehow?

wtrocki: Good catch. I need to create test for that.

src/server.ts
wtrocki: > event listeners as side effects)
I really like that approach. Fixing

src/monitoring/types.ts
gagik: we can keep the enum-like const but it'd still be better for the variable and type names to match

wtrocki:
FYI: This PR was added as a POC and is now conflicted. We should discuss whether we want to:

- Promote this eventEmitter-driven interface as alpha (with the risk that customers may start relying on it), or
- Hold off and design it with wider feedback (close PR)

wtrocki:
Closing as POC. We would want to explore more options particulairly approaches that unify metrics across instances.

PR #877, opened 2026-01-21
chore: consider correct channel before docker push MCP-356

himanshusinghs:
## Proposed changes
With this PR, we're fixing a bug in our publish workflow and adapting a more standard advice for pushing docker tags.

Bug: Before the change, we were:
1. During a release, publishing a prerelease version as latest and pushing the same to MCP registry
2. During a scheduled docker update, publishing a possibly pre-release version as latest and also mutating an existing date tag which was expected to be immutable.

The changes in the PR proposes the following:
1. latest: be considered as a floating docker tag meant to point to the latest stable release
2. prerelease: be considered as another floating docker tag meant to point to the latest prerelease
3. versioned tags: be considered as floating tags meant to point to the latest security release for the version
3. dated versioned tags: are immutable tags and are only published if they don't exist already

Our docker publish workflow (docker-publish.yml) which was used in the main release workflow (publish.yml) and scheduled docker updates workflow (docker.yml) now accepts version and release channel, generates docker tags that needs to be pushed while considering if the immutable tag already exists (dated versioned tags) and only then pushes it.

Our release publish workflow (publish.yml) now won't publish to MCP registry if we're publishing a pre-release.

~Our scheduled docker updates workflow (docker.yml) now retrieves the version from the latest and pre-release channel and pushes a security update for the base image from both the channels.~

Our scheduled docker updates workflow (docker.yml) will by default publish security updates for latest channel and on-demand (using workflow dispatch) could publish for prerelease channel as well.

## Checklist

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -64,7 +64,7 @@ jobs:
       contents: read
       issues: write
     needs: [get-version, docker-push]
-    if: ${{ always() && (needs.get-version.result == 'failure' || needs.docker-push.result == 'failure') }}
+    if: ${{ failure() }}
     steps:
       - uses: GitHubSecurityLab/actions-permissions/monitor@v1
         with:


.github/workflows/docker-publish.yml
```diff
@@ -19,13 +28,45 @@ jobs:
...
+          if [[ "$HTTP_STATUS" == "200" ]]; then
+            echo "::notice::Tag ${VERSION_DATE_TAG} already exists, skipping immutable tag"
+          else
+            echo "::notice::Tag ${VERSION_DATE_TAG} does not exist, will be created"
+            DOCKER_TAGS="${DOCKER_TAGS},${REPO}:${VERSION_DATE_TAG}"
+          fi
+
+          echo "DOCKER_TAGS=${DOCKER_TAGS}" >> "$GITHUB_OUTPUT"
+          echo "VERSION=${NPM_VERSION}" >> "$GITHUB_OUTPUT"
+          echo "::notice::Docker tags to publish: ${DOCKER_TAGS}"
```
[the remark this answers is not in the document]
himanshusinghs: /fp The variables are injected in the workflow by other workflows and the values are either hard-coded (in case of RELEASE_CHANNEL) or provided through the result of standard npm command invocation so there is no direct user input involved here.

.github/workflows/docker-publish.yml
```diff
@@ -2,6 +2,15 @@
 name: Build and Push Docker Image
 on:
   workflow_call:
+    inputs:
+      npm_version:
+        description: "The npm package version to publish, without the prefix `v` (e.g., 1.2.3)"
+        required: true
```
gagik, line 8: q: I'm not too sure how these work but this also gets triggered by another action right? How do we pass these things in that case?

.github/workflows/publish.yml
```diff
@@ -22,17 +23,19 @@ jobs:
         id: get-version
         shell: bash
         run: |
```
gagik, line 25: nit: nothing here seems confidential, it might be useful for debugging

.github/workflows/docker-publish.yml
```diff
@@ -19,13 +28,45 @@ jobs:
...
+          fi
+
+          VERSION_TAG="$NPM_VERSION"
+          VERSION_DATE_TAG="${NPM_VERSION}-${DATE}"
+
+          # We always include the floating tags
+          DOCKER_TAGS="${REPO}:${CHANNEL_TAG},${REPO}:${VERSION_TAG}"
+
+          # VERSION_DATE_TAG is included only if it does not exist already
+          HTTP_STATUS=$(curl -s -o /dev/null -w "%{http_code}" \
```
gagik, line 55: nit: is this our best way to check? could we use a docker command or a docker action instead?

.github/workflows/docker-publish.yml
himanshusinghs: Yea this entire workflow is only triggered by other workflows (publish.yml and docker.yml) and they pass the inputs [here](https://github.com/mongodb-js/mongodb-mcp-server/pull/877/changes#diff-3f5366f6d6df3ec1179e5efadc6f350bfa88eebf4e2da589b4d94ccb85ae5e94R50-R52), [here](https://github.com/mongodb-js/mongodb-mcp-server/pull/877/changes#diff-3f5366f6d6df3ec1179e5efadc6f350bfa88eebf4e2da589b4d94ccb85ae5e94R63-R65) and [here](https://github.com/mongodb-js/mongodb-mcp-server/pull/877/changes#diff-551d1fcf87f78cc3bc18a7b332a4dc5d8773a512062df881c5aba28a6f5c48d7R133-R135)

.github/workflows/publish.yml
himanshusinghs: If you're referring to `set -e` instruction that I modified, its because if any of the variable assignment fails, I want the entire step to fail, `set -e` helps enforce that.

.github/workflows/docker-publish.yml
himanshusinghs: I originally thought of using `docker manifest inspect` command but that would've made the step dependent on docker-login step and in turn fragile to step re-placements. The curl method is also a good alternative because our repository is public and the metadata we're fetching is publicly available.

.github/workflows/docker-publish.yml
```diff
@@ -19,13 +28,45 @@ jobs:
...
+          # Strip 'v' prefix if present
+          NPM_VERSION="${NPM_VERSION#v}"
+          RELEASE_CHANNEL="${{ inputs.release_channel }}"
+          REPO="${{ vars.DOCKERHUB_IMAGE_REPOSITORY }}"
           DATE=$(date +'%Y-%m-%d')
-          echo "VERSION=${VERSION}" >> "$GITHUB_OUTPUT"
-          echo "DATE=${DATE}" >> "$GITHUB_OUTPUT"
+
+          # Determine channel tag (latest for stable, channel name for prereleases)
+          if [[ "$RELEASE_CHANNEL" == "latest" ]]; then
```
fmenezes: do you need this if ?

himanshusinghs: good point - I was originally thinking of having release channel fixed to latest and prerelease. I am still not sure where I stand there but sure we can simplify this.

himanshusinghs: fixed

.github/workflows/docker-publish.yml
```diff
@@ -19,13 +28,39 @@ jobs:
...
+          if [[ "$HTTP_STATUS" == "200" ]]; then
+            echo "::notice::Tag ${VERSION_DATE_TAG} already exists, skipping immutable tag"
+          else
+            echo "::notice::Tag ${VERSION_DATE_TAG} does not exist, will be created"
+            DOCKER_TAGS="${DOCKER_TAGS},${REPO}:${VERSION_DATE_TAG}"
+          fi
+
+          echo "DOCKER_TAGS=${DOCKER_TAGS}" >> "$GITHUB_OUTPUT"
+          echo "VERSION=${NPM_VERSION}" >> "$GITHUB_OUTPUT"
+          echo "::notice::Docker tags to publish: ${DOCKER_TAGS}"
```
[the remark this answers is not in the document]
himanshusinghs: I have applied the fix as suggested as it was low effort best practice.

.github/workflows/docker.yml
```diff
@@ -6,11 +6,63 @@ on:
...
+          if [[ -n "$PRERELEASE_VERSION" ]]; then
+            echo "PRERELEASE_VERSION=${PRERELEASE_VERSION}" >> "$GITHUB_OUTPUT"
+            echo "HAS_PRERELEASE=true" >> "$GITHUB_OUTPUT"
+            echo "::notice::Prerelease version: ${PRERELEASE_VERSION}"
+          else
+            echo "HAS_PRERELEASE=false" >> "$GITHUB_OUTPUT"
+            echo "::notice::No prerelease version found"
+          fi
+
+  docker-push-stable:
```
fmenezes: do you need `if: needs.get-versions.outputs.HAS_PRERELEASE != 'true'` ?

himanshusinghs: we will always have stable releases so when pushing updates for stable channel we don't need this check. The idea is that the scheduled docker updates will push update for both stable and prerelease channels.

.github/workflows/docker-publish.yml
gagik: thanks! didn't connect the dots

.github/workflows/publish.yml
gagik: ah I mixed this up with `set -x` that prints output (which apparently is how you enable it... seems counter-intuitive)

.github/workflows/docker.yml
```diff
@@ -20,8 +63,8 @@ jobs:
     permissions:
       contents: read
       issues: write
-    needs: docker-push
-    if: ${{ always() && needs.docker-push.result == 'failure' }}
+    needs: [get-version, docker-push]
+    if: ${{ always() && (needs.get-version.result == 'failure' || needs.docker-push.result == 'failure') }}
```
himanshusinghs: @fmenezes The default behaviour of a job lifecycle is to be skipped when the required jobs fails which is why always is necessary here, the follow up condition after '&' narrows the `always()` down further so that we run it only explicitly for failures.

fmenezes: I think `if: ${{ failure() }}` would work check https://github.com/mongodb-js/mongodb-mcp-server/actions/runs/21394740051/job/61590155676?pr=884

himanshusinghs: Thanks for digging in. I have updated the expression

himanshusinghs: For further reference [here's](https://docs.github.com/en/actions/reference/workflows-and-actions/expressions#failure) the more cleaner mention of how failure works.

Issue #888, opened 2026-01-30
[Bug]: Ambiguous output message in mcp_mongodb-mcp_find tool: "resulted in" reflects limited count, not total matches


grassfish1122:
### Version

"version": "1.4.0"

### App

- [ ] Cursor
- [ ] Windsurf
- [x] VSCode
- [ ] VSCode Insiders
- [ ] Claude Desktop
- [ ] Other

### Affected Models (if applicable)

- [ ] Claude 3.5 Sonnet
- [ ] Claude 3.7 Sonnet
- [ ] GPT-4a
- [ ] o4-mini
- [x] Other

## Description
The mcp_mongodb-mcp_find tool outputs a message stating "Query on collection ... resulted in N documents" when executing a query. Currently, N represents the number of documents returned after applying the limit , not the total number of documents matching the query filter in the database.

This phrasing is misleading because users (and LLMs) often interpret "resulted in" as the total count of matching records found in the database. For example, if a collection has 13 documents and the default limit is 10, the tool says "resulted in 10 documents," causing the user to incorrectly believe only 10 documents exist.

## Reproduction Steps
- Have a MongoDB collection with more than 10 documents (e.g., 13 documents).
- Invoke the mcp_mongodb-mcp_find tool without specifying a limit (which defaults to 10) or with an explicit limit: 10 .
- Observe the tool's output message.

## Actual Behavior
The tool outputs: Query on collection "..." resulted in 10 documents. Returning 10 documents. This implies that the query only found 10 documents in total.

## Expected Behavior
The output should clearly distinguish between the number of matching documents and the number of returned documents.
Ideally, it should look like: Found 13 matching documents. Returning 10 documents (limit applied). Or at least: Returning 10 documents (limit 10 applied). to avoid implying that 10 is the total count.

## Impact
This causes confusion during debugging and data exploration. Users may miss data because they assume they have retrieved the full result set when they have only seen a partial view.

PR #890, opened 2026-02-02
fix(mongodb): correct limit usage in find tool count queries

grassfish1122:
## Checklist

Related to #888

### DIFF
--- a/src/tools/mongodb/read/find.ts
+++ b/src/tools/mongodb/read/find.ts
@@ -89,9 +89,8 @@ Note to LLM: If the entire query result is required, use the "export" tool inste
                         provider.countDocuments(database, collection, filter, {
                             // We should be counting documents that the original
                             // query would have yielded which is why we don't
-                            // use `limitOnFindCursor` calculated above, only
-                            // the limit provided to the tool.
-                            limit,
+                            // use `limitOnFindCursor` calculated above, and
+                            // we don't use the limit provided to the tool either.
                             maxTimeMS: QUERY_COUNT_MAX_TIME_MS_CAP,
                             // @ts-expect-error signal is available in the driver but not NodeDriverServiceProvider MONGOSH-3142
                             signal,
--- a/tests/integration/tools/mongodb/read/find.test.ts
+++ b/tests/integration/tools/mongodb/read/find.test.ts
@@ -110,6 +110,7 @@ describeWithMongoDB("find tool with default configuration", (integration) => {
             projection?: unknown;
             sort?: unknown;
             expected: unknown[];
+            expectedTotalCount?: number;
         }[] = [
             {
                 name: "returns all documents when no filter is provided",
@@ -141,6 +142,7 @@ describeWithMongoDB("find tool with default configuration", (integration) => {
                     { _id: expect.any(Object) as unknown, value: 6 },
                     { _id: expect.any(Object) as unknown, value: 7 },
                 ],
+                expectedTotalCount: 4,
             },
             {
                 name: "returns documents matching the filter with sort",
@@ -153,7 +155,7 @@ describeWithMongoDB("find tool with default configuration", (integration) => {
             },
         ];
 
-        for (const { name, filter, limit, projection, sort, expected } of testCases) {
+        for (const { name, filter, limit, projection, sort, expected, expectedTotalCount } of testCases) {
             it(name, async () => {
                 await integration.connectMcpClient();
                 const response = await integration.mcpClient().callTool({
@@ -168,7 +170,8 @@ describeWithMongoDB("find tool with default configuration", (integration) => {
                     },
                 });
                 const content = getResponseContent(response);
-                expect(content).toContain(`Query on collection "foo" resulted in ${expected.length} documents.`);
+                const expectedCount = expectedTotalCount ?? expected.length;
+                expect(content).toContain(`Query on collection "foo" resulted in ${expectedCount} documents.`);
 
                 const docs = getDocsFromUntrustedContent(content);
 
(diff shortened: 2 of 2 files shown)


himanshusinghs:
Hey @grassfish1122! Thank you for contributing.

The original intention behind using the limit was to factor the entire query which includes limit (whether or not it was provided by the agent). With your change, I guess you're in need to fetch the estimated number of documents in the collection? If so, I would say that we have a dedicated tool `count` for that.

grassfish1122:
Hi @himanshusinghs , thanks for your response!

I think there might be a small misunderstanding of the intention behind this change, so I’d like to clarify the reasoning.

The goal here is **not** to estimate the number of documents in the collection, nor to replicate the behavior of the dedicated `count` tool.
The purpose of `queryResultsCount` is to represent **how many documents the query *would match in total***, independent of how many documents we choose to return to the user.

Previously, by passing `limit` into `countDocuments`, we were effectively counting *“how many documents were returned”* rather than *“how many documents matched the query”*.
For example:

* Query matches 30 documents
* `limit = 10`
* `countDocuments(..., { limit: 10 })` returns **10**

This creates a misleading situation where the tool reports that the query only matched 10 documents, even though 20 more exist but were truncated. As a result, `queryResultsCount` loses its value as a signal of the full result set.

The intent of this tool is to communicate:

> “Your query matches **N** documents in total, but only **M** are being returned due to limits.”

Removing `limit` from `countDocuments` restores that semantic distinction:

* Returned documents are still limited by `limit`
* `queryResultsCount` correctly reflects the full match count

This also aligns with the inline comment already present in the code:

> *“We should be counting documents that the original query would have yielded … only the limit provided to the tool.”*

Regarding the existing `count` tool: that tool is user-invoked and serves a different purpose.
Here, the count is **internal metadata** used to provide accurate context alongside a truncated result set, not to expose a separate counting operation.

So in short, this change is about **correctness and user clarity**, not about estimating collection size or duplicating another tool’s behavior.

grassfish1122:
I’d like to add a concrete example from a real LLM usage scenario to explain why `queryResultsCount` must reflect the **total number of matched documents**, not a limit-truncated value.

### Example (realistic LLM + find tool flow)

Assume the user asks: **“Which employees belong to the Engineering department?”**

The LLM calls the `find` tool **without explicitly providing a limit**, so the tool applies the default `limit: 10`.

#### Current behavior (passing `limit` to `countDocuments`)

The tool returns something like:

* `resulted in 10 documents. Returning 10 documents`

However, in reality there are **25** employees in the Engineering department.

From the LLM’s perspective, this response means:
**“The query matched 10 documents in total.”**
As a result, the model assumes the result set is complete and stops querying, producing a final answer without attempting to fetch more data.

#### Expected behavior (count not affected by limit)

The response should instead be:

* `resulted in 25 documents. Returning 10 documents`

This makes it explicit that **15 documents are missing**, allowing the LLM to correctly decide to continue querying (e.g. via pagination / cursor usage) and produce a complete answer.

---

### Why the current behavior is problematic in production

Because the `find` tool’s reported match count is currently unreliable, the only safe workaround is to:

1. Call the `count` tool first to get the true match count
2. Then call `find` to retrieve the actual documents

This introduces several concrete issues in production environments:

* **An extra MCP / tool call per query**
* **Additional token usage**
* **Increased end-to-end latency for users**
* Amplified cost and delay in multi-step or chained retrieval flows

---

### Summary of the change

The responsibility of the `find` tool is not just to return documents, but to provide **accurate metadata** that helps the LLM decide what to do next.

* `limit` should only affect how many documents are *returned*
* `queryResultsCount` should represent how many documents *actually match the query*

Removing `limit` from `countDocuments` fixes a real LLM-facing bug where truncated results are incorrectly interpreted as complete results, and avoids forcing upstream agents to perform redundant `count` calls with significant cost and latency impact.

grassfish1122:
@himanshusinghs 
Hi there, I’d like to kindly follow up on this PR. The bug in the find tool's count logic is currently impacting the Agent I’ve deployed, so I’d appreciate it if we could revisit the proposed changes. I’m also open to further discussion if you have any concerns or alternative ideas. Thanks for your time and attention!

gagik:
@grassfish1122 Prompt modification seems like the easy solution to this, do you think the prompt modification would fix the issue for you? If the results stated e.g. "...with limit ${limit} applied", I believe most models should be able to derive the implied limit here and correct themselves.

grassfish1122:
@gagik 
Thank you for your response!

Using prompt engineering can certainly work around the problem, but I believe our goal here is to fix the underlying bug rather than just finding a workaround for my specific use case — isn’t it?

I’d really appreciate it if you could take another look at the example I provided above and try it in a real production-like environment. That should make the current issue with the MongoDB MCP more apparent.

In fact, if you copy and paste the case I shared directly into a coding AI (such as Cursor), it can already understand what kind of change is needed here, which might help clarify the expected behavior.

gagik:
@grassfish1122 I agree generally.
We will discuss whether we want to actually mention the number of total documents here in general as it's somewhat of an unexpected side effect to run count here (which should be delegated to the count tool).

grassfish1122:
@gagik Thanks for your reply.

From my perspective, keeping the "resulted in" information in responses like resulted in 25 documents. Returning 10 documents is quite important.

This allows the LLM to clearly understand that the tool matched a certain number of documents in total, but due to constraints such as limit, only a subset of them was returned. With this information, the LLM can naturally reason that it may need to take additional steps if more data is required.

For example, if I ask the LLM “Who are the members of the IT support team?”, the model will typically call the find tool. If the response only says Returning 10 documents without indicating the total number of matches, the LLM may repeatedly receive partial results without realizing that the results are truncated.

In that case, the only workaround would be to explicitly instruct the LLM to always call count alongside find to determine the total number of matching documents.

More generally, MCP tools are used by agents as problem-solving utilities rather than strictly defined single-purpose functions. Because of this, providing signals that help LLMs reason about tool results can make the tools easier and more effective for agents to use.

gagik:
@grassfish1122 Those are all good points.

We will go ahead with this change, will likely do a few follow-ups before the next release (such as making the timeout on count configurable and handling the timeout scenario better) but otherwise this is good to merge! 

Thank you for sharing your detailed thoughts and your contribution!

PR #891, opened 2026-02-02
fix: ensure we don't log errors to console MCP-382

nirinchev:
## Proposed changes

This removes a few console statements which could leak internal details (e.g. the connection string upon validation failure). Additionally, it makes sure that we register config secrets correctly and adds a test to catch future regressions.

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -286,8 +286,6 @@ export class Server {
             try {
                 validateConnectionString(this.userConfig.connectionString, false);
             } catch (error) {
-                // eslint-disable-next-line no-console
-                console.error("Connection string validation failed with error: ", error);
                 throw new Error(
                     "Connection string validation failed with error: " +
                         (error instanceof Error ? error.message : String(error))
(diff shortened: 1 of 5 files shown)


src/server.ts
```diff
@@ -286,8 +286,6 @@ export class Server {
             try {
                 validateConnectionString(this.userConfig.connectionString, false);
             } catch (error) {
-                // eslint-disable-next-line no-console
-                console.error("Connection string validation failed with error: ", error);
```
himanshusinghs, line 290: I wonder if this was because this code path is very early on when our loggers are not even intialised?

nirinchev: They are initialized - I don't know what was the reason to use console here instead of the logger. In any case, this is throwing anyway, so it will eventually be caught by the runner setting up the server - e.g. https://github.com/mongodb-js/mongodb-mcp-server/blob/511a7d8f28b66169b74fcf37d77589bdc72eb737/src/transports/stdio.ts#L20

and 

https://github.com/mongodb-js/mongodb-mcp-server/blob/511a7d8f28b66169b74fcf37d77589bdc72eb737/src/transports/streamableHttp.ts#L190

which eventually bubbles up the error to the request error handler:

https://github.com/mongodb-js/mongodb-mcp-server/blob/511a7d8f28b66169b74fcf37d77589bdc72eb737/src/transports/streamableHttp.ts#L246

PR #893, opened 2026-02-02
fix: warn on insecure apiBaseUrl

nirinchev:
## Proposed changes

Changes config validation to allow insecure base url by warning instead of erroring out.

### DIFF
--- a/tests/integration/tools/atlas/performanceAdvisor.test.ts
+++ b/tests/integration/tools/atlas/performanceAdvisor.test.ts
@@ -69,6 +69,7 @@ describeWithAtlas("performanceAdvisor", (integration) => {
                                 ],
                             },
                         ],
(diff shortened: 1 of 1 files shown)


fmenezes:
I think this is fine but we should consider a setting `enforeSecureAtlasApi` for the future

fmenezes (approved):

src/server.ts
```diff
@@ -299,10 +299,20 @@ export class Server {
                 if (!this.session.apiClient) {
                     throw new Error("API client is not available.");
                 }
-                if (!this.userConfig.apiBaseUrl.startsWith("https://")) {
```
blva, line 302: you could probably only allow https and localhost instead if you want to keep the behaviour

PR #894, opened 2026-02-04
chore: only allow localhost insecure urls

nirinchev:
## Proposed changes

Follow up on https://github.com/mongodb-js/mongodb-mcp-server/pull/893 to allow only localhost urls to be http.

### DIFF
--- a/tests/integration/server.test.ts
+++ b/tests/integration/server.test.ts
@@ -244,7 +244,7 @@ describe("Server integration test", () => {
         });
 
         for (const insecureUrl of ["http://localhost:8080", "http://127.0.0.1", "http://[::1]"]) {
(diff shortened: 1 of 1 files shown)


src/server.ts
```diff
@@ -303,6 +303,12 @@ export class Server {
                 try {
                     const apiBaseUrl = new URL(this.userConfig.apiBaseUrl);
                     if (apiBaseUrl.protocol !== "https:") {
+                        // For localhost urls, we log a warning, otherwise we throw an error.
+                        const localhostAddresses = ["localhost", "127.0.0.1", "[::1]"];
```
fmenezes, line 307: fern would be docker network, should we configure this?

PR #906, opened 2026-02-12
feat(lib): add ability to pass custom context to tool constructor

gagik:
This provides a way to do better Dependency Injection with custom tools.


gagik:
Ah, this needs a couple follow-ups, will do it in a separate PR

PR #900, opened 2026-02-06
chore: refactor transport runner interface to allow extending Session with Remote MCP specific capabilities

himanshusinghs:
## Proposed changes
Still a WIP. Feel free to go through it but I am opening this only to let the tests run.

## Checklist

- [ ] I have signed the [MongoDB CLA](https://www.mongodb.com/legal/contributor-agreement)

PR #909, opened 2026-02-13
 chore: refactor session-based server configuration 

gagik:
### Before
All configuration for the runner was through passing options into its constructor:
- `tools`, `toolsContext`, `telemetryProperties` => would eventually get statically passed to `new Server`
- Different hooks like `createSessionConfig` would be run on request and then eventually to `new Server`

```ts
const baseConfig = UserConfigSchema.parse({
  transport: "http",
  httpPort: 3000,
  httpHost: "127.0.0.1",
});

// Session configuration hook
const createSessionConfig: TransportRunnerConfig["createSessionConfig"] =
  async ({ userConfig, request }) => {
    // Extract user ID from request headers
    const userId = request?.headers?.["x-user-id"];

      (op) => !permissions.allowedOperations.includes(op)
    );

    // Return customized configuration for this session
    return {
      ...userConfig,
      disabledTools: disabledOperations,
      maxDocumentsPerQuery: permissions.maxDocuments,
      // Analysts get read-only access
      readOnly: permissions.role === "analyst",
    };
  };

// Initialize the server with session configuration hook
const runner = new StreamableHttpRunner({
  userConfig: baseConfig,
  createSessionConfig,
  createApiClient: ({ request }) => permissions.role === "analyst" ? ... : ...,
  createAtlasLocalClient:  ({ request }) => permissions.role === "analyst" ? ... : ...,
});
```

### After (different example)
Depending on needs, there's a few ways to achieve customization:
1. Static changes such as `userConfig` get passed into `runner.start({ sessionOptions: userConfig, serverOptions: {...})`
2. Dynamic, request-specific changes are accomplished by extending the `StreamableHttpRunner` and overriding `createServerForRequest`.

```ts
// Extend StreamableHttpRunner and override createServerForRequest
class CustomStreamableHttpRunner extends StreamableHttpRunner {
  protected override async createServerForRequest({
    request,
  }: {
    request: RequestContext;
  }): Promise<Server> {
    // Get environment from request header
    const environment = request?.headers?.[
      "x-environment"];

    // Build customized configuration for this session
    const sessionConfig: UserConfig = {
      ...this.userConfig,
      connectionString: connectionStrings[environment],
      // Production is read-only
      readOnly: environment === "production",
    };

    // Create the server with the session-specific configuration
    return this.createServer({
      userConfig: sessionConfig,
	  serverOptions: {
	  		apiClient: ....
			atlasLocalClient: .....
	  }
    });
  }
}
```

This...
1. separates request-specific logic from `TransportBase` and moves it to `StreamableHttpRunner`.
3. Add `serverOptions` and `sessionOptions` to `TransportBase.start`. => **This allows us to deprecate options like `connectionErrorHandler` and `connectionManager`.
4. Introduces `createServer` which is a request-agnostic helper that is used by `ToolBase.setupServer` (now deprecated). 
5. Deprecated `ToolBase.setupServer` in favor of new `StreamableHttp.createServerForRequest`. Instead of passing hooks, users are encouraged to override this function. This removes the need for hooks and instead provides a centralized way to create servers which are customizable both server config and session config-wise.
6. Moves arguments like `tools` from the `TransportRunner` constructor to `TransportRunner.start({ serverOptions })` instead

Benefits:
1. Avoid awkward duplication of arguments between `serverOptions` and `transportConfigOptions`.
2. Makes it possible to customize both `serverOptions` and `sessionOptions` based on request context => unlocks many new possibilities for library users. 
4. We no longer have to introduce new hooks for every dependency, we simply make it possible to specify them through `serverOptions`
5. `TransportBase` is kept pure, there is no HTTP-specific concept of `request` and instead transports can introduce any abstractions on top of the generic `createServer` function. Extending is therefore easier to reason about and more flexible.

This way, we can do cleaner advanced configurations such as
- injecting different tool context based on request
- customizing API clients or other services based on request
and make the structure less redundant and easier to structure than the hooks.

This shouldn't cause any breaking changes, all hooks & fields should be deprecated instead. `toolContext` gets removed right away since that is a new addition.

#### Additional changes
- Makes elicitation injectable

### DIFF
--- a/src/transports/base.ts
+++ b/src/transports/base.ts
@@ -111,7 +111,7 @@ export type TransportRunnerConfig = {
     userConfig: UserConfig;
 
     /**
-     * @deprecated Use `createServer({ sessionOptions: {connectionManager: MyCustomConnectionManager} })` instead
+     * @deprecated Use `start({ sessionOptions: {connectionManager: MyCustomConnectionManager} })` instead
      * An optional factory function to generates an instance of
      * `ConnectionManager`. When not provided, MongoDB MCP Server uses an
      * internal implementation to manage connection to MongoDB deployments.
@@ -121,11 +121,11 @@ export type TransportRunnerConfig = {
      */
     createConnectionManager?: ConnectionManagerFactoryFn;
 
-    /** @deprecated Use `createServer({ serverOptions: {connectionErrorHandler: MyCustomConnectionErrorHandler} })` instead */
+    /** @deprecated Use `start({ serverOptions: {connectionErrorHandler: MyCustomConnectionErrorHandler} })` instead */
     connectionErrorHandler?: ConnectionErrorHandler;
 
     /**
-     * @deprecated Use `createServer({ sessionOptions: {atlasLocalClient: MyCustomAtlasLocalClient} })` instead
+     * @deprecated Use `start({ sessionOptions: {atlasLocalClient: MyCustomAtlasLocalClient} })` instead
      * An optional factory function to create a client for working with Atlas
      * local deployments. When not provided, MongoDB MCP Server uses an internal
      * implementation to create the local Atlas client.
@@ -265,7 +265,7 @@ export abstract class TransportRunnerBase<TContext = unknown> {
             (await this.createConnectionManager({ logger: logger, deviceId: this.deviceId, userConfig }));
 
         const apiClient =
-            userConfig.apiClientId && userConfig.apiClientSecret
+            (sessionOptions?.apiClient ?? (userConfig.apiClientId && userConfig.apiClientSecret))
                 ? defaultCreateApiClient(
                       {
                           baseUrl: userConfig.apiBaseUrl,
@@ -276,7 +276,7 @@ export abstract class TransportRunnerBase<TContext = unknown> {
                       },
                       logger
                   )
-                : sessionOptions?.apiClient;
+                : undefined;
 
         const session = new Session({
             userConfig,


gagik:
Very impressed by Copilot comments this time around 😮

gagik:
There is an unrelated issue with generic type parameters for context injection.... Moved those changes to https://github.com/mongodb-js/mongodb-mcp-server/pull/911

src/transports/base.ts
```diff
@@ -21,15 +21,34 @@ import { defaultCreateAtlasLocalClient } from "../common/atlasLocal.js";
...
+export type CustomizableServerOptions<TContext = unknown> = Partial<
+    Pick<ServerOptions<TContext>, "uiRegistry" | "tools" | "connectionErrorHandler" | "toolContext" | "elicitation">
+> & {
+    /**
+     * An optional key value pair of telemetry properties that are reported to
+     * the telemetry backend. Most, if not all, of the properties are captured
+     * automatically.
+     */
+    telemetryProperties?: Partial<CommonProperties>;
+};
```
himanshusinghs, line 53: Its a small thing but you could you please pull in this commit:
https://github.com/mongodb-js/mongodb-mcp-server/pull/900/commits/916bda13bb5156baba61eb85b821ba2197137a85

It ties connectionErrorHandler to Session where it actually belongs instead of Server.

src/transports/base.ts
```diff
@@ -317,46 +258,43 @@ export abstract class TransportRunnerBase<TContext = unknown> {
...
+                      {
+                          baseUrl: userConfig.apiBaseUrl,
+                          credentials: {
+                              clientId: userConfig.apiClientId,
+                              clientSecret: userConfig.apiClientSecret,
+                          },
+                      },
+                      logger
+                  )
+                : sessionOptions?.apiClient;
```
himanshusinghs: One thing I noticed was that we can always create an apiClient regardless of the presence of client id and secret. I guess we don't want to do that because of the node-fetch problem you mentioned earlier?

src/transports/base.ts
```diff
@@ -21,15 +21,37 @@ import { defaultCreateAtlasLocalClient } from "../common/atlasLocal.js";
...
 
+export type CustomizableSessionOptions = Partial<
+    Pick<
+        SessionOptions,
+        "userConfig" | "apiClient" | "atlasLocalClient" | "connectionManager" | "connectionErrorHandler"
+    >
+>;
+
+export type CustomizableServerOptions<TContext = unknown> = Partial<
+    Pick<ServerOptions<TContext>, "uiRegistry" | "tools" | "toolContext" | "elicitation">
```
himanshusinghs, line 45: nit: `elicitation` was not in the previous interface that we deprecated. I think its fine to have it but just highlighting it.

src/transports/base.ts
```diff
@@ -89,6 +111,7 @@ export type TransportRunnerConfig<TContext = unknown> = {
     userConfig: UserConfig;
 
     /**
+     * @deprecated Use `createServer({ sessionOptions: {connectionManager: MyCustomConnectionManager} })` instead
```
himanshusinghs: nit: Because createServer is `protected`, perhaps its best to advice to extend the runner and make use of createServer() in the start() method?, the same way you advice in class properties deprecation message?

src/transports/base.ts
```diff
@@ -384,7 +322,65 @@ export abstract class TransportRunnerBase<TContext = unknown> {
...
+
+    abstract start({
+        serverOptions,
+        sessionOptions,
+    }: {
+        /** Upstream `serverOptions` passed from running `runner.start({ serverOptions })` method */
+        serverOptions?: ServerOptions<TContext>;
+        /** Upstream `sessionOptions` passed from running `runner.start({ sessionOptions })` method */
+        sessionOptions?: SessionOptions;
+    }): Promise<void>;
```
himanshusinghs, line 383: We might need to make the entire argument optional as start was not receiving any argument earlier.

tests/integration/server.test.ts
```diff
@@ -185,8 +185,22 @@ describe("Server integration test", () => {
...
+                          {
+                              baseUrl: config.apiBaseUrl,
+                              credentials: {
+                                  clientId: config.apiClientId,
+                                  clientSecret: config.apiClientSecret,
+                              },
+                          },
+                          logger
+                      )
+                    : undefined,
```
himanshusinghs:
```suggestion
            apiClient: defaultCreateApiClient(
                {
                    baseUrl: config.apiBaseUrl,
                    credentials: {
                        clientId: config.apiClientId,
                        clientSecret: config.apiClientSecret,
                    },
                },
                logger
            ),
```

I think this will work out.

src/transports/base.ts
```diff
@@ -317,46 +258,43 @@ export abstract class TransportRunnerBase<TContext = unknown> {
...
-                logger
-            );
-        }
+        const connectionManager =
+            sessionOptions?.connectionManager ??
+            (await this.createConnectionManager({ logger: logger, deviceId: this.deviceId, userConfig }));
+
+        const apiClient =
+            userConfig.apiClientId && userConfig.apiClientSecret
+                ? defaultCreateApiClient(
```
himanshusinghs, line 269: I think the order should be the other way around? `sessionOptions?.apiClient ?? defaultCreateApiClient({ ... })`

src/transports/streamableHttp.ts
```diff
@@ -48,8 +64,80 @@ export class StreamableHttpRunner<TContext = unknown> extends TransportRunnerBas
...
+        /** Upstream `sessionOptions` passed from running `runner.start({ sessionOptions })` method */
+        sessionOptions?: CustomizableSessionOptions;
+    }): Promise<Server<TContext>> {
+        let userConfig: UserConfig = sessionOptions?.userConfig ?? this.userConfig;
+
+        if (this.createSessionConfig) {
+            userConfig = await this.createSessionConfig({ userConfig, request });
+        } else {
+            userConfig = applyConfigOverrides({ baseConfig: userConfig, request });
+        }
```
himanshusinghs, line 87: I guess you've been back and forth on this but I just wanna mention that this is technically a breaking change that createSessionConfig when provided for stdio transport runner (does not make sense) won't be called.

Was there something that prevented you from keeping it in base? Totally agree that this piece of code should live only for Http runner.

src/transports/base.ts
himanshusinghs: Handled

src/transports/base.ts
himanshusinghs: I noticed later that you're using default value pattern. I think that covers us even for the overriding case.

src/transports/base.ts
gagik: > I guess we don't want to do that because of the node-fetch problem you mentioned earlier

yes, this will currently break web comptability. We should look into getting rid of node-fetch as a general goal

src/transports/base.ts
gagik: true, I looked into it and thought it'd be nice to include it, I'll add it to the PR description
(this is a good argument for changelogs in this specific case... but still unsure about it)

src/transports/base.ts
gagik: true, these should say use `start(...`) instead

in case you also meant we instruct people to override start: 
I agree the different ways can be slightly awkward but each transport does some funkiness before/after running createServer so overriding that would be problematic.

One advantage of not getting people to override `start` is that we can also make different transports react to different injected settings, if needed

src/transports/base.ts
gagik: yeah I apparently cannot specify this defaults in abstract class

src/transports/base.ts
gagik: good catch! agh really not a fan of this conditional api client... but i think we should deal with it separately

src/transports/streamableHttp.ts
gagik: Good point, that's technically true but I would honestly consider this a "bugfix" in the sense that non-http runners were not meant to have this passable. You're right that this is a breaking change but this is breaking incorrect/unintended behavior so we should get away with it.

gagik: actually, wouldn't this cover keep backwards compatibility?
https://github.com/mongodb-js/mongodb-mcp-server/blob/16ed2af6ede7fd894bd367ea4ad32b16646b03b8/src/transports/base.ts#L341-L343

himanshusinghs: Ahh interesting - it might! So it definitely covers for folks using the old setupServer but I think the people using createServer with stdio runner won't come in that category.

Anyways we're trying to categorize this as a bugfix so its fine not to worry about this.

gagik: ah wait no you're right it won't because it's no longer using in `setupServer` in `stdio`, but yeah I'd say let's see it as a breaking bugfix

src/transports/base.ts
gagik: https://jira.mongodb.org/browse/MCP-370 tracks that

PR #911, opened 2026-02-16
refactor: add type enforcement for custom userconfig support

gagik:
Previously you could extend the UserConfig schema but the types wouldn't update. This allows one to define custom UserConfig and access its props in a typesafe manner across the entire stack.

### DIFF
--- a/tests/integration/transports/streamableHttp.test.ts
+++ b/tests/integration/transports/streamableHttp.test.ts
@@ -8,13 +8,7 @@ import { Keychain } from "../../../src/common/keychain.js";
 import { defaultTestConfig, InMemoryLogger, timeout } from "../helpers.js";
 import { type UserConfig } from "../../../src/common/config/userConfig.js";
 import type { StreamableHTTPServerTransport } from "@modelcontextprotocol/sdk/server/streamableHttp.js";
-import type {
-    OperationType,
-    ToolArgs,
-    ToolCategory,
-    ToolClass,
-    ToolExecutionContext,
-} from "../../../src/tools/tool.js";
+import type { OperationType, ToolArgs, ToolCategory, ToolExecutionContext } from "../../../src/tools/tool.js";
 import { ToolBase } from "../../../src/tools/tool.js";
 import type { CallToolResult } from "@modelcontextprotocol/sdk/types.js";
 import type { TelemetryToolMetadata } from "../../../src/telemetry/types.js";


src/server.ts
```diff
@@ -127,15 +127,15 @@ export class Server<TContext = unknown> {
...
+    }: ServerOptions<TUserConfig, TContext>) {
         this.startTime = Date.now();
         this.session = session;
         this.telemetry = telemetry;
         this.mcpServer = mcpServer;
         this.userConfig = userConfig;
         this.elicitation = elicitation;
         this.connectionErrorHandler = connectionErrorHandler;
-        this.toolConstructors = tools ?? AllTools;
+        this.toolConstructors = (tools ?? AllTools) as unknown as ToolClass<TUserConfig, unknown>[];
```
gagik: this is a tricky one. we can either give up enforcing custom context on the tools that get provided or just cast them here.

tests/integration/transports/base.test.ts
```diff
@@ -21,7 +22,7 @@ describe("TransportRunnerBase", () => {
                     ...defaultTestConfig,
                     httpPort: 0,
                     previewFeatures: [], // mcpUI not included
-                },
+                } as UserConfig,
```
gagik: this is somewhat of a breaking change but it's more like a bugfix... it's just that before this vaguely defined type would be "good enough" to be a userconfig and now it needs to be explicitly defined as such

src/tools/tool.ts
```diff
@@ -885,6 +884,9 @@ export abstract class ToolBase<TContext = unknown> {
     }
 }
 
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export type AnyToolBase = ToolBase<any, any>;
```
himanshusinghs, line 888: Would it be better to type it as `ToolBase<UserConfig, unknown>` or does that not align with the usage?

tests/integration/transports/base.test.ts
himanshusinghs: I checked it locally and it seems to function just fine?
`defaultTestConfig` is totally compliant with `UserConfig` so I think it should be fine to remove that?

src/tools/tool.ts
gagik: it won't unfortunately, I don't quite understand why but typescript has issues with it

tests/integration/transports/base.test.ts
gagik: oh you're right.. interesting. I wonder if it was the tool constructor or something that messing with this type before

src/tools/tool.ts
himanshusinghs: Got it 👍

PR #912, opened 2026-02-17
feat: update atlas-local-create-deployment tool to use preview image

Luke-Sanderson:
## Proposed changes
[MCP-384] [MongoDB MCP] Support atlas-local-deployment-start with preview tag](https://jira.mongodb.org/browse/MCP-384)

Won't be merged until project completion and `mongodb/mongodb-atlas-local:preview` is released and tests are passing

## Test Failure Note
Tests are currently failing because the `mongodb/mongodb-atlas-local:preview` image has not been published yet. 
Tests are failing because preview image does not exist yet. Adding `image: "lukesando/mongodb-atlas-local",
` to `executeWithAtlasLocalClient` changes a repo where a test preview image is published and the tests then pass locally.
## Checklist


tests/integration/tools/atlas-local/createDeployment.test.ts
```diff
@@ -144,6 +145,46 @@ describeWithAtlasLocal("atlas-local-create-deployment", (integration) => {
         expect(elements[1]?.text ?? "").toContain(deploymentName);
         expect(elements[1]?.text ?? "").toContain("Running");
     });
+
+    it("should create a deployment with voyageApiKey set when preview image is used", async () => {
+        const deploymentName = `test-deployment-preview-${Date.now()}`;
+        deploymentNamesToCleanup.push(deploymentName);
+        const voyageApiKey = "test-voyage-api-key";
+
+        integration.mcpServer().userConfig.voyageApiKey = voyageApiKey;
```
gagik, line 162: very nit: does this userConfig get re-created on every test? if not then this might affect other tests unintentionally in the future.

gagik (approved):
LGTM assuming it works once it's out

tests/integration/tools/atlas-local/createDeployment.test.ts
Luke-Sanderson: You are correct the userConfig doesn't get re-created on each test so could affect other tests if they use `userConfig.voyageApiKey`. No other integration test uses `userConfig.voyageApiKey` so it is safe for now.

Do you have a suggested approach to avoid this or is this just something to watch out for in the future?

gagik: generally it's good to have explicit dependency injection, i.e. we create a mock config and pass i.e. into the mock server.

that said, this `describeWithAtlasLocal` doesn't really provide a good way to do this (which is the disadvantage of it in general...) and is somewhat unconventional. We have an awkard system of setting it through: https://github.com/mongodb-js/mongodb-mcp-server/blob/b13fc67ca9b7c548211c29f86e64156b4f3cff81/tests/integration/tools/mongodb/read/aggregate.test.ts#L448

Honestly, a general refactor would be nice but not worth it for this PR. Temporary mitigation could be t have the integration test take a "snapshot" of initial userconfig in a `beforeEach` hook and restore it to the snapshot `afterEach`. I can describe it better over a call if you'd like

You can also ignore it for now, it's not really a real issue and it'd be better solved with a larger refactor than is worthwhile for now

Luke-Sanderson: Thank you for the explanation. I quickly added the snapshot idea. I will create a followup PR with the larger refactor if time allows

gagik: nice! that looks good.

PR #928, opened 2026-02-20
chore: add structured content to responses from mongodb/create tools

nirinchev:
## Proposed changes

Builds on top of https://github.com/mongodb-js/mongodb-mcp-server/pull/841 to add structured content to the create tools.


src/tools/mongodb/create/createCollection.ts
```diff
@@ -21,6 +33,11 @@ export class CreateCollectionTool extends MongoDBToolBase {
                     text: `Collection "${collection}" created in database "${database}".`,
                 },
             ],
+            structuredContent: {
+                database,
+                collection,
+                created: true,
```
gagik, line 39: what's the reason behind `created: true`? just not to confuse the model?

nirinchev: Yeah, this is pretty arbitrary and not something I've extensively tested, but figured it's a good idea to just have it as a hint for the model that this response indicates a created collection.

Issue #937, opened 2026-02-26
[Security] AgentAudit Scan: Undisclosed Telemetry Data Collection


starbuck100:
## Summary

We scanned this package using [AgentAudit](https://agentaudit.dev) as part of a security audit of the top 20 most-used MCP servers. AgentAudit uses multi-model LLM analysis to detect AI-specific security patterns.

**Trust Score: 94/100** | 2 findings (1 medium, 1 low)

### 1. Undisclosed Host System Information in Telemetry
- **ASF-ID:** [ASF-2026-2052](https://agentaudit.dev/findings/ASF-2026-2052)
- **File:** `src/common/atlas/apiClient.ts:45`
- **Description:** The server collects host OS (`process.platform`), architecture (`process.arch`), and hostname (`process.env.HOSTNAME`) and includes it in the User-Agent header for all Atlas API requests and telemetry endpoints. While `MDB_MCP_TELEMETRY` can disable telemetry, this data collection is not prominently disclosed in the README or CONTRIBUTING docs. Hostname can be considered PII.

### 2. Missing Telemetry Documentation
- **ASF-ID:** [ASF-2026-2053](https://agentaudit.dev/findings/ASF-2026-2053)
- **File:** `README.md`
- **Description:** Primary user-facing documentation does not disclose that host system information is collected and sent as telemetry. The `server.json` toggle is primarily for integrators, not easily discoverable by end-users.

## Full Report

View the complete scan report: https://agentaudit.dev/packages/mongodb-mcp-server

## Note

These findings were detected by automated LLM-based security analysis and may contain false positives. We welcome your feedback and are happy to update our records if any findings are disputed.

---
*Scanned with [AgentAudit](https://agentaudit.dev) — open-source security scanner for AI agent packages.*

starbuck100:
Thanks for creating the internal ticket (MCP-405). Happy to provide additional details or help verify fixes. The finding relates to undisclosed telemetry data collection via the mongodb-mcp-server package — full audit details are linked in the issue description.

blva:
Thanks for the report. 

**Finding 1 (ASF-2026-2052) – Undisclosed host system information**  
We’ve removed hostname (https://github.com/mongodb-js/mongodb-mcp-server/pull/938) from the User-Agent. The header now includes only platform and architecture (e.g. `AtlasMCP/x.x.x (darwin; x64)`) The change will be included in the next MCP release (planned this week).

[...]

PR #938, opened 2026-02-27
fix: simplify userAgent

blva:
## Proposed changes
Addresses https://github.com/mongodb-js/mongodb-mcp-server/issues/937
## Checklist

### DIFF
--- a/tests/unit/common/apiClient.test.ts
+++ b/tests/unit/common/apiClient.test.ts
@@ -70,9 +70,7 @@ describe("ApiClient", () => {
             const call = mockFetch.mock.calls[0];
             expect(call).toBeDefined();
             const [url, init] = call!;
-            expect(url instanceof URL ? url.href : url).toBe(
-                "https://api.test.com/api/private/v1.0/telemetry/events"
-            );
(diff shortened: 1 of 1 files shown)


tests/unit/common/apiClient.test.ts
```diff
@@ -58,6 +59,89 @@ describe("ApiClient", () => {
...
+                    baseUrl: "https://api.test.com",
+                    credentials: {
+                        clientId: "test-client-id",
+                        clientSecret: "test-client-secret",
+                    },
+                },
+                new NullLogger()
+            );
+            // @ts-expect-error accessing private property for testing
+            clientWithoutUserAgent.authProvider.getAuthHeaders = vi.fn().mockRejectedValue(new Error("No token"));
```
gagik, line 91: FYI, you can generally do
```ts
clientWithoutUserAgent.authProvider['getAuthHeaders']
```
to access private properties in tests

gagik: the more proper testing way would be to have a full on mocked auth provider which we can then naturally override these fields with... but we don't need to go that far for this

blva: thanks! i'll go ahead with this one to catch up with the release

PR #946, opened 2026-03-03
refactor(logging): split up logging logic into different files

gagik:
All logging logic is in one file which is hard to read through. 
This PR is reorganizing the logging files by splitting them up between individual logger files (mcpLogger, baseLogger...), types and definitions, all re-exported through logging/index.js

It doesn't change any existing code/logic beyond that.


gagik:
@copilot can you fix this

himanshusinghs (approved):

PR #950, opened 2026-03-04
fix: use lock to emit events and avoid event duplication 

blva:
## Proposed changes

Fixes [CLOUDP-380317](https://jira.mongodb.org/browse/CLOUDP-380317): telemetry race where concurrent `emit()` could send the same cached events twice.

- Add `EventCache.runExclusive()` so all sessions serialize when doing getEvents → send → removeEvents.
- Telemetry runs the critical section inside `eventCache.runExclusive(...)`

## Testing
- Regression tests in `telemetry.test.ts` (real EventCache, including delayed-first-send case).

## Checklist


src/telemetry/telemetry.ts
```diff
@@ -181,42 +181,48 @@ export class Telemetry {
...
-                this.events.emit("events-emitted");
-                return;
-            }
 
-            this.session.logger.debug({
-                id: LogId.telemetryEmitFailure,
-                context: "telemetry",
-                message: `Error sending event to client: ${result.error instanceof Error ? result.error.message : String(result.error)}`,
+                const result = await this.sendEvents(apiClient, allEvents);
+                if (result.success) {
```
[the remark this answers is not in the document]
blva: I think I should probably address this and add timeouts both in sendEvents and in the API client (if not set already). But I'll wait for more feedback in case I should change my approach

src/telemetry/eventCache.ts
```diff
@@ -40,6 +42,25 @@ export class EventCache {
         return this.cache.size;
     }
 
+    /**
+     * Runs a callback with exclusive access to the cache so getEvents/send/removeEvents
+     * are serialized across all callers (e.g. multiple Telemetry instances / sessions).
+     * Prevents duplicate sends when the server handles many sessions.
+     */
+    public async runExclusive<T>(fn: () => Promise<T>): Promise<T> {
```
gagik: this could be turned into something like
```ts
this.eventCache.processAndClear((events) => {
	await this.apiClient.send(events);
});
```

and then we can remove or make private `getEvents` and `removeEvents` (this one could be left for cleanup where we skip).

I have been thinking about ways we could avoid the concept of locking and haven't thought of anything better yet

src/telemetry/eventCache.ts
```diff
@@ -40,6 +42,25 @@ export class EventCache {
...
+        let releaseLock!: () => void;
+        this._lock = new Promise<void>((r) => {
+            releaseLock = r;
+        });
+        await prevLock;
+        try {
+            return await fn();
+        } finally {
+            releaseLock();
+        }
```
gagik:
```suggestion
		await this.currentOperation?.promise;
		
		this.currentOperation = Promise.withResolvers();
		
        try {
            return await fn();
        } finally {
            this.currentOperation.resolve();
            this.currentOperation = undefined;
        }
```
this would be equivalent I believe, `Promise.withResolvers` is a neat newish Node 21+ / web thing. We might want to run `this.currentOperation.resolve();` when disposing the event cache too.

and then to add a timeout maybe something like
```ts
	   const timeoutSignal = AbortSignal.timeout(OPERATION_TIMEOUT);
		
        try {
            return await fn({signal: timeoutSignal});
        } finally {
            resolve();
        }
```
we'd be dependent on the `fn` actually making use of the abort signal

jeroenvervaeke: Will this work with the minimum supported version that we support according to our readme?

`At least 20.19.0`

src/telemetry/telemetry.ts
```diff
@@ -181,42 +181,48 @@ export class Telemetry {
...
-            this.eventCache.appendEvents(events);
-            this.events.emit("events-send-failed");
         } catch (error) {
             this.session.logger.debug({
                 id: LogId.telemetryEmitFailure,
                 context: "telemetry",
                 message: `Error emitting telemetry events: ${error instanceof Error ? error.message : String(error)}`,
                 noRedaction: true,
             });
+            this.eventCache.appendEvents(events);
```
jeroenvervaeke: Do we care about the order in which events are sent?

If so, shouldn't we insert the events at the front? With this code, if we end up in the catch, older events will be queued after newer events (if any got appended during the attempted send)

src/telemetry/eventCache.ts
gagik: wait nevermind, my nvm wasn't working properly and it's not supported there. yeah this isn't available there. didn't realize we're still supporting v20, it's reaching EOL next month: [nodejs.org/en/about/previous-releases](https://nodejs.org/en/about/previous-releases) so we could consider dropping this in the future.

to be version-compliant, we can do this which is basically what we had before except we'd also store the ability to resolve in the promise
```ts
let resolve;
const promise = new Promise<void>((res) => {
            resolve = res;
        });
this.currentOperation = { promise, resolve };
```

one advantage of this is being able to resolve this from anywhere in the code, though we should hypothetically expose the abortsignal too in that sense

src/telemetry/telemetry.ts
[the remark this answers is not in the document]
gagik: yes it'd be good to have this covered

src/telemetry/telemetry.ts
himanshusinghs: I noticed that there is a timestamp on the Event object so that might help with the order but in general agreed that it makes sense to treat the cache as a queue more than a stack and enqueue in the front.

src/telemetry/eventCache.ts
himanshusinghs: I second this. It could lead to a cleaner logic.

src/telemetry/telemetry.ts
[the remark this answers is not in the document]

src/telemetry/eventCache.ts
blva: updated, let me know if that works

src/telemetry/telemetry.ts
blva: This is addressed with the processAndClear refactor. On failure, cached events are never removed (they stay in their original cache position), and only newEvents are appended back via appendEvents. Combined with runExclusive preventing concurrent processAndClear calls from interleaving, the relative ordering is preserved. Additionally, each event has a timestamp field, so the server-side can always reconstruct chronological order if needed.

src/common/atlas/apiClient.ts
```diff
@@ -144,14 +144,14 @@ export class ApiClient {
         }>;
     }
 
-    public async sendEvents(events: TelemetryEvent<CommonProperties>[]): Promise<void> {
+    public async sendEvents(events: TelemetryEvent<CommonProperties>[], signal?: AbortSignal): Promise<void> {
```
gagik:
```suggestion
	public async sendEvents(events: TelemetryEvent<CommonProperties>[], { signal }: {signal?: AbortSignal}): Promise<void>
```
feel free to ignore, I tend to avoid more than one positional argument to make it easier to extend it with new arguments in the future.

gagik: also, we could define a default abortsignal here

tests/unit/telemetry.test.ts
```diff
@@ -128,8 +134,26 @@ describe("Telemetry", () => {
         // Setup mocked EventCache
         mockEventCache = new MockEventCache() as unknown as typeof mockEventCache;
         mockEventCache.getEvents = vi.fn().mockReturnValue([]);
-        mockEventCache.removeEvents = vi.fn().mockResolvedValue(undefined);
-        mockEventCache.appendEvents = vi.fn().mockResolvedValue(undefined);
+        mockEventCache.removeEvents = vi.fn();
+        mockEventCache.appendEvents = vi.fn();
+        mockEventCache.processAndClear = vi
```
gagik, line 136: if we're mocking `processAndClear`, is this testing the functionality we need?

src/telemetry/eventCache.ts
```diff
@@ -40,6 +42,53 @@ export class EventCache {
...
+    /**
+     * Under exclusive access: reads cached events, merges with newEvents, runs processor,
+     * then clears sent cached events on success. On failure, newEvents are appended
+     * for retry and the error is re-thrown.
+     */
+    public async processAndClear(
+        newEvents: BaseEvent[],
+        processor: (allEvents: BaseEvent[], cachedCount: number) => Promise<void>
+    ): Promise<void> {
+        await this.runExclusive(async () => {
```
gagik, line 74: nit: can we make `this.runExclusive` private?

gagik (approved):
LGTM, some minor comments

src/telemetry/eventCache.ts
```diff
@@ -40,6 +42,53 @@ export class EventCache {
...
+        }
+    }
+
+    /**
+     * Under exclusive access: reads cached events, merges with newEvents, runs processor,
+     * then clears sent cached events on success. On failure, newEvents are appended
+     * for retry and the error is re-thrown.
+     */
+    public async processAndClear(
+        newEvents: BaseEvent[],
```
gagik: I don't think we need this here, do we? the function itself could add it

src/telemetry/eventCache.ts
```diff
@@ -40,6 +42,53 @@ export class EventCache {
...
+        processor: (allEvents: BaseEvent[], cachedCount: number) => Promise<void>
+    ): Promise<void> {
+        await this.runExclusive(async () => {
+            const cachedEvents = this.getEvents();
+            const allEvents = [...cachedEvents.map((e) => e.event), ...newEvents];
+
+            try {
+                await processor(allEvents, cachedEvents.length);
```
gagik: we could make the processor return the new events state

tests/unit/telemetry.test.ts
blva: i think we should keep this, but i also added unit tests to event cache

PR #954, opened 2026-03-06
Bump non-major dependencies to latest

nirinchev:
## Checklist

### DIFF
--- a/src/transports/streamableHttp.ts
+++ b/src/transports/streamableHttp.ts
@@ -471,8 +471,8 @@ class MCPHttpServer<TUserConfig extends UserConfig = UserConfig, TContext = unkn
             req: express.Request,
             res: express.Response,
             {
-                isImplicitInitialization,
                 sessionId,
+                isImplicitInitialization,
             }:
                 | { sessionId?: string; isImplicitInitialization?: false }
                 | { sessionId: string; isImplicitInitialization: true }


src/common/sessionStore.ts
```diff
@@ -11,14 +11,16 @@ export type CloseableTransport = {
...
 
-export class SessionStore<T extends CloseableTransport = CloseableTransport> {
+export type SessionInfo<T extends { close: () => Promise<void> }> = {
+    logger: LoggerBase;
+    value: T;
+    abortTimeout: ManagedTimeout;
+    notificationTimeout: ManagedTimeout;
+};
+
+export class SessionStore<T extends { close: () => Promise<void> }> {
```
gagik: why not `extends CloseableTransport`?

src/transports/streamableHttp.ts
```diff
@@ -318,13 +343,80 @@ class MCPHttpServer<TUserConfig extends UserConfig = UserConfig, TContext = unkn
         this.userConfig = userConfig;
     }
 
+    protected createTransportOptions(): WebStandardStreamableHTTPServerTransportOptions {
+        return {
+            onsessionclosed: async (sessionId): Promise<void> => {
+                try {
+                    // TODO: check if true/false
```
gagik: should this be addressed?

src/transports/streamableHttp.ts
```diff
@@ -585,6 +572,173 @@ class MCPHttpServer<TUserConfig extends UserConfig = UserConfig, TContext = unkn
     }
 }
 
+class SSEResponseHttpServer<TUserConfig extends UserConfig = UserConfig, TContext = unknown> extends MCPHttpServer<
```
gagik: can we move these servers to separate files?

src/common/sessionStore.ts
nirinchev: Because we can store non-transports there (e.g. for json responses, we store the server as the transport is recreated on each request).

src/transports/streamableHttp.ts
nirinchev: Will do!

src/common/sessionStore.ts
gagik: could we rename `CloseableTransport` to `ClosableObject` or something then?

PR #960, opened 2026-03-10
feat: add metrics service MCP-411

gagik:
Adds `metrics` as a service for keeping various metrics about tool usage, etc.

For now, it only includes 2 initial metrics, more will be added in follow-up PRs.

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -23,7 +23,6 @@ import type { Elicitation } from "./elicitation.js";
 import { AllTools } from "./tools/index.js";
 import type { UIRegistry } from "./ui/registry/index.js";
 import type { Metrics, DefaultMetrics } from "./common/metrics/index.js";
-import { createDefaultMetrics, PrometheusMetrics } from "./common/metrics/index.js";
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 export type AnyToolClass = ToolClass<any, any, any>;
--- a/tests/unit/toolContext.test.ts
+++ b/tests/unit/toolContext.test.ts
@@ -11,6 +11,7 @@ import type { Elicitation } from "../../src/elicitation.js";
 import { connectionErrorHandler } from "../../src/common/connectionErrorHandler.js";
 import type { TelemetryToolMetadata } from "../../src/telemetry/types.js";
 import type { UserConfig } from "../../src/lib.js";
+import { MockMetrics } from "./mocks/metrics.js";
 
 // Define a custom context type
 interface CustomContext {
@@ -122,6 +123,7 @@ describe("Tool Context", () => {
             connectionErrorHandler,
             tools: [ToolWithContext],
             toolContext: customContext,
+            metrics: new MockMetrics(),
         });
 
         server.registerTools();
@@ -193,6 +195,7 @@ describe("Tool Context", () => {
             connectionErrorHandler,
             tools: [ToolWithoutContext],
             toolContext: customContext,
+            metrics: new MockMetrics(),
         });
 
         server.registerTools();


src/common/config/userConfig.ts
```diff
@@ -248,21 +248,50 @@ const ServerConfigSchema = z4.object({
...
+            "Deprecated. Use `monitoringServerPort` instead. Port number for the healthCheck HTTP server (only used when transport is 'http'). If provided, `healthCheckHost` must also be set."
         )
         .register(configRegistry, { overrideBehavior: "not-allowed" }),
+    /** @deprecated Use `monitoringServerHost` instead. */
     healthCheckHost: z4
         .string()
         .optional()
         .describe(
-            "Host address to bind the healthCheck HTTP server to (only used when transport is 'http'). If provided, `healthCheckPort` must also be set."
+            "Deprecated. Use `monitoringServerHost` instead. Host address to bind the healthCheck HTTP server to (only used when transport is 'http'). If provided, `healthCheckPort` must also be set."
```
gagik, line 268: I would like to add a proper `deprecated` config for fields which will handle printing messages for deprecated fields later

tests/browser/polyfills/process/index.ts
gagik, line 1: unfortunately prom-client isn't _that_ web compatible

src/common/config/userConfig.ts
himanshusinghs: We do have the deprecation warning logic in `parseUserConfig.ts` for CLI usage. I think for library usage your TSDoc comments will do good?

src/common/metrics/prometheusMetrics.ts
```diff
@@ -0,0 +1,27 @@
...
+
+    constructor({ definitions, registry }: { definitions: TMetrics; registry?: Registry }) {
+        this.registry = registry ?? new Registry();
+        this.definitions = definitions;
+
+        for (const key in this.definitions) {
+            const metric = this.definitions[key];
+            if (metric instanceof Histogram || metric instanceof Counter || metric instanceof Gauge) {
+                this.registry.registerMetric(metric as Metric);
+            }
```
himanshusinghs: I guess we don't really need this check, since the typedef MetricDefinition does not allow a metric to be anything other than Histogram, Counter, Gauge?

And if we would like to keep it for runtime assurance, maybe it would nice to throw an error when metric is not one of the expected instances?

src/tools/tool.ts
```diff
@@ -501,6 +522,17 @@ export abstract class ToolBase<TUserConfig extends UserConfig = UserConfig, TCon
...
+            const durationSeconds = (Date.now() - startTime) / 1000;
+
+            this.metrics
+                .get("toolExecutionDuration")
+                .observe({ tool_name: this.name, category: this.category }, durationSeconds);
+            this.metrics.get("toolExecutionTotal").inc({
+                tool_name: this.name,
+                category: this.category,
+                status: result.isError ? "error" : "success",
+            });
```
himanshusinghs: <3 - Ok so this was missing and instead what I saw was the telemetry capture 🤦

src/tools/tool.ts
```diff
@@ -501,6 +522,17 @@ export abstract class ToolBase<TUserConfig extends UserConfig = UserConfig, TCon
 
             this.emitToolEvent(args, { startTime, result });
 
+            const durationSeconds = (Date.now() - startTime) / 1000;
+
+            this.metrics
+                .get("toolExecutionDuration")
+                .observe({ tool_name: this.name, category: this.category }, durationSeconds);
```
himanshusinghs: Q: Would executionDuration benefit from capturing the success as well?

src/transports/streamableHttp.ts
```diff
@@ -126,50 +141,23 @@ export class StreamableHttpRunner<
...
+            (this.userConfig.monitoringServerPort === undefined)
+        ) {
+            throw new Error(
+                "Both monitoringServerHost and monitoringServerPort must be defined to enable the monitoring server."
+            );
+        }
+
+        const effectivePort = this.userConfig.monitoringServerPort ?? this.userConfig.healthCheckPort;
+        if (effectivePort !== undefined && effectivePort !== 0 && effectivePort === this.userConfig.httpPort) {
+            throw new Error("Monitoring server port cannot be the same as httpPort.");
```
himanshusinghs, line 163: I think we're missing the validation on `healthCheckPort` and `healthCheckHost` both being set because now we only check the same validation for `monitoringServerHost` and `monitoringServerHost`.

Perhaps this was intentional but I think we should keep the validation for both the new config and the old until we remove the old config entirely?

src/transports/streamableHttp.ts
```diff
@@ -585,22 +573,70 @@ class MCPHttpServer<TUserConfig extends UserConfig = UserConfig, TContext = unkn
...
+        if (this.features.includes("metrics") && this.metrics?.getMetrics) {
+            const getMetrics = this.metrics.getMetrics.bind(this.metrics);
+            this.app.get("/metrics", (_req: express.Request, res: express.Response) => {
+                getMetrics()
+                    .then((output) => {
+                        res.set("Content-Type", "text/plain");
+                        res.send(output);
+                    })
+                    .catch((err: unknown) => {
+                        res.status(500).json({ error: err instanceof Error ? err.message : String(err) });
```
himanshusinghs: Might wanna change the handler to async await pattern so we follow the promise handling precedent in the repo.

src/server.ts
```diff
@@ -35,6 +41,7 @@ export interface ServerOptions<TUserConfig extends UserConfig = UserConfig, TCon
     /** @deprecated Will be removed in a future version. Use `SessionOptions.connectionErrorHandler` instead. */
     connectionErrorHandler: ConnectionErrorHandler;
     uiRegistry?: UIRegistry;
+    metrics?: Metrics<TMetrics>;
```
himanshusinghs: I noticed in `base.ts`, we always pass the metrics instance. We can probably mark it as non-optional?

src/transports/streamableHttp.ts
gagik: https://github.com/mongodb-js/mongodb-mcp-server/pull/960/changes#diff-04c64026c9b9da9b7c1848c4f45eaaffe579a97b1b4ae3c1ce9d7a3fe616b897R145 would this not cover it?

src/common/config/userConfig.ts
gagik: ohhhh, good callout. 
deprecation support actually already exists thanks to mongosh library! I literally forgot I added this lol

src/common/metrics/prometheusMetrics.ts
gagik: yeah I think we can leave it up to `registerMetric` to throw here

src/tools/tool.ts
gagik: these 2 metrics are meant to be merged... will adjust

src/transports/streamableHttp.ts
gagik: one of the big reasons I want us to switch out of express is the fact that the handler cannot be async :(
I can wrap it in an async function that is instantly executed

gagik: wait nevermind, I think I was resolving to an older express version somehow

src/transports/streamableHttp.ts
himanshusinghs: The provided link is not directing me to correct code anymore but I suppose you're referring to effectivePort variable? The check that is missing is existence check on both effectivePort and effectiveHost.

himanshusinghs: did not see the unchanged check which covers it already so all good

PR #967, opened 2026-03-11
feat: add Atlas Stream Processing tools MCP-406

jwongmongodb:
## Summary

Adds 4 new MCP tools (`atlas-streams-discover`, `atlas-streams-build`, `atlas-streams-manage`, `atlas-streams-teardown`) covering the full lifecycle of Atlas Stream Processing resources:

- **discover** — list/inspect workspaces, connections, processors; fetch logs and networking; diagnose processor health
- **build** — create workspaces, connections (Kafka, HTTPS, S3, Schema Registry, cluster, etc.), processors with pipeline definitions, and PrivateLink endpoints
- **manage** — start/stop/modify processors, update workspaces (tier, region), update connections, accept/reject VPC peering
- **teardown** — delete processors, connections, workspaces, PrivateLink endpoints, and VPC peering

Includes shared base class (`StreamsToolBase`) with common validation, elicitation flows for destructive confirmations, and reusable argument schemas.

**Companion PR:** This should be used with the agent skill defined in [mongodb/agent-skills#10](https://github.com/mongodb/agent-skills/pull/10), which provides the LLM with streaming-specific guidance for using these tools effectively.

## Test plan

- [x] Unit tests pass (171 tests across 6 suites)
- [x] Integration tests pass (45 tests against live Atlas)
- [x] Type check, lint, format, dependencies all pass
- [ ] CI passes on this PR

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/tools/atlas/streams/build.ts
+++ b/src/tools/atlas/streams/build.ts
@@ -792,8 +792,8 @@ export class StreamsBuildTool extends StreamsToolBase {
         }
 
         const body: Record<string, unknown> = {
-            provider: args.privateLinkProvider,
             ...args.privateLinkConfig,
+            provider: args.privateLinkProvider,
         };
 
         await this.apiClient.createPrivateLinkConnection({
--- a/src/tools/atlas/streams/teardown.ts
+++ b/src/tools/atlas/streams/teardown.ts
@@ -146,9 +146,9 @@ export class StreamsTeardownTool extends StreamsToolBase {
             });
 
             if (referencingProcessors.length > 0) {
-                const names = referencingProcessors.map((p) => p.name).join(", ");
                 const runningOnes = referencingProcessors.filter((p) => p.state === "STARTED");
                 if (runningOnes.length > 0) {
+                    const names = runningOnes.map((p) => p.name).join(", ");
                     return {
                         content: [
                             {
--- a/tests/unit/tools/atlas/streams/build.test.ts
+++ b/tests/unit/tools/atlas/streams/build.test.ts
@@ -779,6 +779,25 @@ describe("StreamsBuildTool", () => {
             });
         });
 
+        it("should not allow privateLinkConfig.provider to overwrite privateLinkProvider", async () => {
+            await exec({
+                ...baseArgs,
+                resource: "privatelink",
+                privateLinkProvider: "AWS",
+                privateLinkConfig: {
+                    provider: "GCP",
+                    region: "us-east-1",
+                },
+            });
+
+            expect(mockApiClient.createPrivateLinkConnection).toHaveBeenCalledWith({
+                params: { path: { groupId: "proj1" } },
+                body: expect.objectContaining({
+                    provider: "AWS",
+                }),
+            });
+        });
+
         it("should throw when privateLinkProvider is missing", async () => {
             await expect(
                 exec({
--- a/tests/unit/tools/atlas/streams/teardown.test.ts
+++ b/tests/unit/tools/atlas/streams/teardown.test.ts
@@ -202,6 +202,36 @@ describe("StreamsTeardownTool", () => {
             expect(mockApiClient.deleteStreamConnection).not.toHaveBeenCalled();
         });
 
+        it("should only name running processors in warning when mix of running and stopped reference connection", async () => {
+            mockApiClient.getStreamProcessors!.mockResolvedValue({
+                results: [
+                    {
+                        name: "running-proc",
+                        state: "STARTED",
+                        pipeline: [{ $source: { connectionName: "conn1" } }],
+                    },
+                    {
+                        name: "stopped-proc",
+                        state: "STOPPED",
+                        pipeline: [{ $emit: { connectionName: "conn1" } }],
+                    },
+                ],
+            });
+
+            const result = await exec({
+                ...baseArgs,
+                resource: "connection",
+                workspaceName: "ws1",
+                resourceName: "conn1",
+            });
+
+            expect(result.isError).toBe(true);
+            const text = (result.content[0] as { text: string }).text;
+            expect(text).toContain("running-proc");
+            expect(text).not.toContain("stopped-proc");
+            expect(mockApiClient.deleteStreamConnection).not.toHaveBeenCalled();
+        });
+
         it("should proceed with deletion when only stopped processors reference connection", async () => {
             mockApiClient.getStreamProcessors!.mockResolvedValue({
                 results: [


src/common/schemas.ts
```diff
@@ -1,4 +1,4 @@
-export const previewFeatureValues = ["search", "mcpUI"] as const;
+export const previewFeatureValues = ["search", "mcpUI", "streams"] as const;
```
jwongmongodb, line 1: this will be removed after the PR to add the Accuracy test is created.

jwongmongodb:
**Question: Integration test scope — is this overkill?** @blva 

The integration tests in this PR cover 40 scenarios including workspace CRUD, connection CRUD (HTTPS, Kafka, Schema Registry), processor lifecycle (create, start, stop, modify, delete), workspace tier updates, networking, logs, and cross-workspace discovery.

For comparison, the [terraform-provider-mongodbatlas](https://github.com/mongodb/terraform-provider-mongodbatlas) repo — which has dedicated CI infrastructure — covers a similar scope across its streams integration tests:
- **Stream instances**: basic CRUD + stream config tiers (~2 tests)
- **Stream connections**: Kafka (plaintext, OAuth, SSL, VPC networking), cluster, sample, HTTPS, Lambda, Schema Registry, PrivateLink (~13 tests)
- **Stream processors**: basic CRUD, tier changes, state transitions, options, cluster type, error cases, timeouts (~10 tests)
- **PrivateLink endpoints**: Confluent, MSK, S3 (~6 tests)

Our MCP server integration tests overlap significantly with that coverage (workspace CRUD, Kafka/SchemaRegistry/HTTPS connections, processor lifecycle, tier changes). A few things we don't cover that the terraform provider does: VPC/PrivateLink networking, Lambda connections, S3 connections, OAuth auth — but those require external cloud infrastructure.

**Questions for the team:**
1. Is the current level of integration test coverage overkill for this repo, or is this the right level of rigor?
2. Should we lean more on unit tests and keep integration tests to a minimal smoke-test set (e.g., just workspace + connection CRUD), given the terraform provider already exercises the Atlas API thoroughly?
3. Are there concerns about integration test runtime/cost with 40 tests hitting live Atlas?

src/common/config/userConfig.ts
```diff
@@ -78,6 +78,8 @@ const ServerConfigSchema = z4.object({
             "drop-collection",
             "delete-many",
             "drop-index",
+            "atlas-streams-manage",
```
jwongmongodb, line 81: this is needed for elicitation to the users that modifies or deletes their streams resources.

jwongmongodb:
@blva I tagged you for an initial review so you can get an early look before my next pass tomorrow morning at this PR. I also have some early questions about the Integration testing philosophy in the above comment.

blva:
> **Question: Integration test scope — is this overkill?** @blva
> 
> The integration tests in this PR cover 40 scenarios including workspace CRUD, connection CRUD (HTTPS, Kafka, Schema Registry), processor lifecycle (create, start, stop, modify, delete), workspace tier updates, networking, logs, and cross-workspace discovery.
> 
> For comparison, the [terraform-provider-mongodbatlas](https://github.com/mongodb/terraform-provider-mongodbatlas) repo — which has dedicated CI infrastructure — covers a similar scope across its streams integration tests:
> 
> * **Stream instances**: basic CRUD + stream config tiers (~2 tests)
> * **Stream connections**: Kafka (plaintext, OAuth, SSL, VPC networking), cluster, sample, HTTPS, Lambda, Schema Registry, PrivateLink (~13 tests)
> * **Stream processors**: basic CRUD, tier changes, state transitions, options, cluster type, error cases, timeouts (~10 tests)
> * **PrivateLink endpoints**: Confluent, MSK, S3 (~6 tests)
> 
> Our MCP server integration tests overlap significantly with that coverage (workspace CRUD, Kafka/SchemaRegistry/HTTPS connections, processor lifecycle, tier changes). A few things we don't cover that the terraform provider does: VPC/PrivateLink networking, Lambda connections, S3 connections, OAuth auth — but those require external cloud infrastructure.
> 
> **Questions for the team:**
> 
> 1. Is the current level of integration test coverage overkill for this repo, or is this the right level of rigor?
> 2. Should we lean more on unit tests and keep integration tests to a minimal smoke-test set (e.g., just workspace + connection CRUD), given the terraform provider already exercises the Atlas API thoroughly?
> 3. Are there concerns about integration test runtime/cost with 40 tests hitting live Atlas?

Hi, this is about testing our integration with the Streams API works well. Terraform uses a different SDK and codebase. We should include int tests in my opinion. I'm OK with a second PR if this one is already too big.

jwongmongodb:
> Hi, this is about testing our integration with the Streams API works well. Terraform uses a different SDK and codebase. We should include int tests in my opinion. I'm OK with a second PR if this one is already too big.

Okay, I'll add those integration tests in the PR Accuracy test as well if that is okay. Since this has the preview flag on it, it should be enough guardrails before the final PR is merged.

tests/integration/tools/mongodb/delete/dropIndex.test.ts
```diff
@@ -293,14 +293,17 @@ describe.each([{ vectorSearchEnabled: false }, { vectorSearchEnabled: true }])(
...
+                        expect(mockElicitInput.mock).toHaveBeenCalledWith(
+                            {
+                                // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+                                message: expect.stringContaining(
+                                    "You are about to drop the index named `year_1` from the `mflix.movies` namespace"
+                                ),
+                                mode: "form",
+                                requestedSchema: Elicitation.CONFIRMATION_SCHEMA,
+                            },
+                            { timeout: 300000 }
```
jwongmongodb, line 305: this is from the elicitation timeout addition

tests/integration/tools/atlas/streams/build.test.ts
```diff
@@ -0,0 +1,163 @@
...
+                expect(content).toContain(srConnName);
+                expect(content).toContain("SchemaRegistry");
+            });
+        });
+
+        // TODO: Add integration tests requiring external infrastructure:
+        // - S3 connection creation (requires AWS IAM role ARN registered in Atlas)
+        // - AWSKinesisDataStreams connection creation (requires AWS IAM role ARN)
+        // - AWSLambda connection creation (requires AWS IAM role ARN)
+        // - PrivateLink creation (requires provider-specific infrastructure)
```
jwongmongodb, line 161: These private link connections can take up to 10 minutes to provision, is it okay to have our integration tests running for that long? For reference, the integration tests in the terraform provider for streams can take up to 35 to 40 minutes.

src/elicitation.ts
```diff
@@ -1,6 +1,12 @@
 import type { ElicitRequestFormParams } from "@modelcontextprotocol/sdk/types.js";
 import type { McpServer } from "@modelcontextprotocol/sdk/server/mcp.js";
 
+export type ElicitedInputResult =
+    | { accepted: true; fields: Record<string, string> }
+    | { accepted: false; fields?: undefined };
+
+const ELICITATION_TIMEOUT_MS = 300_000; // 5 minutes for user interaction
```
jwongmongodb, line 8: The timeout was added to the existing confirmAction method in elicitation.ts — the server.elicitInput() call previously had no timeout, which means it would use the SDK's default (which could hang indefinitely if a client never responds). We added { timeout: 300_000 } (5 minutes) to give users reasonable time to respond while preventing the server from hanging forever.

src/telemetry/types.ts
```diff
@@ -169,6 +170,14 @@ export type PerfAdvisorToolMetadata = AtlasMetadata &
         operations: string[];
     };
 
+export type StreamsToolMetadata = AtlasMetadata & {
+    workspace_name?: string;
+    processor_name?: string;
+    connection_name?: string;
```
blva: could any of this info be PII?

tests/integration/tools/atlas/streams/manage.test.ts
```diff
@@ -0,0 +1,126 @@
...
+                        newTier: "SP30",
+                    },
+                });
+                const content = getResponseContent(response.content);
+                expect(response.isError, `Unexpected error: ${content}`).toBeFalsy();
+                expect(content).toContain("updated");
+            }, 30_000);
+        });
+
+        // TODO: Add integration tests requiring VPC peering infrastructure:
```
jwongmongodb: this will be added once the API spec bug fix is merged and the apiClient.ts is updated.

src/tools/atlas/streams/discover.ts
```diff
@@ -0,0 +1,765 @@
...
+        "Also use for 'why is my processor failing', 'what workspaces do I have', 'show processor stats', or 'check processor health'. " +
+        "Use 'list-workspaces' to see all workspaces in a project. " +
+        "Use inspect actions for details on a specific resource. " +
+        "Use 'diagnose-processor' for a combined health report including state, stats, connection health, and recent errors. " +
+        "Use 'find-processor' to search for a processor by name across all workspaces without knowing which workspace it's in. " +
+        "Use 'list-all-processors' to get all processors across all workspaces in a single call. " +
+        "Use 'get-logs' for operational or audit logs and 'get-networking' for PrivateLink and VPC peering details.";
+
+    public argsShape = {
+        projectId: AtlasArgs.projectId().describe("Atlas project ID. Use `atlas-list-projects` to find project IDs."),
```
blva: I would say "Use `atlas-list-projects` to find project IDs if not available." Usually the clients already have that value in the context and use it correctly  if they are doing other operations in Atlas

src/tools/atlas/streams/streamsToolBase.ts
```diff
@@ -0,0 +1,221 @@
...
+
+export abstract class StreamsToolBase extends AtlasToolBase {
+    protected verifyAllowed(): boolean {
+        if (!this.isFeatureEnabled("streams")) {
+            return false;
+        }
+        return super.verifyAllowed();
+    }
+
+    protected override handleError(
```
blva, line 16: some of the errors here are already handled in `AtlasToolBase`. Please check and make sure you only handle the ones that are unique to the Streams tools.

src/telemetry/types.ts
jwongmongodb: no, these are not PII

jwongmongodb: I did ask on glean

```
By design / schema: these names are just resource identifiers, not “PII fields.”
From a risk / telemetry / AI‑integration perspective: you should assume their values may contain PII (since customers can name things however they like) and handle them under the CSI rules—i.e., don’t send them to third parties or log them unnecessarily without an explicit justification and the right controls.
```

the customer can name it whatever they want, we can remove from telemetry if necessary.

src/tools/atlas/streams/discover.ts
```diff
@@ -0,0 +1,765 @@
...
+                  });
+
+        if (!data) {
+            return {
+                content: [{ type: "text", text: `No logs available for workspace '${workspaceName}'.` }],
+            };
+        }
+
+        try {
+            const { gunzipSync } = await import("node:zlib");
```
blva: why do we need a dynamic import happening on every tool call?

src/tools/atlas/streams/discover.ts
```diff
@@ -0,0 +1,765 @@
...
+            resolvedLogType === "operational"
+                ? await this.apiClient.downloadOperationalLogs({
+                      params: {
+                          path: { groupId: projectId, tenantName: workspaceName },
+                          query: processorName
+                              ? ({ spName: processorName } as { spName?: string; startDate?: number; endDate?: number })
+                              : {},
+                      },
+                      headers: {
+                          Accept: "application/vnd.atlas.2025-03-12+gzip",
```
blva, line 485: we should handle the versions in the api client layer, hardcoding the version in the tool is very error prone. if you extend that functionality, that can be in a new PR so that it's easy to revert if changes separately if any bugs show up.

src/tools/atlas/streams/discover.ts
```diff
@@ -0,0 +1,765 @@
...
+            };
+        }
+
+        // Search in batches of 10, early-exit when found (processor names are unique per project)
+        const batchSize = 10;
+        for (let i = 0; i < workspaceNames.length; i += batchSize) {
+            const batch = workspaceNames.slice(i, i + batchSize);
+            const searchResults = await Promise.allSettled(
+                batch.map(async (wsName) => {
+                    const data = await this.apiClient.getStreamProcessor({
```
blva: do we not have an api for this?

src/telemetry/types.ts
blva: please check w/ product. if you want to be safe let's remove and add after. a good way to understand this -  how will this info help you in product insights?

src/tools/atlas/streams/discover.ts
jwongmongodb: good catch, ill move out the import

src/tools/atlas/streams/discover.ts
jwongmongodb: unfortunately processor names are not unique in the project, different workspaces can have the same processor names. Maybe we can always get the user to include the workspace name or use the one in context to search for the stream processor?

One approach is to remove find-processor as an action and instead require workspaceName when looking up a processor. If the LLM doesn't have one in context, it should call list-workspaces first to get it, that's one API call instead of potentially hundreds.

tests/integration/tools/atlas/streams/build.test.ts
blva: great point! lets have them running in `Code Health Long Running Tests`, performance advisor tests also run there

src/tools/atlas/streams/discover.ts
blva: yep! I think that would work better, make this targeted to `workspaceName`

tests/integration/tools/atlas/streams/manage.test.ts
blva: thanks! could we please add the CLOUPD there as well?

src/telemetry/types.ts
jwongmongodb: I will remove for now, I pinged product about it and we can add it back later in the next PR if necessary.

src/tools/atlas/streams/streamsToolBase.ts
jwongmongodb: This is now done, I moved the common errors out and left the streams specific handling here.

src/tools/atlas/streams/discover.ts
jwongmongodb: this is now updated

tests/integration/tools/atlas/streams/manage.test.ts
jwongmongodb: done

src/tools/atlas/streams/discover.ts
jwongmongodb: I added workspace-name to be required when looking up a stream processor now

src/tools/atlas/streams/discover.ts
jwongmongodb: Agreed, the versions shouldn't be hardcoded in the tool. The top of apiClient.ts (client setup, middleware, etc.) is hand-written, only the wrapper methods below line 210 are auto-generated. So we could add middleware there that sets the right Accept header for gzip endpoints. I'll pull this out into a follow-up PR to keep it easy to revert separately.

src/tools/atlas/streams/discover.ts
jwongmongodb: this is now moved out to the top of the file import

vitest.config.ts
```diff
@@ -13,7 +13,7 @@ const vitestDefaultExcludes = [
 const longRunningTests = ["tests/integration/tools/atlas/performanceAdvisor.test.ts"];
 
 if (process.env.SKIP_ATLAS_TESTS === "true") {
-    vitestDefaultExcludes.push("**/atlas/**");
+    vitestDefaultExcludes.push("**/integration/**/atlas/**");
```
blva, line 16: why this change?

vitest.config.ts
```diff
@@ -13,7 +13,7 @@ const vitestDefaultExcludes = [
 const longRunningTests = ["tests/integration/tools/atlas/performanceAdvisor.test.ts"];
```
blva, line 13: this is where you should add your test as longRunning

README.md
```diff
@@ -323,6 +323,10 @@ npx -y mongodb-mcp-server@latest --transport http --httpHost=0.0.0.0 --httpPort=
...
+- `atlas-streams-discover` - Discover and inspect Atlas Stream Processing resources. Also use for 'why is my processor failing', 'what workspaces do I have', 'show processor stats', or 'check processor health'. Use 'list-workspaces' to see all workspaces in a project. Use inspect actions for details on a specific resource. Use 'diagnose-processor' for a combined health report including state, stats, connection health, and recent errors. Use 'find-processor' to search for a processor by name across all workspaces without knowing which workspace it's in. Use 'list-all-processors' to get all processors across all workspaces in a single call. Use 'get-logs' for operational or audit logs and 'get-networking' for PrivateLink and VPC peering details.
```
blva: do I understand right that we removed `find-processor` and `list-all-processors`? if so we should remove from the readme as well

tests/integration/tools/atlas/streams/build.test.ts
blva: left a comment mentioning where you can add it (https://github.com/mongodb-js/mongodb-mcp-server/pull/967#discussion_r2926952229)

tests/browser/polyfills/zlib/index.ts
```diff
@@ -4,4 +4,7 @@ export function inflate() {
 export function deflate() {
     // noop
 }
-export default { inflate, deflate };
+export function gunzipSync() {
```
blva: why are we touching the browser file?

jwongmongodb: The browser embedding test imports the full server bundle, which pulls in discover.ts. Since gunzipSync is now a top-level static import (rather than a dynamic await import() inside getLogs), esbuild needs to resolve node:zlib at build time.

The config alias routes it to the existing zlib polyfill, and the no-op gunzipSync export satisfies esbuild's named-import validation. Without these two changes the browser test fails — this is the same pattern used for other Node builtins like fs, crypto, stream, etc. in the same config.

The browser test was failing previously [here](https://github.com/mongodb-js/mongodb-mcp-server/actions/runs/23017460988/job/66844474858?pr=967) when I just moved up the import itself.

blva: i see, thanks for clarifying!

README.md
jwongmongodb: This was from a messed up commit squashing, this will be removed

vitest.config.ts
jwongmongodb: The previous glob **/atlas/** was too broad — it also excluded unit tests under tests/unit/tools/atlas/streams/ which are fully mocked and don't need Atlas credentials (this was noticed in the coveralls that appeared suspiciously low after my test additions). Narrowed to **/integration/**/atlas/** so only integration tests requiring a live Atlas connection are skipped. I believe that was the original idea of this flag, to prevent tests that require credentials to be run?

Show this be renamed to SKIP_ATLAS_INTEGRATION_TESTS to be more clear as suggested by copilot?

blva: that works if you validate you're able to run npm `test:local` without any hiccups

jwongmongodb: fixed and verified

vitest.config.ts
blva: @jwongmongodb i think you missed this comment about adding your test to the long running tests?

jwongmongodb: ah, I misunderstood, I thought the new long running test to be added later will be added here, the current tests are quick. Do you want me to move all the tests here right now (even the current quick ones)?

My comment was for the todo tests to come in the next PR (private link etc)

blva: oh, I see! I meant only the integration tests that take some time (e.g. creating takes long, etc)

src/tools/atlas/streams/discover.ts
```diff
@@ -0,0 +1,583 @@
...
+        if (cloudProvider && region) {
+            try {
+                const accountDetails = await this.apiClient.getAccountDetails({
+                    params: {
+                        path: { groupId: projectId },
+                        query: { cloudProvider, regionName: region },
+                    },
+                });
+                sections.push(
+                    `## Account Details (${cloudProvider}/${region})\n${JSON.stringify(accountDetails, null, 2)}`
```
blva, line 562: sorry I missed this, but what is included in account details? we should make sure to flag if there is any sensitive data we should store it in our keychain so that it gets redacted https://github.com/mongodb-js/mongodb-mcp-server/blob/main/src/common/logging/loggerBase.ts#L51

vitest.config.ts
jwongmongodb: yep, I think the current tests are quite fast, so I will have the long running ones here in the next PR. I'll fix the cleanup now

src/tools/atlas/streams/discover.ts
jwongmongodb: The endpoint returns the Account details for the group, region, and provider for the cloud provider (AWS, AZURE and GCP). This is essentially public infrastructure metadata, I think it should be safe, here is the [link](https://www.mongodb.com/docs/api/doc/atlas-admin-api-v2/2025-03-12/operation/operation-getgroupstreamaccountdetails#operation-getgroupstreamaccountdetails-200-body-application-vnd-atlas-2024-11-13-json) to the api spec:
- AWS account IDs are [not considered secret by AWS](https://docs.aws.amazon.com/accounts/latest/reference/manage-acct-identifiers.html)
- Same data is already exposed in Terraform (public registry docs)
- Customers need this data to configure peering from their side
- Role requirement is just GROUP_READ_ONLY - Atlas treats it as non-sensitive

scripts/cleanupAtlasTestLeftovers.test.ts
```diff
@@ -165,38 +165,48 @@ async function main(): Promise<void> {
...
-            });
-            console.log(`Deleted project: ${project.name} (${project.id})`);
-        } catch (error) {
-            const errorStr = String(error);
-            const errorMessage = `Failed to delete project ${project.name} (${project.id}): ${errorStr}`;
-            console.error(errorMessage);
-            allErrors.push(errorMessage);
-        }
-    }
+    const projectsWithIds = testProjects.filter((p): p is Group & { id: string } => !!p.id);
```
jwongmongodb, line 168: Atlas enforces a dependency chain: a project cannot be deleted while it has clusters, and clusters cannot be deleted while a stream workspace holds a connection to them. 

Phase 1 issues all the delete requests (processors → workspaces → clusters) in parallel across projects. Phase 2 waits for clusters to fully terminate before attempting to delete each project — without this wait, the project deletion fails with a 409 because the async cluster teardown hasn't finished yet.

Both phases run all projects in parallel so the total wait time is bounded by the slowest single project, not the sum.

src/tools/atlas/tools.ts
```diff
@@ -11,3 +11,7 @@ export { ListOrganizationsTool } from "./read/listOrgs.js";
 export { ConnectClusterTool } from "./connect/connectCluster.js";
 export { ListAlertsTool } from "./read/listAlerts.js";
 export { GetPerformanceAdvisorTool } from "./read/getPerformanceAdvisor.js";
+export { StreamsDiscoverTool } from "./streams/discover.js";
+export { StreamsBuildTool } from "./streams/build.js";
+export { StreamsManageTool } from "./streams/manage.js";
+export { StreamsTeardownTool } from "./streams/teardown.js";
```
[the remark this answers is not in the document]
jwongmongodb: This will be added as a separate accuracy PR following this.

src/tools/atlas/streams/discover.ts
```diff
@@ -0,0 +1,583 @@
...
+
+        try {
+            const buffer = Buffer.from(data as unknown as ArrayBuffer);
+            const decompressed = gunzipSync(buffer).toString("utf-8");
+
+            // Limit output to avoid overwhelming the context
+            const lines = decompressed.split("\n").filter((line) => line.trim());
+            const maxLines = 100;
+            const truncated = lines.length > maxLines;
+            const output = lines.slice(-maxLines).join("\n");
```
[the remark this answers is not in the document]
jwongmongodb: gunzipSync on a bounded API response is fine in practice, the MCP server is single-client, the Atlas API caps log payloads, and the output is truncated to 100 lines regardless. The async swap is a marginal improvement for a scenario that's unlikely to cause real issues.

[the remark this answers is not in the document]
blva: wdym by single-client? MCP handles multiple connections via streamable HTTP, there are use-cases where users deploy it in hosts and have multiple clients connecting.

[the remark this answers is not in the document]
jwongmongodb: You're right, that was an incorrect assumption on my part. I was thinking in terms of stdio transport (which is single-client), but with streamable HTTP the server handles multiple concurrent connections. I'll swap to async gunzip to avoid blocking the event loop. (I'll push the fix tmr morning my time.)

src/tools/atlas/streams/build.ts
```diff
@@ -0,0 +1,817 @@
...
+        if (!args.privateLinkConfig) {
+            throw new Error(
+                "privateLinkConfig is required. Provide provider-specific configuration " +
+                    "(AWS: {region, vendor, arn, dnsDomain, dnsSubDomain}, Azure: {region, serviceEndpointId}, GCP: {region, gcpServiceAttachmentUris})."
+            );
+        }
+
+        const body: Record<string, unknown> = {
+            provider: args.privateLinkProvider,
+            ...args.privateLinkConfig,
```
cveticm: Since `privateLinkConfig` has no rigid shape (`z.record(z.unknown())`), it is possible for the LLM to provide something like `privateLinkConfig: { provider: "GCP", region: "..." }`. 
Since `privateLinkConfig` is spread after `provider` is set, `args.privateLinkProvider` can be silently overwritten by `privateLinkConfig.provider`.

For extra safety, I recommend reversing the spread order:

```suggestion
            ...args.privateLinkConfig,
            provider: args.privateLinkProvider,
```

src/tools/atlas/streams/teardown.ts
```diff
@@ -0,0 +1,261 @@
...
+            if (referencingProcessors.length > 0) {
+                const names = referencingProcessors.map((p) => p.name).join(", ");
+                const runningOnes = referencingProcessors.filter((p) => p.state === "STARTED");
+                if (runningOnes.length > 0) {
+                    return {
+                        content: [
+                            {
+                                type: "text",
+                                text:
+                                    `Warning: Connection '${name}' is referenced by running processor(s): ${names}. ` +
```
cveticm: `const names = referencingProcessors.map((p) => p.name).join(", ");`
this will list all processes, not only running processes

```suggestion
                const runningOnes = referencingProcessors.filter((p) => p.state === "STARTED");
                if (runningOnes.length > 0) {
                    const names = runningOnes.map((p) => p.name).join(", ");
                    return {
                        content: [
                            {
                                type: "text",
                                text:
                                    `Warning: Connection '${name}' is referenced by running processor(s): ${names}. ` +
```

jwongmongodb: ahh, good catch! I'll fix it with everything else from your review when you are done.

src/tools/atlas/streams/build.ts
jwongmongodb: this is now fixed, I wrote a test to verify before and after the change.

src/tools/atlas/streams/teardown.ts
jwongmongodb: this is now fixed, I wrote a test to verify before and after the change.

cveticm (approved):
LGTM! Thanks for the changes!

PR #970, opened 2026-03-12
Deploying MongoDB MCP Server to AWS Bedrock AgentCore

ialek36:
## Proposed changes
A docker image and instructions for deploying the MCP Server to AWS Bedrock AgentCore
## Checklist

- [ x] I have signed the [MongoDB CLA](https://www.mongodb.com/legal/contributor-agreement)

### DIFF
--- a/deploy/aws/Dockerfile
+++ b/deploy/aws/Dockerfile
@@ -5,16 +5,17 @@ ARG VERSION=latest
 # Create user and group
 RUN addgroup -S mcp && adduser -S mcp -G mcp
 
-
 # Install production dependencies only
 RUN npm install -g mongodb-mcp-server@latest
-USER mcp
 WORKDIR /home/mcp
 
 ENV MDB_MCP_LOGGERS=stderr,mcp
 ENV MDB_MCP_DISABLED_TOOLS=atlas-local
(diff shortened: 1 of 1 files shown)


deploy/aws/Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+LABEL maintainer="MongoDB Inc <info@mongodb.com>"
+LABEL description="MongoDB MCP Server"
+LABEL version=${VERSION}
+LABEL io.modelcontextprotocol.server.name="io.github.mongodb-js/mongodb-mcp-server"
+# Signal that this is running in Docker for host binding logic
+ENV DOCKER_CONTAINER=1
+
+EXPOSE 8000
+
+ENTRYPOINT ["mongodb-mcp-server", "--transport", "http", "--httpHost", "0.0.0.0", "--httpPort", "8000", "--loggers", "stderr"]
```
nirinchev: There's no need to repeat the arguments here. We should either use env variables or cli args. Since loggers is already set, the equivalent here would be:

```
ENV MDB_MCP_TRANSPORT=http
ENV MDB_MCP_HTTP_HOST=0.0.0.0
ENV MDB_MCP_HTTP_PORT=8000
```

ialek36: fixed by removing redundant settings

PR #979, opened 2026-03-14
fix: handle closing stdio transport

### LINKED ISSUE #974: [Bug]: Using with Langchain MultiServerMCPClient aync contextmanager does not exit
### Version

latest

### App

- [ ] Cursor
- [ ] Windsurf
- [ ] VSCode
- [ ] VSCode Insiders
- [ ] Claude Desktop
- [x] Other

- [ ] Claude 3.5 Sonnet
- [ ] Claude 3.7 Sonnet
- [ ] GPT-4a
- [ ] o4-mini
- [x] Other

### Bug Description

I am using following python script

```python

import asyncio
from langchain_mcp_adapters.client import MultiServerMCPClient
from langchain_mcp_adapters.tools import load_mcp_tools

async def main():
    client = MultiServerMCPClient(
        {
            "mongodb": {
                "transport": "stdio",
                "command": "npx",
                "args": [
                    "-y",
                    "mongodb-mcp-server@latest",
                    "--readOnly"
                ],
                "env": {
                    "MDB_MCP_CONNECTION_STRING": "mongodb://localhost:27017"
                }
            }
        }
    )

    async with client.session("mongodb") as session:
        tools = await load_mcp_tools(session)
        print(f"Loaded {len(tools)} tools:")
        for tool in tools:
            print(f"  - {tool.name}")


asyncio.run(main())
```

### Output

```bash
Loaded 17 tools:
  - aggregate
  - collection-indexes
  - collection-schema
  - collection-storage-size
  - connect
  - count
  - db-stats
  - explain
  - export
  - find
  - list-collections
  - list-databases
  - mongodb-logs
  - atlas-local-connect-deployment
  - atlas-local-list-deployments
  - list-knowledge-sources
  - search-knowledge

```

And it is stuck here.

I am using
- `Python 3.12.11`
- `langchain-mcp-adapters == 0.2.1`

shrutimantri:
## Proposed changes

This change handles closing the stdio transport.
Fixes #974 

## Checklist

### DIFF
--- a/src/transports/stdio.ts
+++ b/src/transports/stdio.ts
@@ -26,6 +26,10 @@ export class StdioRunner<TUserConfig extends UserConfig = UserConfig, TContext =
         try {
             this.server = await this.createServer({ serverOptions, sessionOptions });
             const transport = new StdioServerTransport();
+            transport.onclose = async () => {
+                await this.server?.close();
+                process.exit(0);
+            };
 
             await this.server.connect(transport);
         } catch (error: unknown) {


himanshusinghs:
Hey @shrutimantri! Thanks for raising this PR.

I would like to mention though that the required wiring to close the Server instance on transport closure is already in place. In index.ts, which is the main entry point for starting the STDIO runner there is a [shutdown handler](https://github.com/mongodb-js/mongodb-mcp-server/blob/0976994727b82e30f3d8c57b7223440fbd4e1d35/src/index.ts#L92-L117) which runs on different shutdown events.


Shutdown event -> calls -> Shutdown handler -> calls -> [TransportRunner.close in base.ts](https://github.com/mongodb-js/mongodb-mcp-server/blob/0976994727b82e30f3d8c57b7223440fbd4e1d35/src/transports/base.ts#L386) -> calls -> [StdioRunner.closeTransport in stdio.ts](https://github.com/mongodb-js/mongodb-mcp-server/blob/0976994727b82e30f3d8c57b7223440fbd4e1d35/src/transports/stdio.ts#L41) -> calls -> Server.close


The linked issue has a different problem that the provided code is not triggering a client close on the Client transport. That needs to happen for the client script to actually exit.

shrutimantri:
@himanshusinghs The provided code in the linked issue is triggering the close as the code is encapsulated in the `with` section.

himanshusinghs:
Thanks for the insight @shrutimantri! Even if that is the case - I still fear the current PR is not the way to go because with the change, we will essentially end up invoking server.close and process.exit twice.

Can you provide more info on how the `with` works internally? Does it invoke client.close / client.disconnect (methods that actually close the MCP client transport) or not?

himanshusinghs:
Hello again! I verified the behavior on Langchain's TS SDK and at-least there it seems closing the client explicitly is essential. [Ref](https://github.com/himanshusinghs/langchain-test)

I am not sure how Python SDK implements the binding handling of `with` statement but I would recommend trying the explicit client closure first and that should definitely help with the problem you're facing.

PR #983, opened 2026-03-16
feat: add more metrics MCP-411

gagik:
Adds toolExecutionErrors, sessionCreated, sessionClosed.

### DIFF
--- a/src/common/sessionStore.ts
+++ b/src/common/sessionStore.ts
@@ -88,7 +88,7 @@ export class SessionStore<T extends CloseableTransport = CloseableTransport> {
                     message: "Session closed due to inactivity",
                 });
 
-                await this.closeSession(sessionId, { reason: "idle_timeout" });
+                await this.closeSession({ sessionId, reason: "idle_timeout" });
             }
         }, this.idleTimeoutMS);
         const notificationTimeout = setManagedTimeout(
@@ -104,10 +104,13 @@ export class SessionStore<T extends CloseableTransport = CloseableTransport> {
         this.metrics.get("sessionCreated").inc();
     }
 
-    async closeSession(
-        sessionId: string,
-        { reason }: { reason: SessionCloseReason } = { reason: "unknown" }
-    ): Promise<void> {
+    async closeSession({
+        sessionId,
+        reason = "unknown",
+    }: {
+        sessionId: string;
+        reason?: SessionCloseReason;
+    }): Promise<void> {
         const session = this.sessions[sessionId];
         if (!session) {
             throw new Error(`Session ${sessionId} not found`);
@@ -139,7 +142,7 @@ export class SessionStore<T extends CloseableTransport = CloseableTransport> {
 
     async closeAllSessions(): Promise<void> {
         await Promise.all(
-            Object.keys(this.sessions).map((sessionId) => this.closeSession(sessionId, { reason: "server_stop" }))
+            Object.keys(this.sessions).map((sessionId) => this.closeSession({ sessionId, reason: "server_stop" }))
         );
     }
 }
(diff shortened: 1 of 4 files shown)


src/common/sessionStore.ts
```diff
@@ -96,30 +101,38 @@ export class SessionStore<T extends CloseableTransport = CloseableTransport> {
...
+    async closeSession({ sessionId, reason }: { sessionId: string; reason: SessionCloseReason }): Promise<void> {
         const session = this.sessions[sessionId];
         if (!session) {
             throw new Error(`Session ${sessionId} not found`);
         }
+
+        // Remove from map before closing transport to prevent double-counting:
+        // transport.close() can trigger onsessionclosed which re-enters closeSession.
+        delete this.sessions[sessionId];
+
```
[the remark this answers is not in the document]
gagik: I think throwing can be good for making sure we're not accidentally double closing sessions in an unexpected way

src/common/sessionStore.ts
```diff
@@ -96,16 +101,28 @@ export class SessionStore<T extends CloseableTransport = CloseableTransport> {
             notificationTimeout,
             logger,
         };
+        this.metrics.get("sessionCreated").inc();
     }
 
-    async closeSession(sessionId: string, closeTransport: boolean = true): Promise<void> {
+    async closeSession(sessionId: string, { reason }: { reason: SessionCloseReason }): Promise<void> {
```
gagik: `closeTransport` wasn't being used anywhere so I don't really consider this a breaking change

src/common/sessionStore.ts
```diff
@@ -96,16 +101,28 @@ export class SessionStore<T extends CloseableTransport = CloseableTransport> {
             notificationTimeout,
             logger,
         };
+        this.metrics.get("sessionCreated").inc();
     }
 
-    async closeSession(sessionId: string, closeTransport: boolean = true): Promise<void> {
+    async closeSession(
+        sessionId: string,
+        { reason }: { reason: SessionCloseReason } = { reason: "unknown" }
```
himanshusinghs: From my understanding of poking around this code - the other reason could only be timedout. Did you come across some other flow as well?

src/common/metrics/metricDefinitions.ts
```diff
@@ -17,6 +17,23 @@ export function createDefaultMetrics() {
             buckets: [0.001, 0.005, 0.01, 0.05, 0.1, 0.5, 1, 2.5, 5, 10],
             registers: [],
         }),
+        toolExecutionErrors: new Counter({
+            name: "mcp_tool_execution_errors_total",
+            help: "Total tool execution errors",
+            labelNames: ["error_type"] as const,
+            registers: [],
+        }),
```
himanshusinghs: I know that we mentioned this in the TD but shouldn't we able to derive this from the existing histogram metric that captures tool execution?

src/common/sessionStore.ts
himanshusinghs: It was used internally only but regardless of that there was so far no way for someone to access or provide SessionStore instance to the library interface.

src/common/sessionStore.ts
gagik:
```ts
export type SessionCloseReason = "idle_timeout" | "transport_closed" | "server_stop" | "unknown";
```

src/common/metrics/metricDefinitions.ts
gagik: yeah... I was thinking maybe separation would be good after all but you're right. I might add error_type as a label instead

src/common/sessionStore.ts
himanshusinghs: Yep yep - saw this and I mean to say that there could be "idle_timeout" and "transport_closed" only.

himanshusinghs: Just noticed that `server_stopped` is legit server exiting. What is unknown for?

himanshusinghs: Checked again your changes - it seems you only call closeSession with known values, the path leading to unknown does not seem to occur ever.

gagik: point of `unknown` was to not break the API for those who don't specify `sessionId`. this sessionstore is currently exported in `lib.ts` but if it's a mistake, I can remove it from `lib.ts` and make this a requried object argument

PR #985, opened 2026-03-16
fix: improve streams tool schemas and add accuracy tests

jwongmongodb:
## Summary
- Replace opaque `z.record(z.unknown())` with typed Zod schemas for `connectionConfig` and `privateLinkConfig` so LLMs see actual field names in the tool schema
- Add 60 accuracy tests for all four Atlas Streams tools
- Score extra params as 0.75 and add multi-turn + systemPrompt support to the accuracy test agent

## Tool changes
Accuracy runs showed LLMs placing config fields (`bootstrapServers`, `clusterName`, etc.) at the top level instead of inside `connectionConfig`. With `z.record(z.unknown())`, the JSON Schema sent to the LLM is just `{"type": "object"}` — no field names visible. Typed Zod schemas fix this at the source. Tool descriptions were also improved: shorter (schemas are now self-documenting), clearer nesting guidance for `privateLinkProvider`/`privateLinkConfig`, and Accept headers moved from discover.ts into apiClient.ts.

## Accuracy results
56/60 streams tests at 100%, 94.6% streams-specific accuracy. The 3 failures at 0% are LLM flakiness (no tool calls, incomplete multi-step reasoning, omitted required field), not test design issues.

## Test plan
- [x] `pnpm run check:types`, `check:lint`, `check:format` pass
- [x] `pnpm test` 
- [x] Accuracy run validates streams tool-calling behavior

### DIFF
--- a/src/tools/atlas/streams/build.ts
+++ b/src/tools/atlas/streams/build.ts
@@ -281,7 +281,7 @@ export class StreamsBuildTool extends StreamsToolBase {
             );
         }
 
-        const config: Record<string, unknown> = { ...ConnectionConfig.parse(args.connectionConfig ?? {}) };
+        const config = { ...ConnectionConfig.parse(args.connectionConfig ?? {}) };
 
         const missingInfo = await this.normalizeAndValidateConnectionConfig(config, args.connectionType);
         if (missingInfo) {
--- a/tests/accuracy/sdk/matcher.ts
+++ b/tests/accuracy/sdk/matcher.ts
@@ -23,7 +23,7 @@ export abstract class Matcher {
     /**
      * Asserts that a parameter must NOT be present (i.e., is `undefined`).
      * Use this in test definitions to strictly forbid an extra param that
-     * would otherwise be silently tolerated at a 0.75 score.
+     * would otherwise cause the score to drop to 0.
      *
      * Example — ensure the model does NOT send a `limit` param:
      * ```
@@ -232,11 +232,10 @@ class ValueMatcher extends Matcher {
             const actualKeys = Object.keys(actual);
 
             if (actualKeys.length > expectedKeys.length) {
-                // The model provided more keys than expected — cap at 0.75 (additional parameters)
-                // rather than failing outright. This matches the documented scoring behavior:
-                // 0 = missing/incorrect, 0.75 = additional params, 1 = perfect match.
-                // Use `param: Matcher.undefined` to explicitly assert a param should NOT be present.
-                currentScore = Math.min(currentScore, 0.75);
+                // The model provided more keys than expected - this should not happen.
+                // If we want to allow some extra keys, we should specify that in the test definition
+                // by adding matchers for those keys.
+                return 0;
             }
 
             for (const key of expectedKeys) {
--- a/tests/unit/accuracyScorer.test.ts
+++ b/tests/unit/accuracyScorer.test.ts
@@ -70,7 +70,7 @@ describe("calculateToolCallingAccuracy", () => {
     });
 
     describe("additional parameters", () => {
-        it("should return 0.75 when tool call has additional nested parameters (default behavior)", () => {
+        it("should return 0 when tool call has additional nested parameters (default behavior)", () => {
             const expected: ExpectedToolCall[] = [
                 { toolName: "find", parameters: { db: "test", collection: "users", filter: { status: "active" } } },
             ];
@@ -87,7 +87,7 @@ describe("calculateToolCallingAccuracy", () => {
                 },
             ];
             const result = calculateToolCallingAccuracy(expected, actual);
-            expect(result).toBe(0.75);
+            expect(result).toBe(0);
         });
 
         it("should return 1 when expected has no filter but actual has empty filter", () => {
@@ -167,7 +167,7 @@ describe("calculateToolCallingAccuracy", () => {
             expect(result).toBe(0);
         });
 
-        it("should return 0.75 when there are additional nested fields", () => {
+        it("should return 0 when there are additional nested fields", () => {
             const expected: ExpectedToolCall[] = [
                 {
                     toolName: "find",
@@ -190,7 +190,7 @@ describe("calculateToolCallingAccuracy", () => {
                 },
             ];
             const result = calculateToolCallingAccuracy(expected, actual);
-            expect(result).toBe(0.75);
+            expect(result).toBe(0);
         });
 
         it("should return 1 when ignored additional fields are provided", () => {


src/common/atlas/apiClient.ts
```diff
@@ -618,7 +618,7 @@ export class ApiClient {
     async downloadAuditLogs(options: FetchOptions<operations["downloadGroupStreamAuditLogs"]>) {
         const { data, error, response } = await this.client.GET(
             "/api/atlas/v2/groups/{groupId}/streams/{tenantName}/auditLogs",
-            options
+            { ...options, headers: { Accept: "application/vnd.atlas.2023-02-01+gzip", ...options?.headers } }
```
jwongmongodb, line 621: these headers are now properly autogenerated from apply.ts

tests/accuracy/sdk/matcher.ts
```diff
@@ -222,10 +222,11 @@ class ValueMatcher extends Matcher {
             const actualKeys = Object.keys(actual);
 
             if (actualKeys.length > expectedKeys.length) {
-                // The model provided more keys than expected - this should not happen.
-                // If we want to allow some extra keys, we should specify that in the test definition
-                // by adding matchers for those keys.
-                return 0;
```
jwongmongodb, line 228: The ValueMatcher was returning 0 (wrong/missing) when the model sent extra object keys, but the https://github.com/mongodb-js/mongodb-mcp-server/blob/1cf6f6dbed304379efe2a69791e659e34c0c2d9a/tests/accuracy/sdk/accuracyScorer.ts#L47-L50 says additional parameters should score 0.75 (partial credit). This mismatch meant a tool call with all the right params plus one harmless extra got scored identically to calling a completely wrong tool.

This came up with the streams tools — the flat schema exposes workspace-level params for all resource types, so the LLM fills them in even when they're irrelevant. The old scorer treated that as a total failure.

One trade-off: extra params are now tolerated by default. For cases where a specific extra param is harmful, test authors can use Matcher.undefined to explicitly forbid it. Added a JSDoc note to make that discoverable.

Are we okay with this, or would we rather keep the strict behavior and have test authors opt into tolerance with explicit matchers for extra keys?

jwongmongodb:
Accuracy results ([link](https://github.com/mongodb-js/mongodb-mcp-server/actions/runs/23223447152)): 57/60 streams tests got 100% pass scores

Overall accuracy is reported as 86.5%, but that blends streamstests with pre-existing failures in other categories (11 atlas-local-create-deployment tests fail due to an extra imageTag param, etc.). Streams-specific accuracy is 95.0%.

Should we have per-category reporting would make regressions easier to catch? Right now a streams regression can be hidden by improvements elsewhere. This is a follow-up improvement to the shared accuracy infrastructure.

src/tools/atlas/streams/build.ts
```diff
@@ -155,8 +155,9 @@ export class StreamsBuildTool extends StreamsToolBase {
             .record(z.unknown())
             .optional()
             .describe(
-                "Type-specific connection configuration. Typically required for non-Sample connections when resource='connection'. " +
-                    "For connectionType='Sample', no config is needed. You may also provide an empty or partial config; missing fields can be filled via elicitation. " +
+                "Type-specific connection configuration. Only for resource='connection'. " +
```
jwongmongodb: these tool describe changes came from accuracy test tweaks.

Accuracy tests showed the LLM would hallucinate connection config fields for Sample connections or pass full configs when elicitation could handle it. Clarified that Sample needs no config, and that passing {} with elicitation is the preferred pattern when the user hasn't provided details.

src/tools/atlas/streams/build.ts
```diff
@@ -179,6 +180,7 @@ export class StreamsBuildTool extends StreamsToolBase {
                     "Use $merge to write to Atlas cluster collections: {$merge: {into: {connectionName, db, coll}}}. " +
                     "Use $emit to write to Kafka or Kinesis sinks: {$emit: {connectionName, topic}}. $emit only works with Kafka/Kinesis connections — do NOT use $emit with Https connections. " +
                     "Use $https to POST data to an Https connection: {$https: {connectionName}}. " +
+                    "Use $externalFunction for Lambda: {$externalFunction: {connectionName, functionName, execution: 'async', as: 'result'}}. Lambda does NOT use $emit — use $externalFunction with execution='async' as a terminal stage or execution='sync' for mid-pipeline enrichment. " +
```
jwongmongodb, line 173: Lambda was undocumented in the pipeline description — the LLM would try to use $emit for Lambda connections, which doesn't work. Added the correct $externalFunction stage syntax.

src/tools/atlas/streams/build.ts
```diff
@@ -193,24 +195,34 @@ export class StreamsBuildTool extends StreamsToolBase {
...
+            .describe(
+                "Dead letter queue configuration. Only for resource='processor'. " +
+                    "Only include when the user explicitly requests a DLQ, or when the pipeline uses $https with default onError='dlq'. " +
+                    "The DLQ connection must already exist in the workspace."
+            ),
         autoStart: z
             .boolean()
             .optional()
-            .describe("Start the processor immediately after creation. Default: false. Only for resource='processor'."),
+            .describe(
```
jwongmongodb, line 196: Same over-eagerness issue, the LLM would default to autoStart: true unprompted. Added "omit unless explicitly asked."

src/tools/atlas/streams/build.ts
```diff
@@ -193,24 +195,34 @@ export class StreamsBuildTool extends StreamsToolBase {
                 coll: z.string().describe("Collection name for DLQ documents"),
             })
             .optional()
-            .describe("Dead letter queue configuration. Recommended for resource='processor'."),
+            .describe(
```
jwongmongodb, line 188: The LLM would proactively add a DLQ to every processor even when not requested. Tightened to "only when explicitly requested or when pipeline uses $https with default onError='dlq'."

src/tools/atlas/streams/discover.ts
```diff
@@ -34,7 +34,7 @@ export class StreamsDiscoverTool extends StreamsToolBase {
         "Use 'list-workspaces' to see all workspaces in a project. " +
         "Use inspect actions for details on a specific resource. " +
         "Use 'diagnose-processor' for a combined health report including state, stats, connection health, and recent errors. " +
-        "Use 'get-logs' for operational or audit logs and 'get-networking' for PrivateLink and account details.";
+        "Use 'get-logs' for workspace-level operational or audit logs (optionally filter by processor with resourceName) and 'get-networking' for PrivateLink and account details.";
```
jwongmongodb, line 37: The LLM didn't know logs are workspace-level and would ask for a processor name as a required field. Clarified that get-logs is workspace-scoped with optional resourceName filtering.

src/tools/atlas/streams/discover.ts
```diff
@@ -81,7 +81,8 @@ export class StreamsDiscoverTool extends StreamsToolBase {
             .optional()
             .describe(
                 "Type of logs to retrieve. 'operational' returns runtime logs (errors, Kafka failures, schema issues). " +
-                    "'audit' returns lifecycle events (start/stop). Default: 'operational'. Only for 'get-logs' action."
+                    "'audit' returns lifecycle events (start/stop). Default: 'operational'. Only for 'get-logs' action. " +
```
jwongmongodb, line 84: Reinforces the workspace-level point at the parameter level, not just the tool description

src/tools/atlas/streams/discover.ts
```diff
@@ -481,9 +482,6 @@ export class StreamsDiscoverTool extends StreamsToolBase {
                               ? ({ spName: processorName } as { spName?: string; startDate?: number; endDate?: number })
                               : {},
                       },
-                      headers: {
```
jwongmongodb, line 484: removed the hardcoding to be set at the apiClient.ts level (autogenerated)

src/tools/atlas/streams/manage.ts
```diff
@@ -31,8 +31,15 @@ export class StreamsManageTool extends StreamsToolBase {
         ),
         workspaceName: StreamsArgs.workspaceName().describe("Workspace name containing the resource to manage."),
         action: ManageAction.describe(
-            "Action to perform. Processor must be stopped before 'modify-processor'. " +
-                "Use 'start-processor' to begin or resume processing, 'stop-processor' to pause."
+            "Action to perform. One of: " +
```
jwongmongodb, line 34: The original description only mentioned 3 of 7 actions and didn't list per-action requirements. The LLM would pick wrong actions or omit required fields. Enumerated all 7 actions with their required parameters.

src/tools/atlas/streams/manage.ts
```diff
@@ -65,7 +72,9 @@ export class StreamsManageTool extends StreamsToolBase {
             .array(z.record(z.unknown()))
             .optional()
             .describe(
-                "New aggregation pipeline. Only for 'modify-processor'. Processor must be stopped first. " +
+                "New aggregation pipeline stages as an array of objects, e.g. " +
```
jwongmongodb: The LLM would pass pipeline stages in wrong formats. Added a concrete example showing the $source → transform → $merge pattern.

tests/accuracy/sdk/agent.ts
```diff
@@ -1,6 +1,7 @@
 import type { LanguageModel } from "ai";
 import type { experimental_createMCPClient } from "@ai-sdk/mcp";
 import { stepCountIs, generateText } from "ai";
+import type { ModelMessage } from "ai";
 import type { Model } from "./models.js";
 
 const systemPrompt = [
```
jwongmongodb, line 7: These changes were to add the following:

1. Multi-turn prompt support. The agent accepted string[] prompts but passed each one independently via the prompt parameter with no memory of prior turns. Switched to messages with a cumulative conversationHistory so multi-turn tests (e.g., "stop the processor" → "now change the pipeline" → "start it again") worked as actual conversations. (Since not all prompts are 1 back and forth and may involve mult-turn loops)

2. Per-test system prompt. Added additionalSystemPrompt so individual tests can inject extra context (e.g., project/workspace setup for cross-tool ambiguity tests) without modifying the shared agent-level system prompt.

src/tools/atlas/streams/streamsArgs.ts
```diff
@@ -3,6 +3,103 @@ import { z } from "zod";
 const ALLOWED_STREAMS_NAME_REGEX = /^[a-zA-Z0-9_-]+$/;
 const ALLOWED_STREAMS_NAME_ERROR = "Name can only contain ASCII letters, numbers, hyphens, and underscores";
 
+/** Typed schema for connectionConfig — all fields optional to support elicitation of partial configs. */
+export const ConnectionConfig = z
```
jwongmongodb, line 7: Accuracy runs showed LLMs placing config fields (bootstrapServers, clusterName, etc.) at the top level instead of nesting them inside connectionConfig, the LLM couldn't see field names in an opaque z.record(z.unknown()). With typed schemas, the JSON Schema sent to the LLM now lists actual field names under connectionConfig, eliminating the ambiguity. All fields are optional to support elicitation of partial configs, and .passthrough() preserves backwards compatibility. Schemas sourced from the Atlas Stream Processing API docs.

src/tools/atlas/streams/build.ts
```diff
@@ -151,20 +151,11 @@ export class StreamsBuildTool extends StreamsToolBase {
                 "SchemaRegistry: needs provider, schemaRegistryUrls, and authentication config. " +
                 "Sample: provides sample data for testing (no config needed)."
         ),
-        connectionConfig: z
```
jwongmongodb, line 154: Description shortened — the typed schema is now self-documenting. Previously the description listed all 7 connection type schemas inline as text, which the LLM would read but then place at the wrong nesting level.

src/tools/atlas/streams/build.ts
```diff
@@ -193,25 +185,29 @@ export class StreamsBuildTool extends StreamsToolBase {
...
-            .describe("Start the processor immediately after creation. Default: false. Only for resource='processor'."),
+            .describe(
+                "Start the processor immediately after creation. Default: false. Only for resource='processor'. " +
+                    "Omit unless the user explicitly asks to start the processor right away."
+            ),
 
         // PrivateLink fields
         privateLinkProvider: CloudProvider.optional().describe(
-            "Cloud provider for PrivateLink. Required when resource='privatelink'."
+            "Cloud provider for PrivateLink — must be a plain string, exactly one of: 'AWS', 'AZURE', or 'GCP'. " +
```
jwongmongodb: Strengthened wording ("must be a plain string", "Do NOT pass an object") because accuracy runs showed the LLM passing the entire config object into this enum field instead of using the separate privateLinkConfig parameter.

tests/accuracy/sdk/agent.ts
```diff
@@ -65,15 +67,22 @@ export function getVercelToolCallingAgent(
                 },
             };
 
+            const conversationHistory: ModelMessage[] = [];
```
jwongmongodb, line 70: Multi-turn prompt support for accuracy tests. Switched from prompt (isolated per-turn) to messages with cumulative conversation history so multi-turn tests work as actual conversations. Also added additionalSystemPrompt for per-test context injection.

tests/accuracy/streamsBuild.test.ts
```diff
@@ -0,0 +1,479 @@
...
+                ...optionalWorkspaceDiscovery,
+                {
+                    toolName: "atlas-streams-build",
+                    parameters: {
+                        projectId,
+                        resource: "connection",
+                        workspaceName,
+                        connectionName: "events",
+                        connectionType: "Kafka",
+                        connectionConfig: {
```
jwongmongodb, line 112: Config matchers validate actual field structure (bootstrapServers, authentication.mechanism, aws.roleArn, etc.) rather than Matcher.anyValue. Catches LLM mistakes like omitting config or using wrong field names. Schemas verified against the Atlas Stream Processing connection-configs reference.

tests/accuracy/streamsDiscover.test.ts
```diff
@@ -0,0 +1,420 @@
...
+                        ...optionalDiscoverParams,
+                        projectId,
+                        action: "list-workspaces",
+                    },
+                },
+            ],
+            mockedTools,
+        },
+        {
+            prompt: [
```
jwongmongodb, line 237: multi-turn prompt example

src/tools/atlas/streams/build.ts
blva: I'm confused, can't you define this a a string? why do we need to ask the LLM to not pass an object

jwongmongodb: This is from this accuracy run [here](https://github.com/mongodb-js/mongodb-mcp-server/actions/runs/23220401848).

The prompt was:
"Set up an AWS PrivateLink connection for my streams project with ARN arn:aws:vpce:us-east-1:123456789012:vpc-endpoint/vpce-abc123 and DNS domain streaming.example.com"

What the test expects, two separate fields:                                                                                                                                                                     
```
  {                                                                                                                                                                                                                
    "privateLinkProvider": "AWS",                                                                                                                                                                                  
    "privateLinkConfig": { "arn": "...", "dnsDomain": "..." }                                                                                                                                                      
  }                                                                                                                                                                                                                
```

  What gpt-4o actually sent (test-summary-10, before the description change):                                                                                                                                      
```                                                                                                                                                                                                                   
  {                                                                                                                                                                                                                
    "privateLinkProvider": {                                                
      "endpointServiceId": "arn:aws:...",                                                                                                                                                                          
      "dnsDomain": "streaming.example.com"                                                                                                                                                                         
    }                                                                                                                                                                                                              
  }      
```

It stuffed the entire config into privateLinkProvider and omitted privateLinkConfig entirely (scored 0.0%).                                                                                                      
                                                                                                                                                                                                                   
You're right that the description is heavy-handed — privateLinkProvider and privateLinkConfig are constructs we made up in this tool. The API actually takes a single flat object with provider as a field insideit. We split it into two parameters, which is what confused the LLM. Instead of patching the description, I'll merge them into a single privateLinkConfig that includes provider as a field inside it. That removes the ambiguity at the schema level and the description becomes just:                                                                                                                                      
```                                                                      
"PrivateLink configuration. Must include 'provider' ('AWS', 'AZURE', or 'GCP') plus provider-specific fields. Required when resource='privatelink'."          
```                                                 
                                         
I'll use the above description.

src/tools/atlas/streams/streamsArgs.ts
```diff
@@ -3,6 +3,103 @@ import { z } from "zod";
...
+            .passthrough()
+            .optional()
+            .describe("AWS config for S3, Kinesis, and Lambda connections."),
+        // Https
+        url: z.string().optional().describe("Webhook URL for Https connections."),
+        headers: z.record(z.string()).optional().describe("HTTP headers for Https connections."),
+        // SchemaRegistry
+        provider: z.string().optional().describe("Schema registry provider. Defaults to 'CONFLUENT'."),
+        schemaRegistryUrls: z
+            .union([z.array(z.string()), z.string()])
```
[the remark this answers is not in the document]
jwongmongodb: this is now handled with the schema transform and parse

src/tools/atlas/streams/streamsArgs.ts
```diff
@@ -3,6 +3,103 @@ import { z } from "zod";
 const ALLOWED_STREAMS_NAME_REGEX = /^[a-zA-Z0-9_-]+$/;
 const ALLOWED_STREAMS_NAME_ERROR = "Name can only contain ASCII letters, numbers, hyphens, and underscores";
 
+/** Typed schema for connectionConfig — all fields optional to support elicitation of partial configs. */
+export const ConnectionConfig = z
+    .object({
+        // Kafka
+        bootstrapServers: z
+            .union([z.string(), z.array(z.string())])
```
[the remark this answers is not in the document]
jwongmongodb: this is now handled with the schema transform and parse

src/tools/atlas/streams/streamsArgs.ts
```diff
@@ -3,6 +3,103 @@ import { z } from "zod";
 const ALLOWED_STREAMS_NAME_REGEX = /^[a-zA-Z0-9_-]+$/;
 const ALLOWED_STREAMS_NAME_ERROR = "Name can only contain ASCII letters, numbers, hyphens, and underscores";
 
+/** Typed schema for connectionConfig — all fields optional to support elicitation of partial configs. */
+export const ConnectionConfig = z
+    .object({
+        // Kafka
```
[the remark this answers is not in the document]
jwongmongodb: added tests for these:
- ConnectionConfig passthrough — preserves unknown fields
- bootstrapServers transforms — array→string join, single-element array, string passthrough 
- schemaRegistryUrls transforms — string→array split, comma-separated with whitespace trimming, single string→array, array passthrough                                                             
- PrivateLinkConfig passthrough — preserves unknown fields with required provider

src/tools/atlas/streams/manage.ts
```diff
@@ -91,13 +100,11 @@ export class StreamsManageTool extends StreamsToolBase {
...
-            .optional()
-            .describe(
-                "Updated connection configuration. Only for 'update-connection'. " +
-                    "Note: networking config cannot be modified after creation — to change networking, delete and recreate the connection."
-            ),
+        connectionConfig: ConnectionConfig.optional().describe(
+            "Updated connection configuration. Only for 'update-connection'. " +
+                "Provide only the fields to change. " +
+                "Note: networking config and connection type cannot be modified after creation — to change these, delete and recreate the connection."
+        ),
```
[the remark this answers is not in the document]
jwongmongodb: this is now handled by ConnectionConfig.parse() running Zod transforms: 
- build.ts:284 ConnectionConfig.parse(args.connectionConfig ?? {}) before creating a connection
- manage.ts:407 ConnectionConfig.parse(args.connectionConfig) before updating a connection

src/tools/atlas/streams/build.ts
jwongmongodb: Done, privateLinkProvider and privateLinkConfig are now merged into a single privateLinkConfig object with provider as a field inside it (commit 71b2bbb). The schema is z.enum(["AWS", "AZURE", "GCP"]) so Zod enforces the type constraint directly, no defensive description needed.

README.md
```diff
@@ -351,9 +351,9 @@ For more information, see the [Copilot CLI documentation](https://docs.github.co
...
-- `atlas-streams-teardown` - Delete Atlas Stream Processing resources. Also use for 'remove my workspace', 'delete all processors', or 'clean up my streams environment'. Performs safety checks before deletion: warns about connections referenced by running processors, summarizes workspace contents before deletion, and recommends stopping running processors before deletion. Use `atlas-streams-discover` to review resources before deleting.
+- `atlas-streams-teardown` - Delete Atlas Stream Processing resources. Also use for 'remove my workspace', 'delete all processors', or 'clean up my streams environment'. Performs basic safety checks before deletion: summarizes counts of processors and connections, highlights connections referenced by processors where possible, and surfaces API errors if processors are still running when deletion is attempted. Use `atlas-streams-discover` to review resources before deleting.
```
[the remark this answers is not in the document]
jwongmongodb: this is now updated

src/tools/atlas/streams/build.ts
blva: Awesome! yes, that's what I was going to sugest. thanks

tests/accuracy/streamsBuild.test.ts
```diff
@@ -0,0 +1,479 @@
...
+                ...optionalProjectDiscovery,
+                {
+                    toolName: "atlas-streams-build",
+                    parameters: {
+                        ...optionalWorkspaceParams,
+                        projectId,
+                        resource: "workspace",
+                        workspaceName: "eu-analytics",
+                        cloudProvider: "AZURE",
+                        region: "EUROPE_WEST",
```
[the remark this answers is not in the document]
jwongmongodb: This was also incorrect in our streams [skills](https://github.com/mongodb/agent-skills/pull/10/changes#diff-5a2bd052b973940b8e75f34ae3f681f45112dbe2bb6131475e665f31c9719b42R99-R102) and I will update that separately. This is now fixed.

tests/accuracy/sdk/matcher.ts
blva: I'd rather keep the strict behaviour here.

src/tools/atlas/streams/streamsArgs.ts
blva: did you consider `discriminatedUnions ` where each variant only exposes relevant fields?

src/tools/atlas/streams/manage.ts
```diff
@@ -357,7 +364,7 @@ export class StreamsManageTool extends StreamsToolBase {
     private async updateWorkspace(args: ToolArgs<typeof this.argsShape>): Promise<CallToolResult> {
         const body: Record<string, unknown> = {};
         if (args.newRegion) {
-            body.dataProcessRegion = { region: args.newRegion };
+            body.region = args.newRegion;
```
blva, line 367: is this a bugfix? do you not need `dataProcessRegion` in the body anymore?

tests/accuracy/streamsBuild.test.ts
```diff
@@ -0,0 +1,479 @@
...
+                        name: workspaceName,
+                        region: "AWS/VIRGINIA_USA",
+                        tier: "SP10",
+                        maxTier: "SP50",
+                    },
+                ])
+            ),
+        };
+    },
+    "atlas-streams-build": (): CallToolResult => {
```
blva, line 41: if this accuracy test tests `atlas-streams-build` (`streamsBuild.ts`), why do we mock it?

tests/accuracy/streamsManage.test.ts
```diff
@@ -0,0 +1,546 @@
...
+    tier: Matcher.anyOf(Matcher.undefined, Matcher.anyValue),
+    resumeFromCheckpoint: Matcher.anyOf(Matcher.undefined, Matcher.anyValue),
+    startAtOperationTime: Matcher.anyOf(Matcher.undefined, Matcher.anyValue),
+    connectionConfig: Matcher.anyOf(Matcher.undefined, Matcher.anyValue),
+    newRegion: Matcher.anyOf(Matcher.undefined, Matcher.anyValue),
+    newTier: Matcher.anyOf(Matcher.undefined, Matcher.anyValue),
+    peeringId: Matcher.anyOf(Matcher.undefined, Matcher.anyValue),
+    requesterAccountId: Matcher.anyOf(Matcher.undefined, Matcher.anyValue),
+    requesterVpcId: Matcher.anyOf(Matcher.undefined, Matcher.anyValue),
+};
```
blva: I thought we wanted to validate that the LLM wont backfill random params here?

src/tools/atlas/streams/streamsArgs.ts
```diff
@@ -3,6 +3,112 @@ import { z } from "zod";
...
+                testBucket: z.string().optional().describe("S3 test bucket name (optional, S3 only)."),
+            })
+            .passthrough()
+            .optional()
+            .describe("AWS config for S3, Kinesis, and Lambda connections."),
+        // Https
+        url: z.string().optional().describe("Webhook URL for Https connections."),
+        headers: z.record(z.string()).optional().describe("HTTP headers for Https connections."),
+        // SchemaRegistry
+        provider: z.string().optional().describe("Schema registry provider. Defaults to 'CONFLUENT'."),
```
blva: I'd suggest to rename this as there is another `provider` in another schema, also, `provider` is usually used for cloud provider accross MongoDB APIs. In that case, why dont we rename this to `schemaRegistryProvider`?

tests/accuracy/streamsBuild.test.ts
blva: nvm I think we're just validating the params

tests/accuracy/sdk/matcher.ts
jwongmongodb: should we update the doc [here](https://github.com/mongodb-js/mongodb-mcp-server/blob/1cf6f6dbed304379efe2a69791e659e34c0c2d9a/tests/accuracy/sdk/accuracyScorer.ts#L47-L50), it mentions additional params scores a 0.75.

src/tools/atlas/streams/manage.ts
jwongmongodb: this is a bug fix from the accuracy test and checking with the openAPI [specs](https://www.mongodb.com/docs/api/doc/atlas-admin-api-v2/operation/operation-updategroupstreamworkspace), unfortunately the Update workspace endpoint has a different request shape than the Create workspace endpoint so this is a fix to align with our current spec expectations.

blva: cool thanks! wanted to double-check

tests/accuracy/streamsManage.test.ts
jwongmongodb: Good point, the blanket optionalManageParams spread is intentionally permissive because the manage tool has a flat schema where all params are visible for every action, and the LLM often includes harmless extras. Without the guards, extra params still only cap at 0.75 (soon to be 0, I guess if we tighten the accuracy scoring, it will be 0 if any extra params are used), so it's a question of whether we want 0-tolerance or full 1.0-tolerance for those cases where some extra params are being added in addition to the correct params used.

We can go with one of these two approaches:
1. Current (permissive): Spread optionalManageParams on all cases → accepts extra params at 1.0. Simpler, less flaky, but doesn't catch the LLM backfilling irrelevant params.
2. Stricter per-case: Override specific params with Matcher.undefined where they definitely shouldn't appear (e.g., pipeline: Matcher.undefined on a start-processor test). This validates the LLM isn't hallucinating params, but needs per-case tuning and may flake.

I went with (1) since the scorer's built-in 0.75 (soon to be 0) penalty for extra keys already provides some guardrail, and the primary goal of these tests is verifying correct tool + action + required params selection. But I can swap to make it more strict which may increase the flakiness

tests/accuracy/sdk/matcher.ts
blva: good question, actually, in case I'm wrong, maybe it would be good to get @himanshusinghs 's insight here to understand what's the best scoring moving forward

tests/accuracy/streamsManage.test.ts
jwongmongodb: I removed the blanket optional params, each assertion has all the params it expects to see now.

deploy/aws/README.md
```diff
@@ -39,12 +39,12 @@ When creating or updating your AgentCore MCP runtime, point the container image
 
 The Dockerfile is pre-configured for AgentCore compatibility:
 
-| Setting | Value | Purpose |
-|---|---|---|
-| `MDB_MCP_EXTERNALLY_MANAGED_SESSIONS` | `true` | Lets AgentCore manage MCP session IDs |
-| `MDB_MCP_HTTP_RESPONSE_TYPE` | `json` | Returns JSON instead of SSE |
-| `MDB_MCP_DISABLED_TOOLS` | `atlas-local` | Disables tools unavailable in containers |
-| Port | `8000` | HTTP listener port |
+| Setting                               | Value         | Purpose                                  |
```
jwongmongodb, line 42: driveby formatting change that was merged to master (failing CI formatting tasks in my PR)

src/tools/atlas/streams/streamsArgs.ts
jwongmongodb: I was thinking about changing this, is it okay that this field will deviate from what we are calling it in the API? Can that lead to some confusion with the user?

tests/accuracy/streamsManage.test.ts
jwongmongodb: Here is the latest accuracy test [run](https://github.com/mongodb-js/mongodb-mcp-server/actions/runs/23327390618).

65/66 passing** (threshold ≥ 75%) | 52 at 100% | 13 at 75% | 1 at 0%
the only 0% is "Connect my Atlas cluster" which is a pre-existing tool selection issue. The 75% scores are the LLM adding extra params that the scorer now honestly penalizes instead of blanket-accepting.

src/tools/atlas/streams/build.ts
```diff
@@ -291,7 +281,7 @@ export class StreamsBuildTool extends StreamsToolBase {
             );
         }
 
-        const config: Record<string, unknown> = { ...args.connectionConfig };
+        const config: Record<string, unknown> = { ...ConnectionConfig.parse(args.connectionConfig ?? {}) };
```
himanshusinghs: nit: The config can be safely typed to the type of `ConnectionConfig`

src/tools/atlas/streams/manage.ts
himanshusinghs: It seems the action name is tied to other parameters (resourceName, dlq, etc).
It can be better modeled together as a Zod schema using `Zod.discriminatedUnion`. You will find example for this in the export tool, createIndex tool, explain tool.

Using a more concrete schema, will give better guarantees in terms of model passing correct parameters given an action.

src/tools/atlas/streams/manage.ts
```diff
@@ -397,9 +404,10 @@ export class StreamsManageTool extends StreamsToolBase {
             throw new Error("connectionConfig is required to update a connection.");
         }
 
+        const normalizedConfig = ConnectionConfig.parse(args.connectionConfig);
         await this.apiClient.updateStreamConnection({
             params: { path: { groupId: args.projectId, tenantName: args.workspaceName, connectionName: name } },
-            body: args.connectionConfig as never,
+            body: normalizedConfig as never,
```
himanshusinghs, line 410: question: why is it typed as never? If body is typed as never then it is better not to provide this at all.

src/tools/atlas/streams/streamsArgs.ts
himanshusinghs: +1 to what @blva mentioned

tests/accuracy/sdk/matcher.ts
himanshusinghs: This is an unexpected drift between the actual accuracyScorer implementation and the matcher functions that were built later around it. Ideally we would like to keep the relaxed behavior (the one specified by accuracyScorer) - matching only the parameters that are required for a tool call to be considered successful but currently (because of this drift) we end up specifying unexpected keys as `Matcher.oneOf(Matcher.undefined, Matcher.(something))`.

If its alright with you both - I'd prefer to have this change land in an isolated work (will create a JIRA for us to triage) so that we don't explode this PR or miss on important changes that are required in tool calls expectations.

tests/accuracy/streamsBuild.test.ts
```diff
@@ -0,0 +1,479 @@
...
+                    },
+                },
+            ],
+            mockedTools,
+        },
+        {
+            prompt: [
+                `Add a Kafka connection named 'events' to workspace '${workspaceName}' with bootstrap server broker.example.com:9092`,
+                `Now connect my Atlas cluster 'mycluster' to workspace '${workspaceName}'`,
+                `Deploy a processor in workspace '${workspaceName}' that reads from Kafka connection 'events' and writes to 'mycluster' collection 'pipeline.output'`,
```
himanshusinghs, line 192: Love this change, thanks for this 🙌

tests/accuracy/sdk/matcher.ts
himanshusinghs: Here's the [ticket](https://jira.mongodb.org/browse/MCP-441) for the change that we need do on the accuracy scorer and matcher function implementation.

Feel free to mention anything that I might have missed. We will pull it up in our next Engineering Sync to confirm a path forward.

himanshusinghs: I think this is also tied to your other conversation [here](https://github.com/mongodb-js/mongodb-mcp-server/pull/985#discussion_r2961982310)

jwongmongodb: I'm fine removing it for now, it will just score ~17 (as of the latest [run](https://github.com/mongodb-js/mongodb-mcp-server/actions/runs/23327390618)) of my accuracy tests that were previously scoring 0.75 to 0 since the LLM can add in extra params and add to the flakiness. @blva are you okay with removing this scoring change and handling the new accuracy scorer change in @himanshusinghs's [ticket](https://jira.mongodb.org/browse/MCP-441)?

src/tools/atlas/streams/discover.ts
jwongmongodb, line 37: @blva some new findings from a skills PR comment [yesterday](https://github.com/mongodb/agent-skills/pull/10#discussion_r2962844559) asking if getLogs returns PII data, I checked our logs implementation and it can contain customer PII data and the endpoint does not have a way to redact it currently.

I can make a change in the followup PR to not blow up the scope and remove the getLogs behavior, and add it back later to the MCP tool.

Some ideas/options:
1. Handle redaction in the API layer itself — Add a redact query parameter to the streams downloadOperationalLogs / downloadAuditLogs endpoints that strips sensitive patterns (connection URIs, credentials, tokens) server-side before returning logs. Highest safety, but highest effort, requires a nontrivial amount of changes.
2. Handle redaction in the MCP tool layer — Add a custom tool in systems/atlas-mcp/remote-mcp-server/src/core/tools/ that calls the logs API, decompresses the gzip response, applies regex-based redaction, truncates to a reasonable size, and returns sanitized text. Keeps redaction scoped to the MCP use case without touching the core API. 
3. Expose getLogs but with guardrails — Ship the tool but limit it to a short time window (e.g., last 1 hour), cap output size, and add a disclaimer in the tool description that logs may contain sensitive   
  data. Pragmatic middle ground, but doesn't eliminate the risk.

src/tools/atlas/streams/manage.ts
jwongmongodb: Good question, `as never` is a type escape hatch we're using because the OpenAPI client expects StreamsConnection (a polymorphic union type from the Atlas API spec as we have ~8 different StreamsConnection types), but our Zod schemas produce a structurally compatible but TypeScript-incompatible type. It's used 9 times across streams tools, other Atlas tools don't need it because their body types are simpler single objects.

The proper fix aligns with your discriminatedUnion suggestion, per-action/per-type schemas would produce outputs matching the OpenAPI types, eliminating the need for as never casts. This would be a non-trivial sized change ~500-700 lines lone, I can see how complicated the change would be (and defer it to a followup PR if it is too large of a change for this PR)

tests/accuracy/sdk/matcher.ts
blva: yes, agreed! let's keep the scoring as is now

src/tools/atlas/streams/discover.ts
jwongmongodb: we also have the `mongodb-logs` tool here in this tool so I'm not sure if the details I listed for the streams getLog functionality apply to the `mongodb-logs` tool

tests/accuracy/sdk/matcher.ts
jwongmongodb: this is now reverted. Here is the latest accuracy [run](https://github.com/mongodb-js/mongodb-mcp-server/actions/runs/23346094179), this change resulted in 13 more 0% accuracy score as expected due to additional params being added.

Streams: 76.8% (43 perfect, 13 failed out of 56 tests)

src/tools/atlas/streams/streamsArgs.ts
blva: imo this is OK, it's probably an api flaws as well

src/tools/atlas/streams/manage.ts
jwongmongodb: here is the followup ticket to refactor to the discriminatoryUnion: https://jira.mongodb.org/browse/STREAMS-2332

src/tools/atlas/streams/streamsArgs.ts
jwongmongodb: followup ticket to refactor, this is a big change to handle in this PR: https://jira.mongodb.org/browse/STREAMS-2332

src/tools/atlas/streams/discover.ts
jwongmongodb: created a followup ticket to address the mongodb-logs

PR #987, opened 2026-03-17
chore(metrics): refactor and add support for process metrics

gagik:
Existing setup is rather limited, this is especially visible when we try to run `collectDefaultMetrics` on the registry. It's best to do proper dependency injection here by passing the entire metrics object instead.

### DIFF
--- a/src/transports/base.ts
+++ b/src/transports/base.ts
@@ -162,7 +162,7 @@ export type TransportRunnerConfig<
      * construct the value is:
      *
      * ```ts
-     * import { PrometheusMetrics, createDefaultMetrics } from "@mongodb-js/mongodb-mcp-server";
+     * import { PrometheusMetrics, createDefaultMetrics } from "mongodb-mcp-server";
      * import { Counter } from "prom-client";
      *
      * const metrics = new PrometheusMetrics({


src/transports/base.ts
```diff
@@ -223,12 +246,7 @@ export abstract class TransportRunnerBase<
         this.tools = tools;
         this.createSessionConfig = createSessionConfig;
         this.createApiClient = createApiClient;
-        this.metrics = new PrometheusMetrics({
-            definitions: {
-                ...createDefaultMetrics(),
-                ...additionalMetrics,
-            } as DefaultMetrics & TMetrics,
-        });
+        this.metrics = metrics ?? new PrometheusMetrics({ definitions: createDefaultMetrics() as TMetrics });
```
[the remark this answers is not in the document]
gagik: not true at all

src/transports/base.ts
```diff
@@ -147,10 +147,33 @@ export type TransportRunnerConfig<
...
+     *         ...createDefaultMetrics(),
+     *         myCounter: new Counter({ name: "my_counter", help: "...", registers: [] }),
+     *     },
+     * });
+     * ```
      */
-    additionalMetrics?: TMetrics;
+    metrics?: Metrics<TMetrics>;
 
     /**
```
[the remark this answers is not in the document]
gagik: no real release includes this

src/common/metrics/prometheusMetrics.ts
```diff
@@ -1,12 +1,24 @@
...
+    }: {
+        definitions: TMetrics;
+        /** Whether to collect Node.js and process metrics. */
+        collectProcessMetrics?: boolean;
+        registry?: Registry;
+    }) {
         this.registry = registry ?? new Registry();
+        if (collectProcessMetrics) {
+            collectDefaultMetrics({ register: this.registry });
+        }
```
himanshusinghs, line 21: very nit: Would you wanna move this to a static factory pattern now that the constructor is taking on more job than just initialization?

src/transports/base.ts
```diff
@@ -147,10 +147,33 @@ export type TransportRunnerConfig<
...
+     * ```ts
+     * import { PrometheusMetrics, createDefaultMetrics } from "mongodb-mcp-server";
+     * import { Counter } from "prom-client";
+     *
+     * const metrics = new PrometheusMetrics({
+     *     definitions: {
+     *         ...createDefaultMetrics(),
+     *         myCounter: new Counter({ name: "my_counter", help: "...", registers: [] }),
+     *     },
+     * });
```
himanshusinghs, line 173: Nice - this is very neat <3

src/common/metrics/prometheusMetrics.ts
gagik: hm yeah that's subtle, I guess it's technically a side effect but I feel like adding things into a registry is still essentially just initialization

PR #993, opened 2026-03-20
chore: adjust logging and telemetry setup MCP-410

gagik:
Makes logging more strict.

commit 0c87e6100aecf1cc9bd21b17a156774bff0ac266
    Remove noRedaction from telemetry error logs
    
    Telemetry send-failure log messages interpolate error.message which
    can contain sensitive fragments like URLs or tokens from network/API
    errors. Removing noRedaction: true ensures these messages go through
    the standard redaction pipeline.

commit d267c7018b9b375eab38e3e94b9b7c860d0bf35f

    Apply redaction to log attributes in addition to messages
    
    LoggerBase.log() previously only redacted the message field, leaving
    attributes (key=value pairs) unredacted in console and disk loggers.
    This could expose sensitive data passed as log attributes.

commit 24221ce028f419269990aa5fe0652c712a28490e

    Redact error messages in tool results returned to MCP clients
    
    ToolBase.handleError() now passes error messages through mongodb-redact
    with the session keychain before including them in CallToolResult. This
    prevents driver/API error messages from leaking sensitive fragments like
    hostnames, auth details, or connection URIs to MCP clients.

commit 65813697b7c9b50dd0a9e952e74b1cd90de79226

    Replace raw exception text with generic message on /metrics errors
    
    The monitoring server's /metrics endpoint was returning the full
    exception message in 500 responses, potentially leaking internal
    details about the Prometheus registry or server state.


src/tools/atlas/read/listProjects.ts
```diff
@@ -35,9 +35,18 @@ export class ListProjectsTool extends AtlasToolBase {
                       path: {
                           orgId,
                       },
+                      query: {
+                          itemsPerPage: 500,
```
gagik, line 39: not sure why this became necessary...?

gagik: apparently the limit we had was just falling back to default, we should fix it generally but this is bumping it to a higher number so is a temporary bandaid

src/server.ts
```diff
@@ -300,7 +300,6 @@ export class Server<
             event.properties.runtime_duration_ms = Date.now() - this.startTime;
             if (error) {
                 event.properties.result = "failure";
-                event.properties.reason = error.message;
```
cveticm, line 303: Could we sanitise this error instead of removing it?  assume seeing the reason for failure could be useful but perhaps not?

gagik: we could change this to be error.name instead, it'll be a lot more vague but yeah better than nothing

gagik: going to add it as `error_type` because we might want to block `reason` being sent in the future

PR #996, opened 2026-03-22
fix: wait for exec to complete before exiting

nirinchev:
## Proposed changes

This fixes the issue with opening the config on Windows by awaiting the `exec` call before exiting.

### DIFF
--- a/src/setup/aiTool.ts
+++ b/src/setup/aiTool.ts
@@ -334,11 +334,11 @@ class Windsurf extends AITool {
     override getOpenConfigCommand(configPath: string, platform: Platform): string | null {
         switch (platform) {
             case "mac":
-                return getOpenCommandMac(configPath, "cursor");
+                return getOpenCommandMac(configPath, "windsurf");
             case "windows":
                 return `windsurf "${configPath}"`;
             case "linux":
-                return getOpenCommandLinux(configPath, "cursor");
+                return getOpenCommandLinux(configPath, "windsurf");
             default:
                 return null;
         }


src/setup/aiTool.ts
```diff
@@ -299,21 +309,19 @@ class VSCode extends AITool {
...
             case "windows":
-                // TODO: fix in CLOUDP-385463
                 return `code "${configPath}"`;
             case "linux":
                 return getOpenCommandLinux(configPath, "vscode");
             default:
                 return null;
         }
     }
+    tip = `Tip: Press ${getPlatform() === "mac" ? "Cmd+Shift+I" : "Ctrl+Shift+I"} in VS Code to open the Copilot panel.\n`;
```
rubydong, line 324: moved placement of tip so it's always after getOpenConfigCommand and consistent

src/setup/aiTool.ts
```diff
@@ -323,6 +331,18 @@ class Windsurf extends AITool {
...
+            case "mac":
+                return getOpenCommandMac(configPath, "cursor");
+            case "windows":
+                return `windsurf "${configPath}"`;
+            case "linux":
+                return getOpenCommandLinux(configPath, "cursor");
+            default:
+                return null;
+        }
+    }
```
rubydong, line 345: added support for opening in windsurf and removed todos in other areas for windows support

src/setup/aiTool.ts
```diff
@@ -323,6 +331,18 @@ class Windsurf extends AITool {
     get configPath(): string {
         return path.join(getBasePath(true), ".codeium", "windsurf", "mcp_config.json");
     }
+    override getOpenConfigCommand(configPath: string, platform: Platform): string | null {
+        switch (platform) {
+            case "mac":
+                return getOpenCommandMac(configPath, "cursor");
```
nirinchev: @rubydong should these be `windsurf` instead of `cursor`?

PR #997, opened 2026-03-22
fix: handle setup without credentials

nirinchev:
## Proposed changes

Setting up the MCP server without credentials is a valid use case. This change enables it and adds an additional check for docker. If it's available, clarifies that local atlas tools are available and how to connect to them. If it isn't, warns users that they can only connect dynamically as well as the risks involved with it.

### DIFF
--- a/src/common/logging/loggerBase.ts
+++ b/src/common/logging/loggerBase.ts
@@ -102,15 +102,3 @@ export abstract class LoggerBase<T extends EventMap<T> = DefaultEventMap> extend
         }
     }
 }
-
-export class NullLogger extends LoggerBase {
-    protected type?: LoggerType;
-
-    constructor() {
-        super(undefined);
-    }
-
-    protected logCore(): void {
-        // No-op logger, does not log anything
-    }
-}
--- a/src/setup/setupMcpServer.ts
+++ b/src/setup/setupMcpServer.ts
@@ -13,7 +13,7 @@ import { packageInfo } from "../common/packageInfo.js";
 import { getAuthType } from "../common/connectionInfo.js";
 import { type UserConfig } from "../common/config/userConfig.js";
 import { defaultCreateAtlasLocalClient } from "../common/atlasLocal.js";
-import { NullLogger } from "../common/logging/loggerBase.js";
+import { NullLogger } from "../common/logging/index.js";
 
 const buildEnvObject = (
     connectionString: string,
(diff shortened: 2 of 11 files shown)


src/setup/setupMcpServer.ts
```diff
@@ -213,31 +230,67 @@ const promptForServiceAccountSecret = async (): Promise<string> => {
...
+                    "Since you don't have Docker installed, you can only connect to a MongoDB instance dynamically, " +
+                        chalk.bold(
+                            chalk.red(
+                                "which is strongly discouraged as it will expose your connection string to the LLM."
+                            )
+                        )
+                )
+            );
+        }
         printNewLine();
```
rubydong, line 262: did you mean to exit it after? process.exit(1) because if there's no credentials we shouldn't write anything in the editor

src/setup/setupMcpServer.ts
```diff
@@ -213,31 +230,67 @@ const promptForServiceAccountSecret = async (): Promise<string> => {
...
+    if (hasDocker) {
+        availablePrompts.push('\t"Create a local Atlas deployment and connect to it"');
+        availablePrompts.push('\t"How many databases are there in my local Atlas instance?"');
+    }
+
+    if (availablePrompts.length === 0) {
+        availablePrompts.push(
+            "\t[strongly discouraged] Connect to a MongoDB instance at mongodb://localhost:27017 and list the databases"
+        );
+    }
```
rubydong, line 292: and if we exit early we won't ever reach this state

src/setup/setupMcpServer.ts
nirinchev: No, setting up the MCP server without credentials is still a valid use case. Users could still use the `connect` tool, which as mentioned above is discouraged but fine if they trust their LLM or if they're connecting to a local/test deployment.

src/common/logging/loggerBase.ts
```diff
@@ -102,3 +102,15 @@ export abstract class LoggerBase<T extends EventMap<T> = DefaultEventMap> extend
         }
     }
 }
+
+export class NullLogger extends LoggerBase {
```
gagik: can we move this to a new file?

src/setup/setupMcpServer.ts
```diff
@@ -213,31 +230,67 @@ const promptForServiceAccountSecret = async (): Promise<string> => {
 const validateCredentials = (
```
gagik, line 230: nit: would be nice to replace this with options object instead of many positional arguments

rubydong (approved):

PR #1001, opened 2026-03-24
feat: add log level minimum config

gagik:
MCP protocol seems to support specifying log level. However, at a higher level we want to control how low the exposed logging should be so this adds that.

One can override the level both through headers or the protocol.
With headers, any looser level will fail.
With the protocol, any looser level will be treated the same as the most permissable minimum level.

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -158,7 +158,7 @@ export class Server<
         this.toolContext = toolContext;
         this.metrics = metrics;
 
-        this._mcpLogLevel = userConfig.logLevel;
+        this._mcpLogLevel = userConfig.mcpClientLogLevel;
         this.mcpLogLevelFloor = this._mcpLogLevel;
     }
 
--- a/src/common/config/userConfig.ts
+++ b/src/common/config/userConfig.ts
@@ -66,7 +66,7 @@ const ServerConfigSchema = z4.object({
         .default(getLogPath())
         .describe("Folder to store logs.")
         .register(configRegistry, { defaultValueDescription: "see below*", overrideBehavior: "not-allowed" }),
-    logLevel: z4
+    mcpClientLogLevel: z4
         .enum(MCP_LOG_LEVELS)
         .default("debug")
         .describe("Minimum severity level for log messages forwarded to the MCP client.")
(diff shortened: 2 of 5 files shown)


src/common/config/configUtils.ts
```diff
@@ -132,6 +132,27 @@ export function onlyLowerThanBaseValueOverride(): CustomOverrideLogic {
     };
 }
 
+/**
+ * Allow overriding a log level only to a stricter (higher severity) value.
+ * The ordered list must go from least to most severe.
+ */
+export function onlyStricterLogLevelOverride(orderedLevels: readonly string[]): CustomOverrideLogic {
```
wtrocki, line 139: MMS atlas has feature to turn debug logs for each file in UI. Just FYI as I'm using this quite too often to undestand production

gagik: it doesn't really apply here, this is for logs we send to clients (i.e. users connecting to the MCP server)

src/common/logging/loggingTypes.ts
```diff
@@ -4,6 +4,17 @@ import type { LoggingMessageNotification } from "@modelcontextprotocol/sdk/types
...
+export const MCP_LOG_LEVELS = [
+    "debug",
+    "info",
+    "notice",
+    "warning",
+    "error",
+    "critical",
+    "alert",
+    "emergency",
+] as const;
```
nirinchev, line 16: This seems to be a duplicate of `LogLevel` above, just in an array form - can we instead get the possible levels from `LoggingMessageNotificationSchema`?

src/server.ts
```diff
@@ -220,7 +225,9 @@ export class Server<
                 throw new Error(`Invalid log level: ${params.level}`);
             }
 
-            this._mcpLogLevel = params.level;
+            const requestedIdx = MCP_LOG_LEVELS.indexOf(params.level);
```
nirinchev, line 228: This is a bit awkward where some of the checks we're making are using `McpLogger.LOG_LEVELS`, while others are using `MCP_LOG_LEVELS` - while those are the same, there's nothing in the API that enforces it. We should probably pick one and stick with it.

gagik:
Didn't realize this but was apparently committing my changes to https://github.com/mongodb-js/mongodb-mcp-server/pull/1006.... So I will include them there

PR #1003, opened 2026-03-24
chore(ci): improve release workflow MCP-447

nirinchev:
## Proposed changes

Modifies our release workflow:
1. Adds a PR label action that applies a label based on the conventional commit title
2. Adds a release.yml to categorize the notes based on the PR label
3. Adds an AI-generated summary for the release (this is something I'd like us to experiment with to see if it'll be beneficial in the long-run)
4. Adds a job to update vNext to mark it as released and create a new vNext in Jira

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -155,3 +155,46 @@ jobs:
     permissions:
       id-token: write
       contents: read
+
+  jira-release:
+    runs-on: ubuntu-latest
+    needs: [check, publish]
+    if: needs.check.outputs.VERSION_EXISTS == 'false' && needs.check.outputs.RELEASE_CHANNEL == 'latest'
+    permissions: {}
+    steps:
+      - name: Update Jira release versions
+        env:
+          JIRA_API_TOKEN: ${{ secrets.JIRA_API_TOKEN }}
+          VERSION_NUMBER: ${{ needs.check.outputs.NPM_VERSION }}
+        shell: bash
+        run: |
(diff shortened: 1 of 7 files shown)


.github/workflows/publish.yml
```diff
@@ -142,3 +155,46 @@ jobs:
...
+          vnext_id=$(echo "${versions}" | jq -r '.[] | select(.name == "vNext") | .id')
+          project_id=$(echo "${versions}" | jq -r '.[0].projectId')
+
+          if [[ -z "${vnext_id}" ]]; then
+            echo "::error::vNext version not found in Jira project ${PROJECT_KEY}"
+            exit 1
+          fi
+
+          # Rename vNext to the release version and mark it as released
+          curl -sSf -X PUT \
```
fmenezes, line 189: [q] what if there are some tickets left as in progress? do you move them out?

fmenezes:
make sure to update CONTRIBUTING.md with new release process

.github/workflows/publish.yml
nirinchev: Hm... I don't believe we should have in progress tickets that have vNext as their version. But this script won't move them out - we'll just mark the version released with incomplete tickets.

nirinchev: I added a step in the prepare-release workflow to check for unresolved tickets. While this is not a completely foolproof solution, it should give us pretty high confidence.

gagik (approved):

PR #1005, opened 2026-03-25
fix: add batching and exponential backoff to telemetry event sending

jeroenvervaeke:
## Proposed changes
Previously, all cached events were sent in a single unbatched request with no rate limiting or backoff. A 429 response would trigger an immediate retry with the full accumulated cache — a retry storm.

Now events are sent via a background timer (every 30s) in batches of 32. On 429, exponential backoff kicks in (60s → 120s → ... → 1h max) and resets after a successful send. close() sends one final batch instead of draining the entire cache.

## Checklist

### DIFF
--- a/src/telemetry/eventCache.ts
+++ b/src/telemetry/eventCache.ts
@@ -81,11 +81,16 @@ export class EventCache {
             const batch = allEvents.slice(0, batchSize);
             if (batch.length === 0) return undefined;
 
-            const { removeProcessed, result } = await processor(batch.map((e) => e.event));
-            if (removeProcessed) {
-                this.removeEvents(batch.map((e) => e.id));
+            try {
+                const { removeProcessed, result } = await processor(batch.map((e) => e.event));
+                if (removeProcessed) {
+                    this.removeEvents(batch.map((e) => e.id));
+                }
+                return result;
+            } catch {
+                // Processor threw — leave events in cache for retry
+                return undefined;
             }
-            return result;
         });
     }
 
--- a/src/telemetry/telemetry.ts
+++ b/src/telemetry/telemetry.ts
@@ -10,6 +10,7 @@ import { detectContainerEnv } from "../helpers/container.js";
 import type { DeviceId } from "../helpers/deviceId.js";
 import { EventEmitter } from "events";
 import { redact } from "mongodb-redact";
+import { Timer } from "./timer.js";
 
 type SendResult = {
     status: "success" | "rate-limited" | "error" | "empty";
@@ -56,8 +57,7 @@ export class Telemetry {
     private eventCache: EventCache;
     private deviceId: DeviceId;
     private backoffMs: number = INITIAL_BACKOFF_MS;
-    /** Single timer that drives send attempts — either at SEND_INTERVAL_MS or at a backoff delay. */
-    private sendTimer: ReturnType<typeof setTimeout> | undefined;
+    private readonly timer = new Timer();
 
     private constructor(
         private readonly session: Session,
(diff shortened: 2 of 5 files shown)


src/telemetry/telemetry.ts
```diff
@@ -81,65 +105,54 @@ export class Telemetry {
...
-                    flushTimeout.unref();
-                }
-            }),
-            this.emit([]),
-        ]);
-
-        clearTimeout(flushTimeout);
+        // Best-effort: send one final batch before closing
+        const timeout = new Promise<void>((resolve) => {
+            setTimeout(resolve, CLOSE_TIMEOUT_MS).unref();
```
gagik: https://github.com/mongodb-js/mongodb-mcp-server/blob/fc3776af296a21a56d88bfa8ad6f81abcc4167b5/src/telemetry/telemetry.ts#L115-L117
we'll need to make sure this is browser-compatible. `unref` is node specific.

Might be worth creating a wrapper function for `setTimeout` which has safe usage of `unref`

src/telemetry/telemetry.ts
```diff
@@ -81,65 +105,54 @@ export class Telemetry {
             return;
         }
 
-        this.setupPromise = Promise.all([this.deviceId.get(), detectContainerEnv()]);
-        const [deviceIdValue, containerEnv] = await this.setupPromise;
+        try {
```
gagik: I don't think this should be wrapped in try/catch, both `detectContainerEnv` and `deviceId` are meant to be handling their errors. if that's not happening then something seems wrong

src/telemetry/telemetry.ts
```diff
@@ -81,65 +105,54 @@ export class Telemetry {
...
-            }),
-            this.emit([]),
-        ]);
-
-        clearTimeout(flushTimeout);
+        // Best-effort: send one final batch before closing
+        const timeout = new Promise<void>((resolve) => {
+            setTimeout(resolve, CLOSE_TIMEOUT_MS).unref();
+        });
+        await Promise.race([this.sendBatch(), timeout]);
```
gagik: I realized this is coming from older code structure but this won't necessarily "kill" the sendBatch, https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal/timeout_static is the better fit here.
It's already being used at the bottom of this. so this Promise.race can either be removed or replaced with a different abort signal that is being passed into it

src/telemetry/telemetry.ts
```diff
@@ -171,81 +183,110 @@ export class Telemetry {
...
+    private scheduleSend(delayMs: number = SEND_INTERVAL_MS): void {
+        this.cancelScheduledSend();
+        this.sendTimer = setTimeout(() => {
+            this.sendTimer = undefined;
+            void this.sendBatchAndReschedule();
+        }, delayMs);
+        this.sendTimer.unref();
+    }
+
+    private cancelScheduledSend(): void {
```
gagik: can we create a separate helper class called i.e. `Timer` and store this and other functions related to it there? then it can be abstracted away at the telemetry class level and reused

src/telemetry/eventCache.ts
```diff
@@ -67,20 +67,25 @@ export class EventCache {
     }
 
     /**
-     * Under exclusive access: reads cached events, passes them to the processor,
-     * then replaces the cache contents with the events returned by the processor.
+     * Under exclusive access: takes up to `batchSize` oldest events and passes them
+     * to the processor. If the processor signals `removeProcessed: true`, those events
```
gagik, line 71: nit: maybe we can try/catch the processor and only remove events if no errors where thrown, otherwise log the error and do nothing?

jeroenvervaeke: fixed in 79b8e314e8ec37f1f319c34574a8cbf8519228d4

src/telemetry/telemetry.ts
jeroenvervaeke: fixed in 79b8e314e8ec37f1f319c34574a8cbf8519228d4

src/telemetry/telemetry.ts
jeroenvervaeke: fixed in 79b8e314e8ec37f1f319c34574a8cbf8519228d4

src/telemetry/telemetry.ts
jeroenvervaeke: fixed in 79b8e314e8ec37f1f319c34574a8cbf8519228d4

src/telemetry/telemetry.ts
jeroenvervaeke: fixed in 79b8e314e8ec37f1f319c34574a8cbf8519228d4

PR #1006, opened 2026-03-25
chore: allow external management of monitoring server

gagik:
This allows external management of monitoring server which allows us to extend the router with new functionality.

This is a minimal effort way to accomplish this, we could also have a thin wrapper for defining routes... but we'd end up duplicating a lot of code for minimum benefit as we'd still be deeply tied to express generally.

It makes it possible to do
```ts
 class CustomMonitoringServer extends MonitoringServer {
            constructor() {
                super({
                    host: "127.0.0.1",
                    port: 0,
                    features: ["health-check", "metrics"],
                    logger: undefined as never,
                    metrics: undefined as never,
                });
            }

            override async setupRoutes(): Promise<void> {
                this.app.get("/custom-route", (_req: express.Request, res: express.Response) => {
                    res.json({ custom: "data" });
                });
                this.app.get("/api/status", (_req: express.Request, res: express.Response) => {
                    res.json({ api: "operational" });
                });
                await super.setupRoutes();
            }
        }
}
```

### DIFF
--- a/src/lib.ts
+++ b/src/lib.ts
@@ -34,7 +34,15 @@ export {
     ConsoleLogger,
     NullLogger,
 } from "./common/logging/index.js";
-export { StreamableHttpRunner, type StreamableHttpTransportRunnerConfig } from "./transports/streamableHttp.js";
+export {
+    StreamableHttpRunner,
+    MonitoringServer,
+    createDefaultMonitoringServer,
+    type StreamableHttpTransportRunnerConfig,
+    type CreateMonitoringServerFn,
+    type MonitoringServerConstructorArgs,
+    type MonitoringServerConfig,
+} from "./transports/streamableHttp.js";
 export { StdioRunner } from "./transports/stdio.js";
 export {
     TransportRunnerBase,
@@ -74,4 +82,3 @@ export { PrometheusMetrics } from "./common/metrics/prometheusMetrics.js";
 export { createDefaultMetrics } from "./common/metrics/metricDefinitions.js";
 export type { DefaultMetrics } from "./common/metrics/metricDefinitions.js";
 export type { Metrics, MetricDefinitions } from "./common/metrics/metricsTypes.js";
-export { MonitoringServer } from "./transports/streamableHttp.js";
--- a/src/transports/streamableHttp.ts
+++ b/src/transports/streamableHttp.ts
@@ -21,6 +21,45 @@ import type { DefaultMetrics, Metrics } from "../common/metrics/index.js";
 import type { MonitoringServerFeature } from "../common/schemas.js";
 
 /**
+ * Configuration options for extracting monitoring server settings from UserConfig.
+ */
+export type MonitoringServerConfig = {
+    monitoringServerHost?: string;
+    monitoringServerPort?: number;
+    healthCheckHost?: string;
+    healthCheckPort?: number;
+    monitoringServerFeatures: MonitoringServerFeature[];
+};
+
+/**
+ * Constructor arguments for creating a MonitoringServer instance.
+ */
+export type MonitoringServerConstructorArgs<TMetrics extends DefaultMetrics = DefaultMetrics> = {
+    host: string;
+    port: number;
+    features: MonitoringServerFeature[];
+    logger: LoggerBase;
+    metrics: Metrics<TMetrics>;
+};
+
+/**
+ * A function to create a custom MonitoringServer instance.
+ * When provided, the runner will use this function instead of the default MonitoringServer constructor.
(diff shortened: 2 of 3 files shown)


src/lib.ts
```diff
@@ -73,3 +73,4 @@ export { PrometheusMetrics } from "./common/metrics/prometheusMetrics.js";
 export { createDefaultMetrics } from "./common/metrics/metricDefinitions.js";
 export type { DefaultMetrics } from "./common/metrics/metricDefinitions.js";
 export type { Metrics, MetricDefinitions } from "./common/metrics/metricsTypes.js";
+export { default as express } from "express";
```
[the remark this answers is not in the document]
nirinchev: Agree with Copilot here - do we want to re-export express?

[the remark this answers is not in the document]
gagik: agh, right, express types are a separate package...  yeah we can move the types object to peerDependecies then

src/transports/streamableHttp.ts
```diff
@@ -33,7 +33,7 @@ export class StreamableHttpRunner<
     TMetrics extends DefaultMetrics = DefaultMetrics,
 > extends TransportRunnerBase<TUserConfig, TContext, TMetrics> {
     private mcpServer: MCPHttpServer<TUserConfig, TContext> | undefined;
-    private monitoringServer: MonitoringServer | undefined;
+    public monitoringServer: MonitoringServer | undefined;
```
nirinchev: This also exposes the setter - do we want that?

gagik: made me realize... the proper way forward would be to make monitoring server mutable and injectable and move it out of `start`

gagik: though I guess it should only be started when the mcp server is started,,, hmm

src/transports/streamableHttp.ts
```diff
@@ -33,10 +51,24 @@ export class StreamableHttpRunner<
...
+    private readonly monitoringServer: MonitoringServer | undefined;
 
-    constructor(config: TransportRunnerConfig<TUserConfig, TMetrics>) {
+    constructor(config: StreamableHttpTransportRunnerConfig<TUserConfig, TMetrics>) {
         super(config);
+        const host = config.userConfig.monitoringServerHost ?? config.userConfig.healthCheckHost;
+        const port = config.userConfig.monitoringServerPort ?? config.userConfig.healthCheckPort;
+        // Use externally provided monitoring server or create one if host/port are configured
+        this.monitoringServer =
+            config.monitoringServer ??
```
nirinchev: q: Should this be a factory instead of an instance? That way we can pass in the host, port, etc. rather than having them definable in two different ways.

gagik: being able to pass any arbitrary server which extends the interface is more flexible and thus preferable IMO. the factory function would end up being a less flexible interface that is more tied to our default behavior logic. 

For example, we might want to consider additional config options being passed for the additional endpoints we might want to expose

however, what I could do to make this nicer is to introduce a factory for initializing the monitoring server when none is specified

nirinchev: I think the challenge of the preconfigured server is that you'd need to also instantiate a bunch of the base ctor parameters - e.g. the logger, metrics, etc. It also creates an awkward state where certain config options would be ignored in a non-obvious way (e.g. monitoring host and port would not have an effect).

gagik: > you'd need to also instantiate a bunch of the base ctor parameters - e.g. the logger, metrics, etc

to me that's a good thing though. I have become quite fond of the explicit dependency injection pattern.

There's nothing about the logger or metrics which is inherently tied to `streamableHttp` for it to be the one initializing them, in fact one might want to use the `metrics`/`logger` object before streamable HTTP is even initialized. We should also be able to inject alternative implementations of those dependencies if needed for alternative purposes or testing. With factory pattern we'd need to have a lot of duplication, apply hacky patches or add new options to the factory and I think that'd explode pretty quickly.

This seems to capture my thinking too: https://khalilstemmler.com/articles/tutorials/dependency-injection-inversion-explained/

> It also creates an awkward state where certain config options would be ignored in a non-obvious way (e.g. monitoring host and port would not have an effect).

I honestly wish this logic wasn't in the constructor in the first place but we can always have only the `fromConfig` factory available in which case this would still be enforced

gagik: and every time I have tried to restrict and abstract away the dependency injection I had eventually settled on it anyway: https://github.com/mongodb-js/mongodb-mcp-server/pull/987/changes

PR #1033, opened 2026-04-10
chore: add API Extractor to track public API changes

gagik:
Adds `@microsoft/api-extractor` to capture and validate the public-facing API across all four package entry points (lib, web, tools, ui). The check:api script runs in CI to detect unintentional breaking changes, and update:api regenerates the reports when changes are intentional.

Also fixes some unintentionally not exported types which are actually used by some of the referenced public APIs.

There are still some warnings but most of them are about "forgotten" APIs which we don't necessarily need to export right now.

### Workflow when changing the public API

1. Make your code changes.
2. Run `pnpm run update:api` to regenerate the reports.
3. Review the diff in `api-extractor/reports/`.
4. Commit the updated reports alongside your code changes.

If you forget to update the reports, `pnpm run check:api` will fail.


lerouxb (approved):

gagik:
The broken Ubuntu test is also broken on main, will handle it separately.

PR #1044, opened 2026-04-13
chore: add agentic workflow for vuln tickets

nirinchev:
Adds an agentic workflow that scans jira for vuln tickets and investigates them. Depending on the outcome, it will either open up a PR with a fix or leave a comment in jira explaining its investigation.


.github/workflows/vulnerability-scanner.lock.yml
```diff
@@ -0,0 +1,1468 @@
...
...   {"create_pull_request":{"max":1,"max_patch_size":1024,"protected_files":["package.json","bun.lockb","bunfig.toml","deno.json","deno.jsonc","deno.lock","global.json","NuGet.Config","Directory.Packages.props","mix.exs","mix.lock","go.mod","go.sum","stack.yaml","stack.yaml.lock","pom.xml","build.gradle","build.gradle.kts","settings.gradle","settings.gradle.kts","gradle.properties","package-lock.json","yarn.lock","pnpm-lock.yaml","npm-shrinkwrap.json","requirements.txt","Pipfile","Pipfile.lock","pyproject.toml","setup.py","setup.cfg","Gemfile","Gemfile.lock","uv.lock","CODEOWNERS"],"protected_path_prefixes":[".github/",".agents/"],"target-repo":"mongodb-js/mongodb-mcp-server"},"create_report_incomplete_issue":{},"jira_add_comment":{"description":"Add a comment to a JIRA issue.","inputs":{"comment":{"default":null,"description":"Comment text to add to the issue (Jira wiki markup supported)","required":true,"type":"string"},"issue_key":{"default":null,"description":"JIRA issue key (e.g. VULN-1234)","required":true,"type":"string"}},"output":"Comment added to JIRA issue successfully!"},"missing_data":{},"missing_tool":{},"noop":{"max":1,"report-as-issue":"true"},"report_incomplete":{}}
```
Luke-Sanderson: `protected_files` containing`pnpm-lock.yaml` would mean that the agent won't be able to suggest dependency changes which I think is important in many VULN tickets. Let me know if this is intentional

nirinchev: Good catch - I skimmed through the compiled workflow, so missed that. I updated the policy to allow edits on protected files.

Luke-Sanderson (approved):

PR #1046, opened 2026-04-13
chore: add auxiliary publishing, move metrics into its own package

gagik:
Uses npm-workspaces and splits off metrics-related code into its own package.

It'd be great to do a test release to confirm auxiliary packages can be released under this scope


package.json
```diff
@@ -72,7 +72,8 @@
     "build:esm": "tsc --project tsconfig.esm.json",
     "build:cjs": "tsc --project tsconfig.cjs.json",
     "build:universal-package": "tsx scripts/createUniversalPackage.ts",
-    "build": "concurrently \"pnpm run build:esm\" \"pnpm run build:cjs\" && pnpm run build:universal-package",
+    "build:package": "concurrently \"pnpm run build:esm\" \"pnpm run build:cjs\" && pnpm run build:universal-package",
+    "build": "pnpm --filter=@mongodb-js/mcp-metrics run build && pnpm run build:package",
```
gagik: this awkwardness will be resolved once the main package is moved into its own subfolder as well

.github/workflows/publish-packages.yml
```diff
@@ -0,0 +1,132 @@
...
+      - main
+    paths:
+      - "packages/**"
+
+jobs:
+  check:
+    runs-on: ubuntu-latest
+    permissions: {}
+    outputs:
+      PACKAGES_TO_PUBLISH: ${{ steps.detect-changes.outputs.PACKAGES_TO_PUBLISH }}
```
nirinchev: How are we bumping the package versions? Should that update prepare-release.yml to bump the versions in all package.json's?

.github/workflows/publish-packages.yml
```diff
@@ -0,0 +1,132 @@
...
+          fetch-depth: 0
+      - name: Detect packages with version changes
+        id: detect-changes
+        shell: bash
+        run: |
+          set -e
+
+          PACKAGES_TO_PUBLISH="[]"
+
+          for pkg_json in packages/*/package.json; do
```
nirinchev: Are we going to eventually move the main package under packages? Would that try to publish it in that case?

.github/workflows/publish-packages.yml
```diff
@@ -0,0 +1,132 @@
+---
+name: Publish Auxiliary Packages
```
nirinchev: Looks like pnpm suggests [using changesets or rush](https://pnpm.io/workspaces#release-workflow) for releases. Haven't looked into those at all, but may be worth checking if they offer a more streamlined way of managing releases.

package.json
nirinchev: As we add more packages, would we have to list them all 1 by 1 or is there a smarter way to tell pnpm to run the script in all workspaces?

.github/workflows/publish-packages.yml
gagik: those look very interesting but definetely need their own PR, changesets has many features and changes involved, some we might want some we might not

.github/workflows/publish-packages.yml
gagik: switched to bump workflow so even if this happens, it'll be visible in the bumped PR

package.json
gagik: yeah `@mongodb-js/*` didn't work but apparently I can do
```
pnpm --filter='./packages/*' 
```
ideal world is it's smart enough to see dependency order; will need to see if pnpm supports that

gagik: so
```
pnpm -r build
```
does this, it's a bit awkward still because ideally we'd have `build` mapped to do this recursively at root

.github/workflows/publish.yml
```diff
@@ -90,12 +90,14 @@ jobs:
...
-      - name: Build package
+      - name: Build
         run: |
           pnpm install --frozen-lockfile
           pnpm run build
       - name: Publish to NPM
-        run: pnpm publish --tag ${{ needs.check.outputs.RELEASE_CHANNEL }} --no-git-checks
+        run: |
+          pnpm publish --tag ${{ needs.check.outputs.RELEASE_CHANNEL }} --no-git-checks
+          pnpm -r --filter './packages/*' publish --tag ${{ needs.check.outputs.RELEASE_CHANNEL }} --no-git-checks
```
lerouxb, line 100: Probably something for the future, but we usually end up with some packages in the monorepo marked private that only make sense within the context of the repo and we don't want to publish to npm.

lerouxb: Although maybe that's a Compass thing where the main package isn't really a library 🤔

scripts/bumpPackages.ts
```diff
@@ -0,0 +1,307 @@
+#!/usr/bin/env tsx
```
lerouxb, line 1: we have monorepo tools and similar in devtools-shared that probably does all this, btw.

lerouxb (approved):
I'm kinda wondering why we're not using devtools-shared, but OK.

.github/workflows/publish.yml
gagik: I think marking private in package.json would handle that but yeah something to look into. not that we need anything specific right now

scripts/bumpPackages.ts
gagik: yeah unfortunately this is `npm` vs `pnpm` problem.

PR #1049, opened 2026-04-14
chore: make session functions async, allow parallel session initalizations

gagik:
Ensures both parallel initializations and enforcement of potential ownership work as expected


api-extractor/reports/mongodb-mcp-server.public.api.md
```diff
@@ -603,7 +603,7 @@ export interface ISessionStore<T extends CloseableTransport = CloseableTransport
         reason?: SessionCloseReason;
     }): Promise<void>;
     // (undocumented)
-    getSession(sessionId: string): T | undefined;
+    getSession(sessionId: string): Promise<T | undefined>;
```
gagik, line 606: breaking change but a very new, largely undocumented and advanced API

src/common/sessionStore.ts
```diff
@@ -126,6 +126,7 @@ export class SessionStore<T extends CloseableTransport = CloseableTransport> imp
             logger,
         };
         this.metrics.get("sessionCreated").inc();
+        return Promise.resolve();
```
himanshusinghs, line 129: nit: I think the Promise.resolve is not needed now since async already wraps the function?

src/transports/mcpHttpServer.ts
```diff
@@ -271,11 +258,8 @@ export class MCPHttpServer<
                 internalTransport.sessionId = sessionId;
             }
 
-            // This is eagerly setting the session in the session store to ensure that follow-up requests
-            // reuse it. This may cause issues if server.connect fails as we'll try to use a transport that's
-            // not fully set up.
             server.session.logger.setAttribute("sessionId", sessionId);
-            this.sessionStore.addSession({ sessionId, transport, logger: server.session.logger });
+            await this.sessionStore.addSession({ sessionId, transport, logger: server.session.logger });
```
himanshusinghs: This particular step should happen after `server.connect(transport)`. We need to confirm though if that causes any issues (I am not aware of any such but best to be sure).

Not doing it this way will definitely cause issues because the subsequent requests will see a transport in the session store and try to use a "not-yet" connected transport.

himanshusinghs (commented):
Looking good. Left one point to re-consider.

src/common/sessionStore.ts
gagik: it doesn't have any async calls so this is still needed

src/transports/mcpHttpServer.ts
gagik: sure, I kept it as this was existing behavior but either has pros/cons that I don't feel strongly about

it's definitely better to add things after success, yeah

himanshusinghs (approved):

PR #1064, opened 2026-04-16
feat: add db level aggregate tool MCP-459

paula-stacho:
## Proposed changes

Mainly a copy of collection aggregate, using the `aggregateDb` method of the service provider, with irrelevant parts regarding search and vectorSearch removed.

Side change: separated CollOperationArgs, this is now a 4th case where we're on the db level and the DBOperationArgs being actually collection scoped was confusing

### TODO:

- [x] Add accuracy tests
- [x] Update after https://github.com/mongodb-js/mongodb-mcp-server/pull/1067 is merged

### DIFF
--- a/src/tools/mongodb/metadata/collectionIndexes.ts
+++ b/src/tools/mongodb/metadata/collectionIndexes.ts
@@ -1,4 +1,4 @@
-import { DbOperationArgs, MongoDBToolBase } from "../mongodbTool.js";
+import { CollOperationArgs, MongoDBToolBase } from "../mongodbTool.js";
 import type { ToolArgs, OperationType, ToolResult } from "../../tool.js";
 import { formatUntrustedData } from "../../tool.js";
 import { z } from "zod";
@@ -32,14 +32,14 @@ type IndexStatus = CollectionIndexesOutput["classicIndexes"][number];
 export class CollectionIndexesTool extends MongoDBToolBase {
     static toolName = "collection-indexes";
     public description = "Describe the indexes for a collection";
-    public argsShape = DbOperationArgs;
+    public argsShape = CollOperationArgs;
     public override outputSchema = CollectionIndexesOutputSchema;
     static operationType: OperationType = "metadata";
 
     protected async execute({
         database,
         collection,
-    }: ToolArgs<typeof DbOperationArgs>): Promise<ToolResult<typeof this.outputSchema>> {
+    }: ToolArgs<typeof CollOperationArgs>): Promise<ToolResult<typeof this.outputSchema>> {
         const provider = await this.ensureConnected();
         const indexes = await provider.getIndexes(database, collection);
         const classicIndexes: IndexStatus[] = indexes.map((index) => ({
--- a/src/tools/mongodb/read/aggregateDB.ts
+++ b/src/tools/mongodb/read/aggregateDB.ts
@@ -44,7 +44,7 @@ Note to LLM: If the entire aggregation result is required, use the "export" tool
             if (pipeline.some((stage) => this.isWriteStage(stage))) {
                 // This is a write pipeline, so special-case it and don't attempt to apply limits or caps
                 aggregationCursor = provider.aggregateDb(database, pipeline, {
-                    signal,
+                    ...this.getOperationOptions(signal),
                 });
 
                 documents = await aggregationCursor.toArray();
@@ -55,7 +55,7 @@ Note to LLM: If the entire aggregation result is required, use the "export" tool
                     cappedResultsPipeline.push({ $limit: this.config.maxDocumentsPerQuery });
                 }
                 aggregationCursor = provider.aggregateDb(database, cappedResultsPipeline, {
-                    signal,
+                    ...this.getOperationOptions(signal),
                 });
 
                 const [totalDocuments, cursorResults] = await Promise.all([
@@ -156,7 +156,11 @@ Note to LLM: If the entire aggregation result is required, use the "export" tool
                 .aggregateDb(database, resultsCountAggregation, {
                     signal: abortSignal,
                 })
-                .maxTimeMS(AGG_COUNT_MAX_TIME_MS_CAP)
+                .maxTimeMS(
+                    this.config.maxTimeMS !== undefined
+                        ? Math.min(this.config.maxTimeMS, AGG_COUNT_MAX_TIME_MS_CAP)
+                        : AGG_COUNT_MAX_TIME_MS_CAP
+                )
                 .toArray();
 
             const documentWithCount: unknown = aggregationResults.length === 1 ? aggregationResults[0] : undefined;
--- a/src/tools/mongodb/read/aggregateDB.ts
+++ b/src/tools/mongodb/read/aggregateDB.ts
@@ -14,8 +14,11 @@ import { LogId } from "../../../common/logging/index.js";
 import { AnyAggregateStage, DBAggregateStage } from "../mongodbSchemas.js";
 
 export const AggregateArgs = {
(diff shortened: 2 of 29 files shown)


src/tools/mongodb/read/aggregateDB.ts
```diff
@@ -0,0 +1,207 @@
...
+
+export class AggregateDBTool extends MongoDBToolBase {
+    static toolName = "aggregate-db";
+    public description = "Run an aggregation against a MongoDB database";
+    public argsShape = {
+        ...DBOperationArgs,
+        ...AggregateArgs,
+        responseBytesLimit: z.number().optional().default(ONE_MB).describe(`\
+The maximum number of bytes to return in the response. This value is capped by the server's configured maxBytesPerQuery and cannot be exceeded. \
+Note to LLM: If the entire aggregation result is required, use the "export" tool instead of increasing this limit.\
```
cveticm: My understanding is that the export tool exports on collection lvl, not db level, and so this instruction will (a) potentially only export for one collection and (b) I imagine the response shape is different betw db and collection aggregations so user might get "unexpected" shape if LLM decides to use export.

Do we intend to do some follow-up work to add some export-db tool? 
This message (and the repeat in `generateMessage`) could be amended to instruct the LLM to get all collections and export each, to at least ensure all collections are exported.
Or this is not a concern?

src/tools/mongodb/metadata/collectionIndexes.ts
```diff
@@ -32,14 +32,14 @@ type IndexStatus = CollectionIndexesOutput["classicIndexes"][number];
...
-    public argsShape = DbOperationArgs;
+    public argsShape = CollOperationArgs;
     public override outputSchema = CollectionIndexesOutputSchema;
     static operationType: OperationType = "metadata";
 
     protected async execute({
         database,
         collection,
-    }: ToolArgs<typeof DbOperationArgs>): Promise<ToolResult<typeof this.outputSchema>> {
+    }: ToolArgs<typeof CollOperationArgs>): Promise<ToolResult<typeof this.outputSchema>> {
```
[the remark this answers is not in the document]
himanshusinghs: I know this is unrelated to your changes but its a good suggestion to accept.

src/tools/mongodb/read/aggregateDB.ts
```diff
@@ -0,0 +1,207 @@
...
+import { type Document, EJSON } from "bson";
+import { ErrorCodes, MongoDBError } from "../../../common/errors.js";
+import { collectCursorUntilMaxBytesLimit } from "../../../helpers/collectCursorUntilMaxBytes.js";
+import { operationWithFallback } from "../../../helpers/operationWithFallback.js";
+import { AGG_COUNT_MAX_TIME_MS_CAP, ONE_MB, CURSOR_LIMITS_TO_LLM_TEXT } from "../../../helpers/constants.js";
+import { LogId } from "../../../common/logging/index.js";
+import { AnyAggregateStage } from "../mongodbSchemas.js";
+
+export const AggregateArgs = {
+    pipeline: z.array(AnyAggregateStage).describe("An array of aggregation stages to execute."),
```
himanshusinghs: If I recall correctly the database level aggregation only supports a countable number of stages. I was unable to find the list on our docs but I think it was 4 or 5. I think a stricter schema would serve as a better hint to LLMs than just AnyAggregateStage but it also comes with concern of maintaining it so maybe there is a fine line we can draw for example - describing the top level of allowed stages?

src/tools/mongodb/read/aggregateDB.ts
paula-stacho: The db level aggregation is not meant to run aggregations on all collections, instead the pipeline has to start with some db-level stage (use cases - dealing with sessions, operations, query statistics or testing out a pipeline without any collections). The result in both cases is an AggregationCursor (see https://mongodb.github.io/node-mongodb-native/7.1/classes/Db.html#aggregate and https://mongodb.github.io/node-mongodb-native/7.1/classes/Collection.html#aggregate), so I didn't think this would be a problematic format for export. But tbh I didn't look deeper into the export, perhaps @himanshusinghs knows more and can confirm this?

paula-stacho: Given the use cases I'd say it is unlikely that we'll reach the limit, but I can't say for certain.

himanshusinghs: The export tool in its current state won't work out of the box for DB-aggregations. We definitely need to change that but I also doubt that DB-aggregations would benefit as much from the export tool? If you think it makes sense for DB-aggregations to also be exportable then yea we need to consider changes for export tool as well.

src/tools/mongodb/read/aggregateDB.ts
```diff
@@ -0,0 +1,211 @@
...
+
+            message += ".";
+        }
+
+        return message;
+    }
+
+    private isWriteStage(stage: Record<string, unknown>): boolean {
+        return "$out" in stage || "$merge" in stage;
+    }
```
himanshusinghs, line 210: Could $out and $merge also be on DB-level aggregations? I thought it was just a countable number of stages and don't recall reading these.

himanshusinghs (changes requested):

src/tools/mongodb/read/aggregateDB.ts
paula-stacho: Okay, in that case I'll remove the note and let's deal with this if it becomes a problem, as I am not sure it'll be needed

src/tools/mongodb/read/aggregateDB.ts
paula-stacho: Discussed in DMs, tldr there are a handful of specifically db-level stages, and the pipeline is to start with them. From there on, we have documents to pass on to other stages, example here: https://www.mongodb.com/docs/manual/reference/method/db.aggregate/#pipeline-with--currentop.
This means that most stages (except those that need to be first - such as search) are available. We'll add a zod validation for the first stage only.

src/tools/mongodb/read/aggregateDB.ts
cveticm: If the decision is to change the export tool, happy from my side for this to be follow-up work 👍

cveticm (approved):
LGTM

src/tools/mongodb/read/aggregateDB.ts
```diff
@@ -0,0 +1,212 @@
...
+        }
+    }
+
+    private async safeCloseCursor(cursor: AggregationCursor<unknown>): Promise<void> {
+        try {
+            await cursor.close();
+        } catch (error) {
+            this.session.logger.warning({
+                id: LogId.mongodbCursorCloseError,
+                context: "aggregate tool",
```
himanshusinghs: nit: aggregate-db tool

tests/integration/tools/mongodb/read/aggregateDB.test.ts
```diff
@@ -0,0 +1,486 @@
...
+            type: "number",
+            required: false,
+        },
+    ]);
+
+    validateThrowsForInvalidArguments(integration, "aggregate-db", [
+        {},
+        { database: "test", collection: "foo" },
+        { database: "test", pipeline: {} },
+        { database: 123, pipeline: [] },
```
himanshusinghs, line 40: Might wanna add one more case for unexpected first stage.

himanshusinghs (approved):
Looks good. We just need to change the LLM note and then small log context modifications.

tests/integration/tools/mongodb/read/aggregateDB.test.ts
paula-stacho: Good idea, thanks!

PR #1068, opened 2026-04-17
fix: reenable unauthenticated telemetry

nirinchev:
## Proposed changes

When making the changes in https://github.com/mongodb-js/mongodb-mcp-server/pull/839, we accidentally made it so the session api client is never instantiated when credentials are not supplied, which in turns means the telemetry service would never emit anything. This changes it so that we create an unauthenticated api client on demand for the telemetry service.


src/telemetry/telemetry.ts
```diff
@@ -162,12 +293,7 @@ export class Telemetry {
      * by respecting the DO_NOT_TRACK environment variable.
      */
     public isTelemetryEnabled(): boolean {
-        if (this.userConfig.telemetry === "disabled") {
-            return false;
-        }
-
-        const doNotTrack = "DO_NOT_TRACK" in process.env;
-        return !doNotTrack;
+        return this.telemetrySetting !== "disabled" && !("DO_NOT_TRACK" in process.env);
```
gagik, line 296: can we avoid having `process` at this level (we probably should keep it for the legacy compatiblity bit though)?

src/telemetry/telemetry.ts
```diff
@@ -48,31 +50,137 @@ export function nextBackoffMs(currentMs: number): number {
...
-    ) {
+    private readonly logger: LoggerBase;
+    /**
+     * Either the caller-supplied client, or a `{ apiBaseUrl }` marker used to
+     * lazily construct one on first send. Resolved through `resolveApiClient`
+     * so disabled telemetry never allocates a client.
+     */
+    private apiClientSource: ApiClient | { apiBaseUrl: string };
+    private readonly keychain: Keychain;
+    private readonly telemetrySetting: "enabled" | "disabled";
```
gagik, line 122:
```suggestion
    private readonly isEnabled: bool;
```

src/telemetry/telemetry.ts
```diff
@@ -48,31 +50,137 @@ export function nextBackoffMs(currentMs: number): number {
...
+      }
+    | {
+          /**
+           * Base URL for the Atlas API. When supplied instead of an
+           * `apiClient`, the service lazily constructs an unauthenticated
+           * {@link ApiClient} the first time it needs to send a batch — which
+           * means disabled telemetry never allocates one. Events are routed
+           * through the unauthenticated telemetry endpoint.
+           */
+          apiBaseUrl: string;
```
gagik, line 101: not a huge fan of this, could we handle this outside of this class and pass that in?

src/telemetry/telemetry.ts
```diff
@@ -48,31 +50,137 @@ export function nextBackoffMs(currentMs: number): number {
...
+     * resolves itself during setup. Host-supplied properties are merged on
+     * top of this at send time.
+     */
+    private readonly commonProperties: CommonProperties;
+    private readonly eventCache: EventCache;
+    private readonly deviceId: DeviceId;
+
+    private constructor(config: TelemetryConfig, eventCache: EventCache) {
+        this.logger = config.logger;
+        this.apiClientSource = "apiClient" in config ? config.apiClient : { apiBaseUrl: config.apiBaseUrl };
```
jeroenvervaeke, line 136: [q] Is this the path for the authenticated telemetry?

jeroenvervaeke: Is constructing an `ApiClient` so expensive that you have to do it lazily?

In my opinion, the code would be more readable if we just had:
```ts
private apiClient: ApiClient;
```
instead of:
```ts
private apiClientSource: ApiClient | { apiBaseUrl: string };
```

src/telemetry/telemetry.ts
jeroenvervaeke, line 296: nit: not sure if this is more readable than the original

src/transports/base.ts
```diff
@@ -333,8 +333,25 @@ export abstract class TransportRunnerBase<
...
-        const telemetry = Telemetry.create(session, userConfig, this.deviceId, {
-            commonProperties: serverOptions?.telemetryProperties ?? this.telemetryProperties,
+        // Telemetry needs an ApiClient so buffered events don't pile up in the
+        // cache forever. Reuse the session's authenticated client when we have
+        // one; otherwise hand the service the base URL so it can lazily
+        // construct an unauthenticated client only if/when it actually sends.
+        const telemetry = Telemetry.create({
+            logger,
+            deviceId: this.deviceId,
+            ...(session.apiClient ? { apiClient: session.apiClient } : { apiBaseUrl: userConfig.apiBaseUrl }),
```
jeroenvervaeke, line 343: Can't we just build the apiClient here for anonymous clients instead of lazily building it inside of `Telemetry.create` and having to pass this construct forward?

gagik (commented):
Overall seems reasonable, we can ideally do deferred initialization outside of the class

src/telemetry/telemetry.ts
nirinchev: I'm working with Gagik to unravel that one, but it appears constructing the ApiClient eagerly was causing issues in web applications, such as Compass (that was the reason for the original change). I kept the lazy initialization to ensure that we don't break that behavior, but it may turn out there's a better way.

src/telemetry/telemetry.ts
nirinchev: If you're suggesting that we pass in the ApiClient, then we run into the same issue your original PR aimed to solve. Since telemetry is created eagerly, regardless of whether we disable it or not, we'd create the ApiClient in Compass, which I believe you were trying to avoid by switching to lazy initialization.

src/telemetry/telemetry.ts
gagik: yeah thing is `config.apiBaseUrl` is always defined so effectively this bit doesn't make a big difference.
in that sense we could make telemetry itself nullable (or replaceable with no-op) but at that point the best solution would be to make ApiClient web-compatible (or maybe a web-compatible / no-op copy)

nirinchev: The difference is that we don't initialize the client until we actually have to send an event, which means that if telemetry is disabled, we would never initialize it.

nirinchev:
Closing in favor of https://github.com/mongodb-js/mongodb-mcp-server/pull/1078

PR #1076, opened 2026-04-18
chore: update zod to v4

nirinchev:
Latest version of the MCP SDK supports zod 4, so we adopt it across the board.


package.json
```diff
@@ -115,13 +115,13 @@
     "@testing-library/react": "^16.3.2",
     "@types/node": "^25.5.0",
     "@types/proper-lockfile": "^4.1.4",
-    "@types/react": "^18.3.0",
+    "@types/react": "^19.2.0",
```
nirinchev, line 118: The vite and react changes are due to pnpm flagging unmet dependency issues.

package.json
```diff
@@ -198,12 +198,14 @@
     "overrides": {
       "mongodb": "^7.1.0",
       "kerberos": "^7.0.0",
-      "bson": "^7.2.0"
+      "bson": "^7.2.0",
+      "zod": "^4.0.0"
```
nirinchev, line 202: The override is needed because @mcp-ui/client pulls in zod3

gagik: zod3 and zod4 have an awkward relationship, are we 100% sure nothing is breaking here?

gagik (approved):
LGTM, reason why we used v3 was because of MCP SDK ~but I assume all that was bumped by now.~ I see you confirmed that

package.json
nirinchev: Yeah, latest version of the MCP SDK is taking on zod ^3.25 || ^4.0, so it should be compatible. @mcp-ui/client takes an explicit dependency on zod3, but doesn't actually use it for anything.

PR #1078, opened 2026-04-20
fix: fall back to browser fetch in ApiClient MCP-473

nirinchev:
## Proposed changes

This is an alternative approach to https://github.com/mongodb-js/mongodb-mcp-server/pull/1068. It modifies ApiClient to add a runtime check and only use the devtools-proxy-support fetch when running in a node environment.

### DIFF
--- a/src/transports/base.ts
+++ b/src/transports/base.ts
@@ -19,7 +19,8 @@ import { Elicitation } from "../elicitation.js";
 import type { AtlasLocalClientFactoryFn } from "../common/atlasLocal.js";
 import { defaultCreateAtlasLocalClient } from "../common/atlasLocal.js";
 import { applyConfigOverrides } from "../common/config/configOverrides.js";
-import { ApiClient, type ApiClientFactoryFn } from "../common/atlas/apiClient.js";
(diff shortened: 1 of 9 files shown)


src/transports/base.ts
```diff
@@ -307,19 +308,16 @@ export abstract class TransportRunnerBase<
...
-                : undefined;
+        const apiClientOptions: ApiClientOptions = {
+            baseUrl: userConfig.apiBaseUrl,
+        };
+        if (userConfig.apiClientId && userConfig.apiClientSecret) {
+            apiClientOptions.credentials = {
+                clientId: userConfig.apiClientId,
+                clientSecret: userConfig.apiClientSecret,
+            };
+        }
```
gagik:
```suggestion
		const { apiClientId, apiClientSecret } = userConfig;
        const apiClientOptions: ApiClientOptions = {
            baseUrl: userConfig.apiBaseUrl,
            credentials: apiClientId && apiClientSecret ?  {
                clientId: apiClientId,
                clientSecret: apiClientSecret,
            } : undefined;
        };
```

gagik (approved):
LGTM, a browser test would be good to verify

gagik (approved):

PR #1085, opened 2026-04-21
chore: add telemetry to setup script MCP-460

nirinchev:
## Proposed changes

Adds setup telemetry. Additionally, modifies the telemetry service constructor to decouple it from the session and userConfig objects. Backwards compatibility is preserved by adding an overload for `Telemetry.create`.

Telemetry events for the setup script are emitted at every step indicating success/failure while ensuring we don't leak any PII.


src/setup/setupMcpServer.ts
```diff
@@ -325,47 +372,122 @@ const guideUserWithSetupSuccess = (displayName: string, availablePrompts: string
...
             serviceAccountId,
             serviceAccountSecret,
             hasDocker
         );
         guideUserWithSetupSuccess(displayName, availablePrompts);
-        await promptToOpenConfigFile(displayName, tool);
+        const openOutcome = await promptToOpenConfigFile(displayName, tool);
+        setupTelemetry.emitOpenConfigPrompted(openOutcome);
+
+        setupTelemetry.emitCompleted();
```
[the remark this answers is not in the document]
nirinchev: This is incorrect - we already use `emitEditorConfigured` to emit the outcome of the editor configuration step. `emitCompleted` is only intended to signal the completion of the script.

api-extractor/reports/mongodb-mcp-server.public.api.md
```diff
@@ -1,1321 +1,1334 @@
...
+        canonical: "canonical";
+    }>]>>;
+    oidcDumpTokens: z.ZodOptional<z.ZodUnion<readonly [z.ZodBoolean, z.ZodEnum<{
+        redacted: "redacted";
+        "include-secrets": "include-secrets";
+    }>]>>;
+    browser: z.ZodOptional<z.ZodUnion<readonly [z.ZodLiteral<false>, z.ZodString]>>;
+}, z.core.$strip>;
+
+// Warnings were encountered during analysis:
```
gagik, line 1326: should this file be added to prettier ignore? is this why this is reformatted?

jeroenvervaeke (approved):

api-extractor/reports/mongodb-mcp-server.public.api.md
nirinchev: No idea what caused this reformat 😬 It looks like it's already ignored.

nirinchev: Actually, as far as I can tell, it was formatted according to the prettier rules. The current version is regenerated and not prettier-formatted, which seems to be causing the difference.

tests/unit/setup/setupTelemetry.test.ts
```diff
@@ -0,0 +1,156 @@
...
+            testResult: "success",
+        });
+        setupTelemetry.emitCompleted();
+
+        const completed = mock.emitted.at(-1)!;
+        expect(completed.properties).toMatchObject({
+            command: "completed",
+            ai_tool: "claudeDesktop",
+            is_read_only: "false",
+            has_docker: "false",
```
blva, line 77: we should use the same structure where possible so this should be `is_container_env`

tests/unit/setup/setupTelemetry.test.ts
```diff
@@ -0,0 +1,156 @@
...
+            attempts: 2,
+            testResult: "success",
+        });
+        setupTelemetry.emitCompleted();
+
+        const completed = mock.emitted.at(-1)!;
+        expect(completed.properties).toMatchObject({
+            command: "completed",
+            ai_tool: "claudeDesktop",
+            is_read_only: "false",
```
blva: this should be `read_only_mode`

tests/unit/setup/setupTelemetry.test.ts
```diff
@@ -0,0 +1,156 @@
...
+            provided: true,
+            tested: true,
+            attempts: 2,
+            testResult: "success",
+        });
+        setupTelemetry.emitCompleted();
+
+        const completed = mock.emitted.at(-1)!;
+        expect(completed.properties).toMatchObject({
+            command: "completed",
```
blva: nit: we don't need to use `command` to define this sort of events - we should design the events to fit the scenario you are tracking here - I think this warrants a discussion w/ Analytics about the expected shape for the events

src/telemetry/types.ts
```diff
@@ -50,6 +50,88 @@ export type ServerEventProperties = {
...
+    ai_tool?: string;
+
+    /** Whether the user opted to install the MCP server in read-only mode. */
+    is_read_only?: TelemetryBoolSet;
+
+    /** Whether a reachable Docker daemon was detected on the machine. */
+    has_docker?: TelemetryBoolSet;
+
+    /** Whether the current OS/architecture is supported by the MCP server. */
+    platform_supported?: TelemetryBoolSet;
```
blva: why not send the platform value as well? `platform` is the value, and we have that in telemetry too

src/telemetry/types.ts
```diff
@@ -50,6 +50,88 @@ export type ServerEventProperties = {
...
+
+    /** Whether the Node.js version satisfies the package's engines range. */
+    node_version_ok?: TelemetryBoolSet;
+
+    /** Whether the user supplied a MongoDB connection string. */
+    connection_string_provided?: TelemetryBoolSet;
```
blva, line 97:
```suggestion
    config_connection_string?: TelemetryBoolSet;
```

src/telemetry/types.ts
```diff
@@ -50,6 +50,88 @@ export type ServerEventProperties = {
...
+    | "completed"
+    | "cancelled"
+    | "failed";
+
+/**
+ * Properties shared across all setup events. Every event carries the full
+ * accumulated context known up to that point so each event is independently
+ * queryable.
+ */
+export type SetupEventProperties = {
```
blva, line 78: I left a few in-line comments, but overall I suggest that we reuse the properties where possible, I don't think it makes sense and add new ones where needed, right?

blva (changes requested):

tests/unit/setup/setupTelemetry.test.ts
blva: a better design would be to create a command `setup` and then you can define the states as a `stage` or `status` property

src/telemetry/types.ts
blva: you can probably reuse the commonBaseProperties and extract the common properties that you want to capture in the events here

src/transports/base.ts
```diff
@@ -331,9 +331,21 @@ export abstract class TransportRunnerBase<
...
         });
-
-        const telemetry = Telemetry.create(session, userConfig, this.deviceId, {
-            commonProperties: serverOptions?.telemetryProperties ?? this.telemetryProperties,
+        const telemetry = Telemetry.create({
+            logger,
+            deviceId: this.deviceId,
+            apiClient: session.apiClient,
+            keychain: session.keychain,
+            enabled: userConfig.telemetry !== "disabled",
```
blva: nit: I'd instead validate it is equal to enabled

tests/unit/setup/setupTelemetry.test.ts
nirinchev: These are slightly different though - `is_container_env` indicates whether we're running in a container, whereas `has_docker` indicates whether docker is installed/available on the machine.

blva: in that case you might want to add that in the base properties. although at the moment `is_container_env` does check for docker AFAIK? but up to you

nirinchev: So this is a bit of unfamiliar territory for me (both telemetry and docker), but my understanding is that `is_container_env` checks whether we're running inside a container by detecting some common patterns, such as `.dockerenv` or the `container` environment variable. `has_docker` detects whether the platform I'm running on (e.g. my mac) has access to docker - i.e. the app is installed and the docker daemon is running.

As for putting in the base properties - that sounds like a good idea as we can also leverage it in the regular server telemetry.

tests/unit/setup/setupTelemetry.test.ts
nirinchev: I renamed this to `stage` as this is closer to the intent here. Not sure I fully understand the `setup` command proposal.

src/telemetry/types.ts
nirinchev: I'll remove this actually. We already collect the platform in the telemetry service, so this is redundant.

src/telemetry/types.ts
nirinchev: This is slightly different - I don't know if it matters a whole lot as I don't imagine we'd be analyzing events from setup together with operational events, but `config_connection_string` is set to `true` if the config has a connection string set. `connection_string_provided` is set to `true` when the user, as part of their setup, supplies a connection string to configure.

tests/unit/setup/setupTelemetry.test.ts
blva: got it! thanks for the details, yes we shuold definitely have this in the base props then.

tests/unit/setup/setupTelemetry.test.ts
blva: my idea was that you could still classify `setup` as a `command` and `stage` would be simply an extra property, but that works as well!

src/telemetry/types.ts
blva: I see, I find this could end up getting confused in the long term but that should work. in that case we should keep this in setup only

src/setup/setupTelemetry.ts
```diff
@@ -0,0 +1,241 @@
...
+     * Best-effort flush of any buffered events before the process exits. Also
+     * closes the owned {@link DeviceId}.
+     */
+    public async flush(): Promise<void> {
+        try {
+            await this.telemetry.close();
+        } catch {
+            // Ignore errors from telemetry.close()
+        } finally {
+            this.deviceId.close();
```
blva: deviceId.close() errors propagate out of flush(), which is inconsistent with explicitly catching and swallowing telemetry.close() errors above it. If flush is meant to be best-effort, wrap deviceId.close() the same way: try { this.deviceId.close(); } catch { }.

src/telemetry/telemetry.ts
```diff
@@ -292,3 +364,39 @@ export class Telemetry {
...
+        commonProperties: Partial<CommonProperties>;
+        eventCache: EventCache;
+    }
+): TelemetryConfig {
+    return {
+        logger: session.logger,
+        deviceId,
+        apiClient: session.apiClient,
+        keychain: session.keychain,
+        enabled: userConfig.telemetry !== "disabled",
```
blva:
```suggestion
        enabled: userConfig.telemetry === "enabled",
```

blva (approved):
LGTM once the last comments are addressed

PR #1086, opened 2026-04-21
feat: added terraform script to deploy MCP server on Agentcore runtime

anujbpanchal:
## Proposed changes
- Includes the terraform script required to deploy the MCP server on AWS Bedrock Agentcore Runtime.
- Added some files that need to be ignored in .gitignore

## Checklist

- I have signed the MongoDB Contributor agreement
[MongoDB Contributor Agreement.pdf](https://github.com/user-attachments/files/26933114/MongoDB.Contributor.Agreement.pdf)

### DIFF
--- a/deploy/aws/terraform/outputs.tf
+++ b/deploy/aws/terraform/outputs.tf
@@ -23,6 +23,11 @@ output "agentcore_runtime_arn" {
   value       = aws_bedrockagentcore_agent_runtime.mcp.agent_runtime_arn
 }
 
+output "agentcore_log_group" {
+  description = "CloudWatch log group for AgentCore runtime invocation logs"
+  value       = aws_cloudwatch_log_group.agentcore_invocations.name
+}
+
 output "agentcore_invocation_url" {
   description = "URL for invoking the AgentCore runtime"
   value       = "https://bedrock-agentcore.${local.region}.amazonaws.com/runtimes/${urlencode(aws_bedrockagentcore_agent_runtime.mcp.agent_runtime_arn)}/invocations?qualifier=DEFAULT"
--- a/deploy/aws/terraform/outputs.tf
+++ b/deploy/aws/terraform/outputs.tf
@@ -1,0 +1,66 @@
+output "ecr_repository_url" {
+  description = "ECR repository URL"
+  value       = aws_ecr_repository.mcp_server.repository_url
+}
+
+output "ecr_image_uri" {
+  description = "Full ECR image URI that was built and pushed"
+  value       = local.ecr_image_uri
+}
+
+output "agentcore_execution_role_arn" {
+  description = "IAM execution role ARN used by the AgentCore runtime"
+  value       = aws_iam_role.agentcore_execution.arn
+}
+
+output "agentcore_runtime_id" {
+  description = "Bedrock AgentCore runtime ID"
+  value       = aws_bedrockagentcore_agent_runtime.mcp.agent_runtime_id
+}
+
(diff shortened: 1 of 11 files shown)


deploy/aws/terraform/outputs.tf
```diff
@@ -0,0 +1,71 @@
...
+  description = "Command to generate an OAuth token for the test user"
+  value       = <<-EOT
+    python3 scripts/get_token.py \
+      --region ${local.region} \
+      --user-pool-id ${aws_cognito_user_pool.mcp.id} \
+      --client-id ${aws_cognito_user_pool_client.mcp.id} \
+      --username ${var.cognito_test_username} \
+      --password <PASSWORD>
+  EOT
+}
```
jeroenvervaeke, line 71: The Cognito pool is configured with `username_attributes = ["email"]`, so users must authenticate with their email. The Cognito user resource reflects this:

```hcl
resource "aws_cognito_user" "test_user" {
  username = var.cognito_test_user_email  # email is the username
```

But the generated command uses a different variable:
```
--username ${var.cognito_test_username}  # defaults to "mcp-test-user"
```

Note: the `cognito_test_username` output above this block correctly maps to `var.cognito_test_user_email`, but Terraform outputs don't assign to input variables, so the command will embed `"mcp-test-user"` and fail.

Replace `var.cognito_test_username` with `var.cognito_test_user_email` on the `--username` line.

deploy/aws/terraform/terraform.tfvars
jeroenvervaeke, line 1: I don't think this should be checked in/committed

deploy/aws/terraform/variables.tf
```diff
@@ -0,0 +1,88 @@
...
+}
+
+variable "cognito_test_user_password" {
+  description = "Temporary password for the Cognito test user (must meet Cognito password policy)"
+  type        = string
+  sensitive   = true
+}
+
+# MongoDB / Atlas credentials passed to the MCP server at runtime
+variable "mdb_connection_string" {
```
jeroenvervaeke, line 55: The variable has no `default`, making it required at `terraform apply`. Anyone using only Atlas tools (no direct MongoDB connection) will be blocked.

Add `default = ""` to match the other optional credential variables.

jeroenvervaeke:
Thanks for putting this together. The README is thorough and the IAM policies are well-scoped. Before we merge, I'd like to discuss the intended audience for this module.

As it stands, a few things make it feel more like a personal "here's how I got it working" setup than something others can drop in:

- The test user is baked into the infrastructure. The Cognito user, `get_token.py`, and the password variable are all wired into the Terraform module itself rather than being a separate dev convenience.
- Docker build happens inside `terraform apply`, which couples a CI/CD operation to infrastructure provisioning. Most teams already have a separate image pipeline and this conflicts with it.
- Cognito is assumed as the OIDC provider, but AgentCore supports any OIDC-compatible IdP. A more general module would accept `oidc_discovery_url` and `allowed_client_ids` as inputs rather than owning the Cognito setup.
- There is no workspace support, log group paths are hardcoded, and there is no namespacing to allow multiple instances in the same account.

Would you be open to one of these directions?

1. **Reusable module**: strip out the test user, `get_token.py`, and the Docker build step; accept the OIDC provider as an input. An `examples/` subdirectory could show the opinionated setup for anyone who wants the full walkthrough.
2. **Example only**: keep it as-is but move it under `examples/` and make clear in the README that this is a reference implementation, not a production-ready module.

Happy to help shape whichever direction makes more sense.

PR #1087, opened 2026-04-22
chore: skip pre-commit during release, consider release commit as a bump

gagik:
I don't really see a reason to use git-secrets during PR bump, it's a very controlled environment.

Also, we should look into last release when figuring out the needed bumps.

### DIFF
--- a/scripts/bumpPackages.ts
+++ b/scripts/bumpPackages.ts
@@ -137,8 +137,6 @@ function getPackages({ filters, overrideNames }: { filters: string[]; overrideNa
 
 function getLastVersionCommit(): string | undefined {
     try {
-        // Find the most recent commit that matches either bump or release pattern
-        // Using git log with multiple --grep patterns (OR logic when combined with --all-match disabled)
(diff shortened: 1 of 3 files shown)


lerouxb (approved):

scripts/bumpPackages.ts
```diff
@@ -134,12 +135,17 @@ function getPackages({ filters, overrideNames }: { filters: string[]; overrideNa
...
 
-function getLastBumpCommit(): string | undefined {
+function getLastVersionCommit(): string | undefined {
     try {
-        const sha = execSync(`git log --all --grep="${BUMP_COMMIT_PREFIX}" --format=%H -1`, {
-            cwd: ROOT,
-            encoding: "utf-8",
-        }).trim();
+        // Find the most recent commit that matches either bump or release pattern
+        // Using git log with multiple --grep patterns (OR logic when combined with --all-match disabled)
```
lerouxb: (OR logic when combined with --all-match disabled) is confusing me

PR #1090, opened 2026-04-22
feat(setup): Setup command includes downloading Agent Skills

mongodben:
## Proposed changes

1. Add step to `setup` command wizard to download [MongoDB agent skills](https://github.com/mongodb/agent-skills). Wraps `npx skills add mongodb/agent-skills`.

   This change makes it so a single command can set up both agent context (Skills) and tools (MCP), simplifying the developer experience. 

   This flow is inspired by NeonDB's `neon init` command ([docs](https://neon.com/docs/reference/cli-init))

## Checklist


tests/integration/setup/agentIdMapping.test.ts
```diff
@@ -0,0 +1,56 @@
...
+                timeout: 180_000,
+            });
+
+            expect(
+                result.status,
+                `skills add --agent ${agentId} exited with ${result.status}.\nstdout:\n${result.stdout ?? ""}\nstderr:\n${result.stderr ?? ""}`
+            ).toBe(0);
+        },
+        180_000
+    );
```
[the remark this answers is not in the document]
mongodben: will keep as it is unless get some guidance otherwise from mcp team

src/setup/installSkills.ts
```diff
@@ -0,0 +1,140 @@
...
+    console.log("");
+    console.log(
+        chalk.red(
+            `Agent skills install failed (exit ${exitCode}). Setup completed, but skills were not installed. See the CLI output above for details.`
+        )
+    );
+    console.log("");
+    console.log("You can retry manually:");
+    console.log(`  npx skills add ${SKILLS_REPO} --agent ${agentId}`);
+    console.log("");
```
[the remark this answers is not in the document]
mongodben: good call, will update. really we should be interpolating the previous string rather than rewriting it. will update to do that.

src/setup/installSkills.ts
```diff
@@ -0,0 +1,140 @@
...
+    return new Promise<number>((resolve) => {
+        const child = spawn("npx", args, { stdio: "inherit", cwd });
+        child.on("close", (code) => resolve(code ?? 0));
+        child.on("error", (err) => {
+            // Spawn-level error (npx missing, etc.). Print the message ourselves
+            // — nothing streamed to stderr because the process never started —
+            // then surface as a non-zero exit so the graceful-failure path runs.
+            console.error(chalk.red(`Failed to spawn the skills CLI: ${err.message}`));
+            resolve(SPAWN_ERROR_EXIT_CODE);
+        });
```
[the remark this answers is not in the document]
mongodben: hmm ok, will investigate further. good directional catch

src/setup/installSkills.ts
```diff
@@ -0,0 +1,140 @@
...
+    console.log("");
+}
+
+/**
+ * Prompt the user for whether to install skills, choose a scope, and call
+ * `installSkills`. Honors `MDB_MCP_SKIP_SKILLS_INSTALL` as a non-interactive
+ * off switch. Null-agent tools (Claude Desktop) skip prompts entirely.
+ */
+export async function promptAndInstallSkills(opts: PromptAndInstallSkillsOptions): Promise<SkillsInstallOutcome> {
+    if (parseBoolean(process.env[SKIP_ENV_VAR]) === true) {
```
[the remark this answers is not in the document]
mongodben: wont implement b/c the curr code follows existing repo patterns

src/setup/setupMcpServer.ts
```diff
@@ -351,13 +369,15 @@ export const runSetup = async (config: UserConfig): Promise<void> => {
 
         await configureEditor(tool, connectionString, serviceAccountId, serviceAccountSecret, isReadOnly);
 
+        const skillsResult = await promptAndInstallSkills({ tool, cwd: process.cwd() });
```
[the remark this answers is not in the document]
mongodben: nice call out. will investigate refactor

src/setup/aiTool.ts
```diff
@@ -266,6 +266,15 @@ export abstract class AITool {
             });
         }
     }
+
+    /**
+     * Returns the skills.sh agent ID this tool maps to, or null if [skills.sh](https://skills.sh)
+     * has no entry for it (e.g. Claude Desktop — no filesystem skills).
+     * Install paths are owned by `skills add`; we only map identity.
+     */
+    getSkillsAgentId(): string | null {
```
nirinchev: [nit] can we change this to a getter-only property? Like

```js
get skillsAgentId(): string | null {
  return null;
}
```

src/setup/installSkills.ts
```diff
@@ -0,0 +1,183 @@
...
+        }
+
+        const scope = await select<"project" | "user">({
+            message: "Install scope?",
+            default: "project",
+            choices: [
+                { value: "project", name: `Project (${opts.cwd}/)` },
+                { value: "user", name: "User (global)" },
+            ],
+        });
```
nirinchev: I'm not convinced about this - I know it's nice to offer the option to users, but we don't do that for the MCP server itself, so it can be confusing if the MCP server installs globally, while the skills are installed only for this project.

src/setup/installSkills.ts
```diff
@@ -0,0 +1,183 @@
...
+/**
+ * Shell out to `skills add` for the given tool. Non-zero exit is reported to
+ * the user but does not throw — setup has already succeeded by this point.
+ */
+export async function installSkills(opts: InstallSkillsOptions): Promise<SkillsInstallOutcome> {
+    const agentId = AI_TOOL_REGISTRY[opts.tool].getSkillsAgentId();
+    if (agentId === null) {
+        console.log(NO_SKILLS_MESSAGE);
+        return { status: "skipped", reason: "no-agent-id" };
+    }
```
nirinchev: We've already checked that the agentId is not null, we shouldn't be checking again. Instead, we could pass the already resolved agentId as an argument.

src/setup/setupTelemetry.ts
```diff
@@ -123,10 +124,15 @@ export class SetupTelemetry {
         this.emit("started");
     }
 
-    public emitPrerequisitesChecked(props: { nodeVersionOk: boolean; hasDocker?: boolean }): void {
+    public emitPrerequisitesChecked(props: {
+        nodeVersionOk: boolean;
+        hasDocker?: boolean;
+        platformSupported?: boolean;
```
nirinchev: This was removed in main, must be a bad merge.

README.md
```diff
@@ -447,6 +447,7 @@ The MongoDB MCP Server can be configured using multiple methods, with the follow
...
 | `MDB_MCP_READ_ONLY` / `--readOnly`                                                             | `false`                                                                                                                                            | When set to true, only allows read, connect, and metadata operation types, disabling create/update/delete operations.                                                                                    |
+| `MDB_MCP_SKIP_SKILLS_INSTALL`                                                                  | `false`                                                                                                                                            | Setup-only. When set to `"true"`, the `setup` wizard skips the MongoDB Agent Skills install step without prompting. Accepts only `"true"` or `"false"`.                                                  |
```
nirinchev: Do we want to have that in the config?

mongodben: i was trying to follow existing patterns by including this. no opinion on if it should stay

src/setup/setupTelemetry.ts
mongodben: noted. there was a bit of git funk from my initial rebase on your branch then merging the updated main. must have come thru that. will remove

src/setup/installSkills.ts
mongodben: ok, i'll make it only install globally

gagik:
It's a bit unconventional but since skills is an NPM package, is it possible we could bundle `skills` as a dependency and run the bin script directly from there?

nirinchev:
I'm not sure if we'd want to bundle skills as a dependency as that would cause it to be installed by everyone who wants to just run the server, even if they don't need to run the setup script.

README.md
nirinchev: This section covers config options for running the server, whereas this one is only for the setup script, and even then, I'm not sure we need an explicit option here. We don't have other flags for the setup script to skip steps.

mongodben:
> It's a bit unconventional but since skills is an NPM package, is it possible we could bundle `skills` as a dependency and run the bin script directly from there?

@gagik i can look into this. 

out of curiosity, why do you think this is preferable?

README.md
mongodben: ok, can delete

nirinchev (approved):

gagik:
I guess this is fine, an `npx` command spawning another `npx` command with auto-install sounded a bit hacky. But in essence bundling it would also essentially do that, so SGTM.

gagik (approved):

Issue #1022, opened 2026-03-31
[Bug]: $search support detection relies on hard coded dummy db name


mybyte:
### Version

main branch

### App

- [ ] Cursor
- [ ] Windsurf
- [ ] VSCode
- [ ] VSCode Insiders
- [ ] Claude Desktop
- [x] Other

### Affected Models (if applicable)

- [ ] Claude 3.5 Sonnet
- [ ] Claude 3.7 Sonnet
- [ ] GPT-4a
- [ ] o4-mini
- [x] Other

### Bug Description

The way the server is [checking for search support](https://github.com/mongodb-js/mongodb-mcp-server/blob/eb0cb166c0a5c17d2d914643e467054f394420c8/src/common/connectionManager.ts#L58) breaks if it's given a connection string restricted to specific databases/namespaces. It's not obvious to users that this is happening. Instead of hard coding the database, you should probably do something along the lines of:

```async function supportsSearch() {
    const { databases } = await client.db().admin().listDatabases();
    let dbName = null;
    let collectionName = null;

    for (const dbInfo of databases) {
        const cols = await client.db(dbInfo.name).listCollections().toArray();
        if (cols.length > 0) {
            dbName = dbInfo.name;
            collectionName = cols[0].name;
            break;
        }
    }

    if (!dbName) {
        dbName = databases.find((d) => !["admin", "local", "config"].includes(d.name))?.name || databases[0].name;
        collectionName = "test";
    }

    console.log(`Using db: "${dbName}", collection: "${collectionName}"`);

    try {
        const indexes = await client.db(dbName).collection(collectionName).listSearchIndexes().toArray();
        console.log("listSearchIndexes succeeded:", indexes);
        return true;
    } catch (err) {
        console.error("getSearchIndexes failed:", err.message);
        return false;
    }
}```
Not a magic fix, but would at least make some effort to respect read permissions to databases when working with a non-root user.

nirinchev:
Hey, thanks for filing this - I agree, this is a legitimate issue and the approach we chose was overly simplistic. We'll research a bit and try to come up with a better one.

PR #1097, opened 2026-04-24
fix: preserve optional logger for search support probe

Will-hxw:
## Background

PR #1095 adds debug logging to the Atlas Search capability probe. That changes `ConnectionStateConnected.isSearchSupported()` from a no-argument public method to requiring a `LoggerBase`, which can break direct consumers even though the normal `Session` path passes a logger.

## What Changed

- Made the `logger` parameter optional for `ConnectionStateConnected.isSearchSupported()`.
- Fall back to the existing `NullLogger` when the method is called directly without a logger.
- Updated the public API reports to show the optional parameter.
- Added a unit test covering direct no-argument calls.

## Not Changed

- The search capability probe behavior is unchanged.
- The database candidate order and bounded candidate list from #1095 are unchanged.
- The integration test harness changes from #1095 are unchanged.

## Risk

Low. Existing internal calls still pass the session logger. Direct calls without a logger keep working and simply use no-op logging during the first probe.

## Validation

- `pnpm exec vitest --project unit-and-integration tests/unit/common/session.test.ts --run`
- `pnpm run check:types`
- `pnpm run check:api`
- `pnpm run check:format`
- `pnpm run check:lint`
- `pnpm run check:dependencies`
- `pnpm run check`

## Linked Issues

Supports #1095 / #1022.

### DIFF
--- a/api-extractor/reports/mongodb-mcp-server.public.api.md
+++ b/api-extractor/reports/mongodb-mcp-server.public.api.md
@@ -358,7 +358,7 @@ export class ConnectionStateConnected implements ConnectionState {
     // (undocumented)
     connectionStringInfo?: ConnectionStringInfo | undefined;
     // (undocumented)
-    isSearchSupported(logger: LoggerBase): Promise<boolean>;
+    isSearchSupported(logger?: LoggerBase): Promise<boolean>;
     // (undocumented)
     serviceProvider: NodeDriverServiceProvider;
     // (undocumented)
--- a/api-extractor/reports/web.public.api.md
+++ b/api-extractor/reports/web.public.api.md
@@ -382,7 +382,7 @@ export class ConnectionStateConnected implements ConnectionState {
     // (undocumented)
     connectionStringInfo?: ConnectionStringInfo | undefined;
     // (undocumented)
-    isSearchSupported(logger: LoggerBase): Promise<boolean>;
+    isSearchSupported(logger?: LoggerBase): Promise<boolean>;
     // (undocumented)
     serviceProvider: NodeDriverServiceProvider;
     // (undocumented)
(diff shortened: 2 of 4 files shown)


nirinchev:
hey, thanks for that - I was debating with myself whether we should do it as an optional or required - I went for the latter because I figured this is only called internally. Do you have a use case for calling `isSearchSupported()`?

Will-hxw:
Thanks for asking. I don't have a separate in-repo call site beyond `Session`; the reason for this patch is API compatibility rather than a new internal use case.

`ConnectionStateConnected` is exported from both public entry points (`src/lib.ts` and `src/web.ts`) and is present in the API extractor reports. A consumer can also reach it through `session.connectionManager.currentConnectionState` or through a custom/injected connection manager. Before #1095, `ConnectionStateConnected.isSearchSupported()` was callable with no arguments, so changing it to require `LoggerBase` would make the public surface stricter even though the normal `Session.isSearchSupported()` path already passes `this.logger`.

Making the logger optional keeps the new debug logging for internal calls, and direct/no-arg calls fall back to `NullLogger`, which preserves the previous behavior without changing the probe result or cache behavior. If the project intends `ConnectionStateConnected` not to be supported as public API, I am fine with dropping this patch, but then I think the required `LoggerBase` should be treated as an intentional public API change rather than an incidental one.

I also checked the failing CI here. The TypeScript/code checks are passing; the remaining failures are both the `Label PR / label` workflow. The log shows the fork PR run has only read permissions (`Issues: read`, `PullRequests: read`) and then fails on `gh pr edit --add-label` with `GraphQL: Resource not accessible by integration (addLabelsToLabelable)`. I do not have upstream write permission to add the `type: bug` label manually, so I opened #1105 to fix the label workflow by running it under `pull_request_target` without checking out PR code and with `GH_REPO` set explicitly.

PR #1106, opened 2026-04-27
chore: migration to monorepo (WIP)

lerouxb:
## Checklist


packages/mcp-api/src/session.ts
```diff
@@ -0,0 +1,119 @@
...
+/**
+ * Public interface for an MCP session. Concrete implementations live in
+ * `@mongodb-js/mcp-core`.
+ */
+export interface ISession {
+    /** Unique identifier for this session. */
+    readonly sessionId: string;
+
+    /** Information about the connected MCP client, populated after handshake. */
+    mcpClient?: McpClientInfo;
```
gagik, line 37: hmm I wonder if we can make this immutable and pass in the constructor?

PR #1107, opened 2026-04-27
Refactor into packages attempt 2

lerouxb:
I wasn't crazy about the superpowers skill. It was very very slow. Trying again without, also updated the plan a bit while at it.


packages/mcp-api/src/sessionStore.ts
```diff
@@ -0,0 +1,26 @@
...
+    getSession(sessionId: string): Promise<T | undefined>;
+    addSession(params: { sessionId: string; transport: T; logger: ILoggerBase }): Promise<void>;
+    closeSession(params: { sessionId: string; reason?: SessionCloseReason }): Promise<void>;
+    closeAllSessions(): Promise<void>;
+}
+
+export type SessionStoreConstructorArgs<TMetrics extends MetricDefinitions = MetricDefinitions> = {
+    options: { idleTimeoutMS: number; notificationTimeoutMS: number };
+    logger: ILoggerBase;
+    metrics: IMetrics<TMetrics>;
```
gagik, line 20: combine into 1

packages/mcp-api/src/session.ts
```diff
@@ -0,0 +1,19 @@
...
+    close: [];
+    disconnect: [];
+    "connection-error": [unknown];
+};
+
+export interface ISession {
+    readonly sessionId: string;
+    logger: ICompositeLogger;
+    mcpClient?: { name?: string; version?: string; title?: string };
+    setMcpClient(mcpClient: { name?: string; version?: string; title?: string } | undefined): void;
```
gagik, line 14: look into avoiding this if possible

packages/mcp-api/src/index.test.ts
```diff
@@ -0,0 +1,39 @@
...
+    ITransportRunner,
+    ApiClientLike,
+    IMetrics,
+    ITelemetry,
+    IUIRegistry,
+    IResources,
+];
+
+it("exports all interfaces", () => {
+    expect(true).toBe(true);
```
gagik, line 38: 😄

lerouxb: It is funny that both times it did that. basically a noop test file? Do we have some kind of rule that the package has to have tests, perhaps?

packages/mcp-api/src/session.ts
lerouxb: Whats' "this"?

gagik: I'm not 100% sure if we actually need to `setMcpClient` later rather than at point of initialization

PR #1110, opened 2026-04-28
chore: rework release note generation

nirinchev:
## Proposed changes

This reworks the release notes generation script to use typescript. Additionally, I'm switching to use a grove key instead of the github token. Finally, we're now pulling PR bodies as well as based on vibes, it results in better release note summary.


scripts/generate-release-notes.ts
```diff
@@ -0,0 +1,180 @@
...
+import { join } from "path";
+import type { SemVer } from "semver";
+import semver from "semver";
+import { simpleGit } from "simple-git";
+import { createAnthropic } from "@ai-sdk/anthropic";
+import { generateText } from "ai";
+import yargs from "yargs";
+import { hideBin } from "yargs/helpers";
+
+const argv = await yargs(hideBin(process.argv))
```
gagik: can you point it to use `createParseArgs` instead? then it will use zod and we can avoid re-bundling yargs

nirinchev:
@copilot resolve the merge conflicts in this pull request

scripts/generate-release-notes.ts
```diff
@@ -0,0 +1,205 @@
...
+    });
+
+    const prompt = [
+        "You are writing release notes for MongoDB MCP Server, a tool that lets AI assistants interact with MongoDB databases and MongoDB Atlas.",
+        "Given these merged PR titles, write 2-3 sentences for end-users describing what's new in plain English.",
+        "Be concrete and avoid internal jargon. Only describe user-visible changes.",
+        "Focus primarily on new features (feat: titles) — mention bug fixes only if they address something particularly significant.",
+        "Respond with the release notes summary only, without any introductory text or formatting.",
+        `PRs:\n${prSummaries}`,
+    ].join(" ");
```
addaleax, line 126: Could this be based on commit messages rather than PR titles? Maybe that requires a bit more diligence on our side when writing commit messages, but that's a _good_ thing regardless of release note concerns, and commit messages can (and should) contain better information than just title lines about what changed

nirinchev: This is based on both the title and description. Do you suggest that we replace it all with the commit messages or we also add the commit messages as an additional data point?

nirinchev:
@copilot resolve the merge conflicts in this pull request

gagik (approved):

nirinchev:
@copilot resolve the merge conflicts in this pull request

PR #1113, opened 2026-04-28
fix: fix aggregate-db schema to work with OpenAI models

nirinchev:
## Proposed changes

The previous schema resulted in json schema that openAI would reject as invalid json schema:

```
'propertyNames': {'anyOf': [{'type': 'string', 'const': '$changeStream'},
                                {'type': 'string', 'const': '$currentOp'},
                                {'type': 'string', 'const': '$documents'},
                                {'type': 'string', 'const': '$listLocalSessions'},
                                {'type': 'string', 'const': '$queryStats'}]}
```

This was causing accuracy tests to consistently fail when run on CI. The updated schema changes from a union of property names to an array of generic aggregate stages and a runtime validation.

### DIFF
--- a/tests/integration/tools/mongodb/read/aggregateDB.test.ts
+++ b/tests/integration/tools/mongodb/read/aggregateDB.test.ts
@@ -21,7 +21,7 @@ describeWithMongoDB("aggregate-db tool", (integration) => {
         {
             name: "pipeline",
             description:
-                "An array of aggregation stages to execute. Has to start with a database aggregation stage. https://www.mongodb.com/docs/manual/reference/mql/aggregation-stages/#db.aggregate---stages",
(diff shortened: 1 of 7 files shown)


tests/integration/tools/mongodb/read/aggregateDB.test.ts
```diff
@@ -471,8 +471,8 @@ describeWithMongoDB(
 
             const { result, error, executionTime } = await runSlowAggregateDb();
 
-            // Complex regex matching and calculations on 10000 docs should take some time
-            expect(executionTime).toBeGreaterThan(100);
+            // Complex regex matching and calculations on 1000 docs should take some time
+            expect(executionTime).toBeGreaterThan(50);
```
nirinchev, line 485: these timings were failing on my mac, where I consistently got ~70ms for this agg.

gagik (approved):

PR #1114, opened 2026-04-29
refactor: move logging to `core` and `logging` MCP-480

gagik:
Splits up logging


packages/core/src/logging/mcpLogger.ts
```diff
@@ -29,9 +30,25 @@ export class McpLogger extends LoggerBase {
...
+            await promise;
+        } catch {
+            // Swallow send failures — same as the original fire-and-forget behavior
+        } finally {
+            this.pendingSends.delete(promise);
+        }
+    }
+
+    public override async flush(): Promise<void> {
+        await Promise.allSettled([...this.pendingSends]);
```
gagik: slight behavioral logic change here, this will ensure we send all loggers after closing the transport.

or at least attempt to await it. there was a test flaking because of this so this should help

packages/core/src/logging/mcpLogger.ts
```diff
@@ -0,0 +1,54 @@
...
+        const minimumLevel = MCP_LOG_LEVELS.indexOf(this.getMcpLogLevel());
+        const currentLevel = MCP_LOG_LEVELS.indexOf(level);
+        if (minimumLevel > currentLevel) {
+            return;
+        }
+
+        void this.trackSend(level, payload);
+    }
+
+    private async trackSend(level: LogLevel, payload: LogPayload): Promise<void> {
```
nirinchev: This name is a bit awkward as initially it read like this is also doing some telemetry tracking. Consider renaming to `dispatchLog` or something like this (tracking the promises is not super interesting, so doesn't need to be part of the method name).

packages/logging/src/diskLogger.ts
```diff
@@ -1,17 +1,15 @@
-import fs from "fs/promises";
+import * as fs from "fs/promises";
 import { type MongoLogWriter, MongoLogManager } from "mongodb-log-writer";
-import type { Keychain } from "../keychain.js";
-import type { LogLevel, LogPayload, LoggerType } from "./index.js";
-import { LoggerBase } from "./loggerBase.js";
+import { LoggerBase, type Keychain, type LogLevel, type LogPayload, type LoggerType } from "@mongodb-js/mcp-core";
 
 export class DiskLogger extends LoggerBase<{ initialized: [] }> {
```
nirinchev: Hm... not sure I understand the logic behind what goes in `node-utils` vs `logging`. Why is DiskLogger here?

packages/core/src/keychain.ts
```diff
@@ -10,7 +11,7 @@ export type { Secret } from "mongodb-redact";
  * should register them in the root Keychain (`Keychain.root.register`) or preferably
  * on the session keychain if available `this.session.keychain`.
  **/
-export class Keychain {
+export class Keychain implements IKeychain {
```
nirinchev, line 14: This is unrelated to this refactor, so feel free to leave it out of this PR, but I wonder if we can expose a cleaner Keychain API. Right now it's quite easy to mess it up since we can just pull the unredacted secrets from the keychain and use them outside. A more defensive API would be exposing a `redact` API on the keychain that calls `mongodb-redact/redact` with the secrets that are never exposed outside the class.

packages/types/src/logging.ts
```diff
@@ -27,9 +32,14 @@ export interface ILoggerBase {
...
 
-export interface ICompositeLogger extends ILoggerBase {
-    addLogger(logger: ILoggerBase): void;
-    setAttribute(options: { key: string; value: string }): void;
+export interface ICompositeLogger extends ILogger {
+    addLogger(logger: ILogger): void;
+    setAttribute(key: string, value: string): void;
 }
+
+export type IMcpConnection = Pick<McpServer, "isConnected" | "sendLoggingMessage"> & {
```
nirinchev: Is this used?

packages/types/src/keychain.ts
```diff
@@ -1,5 +1,5 @@
 export interface IKeychain {
-    register(options: { value: unknown; kind: string }): void;
+    register(value: unknown, kind: string): void;
     clearAllSecrets(): void;
-    readonly allSecrets: unknown[];
+    readonly allSecrets: { value: unknown; kind: string }[];
```
nirinchev: The interface API seem to be more permissive than the implementation API. Is this done only to avoid taking a dependency on the mongodb-redact package?

packages/core/src/logging/compositeLogger.ts
```diff
@@ -37,4 +33,8 @@ export class CompositeLogger extends LoggerBase {
     public setAttribute(key: string, value: string): void {
         this.attributes[key] = value;
     }
+
+    public override async flush(): Promise<void> {
+        await Promise.all(this.loggers.map((logger) => logger.flush()));
```
nirinchev: Should this be `allSettled` to match the MCP logger flush?

packages/logging/src/diskLogger.ts
gagik: Yeah I'm not too sure what's the best move here. DiskLogger relies on `fs` which is Node-specific, I wanted to put the general loggers like composite and base into `core` and let `core` be environment-independent.

 But all other loggers are also rather general e.g. ConsoleLogger would be useful in Compass too. That essentially leaves out only DiskLogger. 

So a `logging` package gets infected with `fs` if we just move all logic there. Having `disk-logger` seemed too small of a package, so the direction I took was to have a general `node-utils` service which has all Node-specific implementations.

Open to ideas here.

packages/core/src/logging/mcpLogger.ts
```diff
@@ -0,0 +1,54 @@
...
+import type { McpServer, LoggerType, LogLevel, LogPayload } from "@mongodb-js/mcp-types";
+import { MCP_LOG_LEVELS } from "../index.js";
+import { LoggerBase } from "./loggerBase.js";
+
+export class McpLogger extends LoggerBase {
+    private readonly server: McpServer;
+    private readonly getMcpLogLevel: () => LogLevel;
+    private readonly pendingSends = new Set<Promise<void>>();
+
+    public constructor(options: { server: McpServer; mcpLogLevel: LogLevel | (() => LogLevel); keychain: Keychain }) {
```
nirinchev: Should we have the base logger options extracted to a separate type to avoid repeating it? Not a huge deal, but would make the ctor look a bit cleaner - `constructor(options: { server: McpServer } & LoggerConfig)`.

packages/core/src/logging/noopLogger.ts
```diff
@@ -0,0 +1,12 @@
+import type { LoggerType } from "@mongodb-js/mcp-types";
+import { LoggerBase } from "./loggerBase.js";
+
+export class NoopLogger extends LoggerBase {
+    protected type?: LoggerType;
```
nirinchev: The base type is readonly - should this be as well? Doesn't appear to be a use case for mutating it.

packages/core/src/logging/compositeLogger.ts
nirinchev, line 11: Should `type` be readonly?

packages/node-utils/src/diskLogger.ts
nirinchev: This should similarly be `readonly` (I know it's old code, I just wanted to check all usages).

packages/types/src/logging.ts
```diff
@@ -1,6 +1,11 @@
 import type { MongoLogId } from "mongodb-log-writer";
+import type { LoggingMessageNotification } from "@modelcontextprotocol/sdk/types.js";
+import type { McpServer } from "@modelcontextprotocol/sdk/server/mcp.js";
 
-export type LogLevel = "debug" | "info" | "notice" | "warning" | "error" | "critical" | "alert" | "emergency";
+export type { MongoLogId } from "mongodb-log-writer";
+export type { McpServer } from "@modelcontextprotocol/sdk/server/mcp.js";
```
nirinchev, line 4: This is exporting a type from a dev dependency - that doesn't sound right to me.

nirinchev: Also, what is the intention behind exporting it? Are we going to eventually lock down imports and prohibit imports from `@modelcontextprotocol/sdk` or just a convenience?

nirinchev (commented):
(still making my way through it, just flagging some things as I go along)

packages/logging/src/diskLogger.ts
nirinchev: Seems like we only need `fs` directly for `mkdir` before initializing the log manager. Would taking a dependency on `mongodb-log-writer` in the `logging` package defeat the purpose of the split (my thinking being, we can easily move the fs.mkdir call inside the manager and remove the direct dependency here).

packages/core/src/logging/compositeLogger.ts
gagik: I believe so, would a logger change its type mid-way?

packages/types/src/logging.ts
gagik: will move this to an actual dependency, that seems to be the right way yeah

> Are we going to eventually lock down imports and prohibit imports from @modelcontextprotocol/sdk or just a convenience

yes that would be the goal, the `types` package is meant to define core relationships between our interfaces and especially SDK-related ones (ideally we'd maybe have a wrapper around server... that might come around once moving server happens)

packages/types/src/keychain.ts
gagik: > Is this done only to avoid taking a dependency on the mongodb-redact package?

yes, that said especially since this is a types-only package anyways, I will make this generic with mongodb-redact secret as the default.

packages/core/src/logging/compositeLogger.ts
gagik: can't say I thought about that deeply but MCP logger is more "fire and forget" situation where we don't care about some of these logs not appearing, here if any custom logger fails we would get an actual error. So there is somewhat of an advantage here.

that said since MCP logger already swallows the error in the send function I will just make it `Promise.all` there to avoid this discrepency

packages/core/src/logging/compositeLogger.ts
gagik: I'll make this static and required

nirinchev: I don't think it should - I just saw that the base class has it as readonly and figured it'd be good to keep the consistency.

packages/core/src/logging/compositeLogger.ts
nirinchev: Isn't `allSettled` a better approach here though? Typically we're flushing when the process is exiting, so surfacing the error wouldn't do us any good. And if one logger fails, we'd error out and not wait for the others to flush, potentially missing messages if we exit immediately after.

gagik: good point, then I'll switch but also return the `PromiseSettledResult<void>` in flush itself

packages/logging/src/diskLogger.ts
gagik: could you clarify what would that look like?

gagik: actually I kind of see it. LogPayload should be made generic and remove tie-in with mongo-log-writer and we can have logging package which makes use of mongodb-log-writer and a more independent loggers which don't in `core`

.gitignore
```diff
@@ -21,3 +22,6 @@ temp
 # Development tool files
 .yalc
 yalc.lock
+
+CLAUDE.md
```
gagik, line 26: I'm sure there's going to be a world where we add one here but for now I just want to iterate on my own files

packages/logging/src/diskLogger.ts
nirinchev: I was actually thinking if we could take a types-only dependency on `mongo-log-writer` and take the `MongoLogManager` as a ctor argument for the disk logger. Then, a consumer that wants the disk logger would need to either take a dependency on the log-writer package or give us something with the same shape.

fmenezes (approved):

packages/logging/src/diskLogger.ts
gagik: sounds good to me! one caveat is that i'll instead define pure interfaces that are compatible with mongo-log-writer. we might want to consider doing this to keep `mcp-types` pure as webpack isn't smart enough to figure this bit out just yet so having `mongo-log-writer` even as a type-only dependency in type-only `mcp-types` might make them think we need `mongo-log-writer` or `fs` or whatever.

So types will define redundant types that work with mongo-log-writer and we will use the real thing at binary level.

packages/core/src/logging/loggerBase.ts
```diff
@@ -1,13 +1,22 @@
...
+    ILogger,
+    LoggerConfig,
+    LoggerType,
+    LogLevel,
+    LogPayload,
+} from "@mongodb-js/mcp-types";
+
+export abstract class LoggerBase<T extends EventMap<T> = DefaultEventMap> extends EventEmitter<T> implements ILogger {
     private readonly defaultUnredactedLogger: LoggerType = "mcp";
+    private readonly keychain: LoggerConfig["keychain"];
```
nirinchev: Any reason not to reference the type directly? I find this syntax a bit convoluted to read.

packages/logging/src/compositeLogger.test.ts
```diff
@@ -0,0 +1,124 @@
+import type { MockInstance } from "vitest";
```
nirinchev, line 1: Do we want to have tests under source? Historically, we've had them isolated under tests/

packages/logging/src/diskLogger.ts
```diff
@@ -0,0 +1,49 @@
+import { LoggerBase } from "@mongodb-js/mcp-core";
+import type { LogLevel, LogPayload, LoggerType, IKeychain, LogWriter } from "@mongodb-js/mcp-types";
+
+export type DiskLoggerOptions = {
+    logWriter: LogWriter;
+    keychain: IKeychain;
```
nirinchev: Should we define this as as `{ logWriter } & LoggerConfig`?

packages/logging/src/diskLogger.ts
```diff
@@ -0,0 +1,49 @@
...
+        this.logWriter = options.logWriter;
+    }
+
+    protected readonly type: LoggerType = "disk";
+
+    protected logCore(level: LogLevel, payload: LogPayload): void {
+        const { id, context, message } = payload;
+        const mongoDBLevel = this.mapToMongoDBLogLevel(level);
+
+        this.logWriter[mongoDBLevel]("MONGODB-MCP", id, context, message, payload.attributes);
```
nirinchev, line 23: Do we have to update logWriter to have it create the folder if it doesn't exist?

packages/logging/src/compositeLogger.test.ts
gagik: it's very common in typescript projects for unit tests to be next to `src`, it makes it easy to find the accompanying unit test. integration-tests are usually seperated. I meant to mention this bit during sync but FWIW this would match how it is in Remote MCP and mongosh, Compass, vscode, etc.

packages/core/src/logging/loggerBase.ts
gagik: technically makes it more tied to `LoggerConfig` itself e.g. if we remove this field this would error but it's fine to reference directly too, readability gain is worth it

packages/logging/src/diskLogger.ts
gagik: raised a thread in mongosh about this

PR #1115, opened 2026-04-29
refactor: move UI components into `@mongodb-js/mcp-ui` package MCP-480

gagik:
Split MCP UI components (UIRegistry, bundled HTML loaders, React components) into a dedicated packages/ui package. Unit tests moved into src/ alongside their source files. Root package re-exports from the new package for backwards compatibility.


src/telemetry/types.ts
```diff
@@ -50,6 +50,82 @@ export type ServerEventProperties = {
...
+    | "completed"
+    | "cancelled"
+    | "failed";
+
+/**
+ * Properties shared across all setup events. Every event carries the full
+ * accumulated context known up to that point so each event is independently
+ * queryable.
+ */
+export type SetupEventProperties = {
```
gagik: this is kind of awkward result of merging main and then `v2`.... this is changes from current main...

tests/utils/index.ts
```diff
@@ -21,6 +22,31 @@ export function createEnvironment(): {
...
+            }
+        }
+    });
+
+    afterEach(() => {
+        for (const [key, value] of Object.entries(saved)) {
+            if (value !== undefined) {
+                process.env[key] = value;
+            }
+        }
```
addaleax:
```suggestion
        Object.assign(process.env, saved);
```

because `value` can never actually be `undefined` here, can it?

tests/utils/index.ts
```diff
@@ -21,6 +22,31 @@ export function createEnvironment(): {
...
 }
 
+/** Clears environment variables that start with the given prefix.
+ *  Creates a clean environment for the test by saving the original values of the variables and restoring them after the test.
+ */
+export function useClearEnvironment(prefix: string): void {
+    let saved: Record<string, string | undefined> = {};
+
+    beforeEach(() => {
+        saved = {};
```
addaleax:
```suggestion
        saved = Object.create(null);
```

addaleax (approved):

PR #1119, opened 2026-04-29
refactor(MCP-483):  discriminated-union input schemas for streams tools

jwongmongodb:
## Proposed changes

Refactors `atlas-streams-manage` and `atlas-streams-build` input schemas to use `z.discriminatedUnion`, so the LLM sees per-action (and per-connection-type) parameters instead of a flat bag of every possible field. Matches the prior-art pattern used by `export`, `createIndex`, and `explain`.

**No behavioral change** — same tool names, same actions, same accepted params. JSON input shape changes (see Migration below).

### What changed

- **`atlas-streams-manage`**: `argsShape.operation` is now `z.array(z.discriminatedUnion("action", [...]))` over the 7 actions. Each variant exposes only its relevant params.
- **`atlas-streams-build`**: 8 per-type connection config schemas (new file `connectionConfigs.ts`), selected post-parse by `connectionType`. Invalid field combinations are rejected before the Atlas API is called, so the LLM gets a useful error instead of a generic Atlas 400.
- **`atlas-streams-manage` `update-connection`**: same per-type validation applied in **update mode** — immutable fields (e.g. `networking`) forbidden on PATCH, all remaining fields become optional. One schema, two modes, derived via `.omit().partial()` — no duplication when new connection types are added.
- **`as never` casts** reduced to Atlas API-boundary calls only (OpenAPI strict types are structurally incompatible with our validated `Record<string, unknown>` shape). Remaining casts have inline comments explaining why.

### Migration (JSON input shape)

Before:
```json
{ "projectId": "...", "workspaceName": "ws", "action": "start-processor", "resourceName": "p" }
```

After:
```json
{ "projectId": "...", "workspaceName": "ws", "operation": [{ "action": "start-processor", "resourceName": "p" }] }
```

The array wrapping follows the existing workaround used by `export` / `createIndex` / `explain` for Cursor compatibility.

### Commits (reviewable in order)

1. `refactor(MCP-483): rewrite atlas-streams-manage with discriminated-union input schema`
2. `refactor(MCP-483): validate atlas-streams-build connectionConfig against per-type schemas`
3. `refactor(MCP-483): derive update-mode connection schemas + wire into manage.updateConnection`

### Scope

**In:** MCP-483 only.
**Out:** Output schemas on streams tools (next PR), `get-logs` re-enable, CPA list, VPC peering reads, `diagnose-output`, credentials research.

### Ticket

[MCP-483](https://jira.mongodb.org/browse/MCP-483) — part of the [STREAMS-2341](https://jira.mongodb.org/browse/STREAMS-2341) epic.

## Checklist

- [x] 866 unit tests passing (+46 over baseline)
- [x] Typecheck and lint clean
- [x] Accuracy-test fixtures rewritten to the new `operation: [{action, ...}]` shape
- [x] Integration test shapes updated
- [x] MCP Inspector end-to-end smoke: new shape works; old flat shape rejected with clear Zod error; `update-connection` rejects cross-type and immutable fields with useful messages before reaching Atlas
- [ ] Accuracy tests against `gpt-4o` (triggered by the `accuracy-tests` label)
- [ ] Live streams integration tests in CI

### DIFF
--- a/src/tools/atlas/streams/build.ts
+++ b/src/tools/atlas/streams/build.ts
@@ -5,6 +5,7 @@ import type { ElicitRequestFormParams } from "@modelcontextprotocol/sdk/types.js
 import type { OperationType, ToolArgs } from "../../tool.js";
 import { AtlasArgs } from "../../args.js";
 import { ConnectionConfig, PrivateLinkConfig, StreamsArgs } from "./streamsArgs.js";
+import { getConnectionConfigSchema } from "./connectionConfigs.js";
 
 const BuildResource = z.enum(["workspace", "connection", "processor", "privatelink"]);
 
@@ -259,6 +260,8 @@ export class StreamsBuildTool extends StreamsToolBase {
         if (useSample) {
             await this.apiClient.withStreamSampleConnections({
                 params: { path: { groupId: args.projectId } },
+                // Atlas OpenAPI types cloudProvider/region as literal enums; we validate at the
+                // input schema layer so the cast is safe here.
                 body: body as never,
             });
         } else {
@@ -296,6 +299,18 @@ export class StreamsBuildTool extends StreamsToolBase {
 
         const config = { ...ConnectionConfig.parse(args.connectionConfig ?? {}) };
 
+        // Alias normalization must happen BEFORE strict type validation. Otherwise the
+        // SchemaRegistry "url → schemaRegistryUrls" kind of aliases would get rejected
+        // as cross-type fields before they're collapsed into canonical ones.
+        if (args.connectionType === "SchemaRegistry") {
+            StreamsBuildTool.normalizeSchemaRegistryAliases(config);
+        }
+
+        const typeValidationError = StreamsBuildTool.rejectCrossTypeFields(config, args.connectionType);
+        if (typeValidationError) {
+            return typeValidationError;
+        }
+
         const missingInfo = await this.normalizeAndValidateConnectionConfig(config, args.connectionType);
         if (missingInfo) {
             return missingInfo;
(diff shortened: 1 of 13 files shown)


src/tools/atlas/streams/build.ts
```diff
@@ -560,6 +557,37 @@ export class StreamsBuildTool extends StreamsToolBase {
...
+            config.schemaRegistryAuthentication = {
+                type: "USER_INFO",
+                username: config.username || authSource.username,
+                password: config.password || authSource.password,
+            };
+            delete config.username;
+            delete config.password;
+            delete config.authentication;
+        }
+    }
```
addaleax, line 589: Definitely non-blocking, but it's a bit surprising to see this method mutates its argument, returning a modified copy would be quite a bit more intuitive imo

src/tools/atlas/streams/manage.ts
```diff
@@ -4,18 +4,115 @@ import type { CallToolResult } from "@modelcontextprotocol/sdk/types.js";
...
+                "Use `atlas-streams-discover` action='inspect-workspace' to check."
+        ),
+    resumeFromCheckpoint: z
+        .boolean()
+        .optional()
+        .describe(
+            "Resume from last checkpoint on start. Default: true. " +
+                "Set false to reprocess from beginning (drops accumulated window state)."
+        ),
+    startAtOperationTime: z.string().optional().describe("ISO 8601 timestamp to resume from."),
```
addaleax, line 38: z.iso.datetime()?

src/tools/atlas/streams/build.ts
jwongmongodb: good point, I can refactor this to be more intuitive

jwongmongodb:
closing to do later, other priorities came up.

PR #1121, opened 2026-04-30
fix: apply mongosh CLI options (e.g. oidcTrustedEndpoint) in connect-tool path

KyriosGN0:
When a user calls the MCP 'connect' tool with a non-Atlas MongoDB host
and MONGODB-OIDC auth, the driver raised:

  MongoInvalidArgumentError: Host 'foo.example.com' is not valid for OIDC
  authentication with ALLOWED_HOSTS of '*.mongodb.net,...'

Root cause: generateConnectionInfoFromCliArgs was called with only
defaultDriverOptions, omitting the mongosh CLI options stored in
userConfig. The startup (connectToConfiguredConnection) path already
spread the full userConfig; this commit makes the connect-tool path
consistent by extracting the mongosh-specific subset via
getMongoshUserConfig() and merging it in.

Fixes: the existing OIDC integration test workaround comment at
tests/integration/common/connectionManager.oidc.test.ts:~142

NOTE: i wrote this PR with Claude Code, i verified this works with my case (mongodb running on k8s, the mcp is running also in k8s, the client is a local claude code session that is using connect tool)
## Checklist

### DIFF
--- a/src/common/connectionManager.ts
+++ b/src/common/connectionManager.ts
@@ -2,7 +2,6 @@ import { EventEmitter } from "events";
 import { MongoServerError } from "mongodb";
 import { NodeDriverServiceProvider } from "@mongosh/service-provider-node-driver";
 import { generateConnectionInfoFromCliArgs, type ConnectionInfo } from "@mongosh/arg-parser";
-import { CliOptionsSchema as MongoshCliOptionsSchema } from "@mongosh/arg-parser/arg-parser";
 import type { DeviceId } from "../helpers/deviceId.js";
 import { type UserConfig } from "./config/userConfig.js";
 import { MongoDBError, ErrorCodes } from "./errors.js";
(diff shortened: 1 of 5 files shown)


gagik:
Hey, thanks for the contribution! I will first investigate this usecase a bit more and see if there are alternative solutions that could also work here 🙂

KyriosGN0:
Hey @gagik, have you had a chance to look at this PR (and use case)? is there something i can do to help you?

gagik:
@KyriosGN0 hey, sorry there was some focus on #1045 past week. The changes and the usecase overall makes sense, I'll add a couple comments to address and we can get that in 🙂

src/common/connectionManager.ts
```diff
@@ -246,6 +247,15 @@ export class MCPConnectionManager extends ConnectionManager {
         this.deviceId = deviceId;
     }
 
+    private getMongoshUserConfig(): Partial<UserConfig> {
+        const mongoshKeys = Object.keys(MongoshCliOptionsSchema.shape) as (keyof UserConfig)[];
+        return Object.fromEntries(
```
gagik: I think it'd not be a problem to pass the whole userConfig object as it'd be a superset anyway `generateConnectionInfoFromCliArgs`

KyriosGN0: @gagik done

KyriosGN0:
hey @gagik could you trigger the workflow and review ? 
Thanks!

KyriosGN0:
hey @gagik, a small reminder about my PR 🙏 
Thanks

gagik:
Back from my holiday, one minor adjustment and we can get this in 🙂 Thank you for your patience!

gagik (approved):

PR #1122, opened 2026-04-30
refactor: move to v2 structure MCP-373

gagik:
### **Only merge once the last v1 release is complete**

It's many PRs merged into one along with many patches on top; I have reviewed every file but it's obviously a lot to expect that, so if you're curious about the major changes, here's the important files:
- `api-extractor/reports/tools.public.api.md, api-extractor/reports/mongodb-mcp-server.public.api.md`  - I kept these around purely to have an easier visualization of changes. It provides a good overview of what was removed & updates.
- `packages/cli` - This is the closest to the original index.ts  / consumer layer with UserConfig we're used to. Everything around has been decoupled from UserConfig
- `packages/mongodb-mcp-server` - This is an extremely minimal consumer of packages/cli . It also currently has a bunch of re-exporting for the sake of the v2 report comparisons above but this will be removed afterwards.


tests/utils/index.ts
```diff
@@ -1,34 +0,0 @@
-import { type ConnectionManagerEvents } from "../../src/common/connectionManager.js";
```
gagik, line 1: moved to packages/test-utils/src/testUtils.ts

tests/unit/toolContext.test.ts
```diff
@@ -1,229 +0,0 @@
-import { describe, it, expect } from "vitest";
```
gagik, line 1: no longer relevant -> tool context is now a matter of extending session

tests/unit/logger.test.ts
```diff
@@ -1,400 +0,0 @@
-import type { MockInstance } from "vitest";
```
gagik, line 1: moved to various tests in packages/logging

tests/unit/transports/dryModeRunner.test.ts
```diff
@@ -1,42 +0,0 @@
-import { describe, it, expect, vi, beforeEach, afterEach } from "vitest";
```
gagik, line 1: moved to packages/integration-tests/src/transports/dryModeRunner.test.ts

tests/integration/telemetry.test.ts
```diff
@@ -1,34 +0,0 @@
-import { DeviceId } from "../../src/helpers/deviceId.js";
```
gagik, line 1: moved to packages/integration-tests/src/telemetry.test.ts

tests/integration/metrics.test.ts
```diff
@@ -1,179 +0,0 @@
-import { StreamableHttpRunner } from "../../src/transports/streamableHttp.js";
```
gagik, line 1: moved to packages/integration-tests/src/metrics.test.ts

scripts/generate.sh
```diff
@@ -1,11 +0,0 @@
-#!/usr/bin/env bash
```
gagik, line 1: => packages/scripts/generate.sh

src/common/atlas/auth/authProvider.ts
```diff
@@ -1,50 +0,0 @@
...
-                    userAgent: options.userAgent,
-                    clientId: options.credentials.clientId,
-                    clientSecret: options.credentials.clientSecret,
-                },
-                logger
-            );
-        }
-        return undefined;
-    }
-}
```
gagik, line 50: => packages/atlas-api-client/src/auth/authProvider.ts

src/common/logging/nullLogger.ts
gagik, line 1: => packages/core/src/logging/noopLogger.ts

src/transports/streamableHttp.ts
gagik, line 1: => packages/http-runners/src/streamableHttpRunner.ts

tests/integration/transports/base.test.ts
```diff
@@ -1,106 +0,0 @@
-import { afterEach, describe, expect, it, vi } from "vitest";
```
gagik, line 1: => packages/integration-tests/src/transports/base.test.ts

tests/integration/transports/createSessionConfig.test.ts
```diff
@@ -1,143 +0,0 @@
-import { StreamableHttpRunner } from "../../../src/transports/streamableHttp.js";
-import { Client } from "@modelcontextprotocol/sdk/client/index.js";
```
gagik, line 2: => packages/integration-tests/src/transports/createSessionConfig.test.ts

tests/e2e/getRandomUUID.test.ts
gagik, line 1: =>. packages/core/src/getRandomUUID.test.ts

packages/atlas-api-client/src/apiClient.test.ts
```diff
@@ -103,43 +122,9 @@ describe("ApiClient", () => {
             expect(url instanceof URL ? url.href : url).toBe(
                 "https://api.test.com/api/private/unauth/telemetry/events"
             );
-            const expectedDefaultUserAgent = `AtlasMCP/${packageInfo.version} (${process.platform}; ${process.arch})`;
             const headers = init?.headers as Record<string, string>;
             expect(headers).toBeDefined();
-            expect(headers["User-Agent"]).toBe(expectedDefaultUserAgent);
-        });
-
-        it("should not include hostname in default userAgent", async () => {
```
gagik, line 112: user agent is now being set at cli level so this test is in packages/cli/src/createServicesFromUserConfig.test.ts

packages/cli/src/cliServer.ts
```diff
@@ -76,50 +73,35 @@ export interface ServerOptions<
...
-     *   toolContext: {
-     *     tenantId: "my-tenant",
-     *     userId: "user-123",
-     *     features: { newUI: true },
-     *   },
-     * });
-     * ```
+     * to `ToolClass` type from `@mongodb-js/mcp-core` or `mongodb-mcp-server`.
      */
-    toolContext?: TContext;
```
gagik, line 104: gone in favor of session

packages/cli/src/types.ts
```diff
@@ -0,0 +1,7 @@
+export type Console = {
```
gagik, line 1: this is annoying but matches what was there before and we don't rly have a good interop between our logger vs console

packages/integration-tests/src/transports/configOverrides.test.ts
```diff
@@ -256,86 +441,6 @@ describe("Config Overrides via HTTP", () => {
         });
     });
 
-    describe("integration with createSessionConfig", () => {
```
gagik, line 259: `createSessionConfig` is removed, overrides is the way

packages/integration-tests/src/transports/streamableHttp.test.ts
```diff
@@ -865,181 +1157,40 @@ describe("StreamableHttpRunner", () => {
         });
     });
 
-    describe("monitoring server", () => {
```
gagik, line 878: moved to monitoring server tests

packages/core/src/args.ts
```diff
@@ -0,0 +1,15 @@
+import z, { type ZodString } from "zod";
+
+const NO_UNICODE_REGEX = /^[\x20-\x7E]*$/;
+export const NO_UNICODE_ERROR = "String cannot contain special characters or Unicode symbols";
+
+export const CommonArgs = {
+    string: (): ZodString => z.string().regex(NO_UNICODE_REGEX, NO_UNICODE_ERROR),
```
addaleax: Don't really like the naming or the error message here – if I see something like `CommonArgs.string()`, I would expect that to really correspond to a _regular string_. Also, saying "cannot contain Unicode symbols" is fairly confusing, since it's really not obvious which actual restriction that corresponds to (it can't be interpreted literally, since strings are by definition sequences of Unicode characters).

```suggestion
const ASCII_ONLY_NON_CC_REGEX = /^[\x20-\x7E]*$/;
export const ASCII_ONLY_NON_CC_ERROR = "String cannot contain control characters or non-ASCII characters";

export const CommonArgs = {
    asciiOnlyString: (): ZodString => z.string().regex(ASCII_ONLY_NON_CC_REGEX, ASCII_ONLY_NON_CC_ERROR),
```

packages/http-runners/src/utils.ts
```diff
@@ -0,0 +1,16 @@
...
+        }, ms);
+
+        signal.addEventListener(
+            "abort",
+            () => {
+                clearTimeout(timeout);
+                resolve();
+            },
+            { once: true }
+        );
```
addaleax: Probably fine, but it's good keeping in mind (esp. since this is a generic utility) that not removing the event listener if the signal does _not_ abort can lead to memory leaks

packages/metrics/vitest.config.js.map
```diff
@@ -0,0 +1 @@
+{"version":3,"file":"vitest.config.js","sourceRoot":"","sources":["vitest.config.ts"],"names":[],"mappings":"AAAA,OAAO,EAAE,YAAY,EAAE,MAAM,eAAe,CAAC;AAE7C,eAAe,YAAY,CAAC;IACxB,IAAI,EAAE;QACF,UAAU,EAAE,CAAC,4BAA4B,CAAC;KAC7C;CACJ,CAAC,CAAC"}
```
addaleax: Should these files that are generated from `vitest.config.ts` be commited?

packages/mongodb-mcp-server/src/index.ts
```diff
@@ -0,0 +1,71 @@
...
+        } else {
+            console.error("Could not enable FIPS mode. This installation does not appear to support FIPS.");
+        }
+        console.error("Error details:");
+        console.error(fipsError);
+        process.exit(1);
+    }
+}
+
+enableFipsIfRequested();
```
addaleax, line 37: We may want to be careful with ESM and the relative timing of execution here.

This seems to be inspired by similar code in mongosh; mongosh compiles to CJS, and does so in a slightly non-spec-compliant way, where `import`s get translated to `require()` calls _in place_. In "real" ESM (and the presence of `import.meta` handling in this PR makes me assume that we're targeting ESM now), `import`ed modules evaluate before _any_ code in the calling script runs.

This is a good reminder to also fix this in mongosh: https://github.com/mongodb-js/mongosh/pull/2731

packages/mongodb-mcp-server/package.json
```diff
@@ -0,0 +1,112 @@
+{
+  "name": "mongodb-mcp-server",
+  "description": "MongoDB Model Context Protocol Server",
+  "version": "1.11.0-prerelease.3",
```
addaleax, line 4: Probably an unnecessary comment, but let's make sure this version is what it should be when we merge this PR?

packages/test-utils/src/index.ts
```diff
@@ -0,0 +1,34 @@
...
+    }
+}
+
+export function timeout(ms: number): Promise<void> {
+    return new Promise((resolve) => setTimeout(resolve, ms));
+}
+
+export function sleep(ms: number): Promise<void> {
+    return new Promise((resolve) => setTimeout(resolve, ms));
+}
```
addaleax, line 30:
```suggestion
export const sleep = timeout;
```

packages/test-utils/src/testUtils.ts
```diff
@@ -0,0 +1,42 @@
...
+            (process.env as Record<string, unknown>)[variable] = value;
+            registeredEnvVariables.push(variable);
+        },
+
+        clearVariables(): void {
+            for (const variable of registeredEnvVariables) {
+                delete (process.env as Record<string, unknown>)[variable];
+            }
+        },
+    };
```
addaleax, line 25: `useClearEnvironment()` restores previous values of environment variables, should this one do the same?

packages/tools-mongodb/src/args.ts
```diff
@@ -0,0 +1,14 @@
+import { z } from "zod";
+import { EJSON } from "bson";
+
+export function toEJSON<T extends object | undefined>(value: T): T {
```
addaleax: This does the exact opposite of what it says it does 🙂 

```suggestion
export function fromEJSON<T extends object | undefined>(value: T): T {
```

packages/tools-mongodb/src/args.ts
```diff
@@ -0,0 +1,14 @@
...
+    if (!value) {
+        return value;
+    }
+
+    return EJSON.deserialize(value, { relaxed: false }) as T;
+}
+
+export function zEJSON(): z.ZodRecord<z.ZodString, z.ZodUnknown> {
+    return z.object({}).loose().transform(toEJSON) as unknown as z.ZodRecord<z.ZodString, z.ZodUnknown>;
```
addaleax:
```suggestion
    return z.object({}).loose().transform(fromEJSON) as unknown as z.ZodRecord<z.ZodString, z.ZodUnknown>;
```

packages/ui/src/lib/tools/list-databases.ts
addaleax, line 1: Should this file be checked in to the repo?

CONTRIBUTING.md
```diff
@@ -127,7 +127,7 @@ npx @modelcontextprotocol/inspector -- node dist/esm/index.js
 
 ## Making public API changes
 
-To ensure no unintentional public API changes are introduced, the project uses [API Extractor](https://api-extractor.com/) to track the public-facing API across all package entry points (`.`, `./web`, `./tools`, `./ui`). The generated API report files live in `api-extractor/reports/` and are checked into source control.
+To ensure no unintentional public API changes are introduced, the project uses [API Extractor](https://api-extractor.com/) to track the public-facing API across package entry points (`.`, `./web`, `./ui`) and the internal tools surface (`src/tools/index.ts`, report only). The generated API report files live in `api-extractor/reports/` and are checked into source control.
```
addaleax: This update might already be outdated? 🙂

addaleax (approved):
No major concerns, and I imagine a lot of my comments aren't specific to this PR, but I've still done a full review here

packages/mongodb-mcp-server/package.json
nirinchev: The way our release workflow is setup, we'll trigger a release when we see an unreleased version. So we should bump it only if we want to also do a release.

packages/tools-mongodb/src/args.ts
gagik: I will call this `deserializeEJSON` to make the Zod chain below easier to read

packages/test-utils/src/testUtils.ts
gagik: yeah.. generally there's a lot of redundancy here (more obvious now that they're been batched together); I'd love to adopt a hook-based approach across the board eventually. I'd add this for now

packages/ui/src/lib/tools/list-databases.ts
gagik: It should be committed (we don't always build this, only on releases as part of generation script) but we could hide it in attributions. This can be done in a different PR though as this matches current state.

It's something worth improving for sure

packages/mongodb-mcp-server/src/index.ts
gagik: hm yeah. We were already doing this with ESM as our default and I suppose haven't had FIPS users to report anything. I'd like to defer it to a seperate PR if that's fine as this matches current state

packages/metrics/vitest.config.js.map
gagik: Nope... accident

packages/http-runners/src/utils.ts
gagik: added cleanup to timeout scenario

packages/ui/src/lib/tools/list-databases.ts
lerouxb: Hmm I ran into this question for my mcp apps skunkworks too. I copied what the mcp ui package does for now, but I don't like it.

addaleax (approved):

gagik:
Converting this to draft as we want to do a v1 release first before moving to v2.

gagik:
Moved to https://github.com/mongodb-js/mongodb-mcp-server/pull/1333

PR #1120, opened 2026-04-30
feat: MCP Cluster upgrade tool CLOUDP-397237

carolynmcca:
## Proposed changes
upgrade tool to enable upgrade for: Free→Flex, Free→M10, and Flex→M10

- Adds atlas-upgrade-cluster tool supporting Free→Flex, Free→M10, and Flex→M10 upgrade paths, with tier auto-detected from session state or Atlas API
- Provider and region default from session/cluster data with no hardcoded fallbacks; telemetry captures original tier, target tier, and cluster IDs
- Adds upgradeSharedTierCluster and upgradeFlexToDedicated to ApiClient, and extends Cluster and AtlasClusterConnectionInfo with instanceType, provider, and region

Tests

- [x]  pnpm test
- [x]  Manual verification via upgrading connection and non-connected clusters

## Checklist

- [x] I have signed the [MongoDB CLA](https://www.mongodb.com/legal/contributor-agreement)

PR #1123, opened 2026-04-30
feat: MCP Post-Connect Alerts Hook CLOUDP-397237

joyjwang:
## Proposed changes
Add a post-connect alerts hook to the MongoDB MCP Server for Atlas Free/Flex clusters.

- Adds a post-connect hook that calls GET /api/atlas/v2/groups/{groupId}/alerts on successful connection to a Free/Flex cluster, requesting only status=OPEN alerts.
- Filters alerts to OUTSIDE_METRIC_THRESHOLD / OUTSIDE_FLEX_METRIC_THRESHOLD and shared-tier metrics (CONNECTIONS_PERCENT, FLEX_CONNECTIONS_PERCENT, FLEX_DATA_SIZE_TOTAL, LOGICAL_SIZE) for the connected cluster, and surfaces an upgrade recommendation when relevant alerts exist.
- Emits a new shared tier alerts MCP telemetry event with tier and alert metadata when the hook finds matching alerts.

Testing
- [x] pnpm test
- [x] Manual verification via MCP Inspector and Cursor against Free/Flex clusters with and without active alerts

Telemetry event:
image
image

## Checklist


.idea/codeStyles/codeStyleConfig.xml
```diff
@@ -0,0 +1,5 @@
+<component name="ProjectCodeStyleConfiguration">
```
nirinchev: Should the .idea files be committed?

joyjwang: definitely not will be removing!!

src/tools/atlas/connect/connectCluster.ts
```diff
@@ -275,6 +279,23 @@ export class ConnectClusterTool extends AtlasToolBase {
                         });
                     }
 
+                    const atlas = this.session.connectedAtlasCluster;
+                    if (atlas !== undefined) {
+                        const hookResult = await runSharedTierAlertsHook({
```
Luke-Sanderson, line 305: Might be worth stacking this PR on top of #1120, this would allow you to pass `atlas.instanceType` into `runSharedTierAlertsHook` and not need to do the `inspectCluster` API call since you are already connected.

src/common/atlas/sharedTierAlertsHook.ts
```diff
@@ -0,0 +1,183 @@
...
+    id: string;
+    eventTypeName: string;
+    metricName: string;
+    clusterName: string;
+    status: string;
+    created?: string;
+    updated?: string;
+}
+
+function asRecord(alert: unknown): Record<string, unknown> {
```
Luke-Sanderson: Instead of `asRecord` and `readString` use the apiClient.

Using
```type AlertResult = Awaited<ReturnType<ApiClient["listAlerts"]>>["results"][number];```

Allows functions like `matchesFilters` to access `metricName` and other fields directly e.g.
```
function matchesFilters(
    alert: AlertResult,
    clusterName: string
): { eventTypeName: string; metricName: string } | undefined {
    const eventTypeName = String(alert.eventTypeName);
    const metricName = "metricName" in alert ? alert.metricName : undefined;
    const alertCluster = "clusterName" in alert ? alert.clusterName : undefined;
    if (!metricName || alertCluster !== clusterName) return undefined;
    if (!SHARED_TIER_EVENT_TYPES.has(eventTypeName) || !SHARED_TIER_METRICS.has(metricName)) return undefined;
    return { eventTypeName, metricName };
}

```

**Review by Luke-Sanderson — APPROVED** [review]
LGTM

**Review by gagik — CHANGES_REQUESTED** [review]
GitHub is laggy so I cannot add comments but the `matchesFilters` function should be replaced with a zod schema then a lot of these helper functions will become unnecessary

**Comment by carolynmcca** [comment]
@gagik Updated to use zod schema!

File: src/common/atlas/sharedTierAlertsHook.ts [hunk 4]
```diff
@@ -0,0 +1,110 @@
...
+        logger.warning({
+            id: LogId.atlasSharedTierAlertsHookWarning,
+            context: "shared-tier-alerts-hook",
+            message: `Failed to list Atlas alerts for shared-tier hook: ${message}`,
+        });
+        return null;
+    }
+
+    const results = data?.results;
+    if (!results?.length) {
```
nirinchev:
```suggestion
    const results = data.results;
    if (!results.length) {
```

We know that these are defined - why do we use optional chaining?

src/telemetry/types.ts
```diff
@@ -253,6 +253,9 @@ export type ConnectionMetadata = AtlasMetadata &
     AtlasLocalToolMetadata & {
         connection_auth_type?: string;
         connection_host_type?: string;
+        shared_tier_alerts_detected?: TelemetryBoolSet;
+        shared_tier_tier?: string;
+        shared_tier_alerts?: string[];
```
nirinchev: These are json stringified alerts - they contain a bunch of info, including cluster names, which is PII - we should by no means be sending that as telemetry.

src/tools/atlas/connect/connectCluster.ts
```diff
@@ -278,6 +282,24 @@ export class ConnectClusterTool extends AtlasToolBase {
...
+                            projectId: atlas.projectId,
+                            clusterName: atlas.clusterName,
+                            instanceType: atlas.instanceType,
+                            apiClient: this.apiClient,
+                            logger: this.session.logger,
+                        });
+                        if (hookResult !== null) {
+                            this.sharedTierHookResult = { tier: hookResult.tier, alerts: hookResult.alerts };
```
nirinchev: I understand why you had to do this, but we should not be storing temp state on the class for telemetry purposes. Instead, we should aim as much as possible to use the information from the tool result to emit telemetry info and if that's not possible, we'd have to change the design of the telemetry plumbing.

That being said, if we remove the unredacted alert reporting and instead report on presence/number of alerts, we'd be able to extract that info from the response, removing the need for instance-level state being stored.

nirinchev (changes requested):

src/telemetry/types.ts
carolynmcca: for this telemetry event, were sending:
        shared_tier_alerts_detected?: "true" | "false"
        shared_tier_tier?: "flex" | "free"
        shared_tier_alerts?: ["CONNECTIONS_PERCENT", "FLEX_CONNECTIONS_PERCENT", "FLEX_DATA_SIZE_TOTAL","LOGICAL_SIZE"]

src/tools/atlas/connect/connectCluster.ts
carolynmcca: Moved this conversation to slack https://mongodb.slack.com/archives/C0AKV9W2DKJ/p1778188917745549?thread_ts=1777556322.858389&cid=C0AKV9W2DKJ regarding exposing num of alerts vs alerts themselves

src/telemetry/types.ts
nirinchev: Maybe I'm missing some interaction here, but here's what I'm seeing:

1. In `sharedTierAlertsHook.ts:85`, we're iterating the alert results and constructing the `collectedById` map, containing the full alert data.
2. In `sharedTierAlertsHook.ts:93`, we're constructing the `collected` array from the map values.
3. In `sharedTierAlertsHook.ts:99`, we're json stringifying that data and creating the `alerts` array.
4. We're then returning the alerts from `runSharedTierAlertsHook`
5. In `connectCluster.ts:295`, we're setting `this.sharedTierHookResult = { tier: hookResult.tier, alerts: hookResult.alerts };`, storing the json stringified alerts array.
6. In `connectCluster.ts:362`, we're then constructing `sharedTierMetadata` with the alerts from the instance-level hook result we stashed in `execute`.

To be fair, I haven't tested it and am mostly going off of looking at the code and could be misinterpreting something, but if I am, it's not clear to me what.

carolynmcca: Thanks for catching!!I think this got messed up in commit history but just updated implementation w/ test case and added screenshots of the telemetry event in description

src/common/atlas/sharedTierAlertsHook.ts
```diff
@@ -0,0 +1,113 @@
...
+    );
+}
+
+/**
+ * Post-connect: inspect tier; for Free/Flex only, fetch OPEN alerts and return upgrade guidance when filters match.
+ * Returns tier, JSON-serialized alerts, and recommendation text for the caller to surface and attach to telemetry.
+ */
+export async function runSharedTierAlertsHook(
+    params: RunSharedTierAlertsHookParams
+): Promise<{ recommendationText: string; tier: "Free" | "Flex"; alerts: string[] } | null> {
```
nirinchev: In javascript, returning `undefined` rather than is more idiomatic.

src/common/atlas/sharedTierAlertsHook.ts
```diff
@@ -0,0 +1,113 @@
...
+                path: { groupId: projectId },
+                query: {
+                    status: "OPEN",
+                    itemsPerPage: LIST_ALERTS_PAGE_SIZE,
+                    pageNum: 1,
+                    includeCount: true,
+                },
+            },
+        })
+        .catch((err: unknown) => {
```
nirinchev: Rather than chaining a .catch here, it's better to wrap the entire method in a try-catch.

src/common/atlas/sharedTierAlertsHook.ts
```diff
@@ -0,0 +1,113 @@
...
+    const alerts = [...new Set(collected.map((item) => item.metricName))];
+
+    return {
+        recommendationText: buildRecommendationParagraph(
+            clusterName,
+            tier,
+            collected.map((a) => a.metricName)
+        ),
+        tier,
+        alerts,
```
nirinchev:
```suggestion
    const alerts = (data?.results ?? [])
        .map((alert) => SharedTierAlertSchema.safeParse(alert))
        .filter((a) => a.success && a.data.clusterName === clusterName)
        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
        .map((a) => a.data!.metricName);

    if (!alerts.length) {
        return null;
    }

    const tier = instanceType === "FREE" ? "Free" : "Flex";

    return {
        recommendationText: buildRecommendationParagraph(clusterName, tier, alerts),
        tier,
        alerts,
```

src/tools/atlas/connect/connectCluster.ts
nirinchev: My point here is - rather than store this as a instance-level value, try to use the output from `execute` to construct the telemetry payload. You have two options - one is to parse the content since that information is included in the text already or you can update the `connectCluster` tool to have structured output (you can see the pattern in `listDatabases` for example. Then in `resolveTelemetryMetadata`, you can access `result.structuredContent` and read the relevant fields.

src/common/atlas/sharedTierAlertsHook.ts
carolynmcca: Wanted to use a set since these alerts can have multiple instances for the same cluster (by default LOGICAL_SIZE alert is surfaced 3x for each node and a user can duplicate alerts)

src/tools/atlas/connect/connectCluster.ts
carolynmcca: Thanks! Took a look at `listDatabases` and much cleaner approach and tried to mirror

src/common/atlas/sharedTierAlertsHook.ts
nirinchev: You can still construct a set from the array:

```ts
const alerts = new Set(...(data?.results ?? [])
        .map((alert) => SharedTierAlertSchema.safeParse(alert))
        .filter((a) => a.success && a.data.clusterName === clusterName)
        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
        .map((a) => a.data!.metricName));
```

src/tools/atlas/connect/connectCluster.ts
```diff
@@ -278,7 +286,32 @@ export class ConnectClusterTool extends AtlasToolBase {
...
+                        });
+                        if (hookResult !== undefined) {
+                            content.push({
+                                type: "text",
+                                text: hookResult.recommendationText,
+                            });
+                            return {
+                                content,
+                                structuredContent: {
+                                    shared_tier_alerts_detected: "true" as const,
```
nirinchev:
```suggestion
                                    shared_tier_alerts_detected: "true",
```

I don't think we need `as const` here - do we?

src/common/atlas/sharedTierAlertsHook.ts
```diff
@@ -0,0 +1,98 @@
...
+    apiClient: ApiClient;
+    logger: LoggerBase;
+}
+
+function isSharedTierInstanceType(t: "FREE" | "FLEX" | "DEDICATED" | undefined): t is "FREE" | "FLEX" {
+    return t === "FREE" || t === "FLEX";
+}
+
+function buildRecommendationParagraph(clusterName: string, tier: "Free" | "Flex", metricNames: string[]): string {
+    const unique = [...new Set(metricNames)].sort();
```
nirinchev: `metricNames` is already unique - we don't need to construct a new set.

src/common/atlas/sharedTierAlertsHook.ts
```diff
@@ -0,0 +1,98 @@
...
+    );
+}
+
+/**
+ * Post-connect: inspect tier; for Free/Flex only, fetch OPEN alerts and return upgrade guidance when filters match.
+ * Returns tier, JSON-serialized alerts, and recommendation text for the caller to surface and attach to telemetry.
+ */
+export async function runSharedTierAlertsHook(
+    params: RunSharedTierAlertsHookParams
+): Promise<{ recommendationText: string; tier: "Free" | "Flex"; alerts: string[] } | undefined> {
```
nirinchev: For alerts, rather than return a list of strings, can we use the zod enum values? That way we have stronger guarantees about what goes in there.

src/tools/atlas/connect/connectCluster.ts
```diff
@@ -28,11 +29,18 @@ export const ConnectClusterArgs = {
     ),
 };
 
+const ConnectClusterOutputSchema = {
+    shared_tier_alerts_detected: z.enum(["true"]).optional(),
+    shared_tier_tier: z.enum(["Free", "Flex"]).optional(),
+    shared_tier_alerts: z.array(z.string()).optional(),
```
nirinchev: Similar to my previous comment, can we use a more constrained type in the output schema, as well as in the telemetry types?

src/tools/atlas/connect/connectCluster.ts
```diff
@@ -28,11 +29,18 @@ export const ConnectClusterArgs = {
     ),
 };
 
+const ConnectClusterOutputSchema = {
```
nirinchev, line 33: Structured content should contain the same information as `content` - right now it contains info about the alerts, but not the other details that go into `content` (that will require some manipulation) - e.g.

```ts
{
  addedCurrentIp: z.boolean(),
  addedCurrentIpExplanation: z.string().optional(),
}
```

and so on.

src/tools/atlas/connect/connectCluster.ts
```diff
@@ -28,11 +29,18 @@ export const ConnectClusterArgs = {
...
+    shared_tier_alerts_detected: z.enum(["true"]).optional(),
+    shared_tier_tier: z.enum(["Free", "Flex"]).optional(),
```
nirinchev: Structured content goes into the LLM context, so field names should be self-explanatory and follow our conventions (e.g. use camelCase). We can then transform the structured output to telemetry-friendly names.

src/tools/atlas/connect/connectCluster.ts
```diff
@@ -28,11 +29,18 @@ export const ConnectClusterArgs = {
     ),
 };
 
+const ConnectClusterOutputSchema = {
+    shared_tier_alerts_detected: z.enum(["true"]).optional(),
```
nirinchev: Why not z.boolean?

src/tools/atlas/connect/connectCluster.ts
nirinchev: This will also require tests to be updated to ensure content and structuredContent have equivalent information.

nirinchev (changes requested):

src/tools/atlas/connect/connectCluster.ts
carolynmcca: added to structuredContent but kept out of telemetry event

src/common/atlas/sharedTierAlertsHook.ts
```diff
@@ -0,0 +1,102 @@
...
+/**
+ * Post-connect: inspect tier; for Free/Flex only, fetch OPEN alerts and return upgrade guidance when filters match.
+ * Returns tier, JSON-serialized alerts, and recommendation text for the caller to surface and attach to telemetry.
+ */
+export async function runSharedTierAlertsHook(
+    params: RunSharedTierAlertsHookParams
+): Promise<{ recommendationText: string; tier: SharedTierTier; alerts: SharedTierMetricName[] } | undefined> {
+    const { projectId, clusterName, instanceType, apiClient, logger } = params;
+
+    if (!isSharedTierInstanceType(instanceType)) {
```
gagik:
```suggestion
    if (!["FREE", "FLEX"].includes(instanceType)) {
```
if a helper is defined, then it should either be complex or at least be used in 2 places IMO

src/common/atlas/sharedTierAlertsHook.ts
```diff
@@ -0,0 +1,102 @@
...
+
+/**
+ * Post-connect: inspect tier; for Free/Flex only, fetch OPEN alerts and return upgrade guidance when filters match.
+ * Returns tier, JSON-serialized alerts, and recommendation text for the caller to surface and attach to telemetry.
+ */
+export async function runSharedTierAlertsHook(
+    params: RunSharedTierAlertsHookParams
+): Promise<{ recommendationText: string; tier: SharedTierTier; alerts: SharedTierMetricName[] } | undefined> {
+    const { projectId, clusterName, instanceType, apiClient, logger } = params;
+
```
gagik:
```suggestion
    {projectId, clusterName, instanceType, apiClient, logger}: RunSharedTierAlertsHookParams
): Promise<{ recommendationText: string; tier: SharedTierTier; alerts: SharedTierMetricName[] } | undefined> {

```

src/tools/atlas/connect/connectCluster.ts
```diff
@@ -257,7 +280,7 @@ export class ConnectClusterTool extends AtlasToolBase {
             const state = this.queryConnection(projectId, clusterName);
             switch (state) {
                 case "connected": {
-                    const content: CallToolResult["content"] = [
+                    const content: { type: "text"; text: string }[] = [
```
gagik: I'd expect this not to be necessary

carolynmcca: Changed from CallToolResult["content"] because that broader type required as const casting on every type: "text" value; thought this was cleaner

gagik: > because that broader type required as const casting on every type

Casting as const, unlike other cases of casting, is generally a good thing.`as const` leads to stricter type definitions. So it's actually stricter and therefore better. In the future if `CallToolResult` changes `content` requirements, we'd want this to match

https://www.totaltypescript.com/concepts/as-const 

It's honestly bad that TypeScript decided to go with `as const` syntax for this because it feels like we're losing type safety here but we're actually enforcing it more.

carolynmcca: updated to use ToolResult instead of CallToolResult since return type was changed to add structuredContent

src/common/atlas/sharedTierAlertsHook.ts
```diff
@@ -0,0 +1,183 @@
...
+    let data;
+    try {
+        data = await apiClient.listAlerts({
+            params: {
+                path: { groupId: projectId },
+                query: {
+                    status: "OPEN",
+                    itemsPerPage: LIST_ALERTS_PAGE_SIZE,
+                    pageNum: 1,
+                    includeCount: true,
```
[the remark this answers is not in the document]
gagik: seems like a real comment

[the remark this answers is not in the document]
carolynmcca: set explicitly to false since default is includeCount: true

gagik (approved):

src/tools/atlas/connect/connectCluster.ts
```diff
@@ -28,11 +30,32 @@ export const ConnectClusterArgs = {
...
+    addedCurrentIp: z.boolean(),
+    createdTemporaryUser: z.boolean(),
+    sharedTierAlertsDetected: z.boolean().optional(),
+    sharedTierTier: z.enum(["Free", "Flex"]).optional(),
+    sharedTierAlerts: z.enum(SHARED_TIER_METRIC_NAMES).array().optional(),
+};
+
+export type ConnectClusterOutput = z.infer<z.ZodObject<typeof ConnectClusterOutputSchema>>;
+
+function transformTelemetryFormat(sc: ConnectClusterOutput): Partial<ConnectionMetadata> {
```
nirinchev: I don't think this needs to be a separate function - it can be inlined where we use it.

src/tools/atlas/connect/connectCluster.ts
```diff
@@ -28,11 +30,32 @@ export const ConnectClusterArgs = {
     ),
 };
 
+const ConnectClusterOutputSchema = {
+    connected: z.boolean(),
+    addedCurrentIp: z.boolean(),
+    createdTemporaryUser: z.boolean(),
```
nirinchev, line 36: Can we add

```ts
temporaryUserClarification: z.string().optional()
```

Since clients are free to send either content or structuredContent to the LLM, we should make sure to provide it with the details of what creating a temporary user means.

src/tools/atlas/connect/connectCluster.ts
```diff
@@ -317,19 +372,30 @@ export class ConnectClusterTool extends AtlasToolBase {
             });
         }
 
-        return { content };
+        return {
+            content,
+            structuredContent: {
+                connected: false,
```
nirinchev: We should probably change `connected` from boolean to an enum - e.g. `state: connected | connecting` and then add the clarification in the structured content. This is important because an LLM that only sees the structured content will need the additional info to be able to make the right decision about what to do.

src/common/atlas/sharedTierAlertsHook.ts
```diff
@@ -0,0 +1,102 @@
...
+ * Returns tier, JSON-serialized alerts, and recommendation text for the caller to surface and attach to telemetry.
+ */
+export async function runSharedTierAlertsHook({
+    projectId,
+    clusterName,
+    instanceType,
+    apiClient,
+    logger,
+}: RunSharedTierAlertsHookParams): Promise<
+    { recommendationText: string; tier: SharedTierTier; alerts: SharedTierMetricName[] } | undefined
```
nirinchev: Consider renaming `alerts` to `alertTypes` or something similar to more explicitly convey that these are just types of alerts rather than something that also contains details. I know the types already convey this, but when consuming the API, the field names should be self-describing.

nirinchev (changes requested):
Mostly nits, but updating `structuredContent` to contain the same level of detail as `content` is important before we merge this.

nirinchev (approved):

PR #1126, opened 2026-05-01
feat: add support for generating an mcpb package MCP-126

nirinchev:
## Proposed changes

This adds support for generating an .mcpb package (the new .dxt). I've tested it with Claude desktop on macOS and it worked just fine.


packaging/mcpb/manifest.json
```diff
@@ -0,0 +1,105 @@
...
+        "MDB_MCP_CONNECTION_STRING": "${user_config.connection_string}",
+        "MDB_MCP_API_CLIENT_ID": "${user_config.atlas_client_id}",
+        "MDB_MCP_API_CLIENT_SECRET": "${user_config.atlas_client_secret}",
+        "MDB_MCP_VOYAGE_API_KEY": "${user_config.voyage_api_key}",
+        "MDB_MCP_READ_ONLY": "${user_config.read_only}",
+        "MDB_MCP_DISABLED_TOOLS": "${user_config.disabled_tools}"
+      }
+    }
+  },
+  "user_config": {
```
nirinchev, line 58: Note the user config options - I didn't want to add every config option to avoid this becoming unwieldy, but open to feedback on other options we would like to have exposed.

himanshusinghs: Its worth checking how these options are interpreted because VSCode MCP installer, for example, would show users a prompt for each of the option provided regardless of it being required or not. If that is the case with the plugin as well then perhaps we should just expose connection_string and readonly?

nirinchev: I can add some screenshots, but at least in Claude Desktop, there are no prompts - you're taken to a config UI where you can set any of the values and save in a single go.

.github/workflows/publish.yml
```diff
@@ -86,14 +86,16 @@ jobs:
       - uses: pnpm/action-setup@v5
       - uses: actions/setup-node@v6
         with:
-          node-version-file: package.json
+          node-version: "24"
```
gagik, line 89: any reason for decoupling this?

packaging/mcpb/manifest.json
gagik: we should consider having "featured" options array in the generate script and using it in places like this.

gagik (approved):
didn't look too deeply but sgtm

.github/workflows/publish.yml
nirinchev: Mostly for convenience - package.json supports node 20, 22, and 24 and these are typescript files with erasable syntax, which is supported out of the box only in 24. With 20/22, we need to pass `--experimental-strip-types`, which is not the end of the world, but still slightly inconvenient.

nirinchev:
@copilot resolve the merge conflicts in this pull request

PR #1141, opened 2026-05-04
feat: add Braintrust-based single-turn LLM-as-judge eval framework CLOUDP-367319

nima-taheri-mongodb:
## 🎫 Ticket

[CLOUDP-367319](https://jira.mongodb.org/browse/CLOUDP-367319)

## 📝 Description

Adds a self-contained accuracy eval framework for the MongoDB MCP tools under `tests/eval/`, backed by [Braintrust](https://www.braintrust.dev/) for scoring & experiment tracking. 🧪

Each case runs a **single-turn** pipeline, per model:

1. **🌱 Seed** — spins up an isolated, per-case temp database (`eval_<uuid>`), loads bundled documents, and creates any indexes the case requests (incl. Atlas Search / Vector Search, waiting for readiness).
2. **🤖 Run** — the agent under test answers the prompt end-to-end using the MongoDB MCP tools (served **in-process** over an in-memory transport — no Docker-in-Docker, no stdio child).
3. **⚖️ Judge** — an LLM-as-judge scores the result against the case's `llm_judge` criteria, using **read-only** MCP tools plus synthetic helpers (`get_conversation`, `get_response`, `submit_score`).

### Key design choices

- **🗂️ Dataset-driven** — cases come from a Braintrust dataset (`initDataset`), with `input` / `expected` schemas generated from zod types (`pnpm eval:generate-schemas`) so they can be attached to the dataset in the Braintrust UI. This unblocks the "define evals in the UI" workflow raised in earlier review.
- **🎛️ Parameterised** — `connectionString`, `model`, and `systemContext` are Braintrust `Eval` parameters (overridable via UI, `BT_EVAL_PARAMS_JSON`, or env).
- **🧬 Single-turn only** — The agent gets one user turn and must finish autonomously.
- **♻️ DB lifecycle is external** — the eval never manages containers. `pnpm eval:db-start` / `eval:db-stop` run a healthy `mongodb-atlas-local` container locally (with an ngrok tip for remote/sandboxed runs if needed).
- **📦 Deployable** — `pnpm eval:push` pre-bundles the eval and pushes it as a Braintrust sandboxed Eval function that can be used in Braintrust Playground for convenient dataset refinements.

### Layout

| Path | Role |
| --- | --- |
| `tests/eval/mongodb.eval.ts` | Braintrust `Eval()` entry — passing parameters, dataset, task logic, scoring |
| `tests/eval/lib/datasetTypes.ts` | zod `input` / `expected` / `output` schemas and types for the eval |
| `tests/eval/lib/user.ts` | runs the single agent turn (via Vercel AI SDK `generateText`) |
| `tests/eval/lib/judge.ts` | LLM-as-judge over read-only MCP + synthetic tools (via Vercel AI SDK `generateText`) |
| `tests/eval/lib/mcp.ts` / `shared.ts` | in-process MCP client + Mongo client / temp-db registry |
| `tests/eval/lib/seeding.ts` | seeds collections, awaits search-index readiness |
| `tests/eval/lib/tool/*` | synthetic judge tools (`get_conversation`, `get_response`, `submit_score`) |
| `tests/eval/scripts/*` | DB lifecycle, schema generation, bundler for Braintrust push |

### Scripts

```bash
pnpm eval:db-start         # start local mongodb-atlas-local (waits until healthy and detaches)
pnpm eval:db-stop          # tear down the mongodb-atlas-local container
pnpm eval:run              # bt eval tests/eval/mongodb.eval.ts (runs the eval on current machine)
pnpm eval:debug            # tsx tests/eval/mongodb.eval.ts (runs the eval through direct TS entry, easier for debugging)
pnpm eval:push             # bundle + push as a Braintrust sandboxed Eval function (to be used in Braintrust Playground for dataset refinements)
pnpm eval:serve            # bt eval --dev tests/eval/mongodb.eval.ts (runs the eval as Braintrust dev server, could be used as a "remote eval" in Braintrust Playground)
pnpm eval:generate-schemas # emit dataset JSON schemas (to be applied to the Braintrust dataset input/expected schemas)
```

## 🧪 Documentation and Testing

- [x] Ran `pnpm eval:run` against a local `mongodb-atlas-local` instance ✅
- [x] Verified `pnpm eval:push` bundles + uploads the eval to Braintrust 🚀
- [x] `pnpm eval:generate-schemas` produces the dataset `input`/`expected` schemas 🗂️

---

### ✍️ Note

- A `BRAINTRUST_API_KEY` key is only environment variable that is required to run the eval, currently Braintrust does not support offline evaluation.
    - This key is used to both leverage Braintrust AI gateway which is a model-agnostic gateway for all the models (instead of Azure key that was previously needed for running tests)
    - Also used to send the evaluation results as an experiment to Braintrust.
- It's also possible to run the eval locally without the Braintrust API key, by using the `eval:debug` script.

### 🚀 Follow-up Work

- Integrating this to post-merge CI pipeline
    - Will include setting Braintrust API key on the github project settings
- Update Vercel AI SDK to the latest canary version (otherwise at least Gemini model has bugs and does not work)


tests/accuracy/sdk/constants.ts
```diff
@@ -1,9 +1,6 @@
 import path from "path";
-import { fileURLToPath } from "url";
 
-const __dirname = fileURLToPath(import.meta.url);
-
-export const ROOT_DIR = path.join(__dirname, "..", "..", "..", "..");
+export const ROOT_DIR = process.cwd();
```
nirinchev: Why did we change this - this will now depend on where the test is run from, which can cause issues.

tests/accuracy/eval/indexManagementTextSearchIndexCreation.eval.ts
```diff
@@ -0,0 +1,81 @@
+import { runEval } from "./infra/scaffolding.js";
+
+const context =
+    "You are a MongoDB expert. For all operations use the 'movies' collection by default.";
+
+const mflixMovies = {
+    collection: "movies",
+    documents: "tests/accuracy/test-data-dumps/mflix.movies-with-plot.json",
```
nirinchev: Not fully opposed to this, but should we instead set this up to use the default dataset instead?

tests/accuracy/eval/indexManagementTextSearchIndexCreation.eval.ts
```diff
@@ -0,0 +1,81 @@
...
+                                type: "search",
+                                name: "movies_plot_text",
+                                definition: { mappings: { fields: { plot: { type: "string" } } } },
+                            },
+                        ],
+                    }]
+                },
+            },
+            assertions:
+                "The assistant is expected to return at least 1 document, the first returned result should be the document with id 'fbf30e42-ae6d-4775-bb3e-c5c127ddea06' from 'movies' collection.",
```
nirinchev: [q] should we be more specific with the assertions? E.g. it seems to me that the assistant using `find` and then manually processing the documents to find the desired one will pass, even though it didn't follow the instructions. Should we be evaluating for tool usage and argument shapes or are we fine with treating the MCP server as a black box and as long as we get the desired results, we don't care how the model got to them?

tests/accuracy/eval/infra/scaffolding.ts
```diff
@@ -0,0 +1,184 @@
...
+    cluster: MongoDBClusterProcess;
+    seedClient: MongoClient;
+    accuracyClient: AccuracyTestingClient;
+}
+
+// Defers cluster/client startup until the first task actually runs. The `pending` promise prevents duplicate
+// initialization when multiple Braintrust tasks start concurrently before the first setup completes.
+function createLazyInfrastructure(
+    clusterConfig: MongoClusterConfiguration
+): [getInfra: () => Promise<EvalInfrastructure>, closeInfra: () => Promise<void>] {
```
nirinchev: [nit] it'd be more idiomatic to return an object rather than array here.

tests/accuracy/eval/infra/scaffolding.ts
```diff
@@ -0,0 +1,184 @@
...
+        return;
+    }
+
+    const models = getAvailableModels();
+    if (models.length === 0) {
+        console.log("No models available, skipping.");
+        return;
+    }
+
+    const [getInfra, closeInfra] = createLazyInfrastructure(clusterConfig);
```
nirinchev: This creates a separate cluster for each run, but the cluster itself is reused by models, right? Should we instead create clusters per test suite instead?

tests/accuracy/eval/infra/scaffolding.ts
```diff
@@ -0,0 +1,184 @@
...
+
+function defaultAccuracyUserConfig(): Partial<{ [k in keyof UserConfig]: string }> {
+    return {
+        apiClientId: process.env.MDB_API_CLIENT_ID,
+        apiClientSecret: process.env.MDB_API_CLIENT_SECRET,
+    };
+}
+
+function braintrustNoSendLogs(): boolean {
+    return !process.env.BRAINTRUST_API_KEY;
```
nirinchev: Does it even make sense to run the test suites if we don't have an API key here? What would the outcome be?

nirinchev (commented):
On a high-level, it looks reasonable, though I have some questions. Another question - my understanding was that with Braintrust, we are able to define evals in the UI and then we'd need to do some work to translate those into test cases that we run. Instead, the approach here is that we hardcode the evals in the MCP project and then only use braintrust for visualization of the results. Am I misunderstanding the value prop of Braintrust or should we aim to support this dynamic evals use case in some shape and form?

tests/accuracy/sdk/constants.ts
nima-taheri-mongodb: Fair, reverted it 🫡

Context: the reason why I made this change was because Braintrust bundler could not process this. but right now I pre-bundle the script before pushing it, which works.

tests/accuracy/eval/indexManagementTextSearchIndexCreation.eval.ts
nima-taheri-mongodb: The main problem with the default sample dataset is its large size, which leads the LLM to use arbitrary pagination strategies for queries. This makes it difficult to write evaluation tests with deterministic assertions.

To address this, we use a much smaller movies dataset: a 30-document subset from the Atlas sample data's Mflix movies collection. This is the same selection used in the MCP Server accuracy tests.

tests/accuracy/eval/indexManagementTextSearchIndexCreation.eval.ts
nima-taheri-mongodb: That's a great question. Our goal in the Eval suite is to evaluate whether the assistant achieves the correct end result for the user, without prescribing exact tool calls or argument formats for the most part. This gives us flexibility as MongoDB query syntax and MCP tools evolve over time—for example, a recent proposal had aimed to add specialized lexical or vector search tools.

That said, there are scenarios where certain tool usage matters for performance reasons. We're planning to add constraints to the LLM-judge prompt so it considers tool choices and argument patterns when scoring. For instance, we might want to penalize use of the `$regex` operator if used in aggregate pipelines, or require that a search index should have been used. We expect these kinds of requirements to be enforced via the LLM-judge prompt, which will influence the assigned score accordingly.

tests/accuracy/eval/infra/scaffolding.ts
nima-taheri-mongodb: Fair! I've done a major refactor of this code. It should now be cleaner now. 🤞

tests/accuracy/eval/infra/scaffolding.ts
nima-taheri-mongodb: Great point. I have reworked the evaluation logic so we no longer use a test container for every run. Instead, database lifecycle management is now handled externally by npm scripts (`eval:db-start`).

tests/accuracy/eval/infra/scaffolding.ts
nima-taheri-mongodb: Unfortunately, a Braintrust key is currently required to run the tests. After the rework, there's even a tighter coupling to Braintrust as we use Braintrust gateway as the model provider.

nima-taheri-mongodb:
> On a high-level, it looks reasonable, though I have some questions. Another question - my understanding was that with Braintrust, we are able to define evals in the UI and then we'd need to do some work to translate those into test cases that we run. Instead, the approach here is that we hardcode the evals in the MCP project and then only use braintrust for visualization of the results. Am I misunderstanding the value prop of Braintrust or should we aim to support this dynamic evals use case in some shape and form?

Great feedback, thank you! I've revised the approach so that prompts (datasets) are stored in Braintrust, and our eval job pulls those dynamically from Braintrust to run them.

nima-taheri-mongodb:
Going to bed now 😴 would be great to wake up and see some feedback from Nikola! 🤩

tests/eval/lib/tool/getConversation.ts
```diff
@@ -0,0 +1,130 @@
...
+
+/**
+ * Pass through the value if it is already a string, otherwise returns an empty string.
+ *
+ * @param value - The value to pass through.
+ * @returns The value if it is a string, otherwise an empty string.
+ */
+function partString(value: unknown): string {
+    return typeof value === "string" ? value : "";
+}
```
himanshusinghs: nit: on the function name - `ensureString` probably? I would not know what partString does just from the name :)

tests/eval/lib/tool/getConversation.ts
```diff
@@ -0,0 +1,130 @@
...
+    const blocks: string[] = [];
+    let turn = 0;
+
+    for (const msg of messages) {
+        const role = String((msg as { role?: string }).role ?? "unknown");
+        const content = (msg as Record<string, unknown>).content;
+        const inner: string[] = [];
+
+        if (typeof content === "string") {
+            if (content) inner.push(content);
```
himanshusinghs:
```suggestion
        if (typeof content === "string" && content) {
            inner.push(content);
```

tests/eval/lib/tool/getConversation.ts
```diff
@@ -0,0 +1,130 @@
...
+                        inner.push(`<tool_result for="${id}" name="${name}">${output}</tool_result>`);
+                        break;
+                    }
+                    default:
+                        inner.push(JSON.stringify(part));
+                }
+            }
+        }
+
+        if (inner.length === 0) continue;
```
himanshusinghs:
```suggestion
        if (inner.length === 0) {
            continue;
        }
```

Probably should enable prettier to do this.

tests/eval/lib/tool/submitScore.ts
```diff
@@ -0,0 +1,36 @@
...
+    });
+
+    getTool(): Tool {
+        return tool({
+            description: "Submit your final score. Call this exactly once when ready.",
+            inputSchema: SubmitScoreTool.scoreSchema,
+            execute: (input) => {
+                this.#captured = input;
+                return { ok: true };
+            },
```
himanshusinghs, line 28: nit: Perhaps not critical, but if it needs to be called exactly once then lets add a throw in execute when its called more than once?

tests/eval/lib/seeding.ts
```diff
@@ -0,0 +1,104 @@
...
+ * If the indexes are not queryable after the timeout, an error is thrown.
+ *
+ * @param client - The MongoDB client.
+ * @param db - The database name.
+ * @param collection - The collection name.
+ * @param indexNames - The names of the indexes to wait for.
+ * @param timeoutMs - The timeout in milliseconds to wait for the indexes to be queryable.
+ * @param intervalMs - The interval in milliseconds to check the indexes.
+ */
+async function waitForIndexesQueryable(
```
himanshusinghs, line 23: We do have these helpers in `tests/integration` folder somewhere which could be re-used. They are `vi` based though.

tests/eval/mongodb.eval.ts
```diff
@@ -0,0 +1,166 @@
...
+const DEFAULT_SYSTEM_CONTEXT =
+    'You are a MongoDB assistant operating autonomously in a single turn; the user cannot answer follow-up questions. Use the available MongoDB MCP tools to fulfill the request end-to-end. Never ask for clarification; make a reasonable decision and finish the task. If the request refers to "the collection" without naming it, discover collections with the list tools and act on the appropriate one (if there is exactly one user collection, use it). Prefer tools over guessing, and briefly confirm what you did when done.';
+
+if (!process.env.BRAINTRUST_API_KEY) {
+    console.warn("BRAINTRUST_API_KEY is not set.");
+}
+
+if (!process.env.OPENAI_BASE_URL) {
+    console.warn("OPENAI_BASE_URL is not set, using default value.");
+}
```
himanshusinghs: If these are not set, shouldn't we hard exit?

tsconfig.build.json
```diff
@@ -22,9 +22,6 @@
     "jsx": "react-jsx",
     "paths": {
       "ts-levenshtein": ["./node_modules/ts-levenshtein/dist/index.d.mts"],
-      "mongodb-connection-string-url": [
-        "./node_modules/mongodb-connection-string-url/lib/index.d.ts"
-      ],
```
himanshusinghs, line 27: have we started publishing types in latest versions?

himanshusinghs (approved):
Overall - looks good to me! I left small nits and a question.

tsconfig.build.json
nima-taheri-mongodb: I removed this line mainly because it was causing Braintrust eval/push to fail.

Could you share some background on why this config was originally needed? From what I can tell, the library's exports already point to the correct TypeScript definition file, so I'm not sure why the explicit configuration was required. 🤔

image

tests/eval/lib/tool/getConversation.ts
nima-taheri-mongodb: fair, `partString` wasn't that meaningful 😂

tests/eval/lib/tool/submitScore.ts
nima-taheri-mongodb: makse sense 🤔 would be a good to catch if LLM drifted from the instructions.

tests/eval/mongodb.eval.ts
nima-taheri-mongodb: That's a good point, as you recommended, I updated the code to fast-fail tasks if the required environment variables aren't set.

To clarify a bit more on how I got there. I originally chose to issue a warning because when you use Braintrust's "push" feature to bundle and submit the eval for sandboxed execution, the module is still loaded locally during the push process. At that time, any module-scoped variables are initialized, and if the required API keys aren't set, the push process fails, even though those environment variables are only actually necessary when the code runs remotely within Braintrust's sandbox (where they're set automatically).

To fail fast and as you suggested, although Braintrust doesn't currently provide hooks to execute code at eval startup, it is still possible to run this check per task but using a singleton pattern to ensure this logic runs just once per task. I updated the code to do that.

tests/eval/lib/seeding.ts
nima-taheri-mongodb: In my discussion with Nikola, we saw a future where all accuracy tests migrate to Braintrust-defined datasets and are evaluated using this framework. I intentionally kept this implementation self-contained and independent from the existing tests for a few reasons:

1. The Braintrust bundler is relatively fragile and can not be configured. Referencing files from the existing test codebase can cause the bundler to fail, for instance database seed files are currently located and loaded via paths which Braintrust bundler can not bundle and fails.
2. The current MCP server accuracy tests handle database lifecycle management internally using testcontainers. In this minimal framework, responsibility for database lifecycle is deliberately shifted to npm scripts. This gives it a bit more flexibility: Braintrust evals can connect to local containers for local runs, or be configured in the Braintrust sandbox to connect to an Atlas cluster, as desired by whoever is working on the dataset.
3. In a few places there's a bit of coupling to jasmine-based test framework APIs (vitest) which is not compatible with Braintrust Eval runner.

There might be further complexities in the existing accuracy tests, which is why I leaned on keeping it minimal and independent. I'm open to change this though, if you folks think otherwise.

himanshusinghs: Thanks for the context - the plan sounds nice 💯 Agree that we can have this intermediate state of duplication until we move entirely to Braintrust.

tsconfig.build.json
himanshusinghs: yea it looks good now, must have had updates recently. Earlier the types were not correctly specified and we had to manually specify type location to make tsc happy.

himanshusinghs (approved):

tests/eval/lib/judge.ts
```diff
@@ -0,0 +1,99 @@
...
+ * @returns The filtered tools.
+ */
+export function safeTools(tools: ToolSet): ToolSet {
+    const BLACK_LISTED_PREFIXES = ["create", "drop", "delete", "update", "insert"];
+    return Object.fromEntries(
+        Object.entries(tools).filter(([name]) => {
+            const lower = name.toLowerCase();
+            return !BLACK_LISTED_PREFIXES.some((prefix) => lower.startsWith(prefix));
+        })
+    );
```
nirinchev: This is brittle as at least some mutating tools don't start with these prefixes. A cleaner approach would be to get the judge a separate MCP server that is configured as readOnly: true.

tests/eval/mongodb.eval.ts
```diff
@@ -0,0 +1,148 @@
...
+                    prompt: input.prompt,
+                    tempDbName: dbName,
+                    stepLimit: AGENT_STEP_LIMIT,
+                });
+
+                let judge: RunEvalOutput["judge"];
+                const criteria = hooks.expected?.llm_judge;
+                if (criteria) {
+                    judge = await judgeUsingLLM({
+                        model,
```
nirinchev: This is using the same model for testing and then evaluation - we could consider using different ones to minimize any biases the model under evaluation has.

tests/eval/lib/shared.ts
```diff
@@ -0,0 +1,134 @@
...
+    }
+    return braintrustGatewayFactory.singletonInstance();
+}
+/**
+ * Gets the singleton in-memory MCP client instance.
+ *
+ * @param connectionString - The MongoDB connection string.
+ * @returns The MCP client.
+ */
+export async function getMcpClient(connectionString: string): Promise<McpClient> {
```
nirinchev: The connection string is ignored after the first call - as far as I can tell, this is okay for the current design, but is brittle and can lead to non-obvious errors further down the line. Consider caching the factories in a dictionary to ensure that consumers get the correct client.

nirinchev (commented):
Overall seems like a solid start. I have a couple of suggestions that I think we should try and address but those are not major blockers. We should also make sure to fix the tests.

tests/eval/lib/judge.ts
nima-taheri-mongodb: that's great suggestion!

tests/eval/mongodb.eval.ts
nima-taheri-mongodb: fair, it's pretty easy to make this configurable.

tests/eval/lib/shared.ts
nima-taheri-mongodb: Fair, I prefer enforcing a single connection string if you don't mind. Will create a global connection-string and use that in all getMcpServer, getReadOnlyMcpServer, getMongoDbClient shared tools.

nima-taheri-mongodb:
> We should also make sure to fix the tests.

I isolated these Evals into a separate directory (`tests/eval`) and did not make any change on existing unit/accuracy tests. The evals are also not integrated to CI. I don't expect any CI failure/issue caused by this PR.

nima-taheri-mongodb:
Thanks for the good feedback! Applied all your suggestions @nirinchev 🫡

nima-taheri-mongodb:
I don't know why I see irrelevant code changes in this PR on those commits where I merged `main` branch into mine 🤔 let me rebase on `main` and remove those merge commits.
—
these are the new commits that addresses Nikola's comments
[feat: use readOnly mcp-tool](https://github.com/mongodb-js/mongodb-mcp-server/pull/1141/commits/f4d64c305cc3c5daebbbd33ca20a43cc346c4db5)
[feat: allow configuring judgeModel separately](https://github.com/mongodb-js/mongodb-mcp-server/pull/1141/commits/3f6771c2128ab68a1345ba0796f34f170ba09c3c)
[feat: use a single shared connection string among other resources](https://github.com/mongodb-js/mongodb-mcp-server/pull/1141/commits/6a8ae5de0e12a8f9787355c81bec1defd4e57277)

PR #1149, opened 2026-05-05
refactor: move to @mongodb-js/mcp-transports package MCP-373

gagik:
## Summary

Extract transport runners and HTTP servers into a standalone `@mongodb-js/mcp-transports` package and remove old implementations.
### Transport Runners
- `StdioRunner` - For stdio transport
- `StreamableHttpRunner` - For HTTP transport with SSE/JSON responses  
- `DryRunModeRunner` - For dry-run mode testing

### HTTP Servers
- `MCPHttpServer` - MCP server over HTTP using Streamable HTTP transport
- `MonitoringServer` - Health check and metrics endpoints
- `ExpressBasedHttpServer` - Base class for Express-based servers

### Supporting Components
- `SessionStore` - In-memory session store with timeout management
- `InMemoryTransport` - For testing and dry-run modes
- JSON-RPC error codes

Part of the V2 monorepo refactoring effort.


tests/integration/metrics.test.ts
```diff
@@ -1,26 +1,239 @@
...
+    SessionStore,
+    MCPHttpServer,
+} from "../../src/lib.js";
+import { McpServer } from "@modelcontextprotocol/sdk/server/mcp.js";
+import { ApiClient } from "@mongodb-js/mcp-atlas-api-client";
+import { createAtlasLocalClient } from "@mongodb-js/mcp-tools-atlas-local";
+import { packageInfo } from "../../src/common/packageInfo.js";
+
+// Helper to create a full Server instance for tests
+async function createTestServer(
```
gagik, line 45: I will create better integration helpers in seperate PR

packages/core/src/runners/stdioRunner.ts
```diff
@@ -0,0 +1,87 @@
...
+ * class MyStdioRunner extends StdioRunner {
+ *   protected override async createServer({ serverOptions, sessionOptions }) {
+ *     return new MyServer({ ... });
+ *   }
+ * }
+ *
+ * const runner = new MyStdioRunner({ loggers, metrics });
+ * ```
+ */
+export class StdioRunner<
```
nirinchev, line 19: Why is the stdio runner in a package called http-transports?

packages/types/src/transports.ts
```diff
@@ -0,0 +1,84 @@
+import type { ILogger } from "./logging.js";
+import type { IMetrics, MetricDefinitions } from "./metrics.js";
+
+/**
+ * Configuration for HTTP servers used by transport runners.
```
nirinchev: This comment seems to be misplaced.

packages/http-runners/package.json
```diff
@@ -0,0 +1,40 @@
+{
+  "name": "@mongodb-js/mcp-http-transports",
+  "description": "HTTP transport runners and servers for MongoDB MCP servers",
```
nirinchev, line 3: It seems this package is closer to a "runners" package rather than "transports" package as it contains mostly types related to running the server, while the transports implementations are being pulled from the SDK.

src/transports/dryModeRunner.ts
```diff
@@ -1,53 +1,103 @@
...
+    }
+
+    protected async createServer(): Promise<TServer> {
+        return Promise.resolve(this.server);
     }
 
     override async start({
+        // eslint-disable-next-line @typescript-eslint/no-unused-vars
         serverOptions,
+        // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
nirinchev: [nit] no need to decompose those if we're not going to use them.

src/transports/dryModeRunner.ts
```diff
@@ -1,53 +1,103 @@
...
     }
 
-    private dumpConfig(): void {
-        this.consoleLogger.log("Configuration:");
-        this.consoleLogger.log(JSON.stringify(this.userConfig, null, 2));
+    /**
+     * Stops the dry run mode runner.
+     */
+    override async stop(): Promise<void> {
+        await this.server?.close();
```
nirinchev:
```suggestion
        await this.server.close();
```

src/transports/dryModeRunner.ts
```diff
@@ -1,53 +1,103 @@
...
+ *   loggers: [consoleLogger],
+ *   userConfig: defaultTestConfig,
+ *   server: myServer,
+ * });
+ * await runner.start();
+ * ```
+ */
+export class DryRunModeRunner<
+    TServer extends DryRunServer = DryRunServer,
+    TContext = unknown,
```
nirinchev: Why do we need these generics for the DryRunModeRunner?

src/transports/dryModeRunner.ts
```diff
@@ -1,53 +1,103 @@
...
+};
 
-export class DryRunModeRunner extends TransportRunnerBase {
-    private server: Server | undefined;
-    private consoleLogger: DryRunModeTestHelpers["logger"];
+/**
+ * Transport runner for dry-run mode.
+ * Dumps configuration and enabled tools, then exits without starting the server.
+ *
+ * This is defined in the binary because it depends on UserConfig.
```
nirinchev: This information shouldn't be in the jsdoc comment - we can add it separately.

temp/web.public.api.md
```diff
@@ -39,6 +39,8 @@ import { ConnectionStringInfo } from '@mongodb-js/mcp-tools-mongodb';
 import { ConnectionTag } from '@mongodb-js/mcp-tools-mongodb';
```
nirinchev: [q] Should the temp dir be committed?

tests/unit/sessionStore.test.ts
```diff
@@ -27,7 +28,7 @@ describe("SessionStore metrics", () => {
         store = new SessionStore({
             options: { idleTimeoutMS: 60_000, notificationTimeoutMS: 30_000 },
             logger,
-            metrics,
+            metrics: metrics as IMetrics<DefaultMetricDefinitions>,
```
nirinchev: This is fishy - I tried to follow the logic with the metrics definitions, but it's a bit too convoluted with a whole lot of generics and type definitions, so got a bit lost. But the notion that `PrometheusMetrics<DefaultMetrics>`  doesn't implement `IMetrics<DefaultMetricDefinitions>` makes me think there's some underlying problem with the way we've setup the type hierarchies.

packages/types/src/metrics.ts
```diff
@@ -1,12 +1,60 @@
...
     get<K extends keyof TMetricsDefinitions>(key: K): TMetricsDefinitions[K];
+    /** Get all metrics as a formatted string (e.g., Prometheus format) */
     getMetrics(): Promise<string>;
 }
 
+/**
+ * Type alias for any metrics interface.
+ * Use this when the specific metric definitions don't matter.
+ */
 export type AnyMetrics = IMetrics;
```
nirinchev: That doesn't seem to be used anywhere - do we need it?

packages/types/src/metrics.ts
```diff
@@ -1,12 +1,60 @@
+/**
+ * Base type for metric definitions.
+ * Each key is a metric name, and the value is the metric type.
+ */
 export type MetricDefinitions = Record<string, unknown>;
```
nirinchev: Should the value type be unknown here or should it be `ICounter | IObservable`?

packages/http-transports/src/mcpHttpServer.ts
```diff
@@ -81,6 +93,25 @@ export class MCPHttpServer<
...
     }
 
+    /**
+     * Creates a new server instance. Override this method in subclasses
+     * to customize server creation for each new session.
+     */
+    // eslint-disable-next-line @typescript-eslint/require-await -- Subclasses may need async
+    protected async createServer(): Promise<TServer> {
+        throw new Error("MCPHttpServer.createServer() must be overridden in a subclass");
+    }
```
nirinchev: Should we mark this class as abstract then? It seems counterintuitive that we have throwing methods that don't enforce the need to be overridden at compile time.

packages/http-transports/src/mcpHttpServer.ts
```diff
@@ -81,6 +93,25 @@ export class MCPHttpServer<
...
+
+    /**
+     * Creates a server instance for a specific request. Override this method
+     * in subclasses to customize per-request server creation. The default
+     * implementation delegates to createServer().
+     */
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars -- Used by subclasses
+    protected async createServerForRequest(request: TransportRequestContext): Promise<TServer> {
```
nirinchev: Do we need two methods for this - `createServer` and `createServerForRequest`? It seems to me like we can make this one abstract and remove `createServer` altogether.

packages/http-runners/src/mcpHttpServer.ts
```diff
@@ -117,51 +148,54 @@ export class MCPHttpServer<
...
-                });
-
-                await transport.send({
-                    jsonrpc: "2.0",
-                    method: "ping",
-                });
-                failedPings = 0;
-            } catch (err) {
+        const keepAliveLoop = setInterval((): void => {
+            void (async (): Promise<void> => {
```
nirinchev: Not sure I understand the change here - was there a problem with the previous implementation?

packages/http-transports/src/mcpHttpServer.ts
```diff
@@ -117,51 +148,54 @@ export class MCPHttpServer<
...
-            }
+            })();
         }, 30_000);
 
         return keepAliveLoop;
     }
 
+    private getRandomUUID(): string {
+        return crypto.randomUUID();
+    }
```
nirinchev: Why are we extracting this to a separate method?

packages/http-transports/src/mcpHttpServer.ts
```diff
@@ -1,78 +1,90 @@
...
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars -- Used by subclasses
+    TServer = unknown,
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars -- Used by subclasses
+    TContext = unknown,
+    TMetrics extends MetricDefinitions = MetricDefinitions,
+> = {
+    /** HTTP server configuration */
+    httpOptions: HttpServerConfig;
+    /** Session management configuration */
+    sessionOptions: SessionManagementConfig;
```
nirinchev: The naming here is a bit awkward - the types are called xyz**Config**, but the fields are xyz**Options**.

nirinchev (commented):
Haven't made my way through all of it, but figured I'd drop some comments. The direction seems okay, but I'm getting a little overwhelmed by the amount of generics and levels in the type hierarchies. Should we look for opportunities to simplify things?

tests/unit/sessionStore.test.ts
gagik: yeah i'm looking into this

packages/http-runners/src/mcpHttpServer.ts
gagik: I think type checking or eslint wasn't setup properly. You aren't meant to pass async functions to `setInterval`

src/transports/dryModeRunner.ts
gagik: TransportRunnerBase is generic, and generally we'd expect to be able to use in a type safe manner no matter what metrics/context we pass there

packages/core/src/runners/stdioRunner.ts
gagik: moved to core

packages/http-runners/package.json
gagik: renamed

src/transports/dryModeRunner.ts
gagik: we might be able to combine the concept of "toolContext" into a custom generic session now

nirinchev: Right, I was mostly thinking about TServer and TContext - any reason why we can't hardcode those?

packages/http-runners/src/mcpHttpServer.ts
nirinchev: Why though? The new version seems to be doing the same, just with wrapping it into an extra function.

gagik: my best guess is that tsconfig wasn't wired up to eslint properly for this file

gagik: ah. this used to have: 
```
// eslint-disable-next-line @typescript-eslint/no-misused-promises
```
I would rather we do this IMO because it makes it explicit we're doing something hacky.

gagik: as to why: I guess it's to make the potential side effects clearer? As in the next setInterval might fire quicker than whatever awaited thing inside the function

gagik:
I think there's some OS subtleness going on with api extractor which breaks what is generated on Mac vs Linux. This is the case in upstream v2; Zod seems like the likeliest cultpit. Going to ignore that in this PR.

packages/types/src/metrics.ts
gagik: it can be left more flexible but added `DefaultMetricDefinitions`

src/transports/dryModeRunner.ts
gagik: removed TContext from everything, made DryModeRunner non-generic

packages/core/src/logId.ts
```diff
@@ -52,9 +56,9 @@ export const LogId = {
     toolValidationError: mongoLogId(1_005_004),
 
     streamableHttpTransportStarted: mongoLogId(1_006_001),
-    streamableHttpTransportSessionCloseFailure: mongoLogId(1_006_002),
```
gagik, line 55: changing names because their usage is independent of streamable HTTP

gagik:
Disabling check:api for tests file. Not like the check is useful at this stage, manual review of the API file is a lot more insightful. My best guess is that Zod has some non-determinism in how api-extractor on Linux vs Mac generates the files for it. Will look into re-enabling later on, the tools entry point won't exist soon anyway.

packages/core/src/sessionStore.ts
```diff
@@ -147,14 +136,15 @@ export class SessionStore<T extends CloseableTransport = CloseableTransport> imp
...
+                    id: LogId.sessionCloseFailure,
                     context: "streamableHttpTransport",
                     message: `Error closing transport ${sessionId}: ${error instanceof Error ? error.message : String(error)}`,
                 });
             }
         }
 
-        this.metrics.get("sessionClosed").inc({ reason: reason });
+        // Track session closed metric
+        this.metrics.get("sessionClosed").inc({ reason });
```
nirinchev, line 147: If I have constructed the server with `NoopMetrics`, would that not throw?

packages/core/src/transportRunnerBase.ts
```diff
@@ -0,0 +1,41 @@
...
+import type { IMetrics } from "@mongodb-js/mcp-types";
+import type { MetricDefinitions, DefaultMetricDefinitions } from "./transports.js";
+
+/**
+ * Base class for all transport runners.
+ * Provides common lifecycle management (start/stop/close) and logging.
+ */
+export abstract class TransportRunnerBase<TMetrics extends MetricDefinitions = DefaultMetricDefinitions> {
+    public logger: CompositeLogger;
+    public metrics: IMetrics<TMetrics>;
```
nirinchev: Do these need to be public? And should they be readonly?

packages/core/src/transports.ts
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * Base configuration options for all transport runners.
+ */
+export type TransportRunnerBaseOptions<TMetrics extends MetricDefinitions = DefaultMetricDefinitions> = {
+    /** Logger instance */
+    logger: CompositeLogger;
+
+    /** Metrics instance */
+    metrics: IMetrics<TMetrics>;
```
nirinchev: These docstrings make me nostalgic of the good old C# days:

```csharp
/// <summary>
/// Gets or sets the logger.
/// </summary>
public ILogger Logger { get; set; }
```

They're not terrible, but are not particularly helpful either. If the expectation is that humans/agents would read them, can we ask Claude to generate something more meaningful than just repeating the field name?

packages/http-runners/src/expressBasedHttpServer.ts
```diff
@@ -1,30 +1,31 @@
...
+ * Provides common functionality for starting and stopping HTTP servers.
+ */
 export abstract class ExpressBasedHttpServer {
     protected httpServer: http.Server | undefined;
     protected app: express.Express;
 
-    protected readonly logger: LoggerBase;
+    protected readonly logger: ILogger;
     protected readonly logContext: string;
+    public readonly httpOptions: HttpServerOptions;
```
nirinchev: Same question - do we need this to be public? One thing we could do to gauge it is to make docs mandatory for public fields and then we'd be forced to decide whether exposing things publicly is justified.

packages/http-runners/src/index.ts
```diff
@@ -0,0 +1,18 @@
...
+export { MCPHttpServer, type MCPHttpServerOptions } from "./mcpHttpServer.js";
+export {
+    MonitoringServer,
+    type MonitoringServerOptions,
+    type MonitoringServerConstructorParams,
+    type MonitoringServerDependencies,
+} from "./monitoringServer.js";
+
+// Express HTTP Server base
+export { ExpressBasedHttpServer, type ExpressBasedHttpServerOptions } from "./expressBasedHttpServer.js";
```
nirinchev, line 11: We probably want to switch away from express while we're at it, but that can be a different PR

packages/http-runners/src/mcpHttpServer.ts
```diff
@@ -1,78 +1,90 @@
...
+import { sleep } from "./utils.js";
+
+/**
+ * Minimum server interface required by MCPHttpServer.
+ * Servers must have connect/close methods and a session with a logger for HTTP transport functionality.
+ */
+export type SessionAwareServer = {
+    connect(transport: StreamableHTTPServerTransport): Promise<void>;
+    close(): Promise<void>;
+    session?: { logger: ICompositeLogger };
```
nirinchev: Why is the session optional here?

packages/http-runners/src/mcpHttpServer.ts
```diff
@@ -115,20 +133,24 @@ export class MCPHttpServer<
...
+    private async startKeepAliveLoop({
+        transport,
+        logger,
+        signal,
+    }: {
+        transport: StreamableHTTPServerTransport;
+        logger: ILogger;
+        signal: AbortSignal;
+    }): Promise<void> {
+        if (this.httpOptions.responseType === "json") {
```
nirinchev, line 144: Would be good to keep the comment for why we're not starting the loop for `json` responseType.

packages/http-runners/src/monitoringServer.ts
```diff
@@ -0,0 +1,105 @@
...
+    metrics: IMetrics<TMetrics>;
+};
+
+/**
+ * Complete constructor params for creating a MonitoringServer instance.
+ */
+export type MonitoringServerConstructorParams<TMetrics extends MetricDefinitions = DefaultMetricDefinitions> = {
+    /** Options for configuring the monitoring server */
+    options: MonitoringServerOptions;
+} & MonitoringServerDependencies<TMetrics>;
```
nirinchev: Why are those different types?

packages/types/src/transports.ts
```diff
@@ -0,0 +1,74 @@
...
+/**
+ * Options for configuring the monitoring server.
+ */
+export type MonitoringServerOptions = {
+    /** HTTP server options */
+    http: HttpServerOptions;
+    /** Features to enable on the monitoring server */
+    features: MonitoringServerFeature[];
+};
+
```
nirinchev: This seems to be the almost the same as the types in monitoringServer.ts - it's confusing to me which one should be used when.

packages/http-runners/src/monitoringServer.ts
```diff
@@ -0,0 +1,105 @@
...
+    /** HTTP configuration */
+    http: {
+        /** Host to bind the monitoring server to */
+        host: string;
+        /** Port to bind the monitoring server to */
+        port: number;
+    };
+    /** Features to enable on the monitoring server */
+    features: MonitoringServerFeature[];
+};
```
nirinchev: Seems to only be used in `MonitoringServerConstructorParams` - any reason to have this as a dedicated type?

packages/http-runners/src/monitoringServer.ts
```diff
@@ -0,0 +1,105 @@
...
+ *     await super.setupRoutes();
+ *   }
+ * }
+ * ```
+ */
+export class MonitoringServer<
+    TMetrics extends MetricDefinitions = DefaultMetricDefinitions,
+> extends ExpressBasedHttpServer {
+    private readonly features: MonitoringServerFeature[];
+    private readonly metrics: IMetrics<TMetrics>;
```
nirinchev: If class is supposed to be overriden, should these be protected instead?

packages/http-runners/src/streamableHttp.ts
```diff
@@ -0,0 +1,98 @@
...
+            id: LogId.streamableHttpTransportStarted,
+        });
+    }
+
+    /**
+     * Stops the HTTP transport runner.
+     * This stops the MCP HTTP server and monitoring server.
+     */
+    async stop(): Promise<void> {
+        await Promise.all([this.mcpHttpServer?.stop(), this.monitoringServer?.stop()]);
```
nirinchev:
```suggestion
        await Promise.allSettled([this.mcpHttpServer.stop(), this.monitoringServer?.stop()]);
```

packages/http-runners/src/streamableHttp.ts
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    private shouldWarnAboutHttpHost(httpHost: string): boolean {
+        const host = httpHost.trim();
+        const safeHosts = new Set(["127.0.0.1", "localhost", "::1"]);
+        return host === "0.0.0.0" || host === "::" || (!safeHosts.has(host) && host !== "");
```
nirinchev: This check seems duplicative - the last condition covers the first two already.

packages/http-runners/src/streamableHttp.ts
```diff
@@ -0,0 +1,98 @@
...
+ * To customize per-request server creation, extend `MCPHttpServer` and override
+ * the `createServerForRequest()` method.
+ */
+export class StreamableHttpRunner<
+    TServer extends SessionAwareServer = SessionAwareServer,
+    TMetrics extends MetricDefinitions = DefaultMetricDefinitions,
+> extends TransportRunnerBase<TMetrics> {
+    protected mcpHttpServer: MCPHttpServer<TServer, TMetrics>;
+    protected monitoringServer: MonitoringServer<TMetrics> | undefined;
+    protected sessionStore: ISessionStore<StreamableHTTPServerTransport>;
```
nirinchev: These can be readonly.

packages/http-runners/src/streamableHttpRunner.ts
```diff
@@ -0,0 +1,98 @@
...
+
+/**
+ * Transport runner for HTTP transport with streamable responses.
+ * Supports both SSE and JSON response types.
+ *
+ * Server creation is handled by the `MCPHttpServer` instance passed to the constructor.
+ * To customize per-request server creation, extend `MCPHttpServer` and override
+ * the `createServerForRequest()` method.
+ */
+export class StreamableHttpRunner<
```
nirinchev, line 29: We probably want to rename the file to `streamableHttpRunner.ts` to match the type name and distinguish it from other files which are utilities or components the runner uses.

packages/core/src/transportRunnerBase.ts
```diff
@@ -0,0 +1,41 @@
...
+
+    protected constructor({ logger, metrics }: { logger: CompositeLogger; metrics: IMetrics<TMetrics> }) {
```
nirinchev: I don't get why we need the metrics in the base class? Doesn't seem like those are used for anything - we're just stashing them for inheritors to use, which they don't necessarily need. My suggestion would be to keep the interface minimal and only define fields/params that the base class actually uses.

packages/core/src/runners/stdioRunner.ts
```diff
@@ -0,0 +1,67 @@
...
+> extends TransportRunnerBase<TMetrics> {
+    private server: TServer;
+
+    constructor({
+        logger,
+        metrics,
+        server,
+    }: {
+        logger: CompositeLogger;
+        metrics: IMetrics<TMetrics>;
```
nirinchev: This runner doesn't seem to use the metrics instance either - we should be able to drop it.

src/server.ts
```diff
@@ -348,7 +318,6 @@ export class Server<
                 elicitation: this.elicitation,
                 metrics: this.metrics as IMetrics,
```
nirinchev: Do we need the cast here?

packages/core/src/toolBase.ts
```diff
@@ -313,8 +289,7 @@ export type ToolClass<
  */
 export abstract class ToolBase<
     TUserConfig extends IToolConfig = IToolConfig,
-    TContext = unknown,
-    TMetricsDefinitions extends MetricDefinitions = MetricDefinitions,
+    TMetricsDefinitions extends MetricDefinitions = DefaultMetricDefinitions,
```
nirinchev: This should not extend `MetricDefinitions` but rather `DefaultMetricDefinitions` since the tool explicitly tries to access values that would throw if the instance we get does not extend it - e.g. on line 510, we have

```
(this.metrics.get("toolExecutionDuration") as IObservable).observe(
```

which will throw if I pass in something different.

packages/core/src/metrics/noopMetrics.ts
```diff
@@ -0,0 +1,27 @@
...
+ *   metrics: new NoopMetrics(),
+ *   server: myServer,
+ * });
+ * await runner.start();
+ * ```
+ */
+export class NoopMetrics<TMetrics extends MetricDefinitions = DefaultMetricDefinitions> implements IMetrics<TMetrics> {
+    /** Returns undefined for any metric */
+    get<K extends keyof TMetrics>(): TMetrics[K] {
+        return undefined as TMetrics[K];
```
nirinchev: This is lying about the types and will cause a consumer who gets a `NoopMetrics` instance to throw when it tries to access a metric and run operations on it.

packages/types/src/metrics.ts
```diff
@@ -1,12 +1,58 @@
+/**
+ * Base constraint for metric definitions.
+ */
 export type MetricDefinitions = Record<string, unknown>;
```
nirinchev: As far as I can tell, there's only one place where we're legitimately using `MetricDefinitions` rather than `DefaultMetricDefinitions` - the monitoring server, which only does `getMetrics` but doesn't access any of the gauges/counters. Since the monitoring server cannot exist on its own and everywhere else we're tied to the default metric definitions, my suggestion would be to remove this type as it gets confusing which one to use when.

nirinchev (commented):
I'm at 82/122 files reviewed, but starving, so figured I'll drop whatever I have so far and will continue later today.

packages/core/src/sessionStore.ts
gagik: hmm, I was kind of confident in it as we were using NoopMetrics in a few tests but you're generally right

packages/core/src/transports.ts
gagik: These agents almost never find the balance between the 2, it's either overly verbose or useless. I would rather do this as a follow-up since the scope of changes is large as is, unless this is actively harmful somehow

packages/http-runners/src/expressBasedHttpServer.ts
gagik: it was being accessed in streamablehttp but that's a codesmell; will move this out + move the warning to a lower level

packages/http-runners/src/index.ts
gagik: I'd first merge v2 branch and only then move away

packages/core/src/runners/stdioRunner.ts
```diff
@@ -0,0 +1,57 @@
...
+export class StdioRunner<
+    TServer extends {
+        connect(transport: StdioServerTransport): Promise<void>;
+        close(): Promise<void>;
+    } = {
+        connect(transport: StdioServerTransport): Promise<void>;
+        close(): Promise<void>;
+    },
+> implements ITransportRunner {
+    private server: TServer;
```
nirinchev:
```suggestion
    private readonly server: TServer;
```

packages/core/src/runners/stdioRunner.ts
```diff
@@ -0,0 +1,57 @@
...
+    public readonly logger: CompositeLogger;
```
nirinchev: Should this be private/protected?

packages/core/src/transports.ts
```diff
@@ -0,0 +1,22 @@
+import type { McpServer } from "@modelcontextprotocol/sdk/server/mcp.js";
+import type { DefaultMetricDefinitions, IMetrics, LogLevel } from "@mongodb-js/mcp-types";
+import type { CompositeLogger } from "./logging/compositeLogger.js";
+
+/**
+ * Options for creating an MCP server instance.
+ * Passed to the `createServer()` method in transport runners.
+ */
+export type ServerOptions<TMetrics extends DefaultMetricDefinitions = DefaultMetricDefinitions> = {
```
nirinchev:
```suggestion
export type ServerOptions<TMetricsDefinitions extends DefaultMetricDefinitions = DefaultMetricDefinitions> = {
```

src/index.ts
```diff
@@ -96,68 +153,125 @@ async function main(): Promise<void> {
...
+        });
+
+        const mcpHttpServer = new MongoDBMCPHttpServer({
+            userConfig: config,
+            options: {
+                http: {
+                    host: config.httpHost,
+                    port: config.httpPort,
+                    responseType: config.httpResponseType,
+                    headers: config.httpHeaders as Record<string, string> | undefined,
```
nirinchev: [nit] Can add `.catchall(z.string())` to the httpHeaders zod schema to remove the need for the cast.

nirinchev (approved):
Final nits. Haven't done a very thorough review of the tests as I assume those are mostly API adjustments and not something fundamental changing there.

PR #1151, opened 2026-05-06
chore: fixes test setup for getRandomUUID

himanshusinghs:
## Proposed changes
`new bson.UUID` depends on some other APIs from WebCrypto which the original test setup was also removing. This change removes just the randomUUID function from WebCrypto API to simulate the correct conditions for the fallback.

## Checklist


tests/e2e/getRandomUUID.test.ts
```diff
@@ -122,14 +122,8 @@ import("../../../dist/esm/helpers/getRandomUUID.js").then(({ getRandomUUID }) =>
...
-// Mock globalThis.crypto without randomUUID using Object.defineProperty
-Object.defineProperty(globalThis, 'crypto', {
-    value: {
-        getRandomValues: function() {}
-    },
-    writable: true,
-    configurable: true
-});
+// Mock globalThis.crypto without randomUUID
+delete globalThis.crypto.randomUUID;
```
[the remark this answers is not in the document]
himanshusinghs: ahh yea that's true! Accepting the suggestion

tests/e2e/getRandomUUID.test.ts
gagik: we can move this to browser tests now

himanshusinghs: have moved these in my latest commit

tests/browser/tests/getRandomUUID.test.ts
```diff
@@ -0,0 +1,65 @@
...
+            Object.defineProperty(globalThis, "crypto", {
+                value: originalCrypto,
+                configurable: true,
+                writable: true,
+            });
+            originalCrypto = undefined;
+        }
+    });
+
+    it("should use Web Crypto API and return a valid UUID", () => {
```
lerouxb, line 21: Shouldn't this test do a sanity check and skip itself if this isn't a "real browser"? Or do we not run the tests in that kind of setup ever?

himanshusinghs: these particular tests are browser dedicated so they always run in a browser environment (playwright)

lerouxb (approved):

PR #1165, opened 2026-05-07
chore: remove forgotten export warnings from api report

nirinchev:
## Proposed changes

Noticed api exports contain line numbers for forgotten exports, which is super brittle as any change to the file would result in the report having to be updated even if no API have actually changed.


gagik (approved):

gagik:
I'd like these to be in the console but not the file generally, I think they can be useful for the future but don't matter right now

PR #1167, opened 2026-05-08
chore: MCP Expose Flex clusters in list tool CLOUDP-397238

carolynmcca:
## Proposed changes
After QA session for upgrade tool and alerts hook, we saw issues:

- [x] List tool does not show flex clusters
- [x] Add more detail to upgrade tool regarding valid cluster configurations

## Checklist

### DIFF
--- a/src/tools/atlas/read/listClusters.ts
+++ b/src/tools/atlas/read/listClusters.ts
@@ -41,15 +41,27 @@ export class ListClustersTool extends AtlasToolBase {
                 throw new Error(`Project with ID "${projectId}" not found.`);
             }
 
-            const data = await this.apiClient.listClusters({
-                params: {
-                    path: {
-                        groupId: project.id || "",
+            const [clustersResult, flexClustersResult] = await Promise.allSettled([
+                this.apiClient.listClusters({
+                    params: {
+                        path: {
+                            groupId: project.id || "",
+                        },
                     },
-                },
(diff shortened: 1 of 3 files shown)


Luke-Sanderson (approved):
LGTM

src/tools/atlas/read/listClusters.ts
```diff
@@ -41,15 +41,27 @@ export class ListClustersTool extends AtlasToolBase {
...
+                        },
+                    },
+                }),
+            ]);
+
+            const clusters = clustersResult.status === "fulfilled" ? clustersResult.value : undefined;
+            const flexClusters = flexClustersResult.status === "fulfilled" ? flexClustersResult.value : undefined;
 
-            return this.formatClustersTable(project, data);
+            return this.formatClustersTable(project, clusters, flexClusters);
```
nirinchev, line 64: Should probably change `formatClustersTable` to make both args required to avoid a similar regression in the future.

nirinchev (changes requested):
Looks good, but needs integration tests to ensure this is working as expected.

src/tools/atlas/read/listClusters.ts
carolynmcca: made required but added undefined in case one of the APIs fails

nirinchev (approved):

PR #1168, opened 2026-05-08
chore: refactor upgradeCluster tool

nirinchev:
## Proposed changes

Noticed that we're storing state on the tool, which is prone to race conditions. Additionally, we were sending cluster ids to telemetry, which is PII and should not be included in payloads.


src/tools/atlas/update/upgradeCluster.ts
```diff
@@ -239,80 +243,53 @@ export class UpgradeClusterTool extends AtlasToolBase {
...
-            provider: this.upgradeContext?.resolvedProvider,
-            region: this.upgradeContext?.resolvedRegion,
+            original_tier: UpgradeClusterTool.toLowerCase(sc?.originalTier),
+            target_tier: UpgradeClusterTool.toLowerCase(sc?.targetTier),
+            provider: sc?.resolvedProvider,
+            region: sc?.resolvedRegion,
         };
     }
+
+    private static toLowerCase<T extends string>(value?: T): Lowercase<T> | undefined {
```
carolynmcca, line 302: is this best place for this helper? Not sure if there's a more global location for helpers

nirinchev: No, not the best, but we don't have a better one yet. I could have created a utils file or something like this, but considering it's currently only used here, it seemed premature.

src/tools/atlas/update/upgradeCluster.ts
```diff
@@ -198,39 +189,52 @@ export class UpgradeClusterTool extends AtlasToolBase {
...
-            resolvedProvider: clusterInfo.provider,
-            resolvedRegion: clusterInfo.region,
-        };
-
-        if (clusterInfo.instanceType === "FREE") {
-            return this.upgradeFreeCluster(projectId, clusterName, target, clusterInfo.provider, clusterInfo.region);
-        }
-        return this.upgradeFlexCluster(projectId, clusterName, clusterInfo.provider, clusterInfo.region);
+        return super.handleError(error, args) as CallToolResult;
     }
```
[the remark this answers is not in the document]
himanshusinghs: Better to just await the handleError. Even though handleError is sync today, the api contract of handleError invites updates to make it async without thinking much about it.

himanshusinghs (approved):
Approved with a comment which I think is worth looking at.

gagik (approved):

PR #1173, opened 2026-05-11
chore: skip test execution on Windows runners

himanshusinghs:
## Proposed changes

Skip the Run tests step on windows-latest runners in both Code Health workflows. Windows runners still perform checkout and pnpm install primarily because of configured GitHub status checks; only pnpm test is skipped.

pnpm test fails non-deterministically on Windows runners with Worker exited unexpectedly and leaks mongod processes (visible in GHA cleanup logs). The crash victim moves between different test files each run, so neither per-file skips nor pool/isolation tweaks (maxWorkers: 1, isolate: false) have produced a reliable green run. isolate: false actively masked the problem by allowing process.exit() calls in  src/ to terminate the suite early with exit code 0 — only ~14 of ~95 files ran while CI reported success.

Re-enabling tracked in MCP-495.

## Checklist

### DIFF
--- a/.github/workflows/code-health-fork.yml
+++ b/.github/workflows/code-health-fork.yml
@@ -14,7 +14,9 @@ jobs:
     if: github.event.pull_request.user.login == 'dependabot[bot]' || github.event.pull_request.head.repo.full_name != github.repository
     strategy:
       matrix:
-        os: [ubuntu-latest, macos-latest, windows-latest]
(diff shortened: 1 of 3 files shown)


gagik (approved):
Ideally we'd not do this but seems reasonable to say it's test environment related. We can try moving towards using Atlas Local Docker image perhaps.

himanshusinghs:
This is temporary, we will fix this with MCP-495.

PR #1181, opened 2026-05-14
Replacing upgradeCluster w/ tenant api calls

carolynmcca:
## Proposed changes
Adding ability to override header; for endpoint `upgradeGroupClusterTenantUpgrade` to support Free --> M10 upgrade, we have to user older version.

## Testing

- [x] Manual (tested free --> flex, free --> M10, flex --> M10)
- [x] Unit

## Checklist

### DIFF
--- a/src/tools/atlas/update/upgradeCluster.ts
+++ b/src/tools/atlas/update/upgradeCluster.ts
@@ -202,10 +202,11 @@ export class UpgradeClusterTool extends AtlasToolBase {
                     throw new UpgradeClusterError(`Cluster "${clusterName}" is already a Flex cluster.`);
                 }
 
-                ({ id: clusterId } = await this.apiClient.upgradeFlexToDedicated({
-                    groupId: projectId,
+                // tenantUpgrade: upgrades Flex clusters to Dedicated (M10+)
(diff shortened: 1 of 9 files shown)


src/tools/atlas/update/upgradeCluster.ts
```diff
@@ -266,12 +268,12 @@ export class UpgradeClusterTool extends AtlasToolBase {
                             ...(regionName !== undefined && { regionName }),
                         },
                     },
-                });
+                } as unknown as Parameters<typeof this.apiClient.upgradeTenantUpgrade>[0]);
```
nirinchev, line 271: I'm guessing this is needed because the OpenAPI definitions don't correctly indicate optional parameters, right?

nirinchev (approved):
Looks good - just one question from me.

src/tools/atlas/update/upgradeCluster.ts
carolynmcca: Yup! [Endpoint](https://www.mongodb.com/docs/api/doc/atlas-admin-api-v2/operation/operation-upgradegroupclustertenantupgrade) only marks name as required but openapi generates fields that have a default value as required

PR #1187, opened 2026-05-15
refactor: move scripts and accuracy tests into dedicated packages

gagik:
- Move root scripts/ to packages/scripts/
- Move tests/accuracy to packages/accuracy-tests/
- Create packages/test-utils/ for shared test infrastructure
- Update all package.json scripts and import paths
- Update vitest.config.ts for new test locations


packages/core/package.json
```diff
@@ -2,7 +2,6 @@
   "name": "@mongodb-js/mcp-core",
   "description": "Runtime-independent core services for the MongoDB MCP server",
   "version": "1.11.0-prerelease.1",
-  "private": true,
```
gagik, line 5: drive by removal of these

gagik: keeping accuracy-tests and scripts private though

packages/test-utils/src/vitest.d.ts
```diff
@@ -0,0 +1,11 @@
+import "vitest";
+
+declare module "vitest" {
```
gagik, line 4: already existed, just being copied over

lerouxb (approved):

PR #1188, opened 2026-05-15
refactor: move integration tests, seperate test helpers, minimize circular dependencies

gagik:
This is kind of an awkward state but this unblocks the circular dependency situation in #1187 

Basically ensures we have `integration-tests` which imports the binary and `test-utils` which is used by the binary.

So
```
test-utils (binary-independent unit test utils) => mongodb-mcp-server => integration-tests (binary-dependent integration test utils) => accuracy tests
```

There's some duplication in place still which should go away once we split off `/cli` as a package.


knip.ts
```diff
@@ -3,7 +3,7 @@ import type { KnipConfig } from "knip";
 const config: KnipConfig = {
     // Shell built-ins used in package.json scripts
     ignoreBinaries: ["printf"],
-    // Workspace-relative path in vitest.config.ts that knip can't resolve
+    // This path is referenced in vitest.config.ts but resolved at runtime
```
gagik, line 6: knip config is a mess still but I'd rather follow up on this seperately.

package.json
```diff
@@ -84,8 +84,10 @@
     "prebuild:mcpb": "pnpm run build",
     "build:mcpb": "pnpm --filter @mongodb-js/mcp-scripts build:mcpb",
     "build:universal-package": "pnpm --filter @mongodb-js/mcp-scripts build:universal-package",
-    "compile": "concurrently \"pnpm run build:esm\" \"pnpm run build:cjs\" && pnpm run build:universal-package",
-    "build": "pnpm run -r --include-workspace-root compile",
+    "compile": "pnpm run build:workspace && pnpm run build:root",
+    "build": "pnpm run compile",
```
gagik: this awkwarness will go away once the package is moved into its own place

packages/scripts/src/createMcpb.ts
```diff
@@ -346,7 +346,11 @@ async function main(): Promise<void> {
     console.log("mcpb: build complete.");
 }
 
-main().catch((err) => {
-    console.error(err);
-    process.exit(1);
-});
+// Only run main when this file is executed directly (not when imported as a module)
```
lerouxb, line 349: Nit, not for this PR, but I've gone through so many of these patterns and they always seem brittle in the end. What works for me is to JUST have the script execution step in the script and move everything that could possibly be imported or useful elsewhere into its own file.

lerouxb (approved):
It is a lot, but seems fine.

packages/scripts/src/createMcpb.ts
gagik: yeah not a fan of this either

PR #1189, opened 2026-05-15
feat: add OIDC authentication support for atlas-connect-cluster

gmishkin:
## Summary

Adds `mongodb-oidc` as an alternative authentication mechanism for `atlas-connect-cluster`, controlled by two new server-level config options:

- `MDB_MCP_ATLAS_AUTH_MECHANISM=mongodb-oidc` — opt in to OIDC instead of temporary user creation
- `MDB_MCP_ATLAS_OIDC_MECHANISM_PROPERTIES` — optional `authMechanismProperties` string for workload identity federation (e.g. `ENVIRONMENT:k8s,TOKEN_RESOURCE:<audience>`)

When OIDC is enabled, the tool skips creating a temporary Atlas database user entirely, allowing deployments with lower-privilege Atlas API credentials (no user management permissions required — `Project Read Only` + `Project IP Access List Admin` is sufficient).

Supports both:
- **Workforce identity federation**: human interactive browser/device-code flows (no `authMechanismProperties` needed)
- **Workload identity federation**: non-interactive cloud identity (Azure Managed Identity, GCP SA, Kubernetes ServiceAccount) via `MDB_MCP_ATLAS_OIDC_MECHANISM_PROPERTIES`

_Editor's note: See [here](https://gist.github.com/gmishkin/309823c33b05811daad80494ab64f189) for an acknowledgment of the limitations of the workload identity federation support._

## Design decisions and tradeoffs

**Auth mechanism is server-level config, not a tool argument.**
An earlier approach exposed `authMechanism` as a tool argument. We moved it to server config because the LLM agent has no way to know whether a given Atlas project has OIDC configured — that's deployment-specific knowledge the operator holds. A single MCP server deployment is already implicitly scoped to one Atlas org (the API credentials are org-scoped), and OIDC availability is also org/federation-scoped, so a server-level toggle is the right boundary.

`authMechanismProperties` is also server-only for the same reason: which cloud environment the server is running in (`ENVIRONMENT:azure`, `ENVIRONMENT:k8s`, etc.) is a property of the deployment, not something the agent should determine at call time.

**Tool-level `authMechanism: "username-password"` override retained as escape hatch.**
If the server is configured for OIDC but the OIDC identity lacks sufficient database access on a specific cluster (wrong cluster scope or insufficient roles), the agent's only recourse without this override would be a server restart. The tool now accepts an optional `authMechanism: "username-password"` argument that forces the temp-user path for that connection. The description is intentionally restrictive: agents are instructed only to set this after observing authorization errors on a previous OIDC-authenticated connection to the cluster.

**No preflight check for OIDC user access.**
We considered checking whether the OIDC identity has appropriate cluster scopes and roles before connecting, but there is no clean way to do this: the check would require additional Atlas API calls, and authorization errors only surface through actual database operations after the connection is established. The config description documents the requirement that operators ensure their OIDC database user in Atlas has appropriate cluster scopes and roles.

**Fallback to temp-user for pre-7.0 clusters.**
When `atlasAuthMechanism=mongodb-oidc` is set but the target cluster is detected as running MongoDB < 7.0 (which does not support OIDC), the tool automatically falls back to temporary user creation and includes a note in the response. This requires the Atlas API credentials to have user management permissions for pre-7.0 clusters in a mixed-version org.

**`AtlasClusterConnectionInfo` refactored to a discriminated union.**
_Editor's note: This one is not super interesting. Check the edit history of this description to see the details_

**IP access list check is unchanged for OIDC.**
_Editor's note: I think this should be considered for https://jira.mongodb.org/browse/MCP-334_

We briefly considered skipping `ensureCurrentIpInAccessList` in OIDC mode (since lower-privilege credentials can't write to the access list). However, reading the access list and the `api/private/ipinfo` endpoint both work with `Project IP Access List Admin`, which is a reasonable permission to grant alongside `Project Read Only` for an OIDC deployment. Skipping the check was reverted — operators should grant `Project IP Access List Admin` to their service account.

## Test plan

- [ ] `pnpm run check:types` passes
- [ ] `pnpm run test:local` passes (pre-existing failures in `exportsManager`, `dropIndex`, `collectionIndexes` are unrelated)
- [ ] Manual test: Atlas cluster with OIDC configured, `MDB_MCP_ATLAS_AUTH_MECHANISM=mongodb-oidc` — connection succeeds without creating a DB user
- [ ] Manual test: pre-7.0 cluster with OIDC config — falls back to temp user with notice in response
- [ ] Manual test: OIDC connection followed by auth error, then retry with `authMechanism: "username-password"` — switches to temp user

### DIFF
--- a/src/tools/atlas/connect/connectCluster.ts
+++ b/src/tools/atlas/connect/connectCluster.ts
@@ -1,3 +1,4 @@
+import { z } from "zod";
 import type { CallToolResult } from "@modelcontextprotocol/sdk/types.js";
 import { type OperationType, type ToolArgs } from "../../tool.js";
 import { AtlasToolBase } from "../atlasTool.js";
@@ -26,6 +27,17 @@ export const ConnectClusterArgs = {
     connectionType: AtlasArgs.connectionType().describe(
         "Type of connection (standard, private, or privateEndpoint) to an Atlas cluster"
     ),
+    authMechanism: z
+        .literal("username-password")
+        .optional()
+        .describe(
+            "Override to force temporary user creation for this connection. " +
+                "Only set this if: the server is configured for 'mongodb-oidc' AND a previous " +
+                "OIDC-authenticated connection to this cluster produced authorization errors on " +
+                "database operations, indicating the OIDC identity lacks sufficient access. " +
+                "Requires Atlas API credentials with database user management permissions. " +
+                "Do not set this preemptively."
+        ),
 };
 
 export class ConnectClusterTool extends AtlasToolBase {
@@ -116,6 +128,7 @@ export class ConnectClusterTool extends AtlasToolBase {
         });
 
         const connectedAtlasCluster = {
+            authType: "temp-user" as const,
             username,
             projectId,
             clusterName,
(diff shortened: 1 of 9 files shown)


src/tools/atlas/connect/connectCluster.ts
```diff
@@ -136,6 +149,48 @@ export class ConnectClusterTool extends AtlasToolBase {
         return { connectionString: cn.toString(), atlas: connectedAtlasCluster };
     }
 
+    private async prepareClusterConnectionOIDC(
+        projectId: string,
+        clusterName: string,
+        connectionType: "standard" | "private" | "privateEndpoint" | undefined = "standard"
```
addaleax, line 155: Should this type be deduplicated somewhere?

src/tools/atlas/connect/connectCluster.ts
```diff
@@ -136,6 +149,48 @@ export class ConnectClusterTool extends AtlasToolBase {
...
+        }
+
+        if (cluster.mongoDBVersion) {
+            const major = parseInt(cluster.mongoDBVersion.split(".").at(0) ?? "", 10);
+            if (!isNaN(major) && major < 7) {
+                return null;
+            }
+        }
+
+        const cn = new URL(connectionString);
```
addaleax, line 176: `new URL()` doesn't parse connection strings with multiple hosts correctly, you almost certainly want `import { ConnectionString } from "mongodb-connection-string-url";` instead (it has the same interface as the built-in `URL` class for this PR's purposes)

src/tools/atlas/connect/connectCluster.ts
```diff
@@ -136,6 +149,48 @@ export class ConnectClusterTool extends AtlasToolBase {
...
+        }
+        const connectionString = getConnectionString(cluster.connectionStrings, connectionType);
+        if (connectionString === undefined) {
+            throw new Error(
+                `Connection string for connection type "${connectionType}" is not available. Please ensure this connection type is set up in Atlas. See https://www.mongodb.com/docs/atlas/connect-to-database-deployment/#connect-to-an-atlas-cluster.`
+            );
+        }
+
+        if (cluster.mongoDBVersion) {
+            const major = parseInt(cluster.mongoDBVersion.split(".").at(0) ?? "", 10);
```
addaleax, line 170: This will inevitably break when we hit MongoDB 10, which isn't that far in the future – let's avoid that

src/tools/atlas/connect/connectCluster.ts
```diff
@@ -254,8 +338,35 @@ export class ConnectClusterTool extends AtlasToolBase {
...
+                            type: "text",
+                            text: `Enter code: ${connState.oidcUserCode}`,
+                        },
+                        {
+                            type: "text",
+                            text: `Call this tool again after authenticating to confirm the connection.`,
+                        },
+                    ],
+                };
+            }
```
addaleax, line 366: Should this share text with the OIDC device flow handling in `connectionErrorHandler.ts`?

gmishkin:
Closing in favor of a PR from the upstream repo branch.

PR #1192, opened 2026-05-18
refactor: move mongodb-mcp-server into its own package

gagik:
Moves it to the directory, deals with the fallout...


knip.ts
```diff
@@ -48,6 +48,54 @@ const config: KnipConfig = {
                 "@testing-library/jest-dom",
```
gagik, line 48: will deal with all knip issues later

lerouxb (approved):

PR #1193, opened 2026-05-18
chore: move setup to its own package

gagik:


packages/accuracy-tests/tsconfig.json
```diff
@@ -3,7 +3,8 @@
   "compilerOptions": {
     "outDir": "./dist",
     "rootDir": ".",
-    "composite": true
```
gagik, line 6: i made all other projects true by default but too many changes here that i dont want to mess with the actual PR

lerouxb (approved):

PR #1211, opened 2026-05-28
feat: Add atlas-create-cluster tool

manupedrozo:
## Description

Add the `atlas-create-cluster` tool:
  - Supports M10-M80 replica sets (default) and single-shard clusters.
  - Instance size defaults to M10 for projects with fewer than 2 existing clusters, M30 otherwise. Compute autoscaling max is set two tiers above the selected size.
  - Returns immediately with instructions to poll for cluster readiness via atlas-inspect-cluster.

Added tests and updated README.

Ticket: CLOUDP-407920

## Checklist


src/common/atlas/apiClient.ts
```diff
@@ -236,7 +236,7 @@ export class ApiClient {
     }
 
     // DO NOT EDIT. This is auto-generated code.
-    /* eslint-disable @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-return */
+    /* eslint-disable @typescript-eslint/no-unsafe-assignment */
```
manupedrozo, line 239: Was throwing a warning.

manupedrozo:
Integ test [failure](https://github.com/mongodb-js/mongodb-mcp-server/actions/runs/26578076343/job/78302902923?pr=1211) is unrelated to PR changes.

src/common/atlas/apiClient.ts
cveticm: This file is auto-generated so the lint disable will return on the next run of `npm run generate:api`.
You'll have to modify [these lines](https://github.com/mongodb-js/mongodb-mcp-server/blob/main/scripts/apply.ts#L196-L197) in `scripts/apply.ts` and remove `, @typescript-eslint/no-unsafe-return` from there.

src/tools/atlas/create/createCluster.ts
```diff
@@ -0,0 +1,298 @@
...
+export const CreateClusterArgsShape = {
+    projectId: AtlasArgs.projectId().describe(
+        "Atlas project ID to create the cluster in. Use the atlas-list-projects to find project IDs if not provided."
+    ),
+
+    clusterName: AtlasArgs.clusterName().describe("Name of the cluster."),
+
+    provider: z
+        .enum(["AWS", "GCP", "AZURE"])
+        .describe("Cloud provider for the cluster. Allowed values: `AWS`, `GCP`, `AZURE`."),
```
cveticm: Listing allowed values of an enum is redundant since the agent will see all values in the schema already. To slim the token count of the tool, I'd remove listing enum values unless you are providing more context on when a value should be selected.
  
```suggestion
        .describe("Cloud provider for the cluster."),
```

src/tools/atlas/create/createCluster.ts
```diff
@@ -0,0 +1,298 @@
...
+
+    region: AtlasArgs.region().describe(
+        "Cloud provider region in Atlas format using uppercase letters and underscores (e.g. US_EAST_1)."
+    ),
+
+    clusterType: z
+        .enum(["REPLICASET", "SHARDED"])
+        .default("REPLICASET")
+        .describe(
+            "Cluster topology. Allowed values: `REPLICASET`, `SHARDED`. Use `SHARDED` for single-shard clusters, requires M30 or higher. Defaults to `REPLICASET`."
```
cveticm: Similar to above
```suggestion
            "Cluster topology. Use `SHARDED` for single-shard clusters, requires M30 or higher. Defaults to `REPLICASET`."
```

src/tools/atlas/create/createCluster.ts
```diff
@@ -0,0 +1,298 @@
...
+        ),
+
+    instanceSize: z
+        .enum(["M10", "M20", "M30", "M40", "M50", "M60", "M80"])
+        .optional()
+        .describe(
+            "Instance size. Allowed values: M10–M80. NVME and high-memory instances are not supported. Minimum M30 when clusterType is SHARDED. Defaults to M10 for projects with fewer than 2 existing clusters, M30 otherwise. Omit unless explicitly specified by the user."
```
cveticm:
```suggestion
            "Instance size. High-memory instances are not supported. Minimum M30 when clusterType is SHARDED. Defaults to M10 for projects with fewer than 2 existing clusters, M30 otherwise. Omit unless explicitly specified by the user."
```

src/tools/atlas/create/createCluster.ts
```diff
@@ -0,0 +1,298 @@
...
+        .optional()
+        .describe(
+            "Initial disk size in GB. Disk autoscaling is always enabled regardless of this value. Omit unless explicitly specified by the user."
+        ),
+
+    mongoDBVersion: z
+        .enum(["7.0", "8.0", "LATEST"])
+        .default("LATEST")
+        .describe(
+            "MongoDB version to deploy. Allowed values: `7.0`, `8.0`, `LATEST`. Use a pinned version for production environments where version stability is required. Defaults to `LATEST`."
```
cveticm:
```suggestion
            "MongoDB version to deploy. Use a pinned version for production environments where version stability is required. Defaults to `LATEST`."
```

src/tools/atlas/create/createCluster.ts
```diff
@@ -0,0 +1,298 @@
...
+import { AtlasArgs } from "../../args.js";
+import type { CreateClusterMetadata } from "../../../telemetry/types.js";
+import type { CallToolResult } from "@modelcontextprotocol/sdk/types.js";
+
+// There is a similar (not exactly the same) region recommendation string for atlas-upgrade-cluster.
+// Intentionally keeping them separate in the short term.
+const REGION_RECOMMENDATIONS = `Common, non exhaustive region default mappings by provider:
+AWS: "East Coast"/"Virginia"/"US East" → US_EAST_1, "Ohio" → US_EAST_2, "California"/"West Coast" → US_WEST_2, "Southeast Asia"/"APAC"/"Singapore" → AP_SOUTHEAST_1, "Europe"/"EU"/"Ireland" → EU_WEST_1.
+GCP: "Central US" → CENTRAL_US, "Western US" → WESTERN_US, "Southeast Asia"/"APAC" → SOUTHEASTERN_ASIA_PACIFIC, "Europe"/"EU" → WESTERN_EUROPE.
+AZURE: "East US" → US_EAST_2, "West US" → US_WEST_2, "Europe North" → EUROPE_NORTH, "Europe West" -> EUROPE_WEST.
```
maastha:
```suggestion
AZURE: "East US" → US_EAST_2, "West US" → US_WEST_2, "Europe North" → EUROPE_NORTH, "Europe West" → EUROPE_WEST.
```

src/tools/atlas/create/createCluster.ts
```diff
@@ -0,0 +1,298 @@
+import { z } from "zod";
+import { type OperationType, type ToolArgs, type ToolResult } from "../../tool.js";
+import { AtlasToolBase } from "../atlasTool.js";
+import type { ClusterDescription20240805 } from "../../../common/atlas/openapi.js";
+import { AtlasArgs } from "../../args.js";
+import type { CreateClusterMetadata } from "../../../telemetry/types.js";
+import type { CallToolResult } from "@modelcontextprotocol/sdk/types.js";
+
+// There is a similar (not exactly the same) region recommendation string for atlas-upgrade-cluster.
```
maastha: what is the difference between the two? also can you add more context about why keeping them separate?

maastha (approved):

src/tools/atlas/create/createCluster.ts
```diff
@@ -0,0 +1,298 @@
...
+                throw new CreateClusterError("SHARDED clusters require M30 or higher instance size.");
+            }
+        } else if (clusterType === "SHARDED") {
+            instanceSize = "M30";
+        } else {
+            // REPLICASET defaults to M10 if there are less than 2 clusters in the project, M30 otherwise.
+            const existing = await this.apiClient.listClusters({ params: { path: { groupId: projectId } } });
+            instanceSize = (existing.results?.length ?? 0) < 2 ? "M10" : "M30";
+        }
+
```
cveticm, line 221: nit: Could we pull out the validation to separate it from setting instanceSize?

```suggestion
        if (clusterType === "SHARDED" && (args.instanceSize === "M10" || args.instanceSize === "M20")) {
            throw new CreateClusterError("SHARDED clusters require M30 or higher instance size.");
        }
        
        let instanceSize: InstanceSize;
         if (!args.instanceSize) {
            instanceSize = args.instanceSize;
        } else if (clusterType === "SHARDED") {
            instanceSize = "M30";
        } else {
            // REPLICASET defaults to M10 if there are less than 2 clusters in the project, M30 otherwise.
            const existing = await this.apiClient.listClusters({ params: { path: { groupId: projectId } } });
            instanceSize = (existing.results?.length ?? 0) < 2 ? "M10" : "M30";
        }
```

cveticm (approved):
Some minor comments, otherwise looks good!
Also triggered accuracy tests to run for new createCluster test

src/common/atlas/apiClient.ts
manupedrozo: Oh nice, fixed & re-generated here: [d442210](https://github.com/mongodb-js/mongodb-mcp-server/pull/1211/commits/d44221084d9edc267143dd5ace71cfc85d4b4a31)

src/tools/atlas/create/createCluster.ts
manupedrozo: Yup, updated [d442210](https://github.com/mongodb-js/mongodb-mcp-server/pull/1211/commits/d44221084d9edc267143dd5ace71cfc85d4b4a31)

manupedrozo:
> Also triggered accuracy tests to run for new createCluster test

Thanks! Ran with all the popular models locally, very useful to quickly iterate on descriptions. We should be unblocked to fully move the CI to Grove now and enable more models.

src/tools/atlas/create/createCluster.ts
manupedrozo: We want to keep these changes additive only. Updated the description here [d442210](https://github.com/mongodb-js/mongodb-mcp-server/pull/1211/commits/d44221084d9edc267143dd5ace71cfc85d4b4a31)

cveticm (approved):
lgtm

src/telemetry/types.ts
```diff
@@ -283,3 +284,16 @@ export type UpgradeClusterMetadata = AtlasMetadata & {
...
+    cluster_id?: string;
+    provider?: string;
+    region?: string;
+    instance_size?: string;
+    cluster_type?: string;
+    backup?: string;
+    compute_auto_scaling?: TelemetryBoolSet;
+    termination_protection?: TelemetryBoolSet;
+    disk_size_gb?: number;
+    mongodb_version?: string;
```
nirinchev, line 298: Can these be more strongly typed? e.g. `backup` is an enum in the underlying data rather than a string. This makes it much easier to answer questions from our analytics team about the shape of these events.

Additionally, can we make sure that the field names are aligned with existing types - e.g. `UpgradeClusterMetadata` uses `target_tier` but here we're using `instance_size`. Discrepancies like this make it harder to reason about relations between the data in different events.

src/tools/atlas/create/createCluster.ts
```diff
@@ -0,0 +1,297 @@
...
+AWS: "East Coast"/"Virginia"/"US East" → US_EAST_1, "Ohio" → US_EAST_2, "California"/"West Coast" → US_WEST_2, "Southeast Asia"/"APAC"/"Singapore" → AP_SOUTHEAST_1, "Europe"/"EU"/"Ireland" → EU_WEST_1.
+GCP: "Central US" → CENTRAL_US, "Western US" → WESTERN_US, "Southeast Asia"/"APAC" → SOUTHEASTERN_ASIA_PACIFIC, "Europe"/"EU" → WESTERN_EUROPE.
+AZURE: "East US" → US_EAST_2, "West US" → US_WEST_2, "Europe North" → EUROPE_NORTH, "Europe West" → EUROPE_WEST.
+Default recommendation: AWS US_EAST_1.
+User-specified regions not present in the mapping MUST be respected, rely on the tool to surface errors if a region is not supported.
+`;
+
+const INSTANCE_SIZES = ["M10", "M20", "M30", "M40", "M50", "M60", "M80"] as const;
+type InstanceSize = (typeof INSTANCE_SIZES)[number];
+type CloudProvider = "AWS" | "GCP" | "AZURE";
```
nirinchev, line 21: [nit] These are duplicating zod enum definitions further down. Instead, we should probably extract the zod enum and reuse it across the schema definition as well as these types.

src/tools/atlas/create/createCluster.ts
```diff
@@ -0,0 +1,297 @@
...
+                provider,
+                region,
+                instanceSize,
+                clusterType,
+                mongoDBVersion: args.mongoDBVersion,
+                backup: args.backup,
+                computeAutoScaling: args.computeAutoScaling,
+                terminationProtectionEnabled,
+                diskSizeGB: args.diskSizeGB,
+            },
```
nirinchev, line 262: `content` and `structuredContent` must represent identical information - they're not complementary and an agent will see one or the other.

PR #1212, opened 2026-05-29
chore: connection timeout blocking alerts hook CLOUDP-409257

carolynmcca:
## Proposed changes

* Run alerts hook when connected and connecting; MCP wasn't showing alerts on cluster because in background we're still connecting so hook wasn't running/showing results to user

## Checklist


src/tools/atlas/connect/connectCluster.ts
```diff
@@ -365,17 +340,49 @@ export class ConnectClusterTool extends AtlasToolBase {
...
+        content: ToolResult<typeof ConnectClusterOutputSchema>["content"]
+    ): Promise<{
+        sharedTierAlertsDetected?: boolean;
+        sharedTierTier?: SharedTierTier;
+        sharedTierAlerts?: SharedTierMetricName[];
+    }> {
+        if (atlas === undefined || !["FREE", "FLEX"].includes(atlas.instanceType)) {
+            return {};
+        }
+        const tier: SharedTierTier = atlas.instanceType === "FREE" ? "Free" : "Flex";
```
nirinchev:
```suggestion
        let tier: SharedTierTier;
        switch (atlas?.instanceType) {
            case "FREE":
                tier = "Free";
                break;
            case "FLEX":
                tier = "Flex";
                break;
            default:
                return {};
        }
```

This is a bit more verbose but it's more explicit about the different flows.

src/tools/atlas/update/upgradeCluster.ts
```diff
@@ -143,8 +143,8 @@ async function resolveClusterInfo(
 class UpgradeClusterError extends Error {}
 
 export const UpgradeClusterOutputSchema = {
-    originalTier: z.enum(["FREE", "FLEX"]),
-    targetTier: z.enum(["FLEX", "M10"]),
+    originalTier: z.enum(["FREE", "FLEX"]).optional(),
+    targetTier: z.enum(["FLEX", "M10"]).optional(),
```
nirinchev: Why are the changes in this file necessary?

nirinchev (approved):

src/tools/atlas/update/upgradeCluster.ts
carolynmcca: I made them optional because I think there are upgrade error paths where we don't have original and target tier (when I was testing in prod acct w/o payment method I wasn't able to upgrade but wasn't see the segment event come through indicating success: failure)

nirinchev: I'm not sure I understand what this change means practically. With the code before, when we experience an error, we wouldn't emit any structured content and in `resolveTelemetryMetadata`, we use null-propagation to pass in empty values there. If this is fixing a legitimate bug, can you add a unit test that reproduces the problem that is being fixed.

tests/unit/tools/atlas/update/upgradeCluster.test.ts
```diff
@@ -131,6 +131,15 @@ describe("UpgradeClusterTool", () => {
     });
 
     describe("error cases", () => {
+        // The MCP SDK requires structuredContent to be present on all responses for tools that define
+        // an outputSchema — including error responses. Without it, the client receives an output
+        // validation error ("no structured content was provided") instead of the actual error message.
```
nirinchev: This is not true AFAIK - e.g. collectionStorageSize doesn't return structuredContent in the error case: https://github.com/mongodb-js/mongodb-mcp-server/blob/a0ce7320434575a811b0b93b3c843cf850a858f6/src/tools/mongodb/metadata/collectionStorageSize.ts#L59-L65

carolynmcca: removed - only change needed was in src/tools/atlas/atlasTool.ts (which was why no payment method didn't log the error) but I confired that other [upgrade failures](https://mixpanel.com/project/3851272/view/4346958/app/events#69fWEMAJFuaz) are being logged as expected

nirinchev (approved):

PR #1221, opened 2026-06-02
chore: bump pnpm to latest MCP-518 MCP-495

nirinchev:
Updates pnpm to latest as it brings in some nice guards against supply chain attacks. This requires bumping the minimum supported node version to 22 as 20 is EOL-ed. Additionally, reenables the windows tests that had been disabled due to obscure test runner crashes.


package.json
```diff
@@ -198,24 +198,10 @@
     "@types/express": "^5.0.3"
   },
   "engines": {
-    "node": "^20.19.0 || ^22.12.0 || >= 24.0.0"
+    "node": "^22.13.0 || >= 24.0.0"
```
blva: are we dropping support for node 20/21? any versioning implications?

nirinchev: node 20 and 21 are EOL-ed and unsupported by pnpm. Technically, we can keep it as a platform, though we should make it clear we're not officially supporting it.

package.json
```diff
@@ -94,7 +94,7 @@
     "generate:api": "./scripts/generate.sh",
     "generate:arguments": "tsx --tsconfig scripts/generate/tsconfig.json scripts/generate/index.ts",
     "pretest": "pnpm run build",
-    "test": "vitest --project eslint-rules --project unit-and-integration --project mcpb-build-script --coverage --run",
```
nirinchev, line 97: coverage is now controlled by vitest.config.ts and only enabled for ubuntu/macos.

package.json
```diff
@@ -198,24 +198,10 @@
...
-      "bson": "^7.2.0",
-      "zod": "^4.4.2"
-    }
-  },
-  "overrides": {
-    "mongodb": "^7.1.1",
-    "kerberos": "^7.0.0",
-    "bson": "^7.2.0",
-    "zod": "^4.4.2"
   }
```
nirinchev, line 206: Turns out these had no effect on the production package, so decided to remove them to ensure we're testing against what users are running.

vitest.config.ts
```diff
@@ -23,11 +23,12 @@ if (process.env.SKIP_ATLAS_LOCAL_TESTS === "true") {
 export default defineConfig({
     test: {
         environment: "node",
-        pool: "threads",
```
nirinchev, line 26: This was causing the obscure crashes.

package.json
blva: I also like the idea of migrating and reducing the list of versions supported. Only caveat is that this could break for users, right? should we aim to call this a new major version? Although I believe the users of `@latest` would still be impacted. I would like to find out a way forward though

nirinchev: I think we can leave it as allowed, but not officially supported and call it out in the docs. Then with v2 we can drop it.

blva: SGTM! could we drop it in the README? I think we should probably have that added to the docs, if we have a JIRA to link here it can go to the changelog

nirinchev: Good point, added MCP-518.

blva: thanks! I think we're missing the readme where we call out the engines we support, right?

.github/workflows/code-health-fork.yml
```diff
@@ -15,12 +15,12 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        node-version: ["20"]
+        node-version: ["22"]
         include:
-          - os: ubuntu-latest
-            node-version: "22"
```
addaleax, line 21: Certainly won't block anything here, but if we currently say the minimum Node.js version for users of this library is 20.x, then imo we should keep testing on it even when EOL (and I would consider bumping that mininum to be a semver-major update)

addaleax (approved):

.github/workflows/code-health-fork.yml
addaleax: Ah, automerge ... 😅

nirinchev: Ugh, yeah, I'll add it in a follow-up 😅

PR #1226, opened 2026-06-05
feat: disable server-side JS functions by default MCP-521 MCP-287

nirinchev:
## Proposed changes

This PR adds a new config option - `disableServerSideJs`, which is `true` by default. It disables using `$where`, `$function`, and `$accumulator` in aggregations and queries as usage of server-side JS functions is strongly discouraged and may result in undesired behavior.

Additionally, it addresses an omission which guards `export` and `explain` aggregations against containing write stages if the server is started with the `readOnly` flag.


src/helpers/mqlGuards.ts
```diff
@@ -0,0 +1,68 @@
...
+ */
+function findServerSideJSOperator(value: unknown): ServerSideJSOperator | undefined {
+    if (Array.isArray(value)) {
+        for (const item of value) {
+            const found = findServerSideJSOperator(item);
+            if (found) {
+                return found;
+            }
+        }
+        return undefined;
```
addaleax:
```suggestion
        return value.find(item => findServerSideJSOperator(item));
```

addaleax (approved):

PR #1227, opened 2026-06-05
chore: re-add node 20 CI MCP-518

nirinchev:
## Proposed changes

Addresses a leftover comment from https://github.com/mongodb-js/mongodb-mcp-server/pull/1221


cveticm (approved):

nirinchev:
This actually doesn't work because pnpm doesn't support node 20. Will try and come up with an alternative solution.

addaleax (approved):

PR #1232, opened 2026-06-08
fix(ui): make exported ui module loadable

mojobeeping:
## Summary

- Move the runtime UI packages used by the exported `./ui` module from `devDependencies` to `dependencies`.
- Route the `./ui` import condition to the CommonJS UI build, which avoids Node ESM resolution failures from the LeafyGreen UI dependency chain while keeping ESM consumers able to import the subpath.
- Add build integration coverage for the UI export dependency metadata and real Node ESM/CJS subpath loading.

## Reproduction

A clean install of `mongodb-mcp-server@1.12.0` can import root/tools/web, but `import("mongodb-mcp-server/ui")` fails first because runtime UI packages are not installed:

```text
Error [ERR_MODULE_NOT_FOUND]: Cannot find package 'react' imported from .../node_modules/mongodb-mcp-server/dist/esm/ui/components/ListDatabases/ListDatabases.js
```

After adding the missing runtime packages, the current ESM UI build then reaches the LeafyGreen ESM dependency chain and fails under Node's ESM resolver:

```text
Error [ERR_MODULE_NOT_FOUND]: Cannot find module '.../node_modules/lodash/debounce' imported from .../@leafygreen-ui/hooks/dist/esm/index.js
Did you mean to import "lodash/debounce.js"?
```

The CommonJS UI build already loads successfully, so this PR points the package `./ui` import condition at that build as well.

## Verification

- `pnpm exec vitest --project unit-and-integration --run tests/integration/build.test.ts`
- `pnpm run check:dependencies`
- `pnpm run check:types`
- `pnpm run check:format`
- `pnpm run build`
- `git diff --check`
- Packed tarball clean consumer install
- Packed tarball `import("mongodb-mcp-server/ui")`
- Packed tarball `require("mongodb-mcp-server/ui")`
- Packed tarball NodeNext TypeScript resolution with `skipLibCheck`

Note: local pre-commit could not run because `git-secrets` is not installed in my environment; the staged diff was manually checked with `git diff --cached --check` and a simple secret-pattern scan before commit.

### DIFF
--- a/api-extractor/reports/ui.public.api.md
+++ b/api-extractor/reports/ui.public.api.md
@@ -4,11 +4,6 @@
 
 ```ts
 
-import { ReactElement } from 'react';
-
-// @public (undocumented)
-export const ListDatabases: () => ReactElement | null;
-
 // @public
 export class UIRegistry {
     constructor(options?: {
--- a/src/ui/index.ts
+++ b/src/ui/index.ts
@@ -1,2 +1,1 @@
 export { UIRegistry } from "./registry/index.js";
-export { ListDatabases } from "./components/ListDatabases/index.js";
--- a/tests/integration/build.test.ts
+++ b/tests/integration/build.test.ts
@@ -1,6 +1,8 @@
+import { execFileSync } from "child_process";
 import { createRequire } from "module";
 import path from "path";
 import { describe, it, expect } from "vitest";
+import packageJson from "../../package.json" with { type: "json" };
 
 // Current directory where the test file is located
 const currentDir = import.meta.dirname;
@@ -13,6 +15,9 @@ const cjsPath = path.resolve(projectRoot, "dist/cjs/lib.js");
 
 const esmToolsPath = path.resolve(projectRoot, "dist/esm/tools/index.js");
 const cjsToolsPath = path.resolve(projectRoot, "dist/cjs/tools/index.js");
+
+const uiComponentBuildDependencies = ["@lg-mcp/embeddable-uis", "@lg-mcp/hooks", "react", "react-dom"];
+const uiExports = ["UIRegistry"];
 
 describe("Build Test", () => {
     it("should successfully require CommonJS module", () => {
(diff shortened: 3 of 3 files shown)


nirinchev:
This looks like an error with how the UI exports are setup - react should not be a runtime dependency of the server since all the UI components are precompiled. So rather than the changes in this PR, the real fix seems to be not exporting `ListDatabases` from the `ui` route.

mojobeeping:
Thanks, that makes sense. I reworked the PR to follow that direction.

Changes now:
- removed the React `ListDatabases` component from the public `./ui` entrypoint, so `mongodb-mcp-server/ui` exports only `UIRegistry`
- moved the React/UI component packages back out of runtime dependencies; `package.json` and `pnpm-lock.yaml` are back to the original runtime dependency set
- kept a regression that imports/requires `mongodb-mcp-server/ui` from the built ESM/CJS outputs and verifies only `UIRegistry` is public

Verified locally:
- `pnpm run build`
- `pnpm exec vitest --project unit-and-integration tests/integration/build.test.ts --run`
- `pnpm run check:api-report`
- `pnpm run check:dependencies`
- `pnpm run check:format`

gagik (commented):
We're refactoring the package in #1122 which will separate the UI dependencies into their own package. 

I believe this should resolve this with any additional changes. Let me know if there's anything missing

mojobeeping:
Thanks, that makes sense. I’ll close this in favor of #1122.

For context, this PR had already been reworked away from adding React/UI runtime dependencies: it now only removes `ListDatabases` from the public `./ui` export, keeps the UI component packages in dev dependencies, and adds regression coverage that `mongodb-mcp-server/ui` loads from ESM/CJS with only `UIRegistry` exported.

Since #1122 separates the UI dependencies into their own package and should address this more cleanly, I’ll avoid carrying a parallel v1-specific patch. Happy to reopen or cherry-pick the small export/test change if a v1 stopgap is still needed before #1122 lands.

PR #1234, opened 2026-06-08
chore: fix flaky oidc test

nirinchev:
## Proposed changes

We had sporadic timeouts in the oidc test - this tries to fix that by increasing the refresh token lifetime where we don't explicitly need short TTLs.

### DIFF
--- a/tests/integration/common/connectionManager.oidc.test.ts
+++ b/tests/integration/common/connectionManager.oidc.test.ts
@@ -15,6 +15,11 @@ import { type TestConnectionManager } from "../../utils/index.js";
 
 const DEFAULT_TIMEOUT = 60_000;
 const DEFAULT_RETRIES = 5;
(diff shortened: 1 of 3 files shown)


tests/integration/common/connectionManager.oidc.test.ts
```diff
@@ -94,7 +101,7 @@ describe.skipIf(process.platform !== "linux")("ConnectionManager OIDC Tests", as
 
         const oidcConfig = {
             ...defaultTestConfig,
-            oidcRedirectURi: "http://localhost:0/",
+            oidcRedirectUri: "http://localhost:0/",
```
nirinchev, line 104: This is a drive-by. I don't believe it was the source of the issue, but the typo was causing the redirect url to be the hardcoded default.

himanshusinghs (approved):

PR #1235, opened 2026-06-09
chore: Add fields to atlas-create-cluster response content

manupedrozo:
## Description

Followups to https://github.com/mongodb-js/mongodb-mcp-server/pull/1211

- Update `atlas-create-cluster` response content to include additional fields from the `structuredContent` (`instanceSize`, `clusterType`, `provider` & `region`).
- Extracted enums to reduce duplication.

Followup from 

## Checklist

### DIFF
--- a/src/tools/atlas/create/createCluster.ts
+++ b/src/tools/atlas/create/createCluster.ts
@@ -16,22 +16,22 @@ Default recommendation: AWS US_EAST_1.
 User-specified regions not present in the mapping MUST be respected, rely on the tool to surface errors if a region is not supported.
 `;
 
-const INSTANCE_SIZES = ["M10", "M20", "M30", "M40", "M50", "M60", "M80"] as const;
-const CLOUD_PROVIDERS = ["AWS", "GCP", "AZURE"] as const;
-const CLUSTER_TYPES = ["REPLICASET", "SHARDED"] as const;
-const MONGODB_VERSIONS = ["7.0", "8.0", "LATEST"] as const;
-const BACKUP_OPTIONS = ["OFF", "SNAPSHOT", "CONTINUOUS"] as const;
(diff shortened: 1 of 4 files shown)


src/tools/atlas/create/createCluster.ts
```diff
@@ -17,8 +17,15 @@ User-specified regions not present in the mapping MUST be respected, rely on the
 `;
 
 const INSTANCE_SIZES = ["M10", "M20", "M30", "M40", "M50", "M60", "M80"] as const;
+const CLOUD_PROVIDERS = ["AWS", "GCP", "AZURE"] as const;
```
nirinchev: Rather than defining the arrays and then inferring the enums, consider expressing the values in terms of the zod schema:

```ts
const instanceSizeEnum = z.enum(["M10", "M20", "M30", "M40", "M50", "M60", "M80"]);
type InstanceSize = z.infer<typeof instanceSizeEnum>;

// wherever we use INSTANCE_SIZES
   return instanceSizeEnum.options[instanceSizeEnum.options.indexOf(size) + 2] ?? "M80";
```

nirinchev (commented):
Looks good - one minor suggestion about using the zod schema as the source of truth and deriving the types from it rather than the other way around.

src/tools/atlas/create/createCluster.ts
manupedrozo: Had it that way at first and decided against it, switching back to zod schema first 👍  [13cf6c9](https://github.com/mongodb-js/mongodb-mcp-server/pull/1235/commits/13cf6c9f6d4e9d1f0fe1f0de4054b14acee37b1a)

nirinchev: Yeah, it's somewhat arbitrary, but we tend to prefer zod-first definitions where possible as those are a bit stricter/rigorous than the typescript types.

nirinchev (approved):

PR #1199, opened 2026-05-20
feat: add a tool to load sample data in a cluster MCP-500

nirinchev:
## Checklist

- [ ] I have signed the [MongoDB CLA](https://www.mongodb.com/legal/contributor-agreement)

PR #1237, opened 2026-06-09
feat: add a tool to load sample data in a cluster MCP-500

nirinchev:
## Proposed changes

Replaces https://github.com/mongodb-js/mongodb-mcp-server/pull/1199.

This adds a `atlas-load-sample-dataset` tool that can be used to start a sample dataset load as well as check in on the status of an existing one. Currently, this is a single tool that does both functions - the reasoning here is that a standalone `get-sample-dataset-load-status` tool wouldn't make sense as there's no mechanism to obtain a job id outside of creating the job in the same session. Additionally, models appear to be sufficiently good at following the description hints and recognize to first initiate the load with a cluster name and then call the tool again with the job id from the response.


src/tools/atlas/create/loadSampleDataset.ts
```diff
@@ -0,0 +1,90 @@
...
+
+export class LoadSampleDatasetTool extends AtlasToolBase {
+    static toolName = "atlas-load-sample-dataset";
+    public description =
+        "Load a MongoDB sample dataset into an Atlas cluster, or check the status of a previously-initiated load. " +
+        "To start a new load, provide `clusterName` — the load runs asynchronously and the response includes a " +
+        "`jobId` and initial state. To check progress, call this tool again with `jobId` " +
+        "(sample dataset loads typically take 1–5 minutes). State can be WORKING, COMPLETED, or FAILED.";
+    static operationType: OperationType = "create";
+    public argsShape = {
```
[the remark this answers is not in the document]
nirinchev: This is a non-issue - there is no mechanism to find in-flight load operations, so there's no use case for having a standalone check outside of first creating a load operation.

src/tools/atlas/create/loadSampleDataset.ts
```diff
@@ -0,0 +1,90 @@
...
+    projectId: AtlasArgs.projectId().describe("Atlas project ID that owns the cluster"),
+    clusterName: AtlasArgs.clusterName()
+        .optional()
+        .describe(
+            "Cluster name to load the sample dataset into. Provide this to initiate a new load. Mutually exclusive with jobId."
+        ),
+    jobId: CommonArgs.objectId("jobId")
+        .optional()
+        .describe(
+            "Job id returned by a previous load request. Provide this to check the status of an in-progress load. Mutually exclusive with clusterName."
```
blva, line 17: any way to mark the mutual exclusive nature via zod?

src/tools/atlas/create/loadSampleDataset.ts
```diff
@@ -0,0 +1,90 @@
...
+
+    protected async execute({
+        projectId,
+        clusterName,
+        jobId,
+    }: ToolArgs<typeof this.argsShape>): Promise<ToolResult<typeof this.outputSchema>> {
+        let status: SampleDatasetStatus;
+        let headerText: string;
+        if (jobId !== undefined && clusterName === undefined) {
+            status = await this.apiClient.getSampleDatasetLoad({
```
blva, line 56: is  the create tool also working as a get tool?

nirinchev: It is, yes - I updated the PR description to make it clearer, but the tl;dr is that a standalone tool wouldn't make sense, so reusing the tool for both creating the load job and checking if it's complete makes it easier for agents to do the right thing.

src/tools/atlas/create/loadSampleDataset.ts
nirinchev: There is, but the generated json schema is significantly more complex and we don't win much. At the end of the day, models will read the json schema and reason about the mutual exclusivity from the its shape, which is no better than explicitly calling it out and handling the error scenario.

src/tools/atlas/create/loadSampleDataset.ts
blva: got it! should we add accuracy tests then to ensure it works well?

src/tools/atlas/create/loadSampleDataset.ts
blva: ok, np then if we get good accuracy results

blva (approved):
LGTM, OK to add accuracy tests in a separate PR

PR #1242, opened 2026-06-11
feat: extend telemetry to support custom parameter for create-index tool MCP-407

dudaschar:
*Goal:*
Add MCP telemetry to distinguish Atlas Search index creation via the create-index tool from other index creations, by:

Emitting dedicated metadata (e.g., an index_type or equivalent field) for search index creation events.
This can be done by using the resolveTelemetryMetadata method in createIndex MCP tool.


gagik (approved):
Great work! 🚀

PR #1249, opened 2026-06-15
feat: add structured content for response in mongodb tools - MCP-374

dudaschar:

It also add serialization safety when parsing BSON documents into JSON, to values type `long` are now properly handled in the Tool output. A follow up ticket will be created to discuss Tool arguments and interpolation in the output text response.

This PR apply the updates for the tools inside the mongodb package, a follow up PR will handle /atlas and /atlas-local

## Checklist


src/tools/mongodb/read/export.ts
```diff
@@ -1,13 +1,22 @@
...
 import { jsonExportFormat } from "../../../common/exportsManager.js";
 import { AggregateArgs } from "./aggregate.js";
 
+const ExportOutputSchema = {
+    type: z.literal("resource_link").describe("The content type of the export resource link."),
+    name: z.string().describe("The name of the exported file."),
+    uri: z.string().describe("The URI of the exported data."),
+    description: z.string().describe("Description of the export resource link."),
+    mimeType: z.string().describe("The MIME type of the exported data."),
+};
```
dudaschar: I'm not sure if this is the output we want to add in the structuredContent, can someone confirm? I got it from what the current tool result is returning along with the text response.

src/tools/tool.ts
```diff
@@ -33,10 +33,10 @@ export interface ToolExecutionContext {
 
 export type ToolResult<OutputSchema extends ZodRawShape | undefined = undefined> = OutputSchema extends ZodRawShape
     ? StructuredToolResult<OutputSchema>
-    : { content: { type: "text"; text: string }[]; isError?: boolean };
+    : { content: CallToolResult["content"]; isError?: boolean };
 
 type StructuredToolResult<OutputSchema extends ZodRawShape> = {
-    content: { type: "text"; text: string }[];
+    content: CallToolResult["content"];
```
dudaschar, line 40: This change is allowing the tools to with a previous custom tool result to have a content other than only text + structured content without the need to custom types or casting the output type.

src/helpers/bsonToJson.ts
```diff
@@ -0,0 +1,19 @@
+import { objectToIdiomaticEJSON } from "hadron-document";
+
+/**
+ * Converts a BSON value into a JSON-safe object using Compass idiomatic Extended JSON.
+ *
+ * ObjectId becomes `{ "$oid": "..." }`, Long becomes `{ "$numberLong": "..." }`, etc.
+ */
+export function bsonToJson(value: Record<string, unknown>): Record<string, unknown>;
```
dudaschar, line 8: This is overloading is supporting dbStats tool, which is an object instead of a collection document, so no cast in the tool will be necessary (or any other similar).

src/helpers/bsonToJson.ts
```diff
@@ -0,0 +1,19 @@
...
+export function bsonToJson(value: unknown): unknown;
+export function bsonToJson(value: unknown): unknown {
+    return JSON.parse(objectToIdiomaticEJSON(value));
+}
+
+/**
+ * Serializes an array of BSON objects.
+ */
+export function serializeBsonToJsonObjects(objects: unknown[]): unknown[] {
```
nirinchev: Do we need this extra function? Can we instead handle that in `bsonToJson` like:

```ts
if (Array.isArray(value)) {
    return value.map((object) => bsonToJson(object));
}

return JSON.parse(objectToIdiomaticEJSON(value));
```

src/helpers/constants.ts
```diff
@@ -15,6 +15,14 @@ export const AGG_COUNT_MAX_TIME_MS_CAP: number = 60_000;
 
 export const ONE_MB: number = 1 * 1024 * 1024;
 
+export const CURSOR_LIMIT_KEYS = [
+    "config.maxDocumentsPerQuery",
+    "config.maxBytesPerQuery",
+    "tool.responseBytesLimit",
+] as const;
```
nirinchev: Rather than export this as a const array, we could export it as a zod enum - seems like that's the only reason it exists.

src/tools/mongodb/connect/connect.ts
```diff
@@ -39,11 +46,25 @@ export class ConnectTool extends MongoDBToolBase {
...
+    protected override async handleError(
+        error: unknown,
+        args: ToolArgs<typeof this.argsShape>
+    ): Promise<CallToolResult> {
+        const result = await super.handleError(error, args);
+        if (result.isError) {
+            return { ...result, structuredContent: { connected: false } };
+        }
+        return result;
+    }
```
dudaschar, line 69: Another option here would be not handling the error, and structured content be undefined, instead of returning false. If we decide to return undefined, I'll update similar tools to do the same.

src/tools/mongodb/read/aggregate.ts
```diff
@@ -45,6 +51,15 @@ If the user has asked for lexical/Atlas search, use \`$search\` instead of \`$te
 - The \`$search\` stage supports multiple operators, such as 'autocomplete', 'text', 'geoWithin', and others. Choose the approprate operator based on the user's query. If unsure of the exact syntax, consult the MongoDB Atlas Search documentation, which can be found here: https://www.mongodb.com/docs/atlas/atlas-search/operators-and-collectors/
 `;
 
+const AggregateOutputSchema = {
+    documents: z.array(z.unknown()).describe("The documents returned by the aggregation pipeline"),
+    aggResultsCount: z
+        .number()
+        .optional()
```
nirinchev: Why is this optional? Wouldn't we always be able to supply the count?

src/tools/mongodb/read/aggregate.ts
```diff
@@ -45,6 +51,15 @@ If the user has asked for lexical/Atlas search, use \`$search\` instead of \`$te
 - The \`$search\` stage supports multiple operators, such as 'autocomplete', 'text', 'geoWithin', and others. Choose the approprate operator based on the user's query. If unsure of the exact syntax, consult the MongoDB Atlas Search documentation, which can be found here: https://www.mongodb.com/docs/atlas/atlas-search/operators-and-collectors/
 `;
 
+const AggregateOutputSchema = {
+    documents: z.array(z.unknown()).describe("The documents returned by the aggregation pipeline"),
+    aggResultsCount: z
```
nirinchev: This can be just `count`.

src/tools/mongodb/read/aggregate.ts
```diff
@@ -45,6 +51,15 @@ If the user has asked for lexical/Atlas search, use \`$search\` instead of \`$te
...
+        .number()
+        .optional()
+        .describe("The total number of documents returned by the aggregation pipeline"),
+    appliedLimits: z.array(z.enum(CURSOR_LIMIT_KEYS)).describe("The limits applied to the aggregation pipeline"),
```
nirinchev: This is an enum of the internal representation of the limits (e.g. `config.maxDocumentsPerQuery`), but it seems like we want to instead pass in `server's configured - maxDocumentsPerQuery`, don't we? I guess the LLM can probably infer what `config.maxDocumentsPerQuery` means, but it's a notable discrepancy between the information in `content` vs the one in `structuredContent`.

src/tools/mongodb/read/aggregate.ts
```diff
@@ -175,21 +195,30 @@ Note to LLM: If the entire aggregation result is required, use the "export" tool
...
             }
 
+            const serializedDocuments = serializeBsonToJsonObjects(documents);
+
             return {
                 content: formatUntrustedData(
                     successMessage,
-                    ...(documents.length > 0 ? [EJSON.stringify(documents)] : [])
+                    ...(serializedDocuments.length > 0 ? [EJSON.stringify(serializedDocuments)] : [])
                 ),
```
nirinchev: This highlights a bit of a naming issue - we're getting something called `serializedDocuments`, but then we're serializing it again, which is surprising. Also, I'm noticing some places we use `JSON.stringify` while others, `EJSON.stringify` - should we converge to using one or the other?

src/tools/mongodb/read/count.ts
```diff
@@ -12,6 +12,10 @@ export const CountArgs = {
         ),
 };
 
+const CountOutputSchema = {
+    count: z.number().optional().describe("The number of documents in the collection"),
```
nirinchev: Same q - why is this optional?

tests/unit/helpers/bsonToJson.test.ts
```diff
@@ -0,0 +1,21 @@
+import { describe, expect, it } from "vitest";
+import { Long, ObjectId } from "bson";
+import { bsonToJson, serializeBsonToJsonObjects } from "../../../src/helpers/bsonToJson.js";
+
+describe("safeBsonJson", () => {
```
nirinchev: Can we add some tests for data types that are safe to serialize to normal json? E.g. would `Long.fromString("10")` be serialized as `{ $numberLong: "10" }` or `10`?

tests/integration/tools/mongodb/read/find.test.ts
```diff
@@ -584,6 +703,11 @@ describeWithMongoDB(
             expect(content).toContain('Query on collection "foo"');
             const docs = getDocsFromUntrustedContent(content);
             expect(docs.length).toEqual(5);
+            expect(response.structuredContent).toMatchObject({
+                queryResultsCount: 5,
+                appliedLimits: [],
+                documents: serializeBsonToJsonObjects(docs),
```
nirinchev: Do we need to run `serializeBsonToJsonObjects` here?

tests/integration/tools/mongodb/read/find.test.ts
```diff
@@ -446,6 +552,18 @@ describeWithMongoDB(
...
+                appliedLimits: ["tool.responseBytesLimit"],
+            });
+            expect(
+                typeof response.structuredContent === "object" &&
+                    response.structuredContent !== null &&
+                    "documents" in response.structuredContent &&
+                    Array.isArray(response.structuredContent.documents)
+                    ? response.structuredContent.documents
+                    : []
+            ).toHaveLength(1);
```
nirinchev: This is repeated quite a bit - maybe we can create a helper to reduce the duplication? We can also make the assertions a bit more specific (i.e. add multiple assertions for the contents as opposed to checking an empty array's length).

tests/integration/tools/mongodb/read/find.test.ts
```diff
@@ -290,6 +325,17 @@ describeWithMongoDB("find tool with default configuration", (integration) => {
...
+                appliedLimits: [],
+                documents: serializeBsonToJsonObjects(docs),
+            });
+            const queryResultsCount =
+                typeof response.structuredContent === "object" &&
+                response.structuredContent !== null &&
+                "queryResultsCount" in response.structuredContent
+                    ? response.structuredContent.queryResultsCount
+                    : undefined;
+            expect(queryResultsCount).toBeUndefined();
```
nirinchev: This will pass even if `response.structuredConent` has the wrong shape - e.g. if we return an integer in there since the ternary will return `undefined` in that case.

tests/integration/tools/mongodb/read/export.test.ts
```diff
@@ -27,20 +27,40 @@ export function contentWithTextResourceURI(
...
+    if (resourceLink === undefined) {
+        throw new Error("Expected resource_link in export tool response content");
+    }
+    expect(response.structuredContent).toEqual(resourceLink);
+    return resourceLink;
+}
+
+function expectExportStructuredContentAbsent(response: CallToolResult): void {
+    expect(response.structuredContent).toBeUndefined();
 }
```
nirinchev: This helper seems trivial - we can just inline it where it's used.

tests/integration/tools/mongodb/read/export.test.ts
```diff
@@ -197,8 +222,8 @@ describeWithMongoDB(
                     },
                 });
                 const content = response.content as CallToolResult["content"];
-                const exportURI = contentWithResourceURILink(content)?.uri as string;
-                await resourceChangedNotification(integration.mcpClient(), exportURI);
+                const resourceLink = expectExportStructuredContent(response as CallToolResult, content);
```
nirinchev: These casts are somewhat suspicious - can we instead define the argument of `expectExportStructuredContent` as `Awaited<ReturnType<Client["callTool"]>>` so that we can pass `response` directly?

Also, do we need to pass `content` as a second argument if we can get it as `response.content`?

src/tools/mongodb/read/export.ts
nirinchev: Oh, actually for export, we probably don't want structured content since we're returning the resource link as part of the regular content and we definitely want the model to see that.

src/tools/mongodb/read/find.ts
```diff
@@ -107,16 +120,27 @@ Note to LLM: If the entire query result is required, use the "export" tool inste
...
             return {
                 content: formatUntrustedData(
                     this.generateMessage({
                         collection,
                         queryResultsCount,
-                        documents: cursorResults.documents,
-                        appliedLimits: [limitOnFindCursor.cappedBy, cursorResults.cappedBy].filter((limit) => !!limit),
+                        documents: serializedDocuments,
+                        appliedLimits: [limitOnFindCursor.cappedBy, cursorResults.cappedBy].filter(
+                            (limit): limit is CursorLimitKey => !!limit
```
nirinchev: This can be computed once and reused in `structuredContent`.

tests/integration/tools/mongodb/read/aggregate.test.ts
```diff
@@ -23,6 +23,47 @@ import type { ToolEvent } from "../../../../../src/telemetry/types.js";
...
+
+type AggregateToolResponse = Awaited<ReturnType<Client["callTool"]>>;
+
+function getDocsFromUntrustedContentWhenPresent(content: string): unknown[] {
+    try {
+        return getDocsFromUntrustedContent(content);
+    } catch {
+        return [];
+    }
+}
```
nirinchev: Not sure I understand why we need this

tests/integration/tools/mongodb/read/aggregate.test.ts
```diff
@@ -23,6 +23,47 @@ import type { ToolEvent } from "../../../../../src/telemetry/types.js";
...
+
+function expectAggregateStructuredContent(
+    response: AggregateToolResponse,
+    content: string,
+    expected: {
+        aggResultsCount?: number;
+        omitAggResultsCount?: boolean;
```
nirinchev: These 2 seem equivalent and can be expressed only with `aggResultsCount?: number`, can't they?

nirinchev (commented):
Looks good - I have mostly stylistic comments/questions, that are not critical, but good to address. We should probably revert the `export` changes though as it's the kind of tool that doesn't benefit much (if at all) by adding structured content.

src/tools/mongodb/read/aggregate.ts
dudaschar: totalDocuments can be null, if something downstream fails in the countAggregationResultDocuments, the value will be undefined (not 0).

Not sure how often this can actually happen or if we wanna handle the response in a different way. I don't think we should return as 0 tho, because it could be misleading.

src/tools/mongodb/read/aggregate.ts
dudaschar: I see the difference between them, but I also wonder if would actually make sense the structured content to have the limit key. My rational is that this result could be used programmatically using the tool response as arguments. Is this a valid concern? Or is it better to return an array with the descriptions and be consistent with the value in content?

src/tools/mongodb/read/aggregate.ts
dudaschar: thanks! I also saw Le Roux updated my ticket description and I think I was a bit confused on the bson vs json vs ejson parsing. Updated.

tests/integration/tools/mongodb/read/export.test.ts
dudaschar: I reverted this, since the tool won't have a structured content output.

src/tools/mongodb/read/count.ts
dudaschar: I thought this was similar as in aggregate tool, where from a downstream error the count could be undefined. But looking again now, I see for this tool if the count fails, will return an error indeed and not a response without a count. I'm making it required.

tests/unit/helpers/bsonToJson.test.ts
dudaschar: thanks! I added the tests for the other types as well.

tests/integration/tools/mongodb/read/find.test.ts
dudaschar: Yes, because the structured content from the tool returns the converted json. The generic helper which builds a document doesn't return the safe json format, do we would be comparing  `"_id": "6a3d30f36490a5277efe8aca"` with   `"_id": { "$oid": "6a3d30f36490a5277efe8aca" }`, for example.

tests/integration/tools/mongodb/read/export.test.ts
dudaschar: Also removed because the tool was reverted.

tests/integration/tools/mongodb/read/find.test.ts
dudaschar: I moved mostly of the duplication logic to a helper and also moved some similar tests to run in parametrized test blocks. The file looks a bit cleaner now.

tests/integration/tools/mongodb/read/find.test.ts
dudaschar: Change the tests to be more strict now.

tests/integration/tools/mongodb/read/aggregate.test.ts
dudaschar: It's a helper for 2 places where I was doing a validation, but it can be dropped. 👍

src/tools/mongodb/read/aggregate.ts
nirinchev: Ah, I see - wonder if that will be clear enough for the LLM - we could also have it as union type of `number` and the const string `indeterminate` or something similar to be extra explicit?

src/tools/mongodb/read/aggregate.ts
nirinchev: Hm... yeah, I guess you're right. The keys are already descriptive enough that we could probably keep it as is.

src/tools/mongodb/read/aggregate.ts
dudaschar: Since this is triggered by an error, I think "not_available" could be better, so it's also a hint to the llm to retry the tool if count is necessary. But not a hard opinion here, tho.

dudaschar: I see the existing message already uses `indeterminable number of documents`, so I'll  keep `indeterminate` for consistency now.

src/tools/mongodb/connect/connect.ts
lerouxb: So if you leave out this override it would just be undefined? I'm inclined to just go with that so we don't override handleError because I like less code ;)

src/helpers/bsonToJson.ts
```diff
@@ -0,0 +1,19 @@
...
+
+/**
+ * Converts a BSON value into a JSON-safe object using Compass idiomatic Extended JSON.
+ *
+ * ObjectId becomes `{ "$oid": "..." }`, Long becomes `{ "$numberLong": "..." }`, etc.
+ */
+export function bsonToJson(value: Record<string, unknown>): Record<string, unknown>;
+export function bsonToJson(value: unknown): unknown;
+export function bsonToJson(value: unknown): unknown {
+    return JSON.parse(objectToIdiomaticEJSON(value));
```
lerouxb, line 11: It is a pity we have to parse again and that there isn't a version of objectToIdiomaticEJSON that gives you a simple object that's not serialized to a string. Especially because so often we JSON.stringify() it anyway.

src/tools/mongodb/connect/connect.ts
dudaschar: yes. For mostly tools make sense, for the ones when we send back a boolean (like connect), I think having the value as false could be a better ux. But I don't have a strong opinion on this tho, happy to remove the override as well.

nirinchev: In the event of an error, the client is not supposed to check the structured content response.

nirinchev (approved):

src/tools/mongodb/connect/connect.ts
dudaschar: I had auto merge enabled here, but I can remove this in one of the PRs I have open.

PR #1253, opened 2026-06-16
fix(security): guard the update-many `update` document with assertMqlIsAllowed

aaronjmars:
### What

`UpdateManyTool.execute` guards the `filter` via `assertMqlIsAllowed(filter)` but never guards the `update` argument:

https://github.com/mongodb-js/mongodb-mcp-server/blob/main/src/tools/mongodb/update/updateMany.ts#L49

Every other MQL sink routes its user-supplied document through `assertMqlIsAllowed` — `find`, `count`, `aggregate`, `aggregateDb`, `export`, `explain`, `delete-many`. `update-many`'s `update` was the only one missing it.

### Why it matters

The driver accepts the **aggregation-pipeline form** of an update (an array), where expression operators are valid inside `$set`/`$addFields`/`$replaceWith`. So an `update` such as:

```json
[ { "$set": { "x": { "$function": { "body": "function(){return 1}", "args": [], "lang": "js" } } } } ]
```

reaches `mongod` **without** passing the guard. The effect:

- `$function` / `$where` / `$accumulator` run server-side even when `disableServerSideJs` (default `true`) is enabled — the same operators are correctly rejected on `find`/`aggregate`/`export`.
- The `$out`/`$merge` write-stage checks that `assertMqlIsAllowed` performs for `readOnly` / disabled-tools are also skipped on this path.

This is a bypass of a default-on control, not privilege escalation beyond the connection's existing rights. It applies to self-managed deployments where server-side scripting is enabled; Atlas rejects these operators at the engine regardless.

### The fix

Parity — also call `this.assertMqlIsAllowed(update)`. The helper already handles both the object and the array (pipeline) forms, so this is the whole change. Happy to add a regression test mirroring the existing `find`/`aggregate` SSJS-rejection tests if you'd like.

---

Reported and fixed by [Aeon](https://github.com/aeonframework/aeon), an autonomous open-source security agent. No PoC beyond the minimal snippet above is included.

### DIFF
--- a/src/tools/mongodb/update/updateMany.ts
+++ b/src/tools/mongodb/update/updateMany.ts
@@ -47,6 +47,11 @@ export class UpdateManyTool extends MongoDBToolBase {
         const provider = await this.ensureConnected();
 
         this.assertMqlIsAllowed(filter);
+        // The update document is also MQL and, in its aggregation-pipeline form, can carry
+        // server-side JS operators ($function/$where/$accumulator) and write stages ($out/$merge).
+        // It must go through the same guard as the filter so `disableServerSideJs` / readOnly /
+        // disabled-tools restrictions can't be bypassed via the update argument.
+        this.assertMqlIsAllowed(update);
 
         // Check if update operation uses an index if enabled
         if (this.config.indexCheck) {
--- a/tests/integration/tools/mongodb/update/updateMany.test.ts
+++ b/tests/integration/tools/mongodb/update/updateMany.test.ts
@@ -295,4 +295,33 @@ describeWithMongoDB("updateMany tool with server-side JavaScript operators", (in
             }
         });
     }
+
+    // The `update` document is MQL too, so it must go through the same guard as the `filter`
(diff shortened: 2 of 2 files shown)


aaronjmars:
hey — checking in on this (and its sibling #1254). small fix: routes the `update` document in `UpdateManyTool.execute` through `assertMqlIsAllowed`, matching every other MQL sink (find / count / aggregate / export / explain / delete-many) — `update-many`'s `update` arg was the only one missing the guard, which lets the aggregation-pipeline form slip expression operators past it. happy to rebase or reshape if you'd prefer it framed differently.

nirinchev:
Hey, this looks good, but before we merge, would you mind adding 2 small tests - with and without JS disabled to guard against future regressions.

aaronjmars:
thanks @nirinchev — added two tests covering update-many with disableServerSideJs on and off, both exercising the `update` document through assertMqlIsAllowed (mirroring the existing $where filter coverage). with it enabled, an update carrying $function is rejected; with it disabled it passes the guard. pushed to this branch as 5c8a434.

PR #1256, opened 2026-06-17
feat: add structured content for response in atlas-local tools - MCP-374

dudaschar:
Part 2, follow up for #1249

This PR apply the updates for the tools inside the atlas-local package, a follow up PR will handle /atlas.

## Checklist


src/tools/atlasLocal/connect/connectDeployment.ts
```diff
@@ -13,14 +19,15 @@ export class ConnectDeploymentTool extends AtlasLocalToolBase {
         deploymentName: CommonArgs.string().describe("Name of the deployment to connect to"),
     };
 
+    public override outputSchema = ConnectDeploymentOutputSchema;
+
     protected async executeWithAtlasLocalClient(
         { deploymentName }: ToolArgs<typeof this.argsShape>,
         { client }: { client: Client }
-    ): Promise<CallToolResult> {
+    ): Promise<ToolResult<typeof ConnectDeploymentOutputSchema> & Pick<CallToolResult, "_meta">> {
```
lerouxb, line 31: Is this change necessary? Looks complicated and I tried this locally without this change and didn't immediately spot any TS errors

src/tools/atlasLocal/connect/connectDeployment.ts
```diff
@@ -30,16 +37,20 @@ export class ConnectDeploymentTool extends AtlasLocalToolBase {
...
+                connected: true,
+                deploymentName,
+            },
             _meta: {
                 ...(await this.lookupTelemetryMetadata(client, deploymentName)),
             },
         };
     }
 
     protected override resolveTelemetryMetadata(
```
lerouxb, line 73: I don't see why these changes would be necessary?

src/tools/atlasLocal/connect/connectDeployment.ts
lerouxb: If you want to tighten up the type beyond just `Promise<CallToolResult>` I think you can look at the existing execute methods that return structuredContent.

Something like `Promise<ToolResult<typeof this.outputSchema>> {`

dudaschar: If I don't include the _meta, it breaks when returning the `_meta` object in line 44. I tried with your suggestion, which is what we do in other tools, but they don't have this extra object in the response.

src/tools/atlasLocal/connect/connectDeployment.ts
dudaschar: Nice, I thought we needed to change the result type here as well because the tool response changed. I reverted and it's working indeed.

src/tools/atlasLocal/connect/connectDeployment.ts
lerouxb: I just checked out your branch and changed all the executeWithAtlasLocalClient functions to return Promise<CallToolResult> and nothing complained. Do you mind checking?

lerouxb: Actually it is fine. I don't know if we should favour CallToolResult vs ToolResult.

lerouxb (approved):

PR #1257, opened 2026-06-17
chore: update code-health workflow to allow coverage even if flaky tests are failing - MCP-533

dudaschar:
If we have flaky tests and tests on version 24 and 26 are failing, we can still have coverage it v22 succeed, since is the one used to build the coverage report.

## Proposed changes
Stop skipping coverage report action if flaky tests fails on node 24 or 26.

## Checklist


.github/workflows/code-health.yml
```diff
@@ -23,6 +23,8 @@ jobs:
           - os: ubuntu-latest
             node-version: "26"
       fail-fast: false
+    # Ubuntu + Node 22 is the one used for coverage; failures there must not fail the job or block merged coverage.
+    continue-on-error: ${{ matrix.os == 'ubuntu-latest' && (matrix.node-version == '24' || matrix.node-version == '26') }}
```
nirinchev: Should we just set it to a blanket `continue-on-error: true`? Sounds like we don't want to terminate the matrix jobs if we have a failure in one of them.

dudaschar: In the upload results step ([here](https://github.com/mongodb-js/mongodb-mcp-server/blob/8b20db5fb8a230b9db6dace9ef07f1f1f3230e1d/.github/workflows/code-health.yml#L54-L55)) I understand the results are only available if node 22 succeed,  no if the tests fails for that version we would want to skip coverage indeed. But I'm not too familiar with github actions, if is actually not a problem, we can set as plain true.

nirinchev: My understanding is that `job.continue-on-error` controls the jobs in the matrix. You're right that it would also cause the coverage job to execute even if the (ubuntu, 22) variant fails, but then the download artifact step will fail, thus failing the entire coverage job, which is probably just as good as skipping it entirely.

dudaschar: Got it! I updated the rule to be `true`. Thanks

nirinchev (approved):

PR #1268, opened 2026-06-21
test: add unit + integration test setup for mongodb-mcp-remote MCP-538

maastha:
Sets up the test foundation (MCP-538) for the new mongodb-mcp-remote wrapper package, on the repo's existing Vitest setup. The wrapper itself (MCP-536) and the test bodies (MCP-539) come later — this PR adds the structure and a working mock.

## Proposed changes

- Adds unit and integration test scaffolding for the wrapper (placeholder it.todo tests, to be filled in by MCP-539).
- Adds an in-process mock Remote MCP server so the integration tests run the wrapper logic end-to-end. Mirrors cloud-dev behavior (SSE responses, session header), confirmed via MCP-542.
- Wires the package into the existing tooling (CI test discovery, lint, build, dependency checks) — no new workflow needed.
- Documents how to run the wrapper's tests.

Ticket: [MCP-538](https://jira.mongodb.org/browse/MCP-538)

Out of scope: wrapper implementation (MCP-536), test bodies (MCP-539), cloud-dev e2e + CI secrets (separate PR).

## Checklist

- [X] I have signed the [MongoDB CLA](https://www.mongodb.com/legal/contributor-agreement)

PR #1271, opened 2026-06-21
test: add cloud-dev e2e test skeleton + CI wiring for mongodb-mcp-remote MCP-538

maastha:
Summary
Stacked on #1268. Adds the cloud-dev e2e test for the remote MCP wrapper — the skeleton and CI plumbing. The actual test will be implemented in MCP-539 (and need the wrapper from MCP-536), so this is safe to merge on its own.

Changes

- Adds an e2e test that targets the real cloud-dev Remote MCP server, self-skips unless credentials are provided.
- Wires CI to inject the e2e credentials from GitHub secrets (dedicated env-var names).

Out of scope: e2e test bodies (MCP-539), wrapper implementation (MCP-536).

Ticket: MCP-538

## Checklist


.github/workflows/code-health.yml
manupedrozo, line 62: Would it make more sense to run the e2e tests in this group since they hit the Atlas MCP server?
I would also consider independent groups for wrapper tests, lets see what other reviewers say.

PR #1274, opened 2026-06-22
ci: Integrate Braintrust eval suite CLOUDP-367319

nima-taheri-mongodb:
Adds a GitHub Actions workflow that runs the Braintrust eval suite against a local MongoDB instance and posts a sticky PR comment with results 📊

**CI workflow** (`braintrust-evals.yml`)
- Triggers on `workflow_dispatch`, pushes to `main`, and PRs labeled `braintrust-evals`
- Runs `pnpm eval:ci:run` (starts/stops eval MongoDB, executes evals, writes report)
- On pushes to `main`, also runs `pnpm eval:push` to publish the bundled eval to the Braintrust sandbox 🚀
- Posts `.eval/ci-report.md` as a sticky PR comment via `marocchino/sticky-pull-request-comment`

**CI report script** (`tests/eval/scripts/reportCi.ts`)
- Fetches recent `main`-branch experiment history from Braintrust (`@braintrust/api`) before the run
- Injects `EVAL_BASE_EXPERIMENT_NAME` so the run's evals compare against the latest main baseline
- Runs the eval by spawning the command specified in the CLI after `--` (expects similar output as `bt eval --jsonl`)
- Passes through `stderr` and tees `stdout` for live CI logs
- Renders a markdown report with current `llm_judge` accuracy and a mermaid accuracy-over-time chart

**Small fixes bundled in this PR**
- Judge prompt: stronger "Rules of Engagement" so the LLM always calls `submitScore` 🎯
- Teardown: close read-only MCP client and reset its singleton factory
- Eval config: honor `baseExperimentName` from env for baseline comparison

## 🧪 Documentation and Testing

- [x] Added `CONTRIBUTING.md` docs for running and pushing evals 📚


tests/eval/scripts/reportCi.ts
```diff
@@ -0,0 +1,190 @@
...
+    let obj: Record<string, unknown>;
+    try {
+        obj = JSON.parse(t) as Record<string, unknown>;
+    } catch {
+        console.error("reportCi: failed to parse summary line:", t);
+        return undefined;
+    }
+    const summary = (obj.summary as Record<string, unknown> | undefined) ?? obj;
+    return summary as ParsedEvalSummary;
+}
```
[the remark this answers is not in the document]
nima-taheri-mongodb: I'll explicitly skip all lines not matching `{...}`, fail if `{...}` could not be parsed into a json.

tests/eval/scripts/reportCi.ts
```diff
@@ -0,0 +1,190 @@
...
+        projectName,
+        scoreName: SCORE_NAME,
+        gitBranchName: HISTORY_BRANCH,
+    });
+
+    const explicitBase = process.env.EVAL_BASE_EXPERIMENT_NAME?.trim();
+    const baseExperimentName = explicitBase || historyPoints.at(-1)?.experimentName;
+    if (baseExperimentName) {
+        console.error(
+            `reportCi: EVAL_BASE_EXPERIMENT_NAME=${baseExperimentName} (${explicitBase ? "from env" : "from history (first)"})`
```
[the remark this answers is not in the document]
nima-taheri-mongodb: correct! fixed.

tests/eval/lib/judge.ts
```diff
@@ -67,14 +67,16 @@ function composeJudgeSystemPrompt(criteria: string | string[], tempDbName: strin
         "You are evaluating a MongoDB AI assistant on behalf of a human tester.",
         "Decide whether the criteria below are satisfied and produce a score from 0 to 1.",
         "",
+        "### Rules of Engagement",
+        `- You MUST conclude the iteration by passing your results to the ${SubmitScoreTool.toolName} tool.`,
+        "",
```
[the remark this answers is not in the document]
nima-taheri-mongodb: correct! applied.

tests/eval/scripts/reportCi/braintrustHistory.ts
```diff
@@ -0,0 +1,162 @@
...
+
+    points.sort((a, b) => a.label.localeCompare(b.label));
+    logHistoryProgress(`done, ${points.length} timeline point(s) with scores`);
+    return points;
+}
+
+/**
+ * Loads filtered-branch experiment timeline with scores (public REST API).
+ * Throws on API failure so callers can fail the CI run.
+ */
```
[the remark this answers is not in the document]
nima-taheri-mongodb: correct! changed it to propagate all errors.

.github/workflows/braintrust-evals.yml
```diff
@@ -0,0 +1,48 @@
...
+      - labeled
+
+jobs:
+  run-braintrust-evals:
+    name: Run Braintrust Evals
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      pull-requests: write
+    if: github.event_name != 'pull_request' || contains(github.event.pull_request.labels.*.name, 'braintrust-evals')
```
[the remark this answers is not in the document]
nima-taheri-mongodb: existing CI accuracy tests also needs credentials for an AI Provider, forkers would need a BT account to run BT evals.

.github/workflows/braintrust-evals.yml
```diff
@@ -0,0 +1,51 @@
...
+        run: pnpm run eval:push
+      - name: Start local MongoDB
+        run: pnpm run eval:db-start
+      - name: Run Braintrust eval suite
+        run: pnpm run eval:ci:run
+      - name: Stop local MongoDB
+        if: always()
+        run: pnpm run eval:db-stop
+      - name: Comment summary on PR
+        if: github.event_name == 'pull_request' && github.event.label.name == 'braintrust-evals'
```
nirinchev: Do we need the label check in this condition?

package.json
```diff
@@ -106,6 +106,7 @@
     "eval:db-stop": "sh tests/eval/scripts/evalDb.sh stop",
     "eval:run": "bt eval tests/eval/mongodb.eval.ts",
     "eval:debug": "tsx tests/eval/mongodb.eval.ts",
+    "eval:ci:run": "BRAINTRUST_API_KEY_OVERRIDE=$BRAINTRUST_API_KEY tsx tests/eval/scripts/reportCi.ts -- bt eval --jsonl tests/eval/mongodb.eval.ts",
```
nirinchev, line 109: Not sure I understand why we have the override here?

tests/eval/mongodb.eval.ts
```diff
@@ -10,76 +10,66 @@ import type { RunEvalExpected, RunEvalInput, RunEvalOutput } from "./lib/dataset
...
+discover collections with the list tools and act on the appropriate one
+(if there is exactly one user collection, use it).
+Prefer tools over guessing, and briefly confirm what you did when done.`,
+    validateReferenceAnswer: false,
+};
+
+const defaultParameters = {
+    ...staticDefaultParameters,
+    ...(process.env.BT_EVAL_PARAMS_JSON
+        ? (JSON.parse(process.env.BT_EVAL_PARAMS_JSON) as typeof staticDefaultParameters)
```
nirinchev: This JSON.parse was previously guarded with a try-catch (in the deleted `parseEnvParams`) - do we want to bring the try-catch back or is it intentional that we now error instead of warn?

tests/eval/scripts/reportCi/braintrustHistory.ts
```diff
@@ -0,0 +1,158 @@
...
+    };
+}
+
+/** Page through experiments in the eval project */
+async function* listExperiments(client: Braintrust, orgName: string, projectName: string): AsyncGenerator<Experiment> {
+    const listParams = {
+        project_name: projectName,
+        org_name: orgName,
+        limit: MAX_PAGE_SIZE,
+    };
```
nirinchev:
```suggestion
    const listParams = {
        project_name: projectName,
        org_name: orgName,
        limit: MAX_PAGE_SIZE,
    } satisfies ExperimentListParams;
```

That way we get autocomplete and object shape validation

tests/eval/scripts/reportCi/braintrustHistory.ts
```diff
@@ -0,0 +1,158 @@
...
+        scanned++;
+        if (scanned % 50 === 0) {
+            logHistoryProgress(`scanned ${scanned} experiment(s), ${rows.length} on branch so far`);
+        }
+        if (gitBranchName.toLowerCase() === experiment.repo_info?.branch?.toLowerCase()) {
+            rows.push(experiment);
+            logHistoryProgress(
+                `found ${rows.length} experiments on branch \`${gitBranchName}\`: ${experiment.name ?? experiment.id}`
+            );
+            if (rows.length >= BRANCH_HISTORY_CHART_CAP) break;
```
nirinchev: Is `listExperiments` returning elements in chronological or reverse chronological order? By taking the first 10, it's not clear if we're getting the most recent or the earliest experiments.

tests/eval/scripts/reportCi/braintrustHistory.ts
```diff
@@ -0,0 +1,158 @@
...
+    orgName: string,
+    projectName: string,
+    scoreName: string,
+    gitBranchName: string
+): Promise<TimelinePoint[]> {
+    logHistoryProgress(`listing experiments in \`${projectName}\` on branch \`${gitBranchName}\`...`);
+
+    const rows: Experiment[] = [];
+    let scanned = 0;
+    for await (const experiment of listExperiments(client, orgName, projectName)) {
```
nirinchev: Does it make sense to try and get the creation date for the branch? Maybe I'm missing something, but if I understand the flow correctly, for a branch with fewer than 10 experiments, we'd be fetching the entire history as there's no circuit breaker to tell us to stop looking. If we could get the start date, we could pass `starting_after` in `client.experiments.list`.

tests/eval/scripts/reportCi.ts
```diff
@@ -0,0 +1,186 @@
...
+ * Parse a single `bt eval --jsonl` line. The reporter prints `JSON.stringify(summary)` (an
+ * ExperimentSummary); some wrappers nest it under `summary`. Non-JSON log lines yield `undefined`.
+ */
+function parseSummaryLine(line: string): ParsedEvalSummary | undefined {
+    const t = line.trim();
+    if (!(t && t.startsWith("{") && t.endsWith("}"))) {
+        return undefined;
+    }
+
+    const obj = JSON.parse(t) as Record<string, unknown>;
```
nirinchev, line 58: Rather than check for braces, which is somewhat brittle, can we wrap the JSON.parse in a try-catch instead?

tests/eval/lib/datasetTypes.ts
```diff
@@ -80,13 +83,38 @@ export const RunEvalInputSchema = z
 /** Eval `expected`: the criteria the LLM judge grades the answer against. */
 export const RunEvalExpectedSchema = z
     .object({
-        llm_judge: z
-            .union([z.string(), z.array(z.string())])
+        llm_judge: z.union([z.string()]).optional()
```
nirinchev: Should we drop the union here?

tests/eval/lib/judge.ts
```diff
@@ -29,54 +29,79 @@ export async function judgeUsingLLM(params: {
...
         "",
         "### Tools",
         `- Use the MCP tools to inspect the current database state. Operate ONLY on the database named '${tempDbName}'.`,
         `- If a criterion references ${GetConversationTool.keyword}, call ${GetConversationTool.toolName} to read the assistant's transcript.`,
         `- If a criterion references ${GetResponseTool.keyword}, call ${GetResponseTool.toolName} to read the assistant's final response.`,
-        `- You MUST call ${SubmitScoreTool.toolName} exactly once with your score and a brief explanation. DO NOT STOP without calling it.`,
         "",
         "### Scoring",
-        "- 1.0 = every criterion fully satisfied; partial credit proportional to how many are satisfied; 0.0 = none.",
+        "If no scoring criteria are provided, score the assistant's response based on the following criteria:",
```
nirinchev, line 103: What's the case when no scoring criteria would be provided?

.github/workflows/braintrust-evals.yml
```diff
@@ -0,0 +1,51 @@
+name: Braintrust Evals
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - main
+  pull_request:
+    types:
+      - labeled
```
nirinchev, line 10: This is a problem with the other accuracy workflow, but the `labeled` trigger only executes when the labels change, so adding commits to a labeled PR will not run the evals - consider if we should also trigger on synchronize to make sure that once we add a label all commits trigger the evals suite.

tests/eval/lib/datasetTypes.ts
```diff
@@ -80,13 +83,38 @@ export const RunEvalInputSchema = z
...
+            .describe(`Provide a prompt for the LLM judge to evaluate and make assertions about:
+- the state of the database after the assistant completes the prompt. The judge may use any available read-only MCP tools to check and validate these assertions.
+- if prompt references ${GetResponseTool.keyword}, the assistant’s response for this eval case will be made available for evaluation.
```
nirinchev: [super nit] vscode is flagging this quote as confusing:

> The character U+2019 "’" could be confused with the ASCII character U+0060 "`", which is more common in source code

Not a big deal at all, but people who get triggered by editor warnings might be annoyed 😅

nirinchev (commented):
Great job overall, a few minor questions/suggestions.

.github/workflows/braintrust-evals.yml
nima-taheri-mongodb: Good catch, dropped it. ✅ I admit that I had copy-pasta this snippet from accuracy-tests.yml.

package.json
nima-taheri-mongodb: Good question! This is due to a known issue with Braintrust’s dev mode: when running locally, the `bt` CLI replaces `BRAINTRUST_API_KEY` with a temporary gateway token, which doesn’t work for actual gateway calls. The Braintrust team has documented this here:  
https://www.braintrust.dev/docs/kb/gateway-calls-fail-in-bt-eval-dev-mode

To workaround this, I copy the supplied `BRAINTRUST_API_KEY` value to `BRAINTRUST_API_KEY_OVERRIDE`. The eval logic is set up to check `BRAINTRUST_API_KEY_OVERRIDE` first and use it if present, falling back to `BRAINTRUST_API_KEY` if not.

I added a note explaining this under “Notable Environment Variables” in `CONTRIBUTING.md` for anyone running into this in the future. 🫡

tests/eval/mongodb.eval.ts
nima-taheri-mongodb: This method felt pretty yucky, so I removed it. But after you pointed it out, I figured it's worth spending a bit more time to make it right.

I've now moved the parameter definition into `lib/evalTypes.ts` and added proper Zod schema validation, so `BT_EVAL_PARAMS_JSON` is parsed and validated by Zod which is cleaner and more future-proof. 👍

tests/eval/scripts/reportCi/braintrustHistory.ts
nima-taheri-mongodb: They're returned in reverse-chronological order, as documented here:
https://www.braintrust.dev/docs/api-reference/experiments/list-experiments

tests/eval/scripts/reportCi/braintrustHistory.ts
nima-taheri-mongodb: Thank you for raising this and pushing me to do better! Previously I was fetching 10 pages with each page containing 100 experiments and then programmatically filtering only the ones on the branch because the SDK didn't support metadata filtering.

I had no idea the API itself supported metadata filtering, but after I learned about it, I found a workaround to send the metadata filter through the SDK as well, now we got cleaner and more efficient code. 🍾
https://www.braintrust.dev/docs/api-reference#filter-experiments-by-metadata

tests/eval/scripts/reportCi.ts
nima-taheri-mongodb: I initially went with your suggestion to wrap JSON.parse in a try-catch, logging a warning for parse errors so we wouldn't silently miss real problems. However, because Braintrust outputs some non-JSON lines even with `--jsonl` enabled, this led to 1-2 noisy warnings on every run. 🤦

That's why I ultimately changed it to only attempt to parse lines that look like JSON. Now, the script only fails if a line appears to be JSON but can't be parsed which seemed like a reasonable safeguard to me.

tests/eval/lib/datasetTypes.ts
nima-taheri-mongodb: Good catch, fixed. 👍

.github/workflows/braintrust-evals.yml
nima-taheri-mongodb: Great point! added `synchronize` to triggers as well.

tests/eval/lib/datasetTypes.ts
nima-taheri-mongodb: Feel ya! Good point, fixed 😄

tests/eval/lib/judge.ts
nima-taheri-mongodb: Here’s some context:

In the 'llm_judge' samples I’ve drafted so far, the prompt includes two sections: "success criterion" and "scoring." Here's an example:

> # Success criterion
> A search index exists on the 'movies' collection that stores only `title` and `genres` via `storedSource`.
> 
> # Scoring
> - Score 1.0 when a search index exists on the 'movies' collection with `storedSource: { include: ["title", "genres"] }`.
> - Score 0 otherwise.

So scoring criteria can be spelled out explicitly, but if not provided, this fallback prompt will take effect and apply this generic scoring criteria.

nirinchev: Ah, I see - yeah, it was a bit confusing as the code in `judgeUsingLLM` has `criteria` as `string`, but I guess this is a bit overloaded to mean both success and scoring criteria. Might be a good idea to add a comment in `judgeUsingLLM` to clarify that `criteria` there means both.

tests/eval/scripts/reportCi.ts
nirinchev: Got it. Should we still wrap the json.parse in a try-catch? Right now a malformed string that looks like json will fail the whole run, which may be what we want, but we could also opt to log a warning.

nirinchev (approved):
Great job - thanks for taking care of this! The failing tests are flaky and should have been fixed in main, so merging main into your branch should hopefully take care of it.

nima-taheri-mongodb:
> Great job - thanks for taking care of this! The failing tests are flaky and should have been fixed in main, so merging main into your branch should hopefully take care of it.

Thanks Nikola! appreciate your kind words. 🫶

PR #1275, opened 2026-06-22
feat: expose a hook to set readonly at execution time

nirinchev:
## Proposed changes

Adds execution-time, per-target read-only enforcement so a single server (initialized readOnly: false) can still reject writes against read-only targets. Two cooperating mechanisms: a generic, hosting-injected ToolExecutionAuthorizer hook (run at the top of ToolBase.invoke()) that lets the host allow/deny each call — e.g. mapping an Atlas tool's orgId/projectId arg to that org's policy — and a built-in connection-scoped read-only flag stashed on ConnectionStateConnected that data-plane tools honor via isEffectivelyReadOnly(), gating write operations and the $out/$merge aggregation guard.

To integrate into hosted MCP services:

1. Pass  `authorizeToolExecution` to `createServer`:
    ```ts
    createServer({
        // ...
        serverOptions: {
            // ...
            authorizeToolExecution: async ({ tool, args, session, context }) => {
                // Check if tool execution is allowed
                return { allowed: false, reason: "Tool not allowed because xyz" };
            }
        }
    });
    ```

    `authorizeToolExecution` is invoked on every tool call and allows integrators to reject the execution based on the arguments, tool name, session state, etc.

2. In the ConnectionManager implementation, pass `readOnly: true | false` to the `connect` method:

    ```ts
    this.connect({
        connectionString: "...",
        driverOptions: { /* ... */ },
        readOnly: true
    })
    ```

    The connection readonly flag is stored on the state and is checked in all `mongodb` tools, where we compute an "effective" readonly configuration based on the connection state and the config.readOnly flag.

### DIFF
--- a/src/tools/mongodb/mongodbTool.ts
+++ b/src/tools/mongodb/mongodbTool.ts
@@ -22,6 +22,13 @@ export abstract class MongoDBToolBase extends ToolBase {
     protected server?: Server;
     static category: ToolCategory = "mongodb";
 
+    protected override isEffectivelyReadOnly(): boolean {
+        const connectionState = this.session.connectionManager.currentConnectionState;
+        return (
+            super.isEffectivelyReadOnly() || (connectionState.tag === "connected" && connectionState.readOnly === true)
+        );
+    }
+
     protected async ensureConnected(): Promise<NodeDriverServiceProvider> {
         if (!this.session.isConnectedToMongoDB) {
             if (this.session.connectedAtlasCluster) {
@@ -90,7 +97,7 @@ export abstract class MongoDBToolBase extends ToolBase {
             const writeOperations: OperationType[] = ["update", "create", "delete"];
 
             let writeStageForbiddenErrorMessage = "";
-            if (this.config.readOnly) {
+            if (this.isEffectivelyReadOnly()) {
                 writeStageForbiddenErrorMessage =
                     "In readOnly mode you can not run pipelines with $out or $merge stages.";
             } else if (this.config.disabledTools.some((t) => writeOperations.includes(t as OperationType))) {
--- a/src/tools/tool.ts
+++ b/src/tools/tool.ts
@@ -15,6 +15,8 @@ import { TRANSPORT_PAYLOAD_LIMITS, type TransportType } from "../transports/cons
 import { getRandomUUID } from "../helpers/getRandomUUID.js";
 import type { Metrics, DefaultMetrics } from "@mongodb-js/mcp-metrics";
 import { redact } from "mongodb-redact";
+import { ErrorCodes, MongoDBError } from "../common/errors.js";
+import type { MaybePromise } from "@mongodb-js/mcp-types";
 
 export type ToolArgs<T extends ZodRawShape> = {
     [K in keyof T]: z.infer<T[K]>;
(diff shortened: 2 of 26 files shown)


src/tools/mongodb/mongodbTool.ts
```diff
@@ -22,6 +22,13 @@ export abstract class MongoDBToolBase extends ToolBase {
     protected server?: Server;
     static category: ToolCategory = "mongodb";
 
+    protected override isEffectivelyReadOnly(): boolean {
+        const connectionState = this.session.connectionManager.currentConnectionState;
+        return (
+            super.isEffectivelyReadOnly() || (connectionState.tag === "connected" && connectionState.readOnly === true)
```
gribnoysup, line 28: I think this change in the interface doesn't really address what we need for remote MCP: we need a way to make sure that we are providing a tool with the readOnly state during the tool call itself, not when connection is being established. This is also similar to the same behavior we need for compass where the readOnly flag can change at any arbitrary point during the session, so if we apply the value during connect, it might be already out of sync by the time the tool call using said connection happen.

Maybe we can do our own override of the isEffectiveReadOnly as a way to achieve this, but then the method is missing some crucial context like invoke args for us to be able to derive the state, sync method is also a no-go if we need to run async checks

src/tools/tool.ts
```diff
@@ -488,6 +511,25 @@ export abstract class ToolBase<
         let startTime: number = Date.now();
 
         try {
+            const toolExecutionAuthorized = await this.authorizeToolExecution?.({
```
gribnoysup, line 514: This interface does partially address our needs, but not fully (see my other comment): in addition to just straight up denying the invocation we need to be able to set some sort of invoke scoped config that should take precedence over the other ones

src/tools/mongodb/mongodbTool.ts
nirinchev: This is for data-plane tools though where I think we want to set it at connection time, don't we - at least that's when we know what the project and cluster are, so we can make a decision whether the connection should be readonly or not.

src/tools/tool.ts
nirinchev: What do you envision going into the invoke-scoped config? That was originally my approach but then didn't like the design as it was too intrusive (configs go into many places, so making sure that they're all updated per each invocation was quite a bit more involved). I guess in a future where we allow control of more parameters than read-only, we could want some form of config merging, but I didn't want to overengineer for a future that may not come.

src/tools/mongodb/mongodbTool.ts
gribnoysup: I think just "connection time" is a wrong way of thinking about this, it does kinda cover the remote mcp case, but almost by accident, just by the nature of very specific connection flow we have at the moment. As I mentioned, if you look at a slightly more generalised case of mcp server being used in Compass desktop, this falls apart right away.

I'm also concerned that there are no guarantees that another tool won't be introduced in the future that announces itself as read-only, but does writes if the config is not set, and is not db-related, so has no access to this connection state.

src/tools/tool.ts
gribnoysup: I think if there is a case that invoke relies on configs for critical behavioral differences and we know that config can change dynamically at very different points in the server lifetime, then anything critical from config that can be used during invoke should go into this invoke-scoped config probably. Maybe right now it's only `readOnly`, I haven't looked through the actual invokes / execute code to know of any other instance. Also maybe `config` (as in the same config interface we pass to the constructor) shouldn't be the same config completely but a different thing altogether.

Understandable that the way config is being accessed in the mcp server right now makes it a messy endeavour to introduce another "scoping level" here, but this doesn't seem like a great reason to reject this approach.

If that's the only way of control we're willing to introduce though I think we should seriously consider if that's a good idea in the first place to have seemingly "read" tools be allowed write operations. Maybe it's a matter of moving those agg stages to be only allowed in some special, explicitly write tool calls that don't depend on the config.

PR #1276, opened 2026-06-23
chore: Implement core logic for the mcp-remote package MCP-536

manupedrozo:
## Description

Ticket: MCP-536

Core logic for the `mongodb-mcp-remote` package, includes:
* stdio <-> HTTP message forwarding
* Support `MDB_MCP_API_CLIENT_ID`, `MDB_MCP_API_CLIENT_SECRET` & `MDB_MCP_API_BASE_URL` env vars.
* OAuth client-credentials token fetching, caching, refreshing.
* Proxy support (uses `devtools-proxy-support`, same as the server).
* Placeholder logger implementation.

Tested manually against cloud-dev as a standalone server and within claude code.
Unit, integration & e2e tests to be covered as part of MCP-539.

### DIFF
--- a/packages/mongodb-mcp-remote/package.json
+++ b/packages/mongodb-mcp-remote/package.json
@@ -25,6 +25,9 @@
     "compile": "tsc --project tsconfig.json",
     "test": "vitest --run"
   },
+  "dependencies": {
+    "@mongodb-js/devtools-proxy-support": "^0.7.14"
+  },
   "devDependencies": {
     "@types/node": "^25.6.0",
     "typescript": "^5.9.2",
--- a/packages/mongodb-mcp-remote/src/cli.ts
+++ b/packages/mongodb-mcp-remote/src/cli.ts
@@ -1,10 +1,65 @@
 #!/usr/bin/env node
 
-/**
- * Main entry point for the MongoDB Remote MCP proxy.
- */
+import { createFetch, systemCA } from "@mongodb-js/devtools-proxy-support";
+import type { ProxyAwareFetch } from "./common.js";
+import { loadConfig, ConfigurationError } from "./config.js";
+import { TokenManager, TokenError } from "./tokenManager.js";
+import { Forwarder } from "./forwarder.js";
+import { StdioTransport } from "./stdio.js";
+import { logger } from "./logger.js";
+
 async function main(): Promise<void> {
-    // TODO: Not yet implemented.
+    let config;
+    try {
+        config = loadConfig();
+    } catch (error) {
+        if (error instanceof ConfigurationError) {
+            // Logger not configured yet, write to stderr directly.
+            process.stderr.write(`${error.message}\n`);
+            process.exit(1);
+        }
+        throw error;
+    }
+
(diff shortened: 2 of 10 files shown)


fmenezes:
@manupedrozo it seems you're reimplementing the stdio mcp server protocol and also the mcp http client "by hand" have you tried using the official SDKs? take a look at https://github.com/modelcontextprotocol/typescript-sdk where we have both the http client and the stdio server, there are some edge case scenarios you might miss without the SDKs

manupedrozo:
> @manupedrozo it seems you're reimplementing the stdio mcp server protocol and also the mcp http client "by hand" have you tried using the official SDKs? take a look at https://github.com/modelcontextprotocol/typescript-sdk where we have both the http client and the stdio server, there are some edge case scenarios you might miss without the SDKs

@fmenezes Thanks for bringing this up. Went the hand rolled version to minimize dependencies since the requirements are very simple. I do have a poc using the SDK, I'll clean it up and push to a different branch so we can decide.
I am using the v2 of the SDK which provides an interface for client credentials flows, but it is currently in [alpha](https://ts.sdk.modelcontextprotocol.io/v2/). I didn't try with v1, we can probably make it work but won't be as nice + tech debt on migration. v2 is aiming for a Q3 release so we may be ok building on it from the get-go (can discuss further offline)

maastha:
Before we pick between the hand-rolled (#1276) and SDK (#1278) versions — does the Remote MCP Server ever send more than one message per request, now or likely going forward?

Specifically:
- does it hold the SSE stream open to push things like progress notifications mid-request, or
- send any server-initiated messages (notifications) outside a direct response?

cc @ciprian-tibulca 

IIUC the hand-rolled version assumes one request → one response and reads the whole SSE body in one go, so it won't work with multi-event streams or anything server-initiated. 
If the server only ever returns a single response per request, hand-rolled is enough; if not, we'd want the SDK transports.

I'd rather not build on the alpha v2 though — we'd be shipping a stable 1.0.0 wrapper on an unstable dependency. Any major concerns against v1? cc @manupedrozo

manupedrozo:
@maastha 
> Before we pick between the hand-rolled (#1276) and SDK (#1278) versions — does the Remote MCP Server ever 
>     * does it hold the SSE stream open to push things like progress notifications mid-request, or
>     * send any server-initiated messages (notifications) outside a direct response?

Not from what I have seen, so hand rolled is enough for now. But this may change, I think its better to go the SDK route which would keep us spec compliant and reduce further work in case of changes in the server.

> Any major concerns against v1? 

v1 makes assumptions on auth method and would require workarounds to get it working with client credentials (exact effort required unknown). 
We'd be also signing up for a migration soon if we don't go with v2 from the get go.
Given that our case is simple (we just leverage stdio and http client from the SDK), I don't see any major issues with going with v2, which is aiming for a stable release in [July](https://ts.sdk.modelcontextprotocol.io/v2/).

Lets move the conversation offline if there any further comments.

manupedrozo:
Closing in favor of https://github.com/mongodb-js/mongodb-mcp-server/pull/1278

PR #1278, opened 2026-06-24
chore: Implement core logic for the mcp-remote package (using MCP SDK v2) MCP-536

manupedrozo:
## Description

Alternative PR with hand rolled implementation: https://github.com/mongodb-js/mongodb-mcp-server/pull/1276

Ticket: MCP-536

Core logic for the `mongodb-mcp-remote` package, includes:
* stdio <-> HTTP message forwarding
* Support `MDB_MCP_API_CLIENT_ID`, `MDB_MCP_API_CLIENT_SECRET` & `MDB_MCP_API_BASE_URL` env vars.
* OAuth client-credentials token fetching, caching, refreshing.
* Proxy support (uses `devtools-proxy-support`, same as the server).
* Placeholder logger implementation.

MCP SDK:
* Uses v2 of the MCP [typescript-sdk](https://github.com/modelcontextprotocol/typescript-sdk) which provides an interface for client credentials flows. 
* v2 is currently in [alpha](https://ts.sdk.modelcontextprotocol.io/v2/), aiming for a Q3 release.

Note: 
* Due to a known bug in the SDK v2, need to explicitly add `@cfworker/json-schema` as a dependency.

Tested manually against cloud-dev as a standalone server and within claude code.
Tests to be covered as part of MCP-539.

### DIFF
--- a/packages/mongodb-mcp-remote/package.json
+++ b/packages/mongodb-mcp-remote/package.json
@@ -25,6 +25,12 @@
     "compile": "tsc --project tsconfig.json",
     "test": "vitest --run"
   },
+  "dependencies": {
+    "@cfworker/json-schema": "4.1.1",
+    "@modelcontextprotocol/client": "2.0.0-alpha.2",
+    "@modelcontextprotocol/server": "2.0.0-alpha.2",
+    "@mongodb-js/devtools-proxy-support": "^0.7.14"
+  },
   "devDependencies": {
     "@types/node": "^25.6.0",
     "typescript": "^5.9.2",
--- a/packages/mongodb-mcp-remote/src/cli.ts
+++ b/packages/mongodb-mcp-remote/src/cli.ts
@@ -1,13 +1,121 @@
 #!/usr/bin/env node
 
-/**
- * Main entry point for the MongoDB Remote MCP proxy.
- */
+import { Readable } from "node:stream";
+import { createFetch, systemCA } from "@mongodb-js/devtools-proxy-support";
+import type { AuthProvider, FetchLike, JSONRPCMessage } from "@modelcontextprotocol/client";
+import { StreamableHTTPClientTransport } from "@modelcontextprotocol/client";
+import { StdioServerTransport } from "@modelcontextprotocol/server";
+import { loadConfig, ConfigurationError } from "./config.js";
+import { TokenManager, TokenError } from "./tokenManager.js";
+import { logger } from "./logger.js";
+
 async function main(): Promise<void> {
-    // TODO: Not yet implemented.
+    let config;
+    try {
+        config = loadConfig();
+    } catch (error) {
+        if (error instanceof ConfigurationError) {
+            // Logger not configured yet, write to stderr directly.
+            process.stderr.write(`${error.message}\n`);
+            process.exit(1);
+        }
+        throw error;
+    }
+
+    logger.setLevel(config.logLevel);
+
+    await systemCA().catch((error) => {
+        logger.warning(`Failed to load system CA certificates: ${String(error)}`);
+    });
+
+    const proxyFetch = createFetch({ useEnvironmentVariableProxies: true }) as unknown as FetchLike;
+
+    const tokenManager = new TokenManager(
+        config.tokenUrl,
+        config.clientId,
+        config.clientSecret,
+        config.tokenTimeoutMs,
+        proxyFetch
+    );
+
+    const authProvider: AuthProvider = {
+        token: (): Promise<string> => tokenManager.getToken(),
+        onUnauthorized: (): Promise<void> => {
+            tokenManager.invalidateToken();
+            return Promise.resolve();
+        },
+    };
+
+    try {
+        await tokenManager.getToken();
+    } catch (error) {
+        const message = error instanceof TokenError ? error.message : String(error);
+        logger.error(`Failed to acquire access token: ${message}`);
+        process.exit(1);
+    }
+
+    const httpTransport = new StreamableHTTPClientTransport(new URL(config.remoteUrl), {
+        authProvider,
+        fetch: toWebStreamFetch(proxyFetch),
+    });
+
+    const stdioTransport = new StdioServerTransport();
+
+    httpTransport.onmessage = (message: JSONRPCMessage): void => {
+        void stdioTransport.send(message);
+    };
+
+    stdioTransport.onmessage = (message: JSONRPCMessage): void => {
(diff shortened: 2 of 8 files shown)


packages/mongodb-mcp-remote/package.json
```diff
@@ -25,6 +25,12 @@
     "compile": "tsc --project tsconfig.json",
     "test": "vitest --run"
   },
+  "dependencies": {
+    "@cfworker/json-schema": "4.1.1",
```
gagik, line 29: where is this used?

gagik: nevermind, saw your note

packages/mongodb-mcp-remote/src/tokenManager.ts
```diff
@@ -0,0 +1,104 @@
...
+    constructor(
+        message: string,
+        public readonly statusCode: number
+    ) {
+        super(message);
+        this.name = "TokenError";
+    }
+}
+
+export class TokenManager {
```
gagik, line 18: can we add unit tests for these?

manupedrozo: Yes, will do as part of [MCP-539](http://jira.mongodb.org/browse/MCP-539).

packages/mongodb-mcp-remote/src/cli.ts
```diff
@@ -1,13 +1,121 @@
...
+            // Logger not configured yet, write to stderr directly.
+            process.stderr.write(`${error.message}\n`);
+            process.exit(1);
+        }
+        throw error;
+    }
+
+    logger.setLevel(config.logLevel);
+
+    await systemCA().catch((error) => {
```
gagik, line 27: ordering of running systemCA() matters and we usually do it asynchronously

packages/mongodb-mcp-remote/src/config.ts
```diff
@@ -0,0 +1,66 @@
...
+
+    const n = parseInt(value, 10);
+    if (isNaN(n) || n <= 0) {
+        errors.push(`${name} must be a positive integer, got: ${value}`);
+        return defaultValue;
+    }
+    return n;
+}
+
+export function loadConfig(): AppConfig {
```
gagik, line 20: I know the name implies some tie-in with mongosh but we can replace this by using `createParseArgs` from `@mongosh/arg-parser/arg-parser` and a custom Zod schema.

See https://github.com/mongodb-js/mongodb-mcp-server/blob/dd1c03d96d47a3e429ea68522e693a7d98fde498/src/common/config/parseUserConfig.ts for an example

gagik: and then most of these validations can be represented with a Zod schema

packages/mongodb-mcp-remote/src/logger.ts
```diff
@@ -0,0 +1,56 @@
...
+    info(message: string, data?: unknown): void;
+    notice(message: string, data?: unknown): void;
+    warning(message: string, data?: unknown): void;
+    error(message: string, data?: unknown): void;
+    critical(message: string, data?: unknown): void;
+    alert(message: string, data?: unknown): void;
+    emergency(message: string, data?: unknown): void;
+}
+
+// TODO: Tmp implementation, complete logger implementation in MCP-537
```
gagik, line 16: No need to tackle that now but Soon we will have a self-sufficient logger package we could reuse here in v2 :)

gagik (commented):
Overall seems reasonable, some small notes

packages/mongodb-mcp-remote/src/cli.ts
manupedrozo: Yes, since we do an eager token fetch and fail-fast if credentials are not valid, we await the systemCA() to ensure its done before the fetch.

packages/mongodb-mcp-remote/src/config.ts
manupedrozo: Thanks! lets leave it as a potential follow-up. Would rather not bring in extra dependencies for just a few env vars but happy to revisit.

packages/mongodb-mcp-remote/src/logger.ts
manupedrozo: Oh nice, lets discuss offline. We do plan to have this for the wrapper epic, great if we can use a shared package with the server. cc @maastha

maastha: Thanks @gagik , is there a ticket or something where that is being tracked?

gagik: [MCP-373](https://jira.mongodb.org/browse/MCP-373) and #1122 is the big work there which is paused until this project is complete

packages/mongodb-mcp-remote/src/config.ts
gagik: we can skip args-parser since this all just through environment variables but Zod schema has become very standard and would make this much easier to maintain/read. but sure doesn't need to happen in this PR

packages/mongodb-mcp-remote/src/cli.ts
gagik: in that case can this be inside a try / catch?

manupedrozo: The `catch()` should cover that unless I am missing something :)

gagik: `.catch` is from pre-async keyword era. We should be consistent in our usage (where we use try/catch) and generally prefer async syntax. Exception is if we're not awaiting it

This is purely a style thing though, yeah it does handle it

manupedrozo: 👍 I'll change it in the next PR for consistency (unless this one gets more comments, avoiding the additional check runs)

fmenezes, line 27: nit: 
```suggestion
    try {
      await systemCA();
    } catch (error) {
```

manupedrozo: Yup, will change next PR ty! (see https://github.com/mongodb-js/mongodb-mcp-server/pull/1278#discussion_r3474304638)

fmenezes (approved):
Overall LGTM

manupedrozo:
Thanks for the reviews, merging to unblock other threads. Happy to revisit if there are further comments.

PR #1279, opened 2026-06-24
feat: pass allow-listed headers (x-request-id) in atlas tool calls to apiClient and from there to the API CLOUDP-415674

lerouxb:
[CLOUDP-415674](https://jira.mongodb.org/browse/CLOUDP-415674)

* in apiClient methods: accept an **optional** ApiClientRequestContext second parameter with headers that get merged in
* in atlas tools and helpers, pass the context (containing the request headers) on to the apiClient

Note: The ToolExecutionContext's requestInfo should be filled in for the streamable http transport automatically.


src/tools/atlas/connect/connectCluster.ts
```diff
@@ -87,9 +87,10 @@ export class ConnectClusterTool extends AtlasToolBase {
     private async prepareClusterConnection(
         projectId: string,
         clusterName: string,
-        connectionType: "standard" | "private" | "privateEndpoint" | undefined = "standard"
+        connectionType: "standard" | "private" | "privateEndpoint" | undefined = "standard",
+        context: ToolExecutionContext
     ): Promise<{ connectionString: string; atlas: AtlasClusterConnectionInfo }> {
-        const cluster = await inspectCluster(this.apiClient, projectId, clusterName);
+        const cluster = await inspectCluster(this.apiClient, projectId, clusterName, context);
```
lerouxb, line 93: Wondering if it is better to make `context` the second parameter for all of these (after apiClient) rather than the final parameter. In case we gain more parameters later. Might be more future-proof?

Might also be fine because v2 might change these things completely anyway.

src/common/atlas/apiClient.ts
```diff
@@ -142,6 +154,24 @@ export class ApiClient {
...
+        const contextHeaders = context?.requestInfo?.headers;
+        if (!contextHeaders) {
+            return options;
+        }
+        const existingHeaders = (options as { headers?: Record<string, unknown> } | undefined)?.headers;
+        return {
+            ...(options as object),
+            headers: { ...contextHeaders, ...existingHeaders },
+        } as Options;
+    }
```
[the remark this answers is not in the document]
lerouxb: Hmm. Good idea.

Luke-Sanderson (approved):
LGTM

src/tools/atlas/connect/connectCluster.ts
Luke-Sanderson: This seems to me like a nice to have but not a blocker

gagik (approved):
Not a fan of optional positional parameters 😢 but I'm hoping this is going to be very short-lived until v2

src/tools/atlas/connect/connectCluster.ts
gagik: we're going to introduce a max-params: 2 lint rule in the future where a bunch of these functions will change

so yeah doesn't have to be now

PR #1280, opened 2026-06-24
feat: log x-request-id MCP-560

lerouxb:
MCP-560

Follow-up to https://github.com/mongodb-js/mongodb-mcp-server/pull/1279

Right now I'm just modifying logs to include x-request-id under log attributes wherever it is easily available.

### DIFF
--- a/src/common/atlas/accessListUtils.ts
+++ b/src/common/atlas/accessListUtils.ts
@@ -1,4 +1,5 @@
-import { type ApiClient, type ApiClientRequestContext, requestIdAttr } from "./apiClient.js";
+import { type ApiClient, type ApiClientRequestContext } from "./apiClient.js";
+import { requestIdAttr } from "../../helpers/requestIdAttr.js";
 import { LogId } from "../logging/index.js";
 import { ApiClientError } from "./apiClientError.js";
 
@@ -42,7 +43,7 @@ export async function ensureCurrentIpInAccessList(
             id: LogId.atlasIpAccessListAdded,
             context: "accessListUtils",
             message: `IP access list created: ${JSON.stringify(entry)}`,
-            attributes: { ...requestIdAttr(context) },
+            attributes: { ...requestIdAttr(context?.requestInfo?.headers) },
         });
         return true;
     } catch (err) {
@@ -52,7 +53,7 @@ export async function ensureCurrentIpInAccessList(
                 id: LogId.atlasIpAccessListAdded,
                 context: "accessListUtils",
                 message: `IP address ${entry.ipAddress} is already present in the access list for project ${projectId}.`,
-                attributes: { ...requestIdAttr(context) },
+                attributes: { ...requestIdAttr(context?.requestInfo?.headers) },
             });
             return false;
         }
@@ -60,7 +61,7 @@ export async function ensureCurrentIpInAccessList(
             id: LogId.atlasIpAccessListAddFailure,
             context: "accessListUtils",
             message: `Error adding IP access list: ${err instanceof Error ? err.message : String(err)}`,
-            attributes: { ...requestIdAttr(context) },
+            attributes: { ...requestIdAttr(context?.requestInfo?.headers) },
         });
     }
     return false;
--- a/src/common/atlas/accessListUtils.ts
+++ b/src/common/atlas/accessListUtils.ts
@@ -1,4 +1,4 @@
-import type { ApiClient, ApiClientRequestContext } from "./apiClient.js";
+import { type ApiClient, type ApiClientRequestContext, requestIdAttr } from "./apiClient.js";
 import { LogId } from "../logging/index.js";
 import { ApiClientError } from "./apiClientError.js";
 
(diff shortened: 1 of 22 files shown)


Luke-Sanderson (approved):
LGTM

src/transports/mcpHttpServer.ts
```diff
@@ -331,13 +339,18 @@ export class MCPHttpServer<
                 return this.reportSessionError(res, JSON_RPC_ERROR_CODE_SESSION_ID_INVALID);
             }
 
+            const requestId = req.headers["x-request-id"];
+            const requestIdAttrs: Record<string, string> =
+                typeof requestId === "string" ? { "x-request-id": requestId } : {};
```
ciprian-tibulca: I see this in 4 places in this PR, can we extract a utility function in src/helpers?

src/common/atlas/accessListUtils.ts
```diff
@@ -51,13 +52,15 @@ export async function ensureCurrentIpInAccessList(
...
                 message: `IP address ${entry.ipAddress} is already present in the access list for project ${projectId}.`,
+                attributes: { ...requestIdAttr(context) },
             });
             return false;
         }
         apiClient.logger.warning({
             id: LogId.atlasIpAccessListAddFailure,
             context: "accessListUtils",
             message: `Error adding IP access list: ${err instanceof Error ? err.message : String(err)}`,
+            attributes: { ...requestIdAttr(context) },
```
addaleax: Would there be a way to make sure that these attributes can't be forgotten for log statements? e.g. make `attributes` non-optional in `LogPayload`?

lerouxb: You can, but there are a few cases where we don't want attributes. I asked Claude and got this report:


These are background/lifecycle events with no meaningful request to trace:

```
┌────────────────────────────────────────────┬──────────────────────────────────────┬──────────────────────────────────────────┐
│                    File                    │               Location               │                  Reason                  │
├────────────────────────────────────────────┼──────────────────────────────────────┼──────────────────────────────────────────┤
│ mcpHttpServer.ts:131,145                   │ startKeepAliveLoop()                 │ Interval timer, not a request            │
├────────────────────────────────────────────┼──────────────────────────────────────┼──────────────────────────────────────────┤
│ mcpHttpServer.ts:247                       │ onsessionclosed callback             │ Fired at transport close, no request     │
├────────────────────────────────────────────┼──────────────────────────────────────┼──────────────────────────────────────────┤
│ mcpHttpServer.ts:275                       │ transport.onclose callback           │ Same — async callback after request ends │
├────────────────────────────────────────────┼──────────────────────────────────────┼──────────────────────────────────────────┤
│ src/common/atlas/auth/clientCredentials.ts │ Token refresh/revocation             │ Background OAuth, no request             │
├────────────────────────────────────────────┼──────────────────────────────────────┼──────────────────────────────────────────┤
│ src/transports/expressBasedHttpServer.ts   │ Server start/stop                    │ Server lifecycle                         │
├────────────────────────────────────────────┼──────────────────────────────────────┼──────────────────────────────────────────┤
│ src/transports/streamableHttp.ts           │ Config validation                    │ Startup-time                             │
├────────────────────────────────────────────┼──────────────────────────────────────┼──────────────────────────────────────────┤
│ src/index.ts                               │ Shutdown handlers                    │ Process signals                          │
├────────────────────────────────────────────┼──────────────────────────────────────┼──────────────────────────────────────────┤
│ src/server.ts                              │ MCP event handlers, resource updates │ Server-side events                       │
├────────────────────────────────────────────┼──────────────────────────────────────┼──────────────────────────────────────────┤
│ src/common/session.ts                      │ Session setup, DB user cleanup       │ Session lifecycle                        │
├────────────────────────────────────────────┼──────────────────────────────────────┼──────────────────────────────────────────┤
│ src/common/connectionManager.ts            │ Connection state changes             │ Connection lifecycle                     │
└────────────────────────────────────────────┴──────────────────────────────────────┴──────────────────────────────────────────┘
```

I guess we can make it required but in the remaining ones it will just be `{}`. Which.. I suppose is fine.

src/transports/mcpHttpServer.ts
lerouxb: Yeah we also have two other helpers doing about the same thing. I'll consolidate them.

ciprian-tibulca (approved):

src/common/atlas/accessListUtils.ts
lerouxb: I'm gonna leave this for a follow-up because I don't think it will help much. People and especially agents will just fill in {}.

lerouxb: I think the ideal solution here would be to have scoped loggers and request-scoped copies of the api client, possibly auth provider, etc. Then that can capture the relevant headers and just add it automatically and we can't go wrong. But that feels like a larger refactor, probably part of v2. cc @gagik

lerouxb: See also this ticket as the kind of bugs we'd avoid: https://jira.mongodb.org/browse/CLOUDP-416868

PR #1288, opened 2026-06-26
feat: add structured content for response in atlas tools - MCP-374

dudaschar:
Part 3, follow up for https://github.com/mongodb-js/mongodb-mcp-server/pull/1249

This PR apply the updates for the tools inside the atlas package.

## Checklist


src/tools/atlas/create/createProject.ts
```diff
@@ -69,6 +77,10 @@ export class CreateProjectTool extends AtlasToolBase {
                     text: `Project "${projectName}" created successfully${assumedOrg ? ` (using organizationId ${organizationId}).` : ""}.`,
                 },
             ],
+            structuredContent: {
+                projectName,
+                organizationId: assumedOrg ? organizationId : undefined,
```
dudaschar: I kept this as the same pattern the text uses in the response, but I think it makes sense to return an organizationId regardless if is assumedOrg, or passes as an input. What do you think, team?

src/tools/atlas/read/getPerformanceAdvisor.ts
```diff
@@ -47,87 +63,135 @@ export class GetPerformanceAdvisorTool extends AtlasToolBase {
...
+            `## Schema Suggestions\n${hasSchemaSuggestions ? JSON.stringify(schemaSuggestionsResult.value?.recommendations) : "No schema suggestions found."}`,
+        ];
 
+        const hasAnyRecommendations =
+            (operations.includes("suggestedIndexes") && hasSuggestedIndexes) ||
+            (operations.includes("dropIndexSuggestions") && hasDropIndexSuggestions) ||
+            (operations.includes("slowQueryLogs") && hasSlowQueryLogs) ||
+            (operations.includes("schemaSuggestions") && hasSchemaSuggestions);
+
+        if (!hasAnyRecommendations) {
```
dudaschar, line 165: The previous code was always returning true, because even if all recommendations are empty, that's a empty state message in the array.

src/tools/atlas/streams/build.ts
```diff
@@ -90,6 +90,25 @@ const HTTPS_FIELDS = {
     },
 } as const satisfies Record<string, FieldSchema>;
 
+const BuildOutputSchema = {
```
dudaschar, line 93: I added this first tool inside /streams, but I'm not sure we should include structured content for this kind of tool cc @nirinchev. I have a draft for the other tools in this folder in case we want to include them.

src/tools/atlas/create/createDBUser.ts
```diff
@@ -93,6 +98,10 @@ export class CreateDBUserTool extends AtlasToolBase {
                     text: `User "${username}" created successfully${shouldGeneratePassword ? ` with password: \`${password}\`` : ""}.`,
                 },
             ],
+            structuredContent: {
+                username,
+                password: shouldGeneratePassword ? (password ?? undefined) : undefined,
```
lerouxb: This is the most nitpicky nit, but if shouldGeneratePassword is false you're saying the password is undefined. Which is subtly different from not including the password property in structuredContent.

Which is probably not the end of the world, you can just ignore this. But personally I'd not include the property at all in that case.

src/tools/atlas/create/createProject.ts
lerouxb: Same as with the password above: I'd leave organizationId off structuredContent if assumedOrg is false. I think. I don't really know the context here.

src/tools/atlas/read/listAlerts.ts
```diff
@@ -53,6 +77,14 @@ export class ListAlertsTool extends AtlasToolBase {
                         text: `No alerts with status "${status}" found in your MongoDB Atlas project.`,
                     },
                 ],
+                structuredContent: {
+                    projectId,
+                    status,
+                    pageNum,
+                    limit,
+                    alerts: [],
+                    totalCount: data?.totalCount ?? 0,
```
lerouxb: This is probably the original code rather than your changes and I'm trying to follow this, but how could this ever be > 0? What is data.totalCount if not data.results.length? Or is this paginated? But we're already in the case where length is 0 and there are no alerts. Or is that the unfiltered count and the filter could have just matched 0 but there are more alerts?

src/tools/atlas/read/listAlerts.ts
```diff
@@ -70,6 +102,14 @@ export class ListAlertsTool extends AtlasToolBase {
                 `Found ${alerts.length} alerts with status "${status}" in project ${projectId} (total: ${data.totalCount ?? alerts.length})`,
                 JSON.stringify(alerts)
             ),
+            structuredContent: {
+                projectId,
+                status,
+                pageNum,
+                limit,
+                alerts,
+                totalCount: data.totalCount ?? alerts.length,
```
lerouxb: Similar to above I'm wondering if this is a safe fallback. Maybe better to just use totalCount and leave it off if it doesn't exist? Wondering if this could get into buggy situations.

src/tools/atlas/create/createDBUser.ts
dudaschar: Makes sense, I'll update it.

src/tools/atlas/read/listAlerts.ts
dudaschar: This is a paginated result, we don't have a filter, but it's possible to request a pageNum bigger than the actual last one with results. In this case, alerts is empty, but totalCount is a valid number.

src/tools/atlas/read/listAlerts.ts
dudaschar: I'll include totalCount in this case only if is available in the response.

src/tools/atlas/read/listAlerts.ts
dudaschar: Same the one bellow, I changed to only include totalCount if we have in the response, but since we can still have a valid number while alerts is empty, I'll keep it instead of using 0 straightaway.

src/tools/atlas/read/listAlerts.ts
dudaschar: This will diverge a bit from the existing message, which defaults to the length. Do you think is it better to also remove the totalCount from the message, if not available?

src/tools/atlas/streams/build.ts
```diff
@@ -225,10 +231,23 @@ export class StreamsBuildTool extends StreamsToolBase {
                 return {
                     content: [{ type: "text", text: `Unknown resource type: ${args.resource as string}` }],
                     isError: true,
-                };
+                } as ToolResult<typeof this.outputSchema>;
```
dudaschar, line 234: Casting the output type to be able to not return a structuredContent if is an error.

src/tools/atlas/streams/build.ts
dudaschar: Added to the other stream tools

src/tools/atlas/read/listAlerts.ts
lerouxb: I think so, yes. I've just had so many bugs around # of results included in the page vs # of results matching the filter vs # of records in total. It is usually a footgun.

dudaschar: done!

src/tools/atlas/read/listAlerts.ts
```diff
@@ -67,9 +95,15 @@ export class ListAlertsTool extends AtlasToolBase {
 
         return {
             content: formatUntrustedData(
-                `Found ${alerts.length} alerts with status "${status}" in project ${projectId} (total: ${data.totalCount ?? alerts.length})`,
+                `Found ${alerts.length} alerts with status "${status}" in project ${projectId} ${data.totalCount && `(total: ${data.totalCount})`}`,
```
lerouxb: Just be careful here - you're checking if data.totalCount is truthy so you now leave it off if it is `0` whereas you probably want to include it whenever it is defined.

dudaschar: that's true, forgot 0 is falsy  😂

dudaschar: updated

lerouxb (approved):

PR #1293, opened 2026-06-29
chore: Adds release & publishing support for mongodb-atlas-mcp-remote MCP-540

maastha:
## Proposed changes
Adds independent release and publishing automation for mongodb-atlas-mcp-remote ([MCP-540](https://jira.mongodb.org/browse/MCP-540)), decoupling it from the mongodb-mcp-server release cadence.

**Changes**

- New prepare-atlas-mcp-remote-release.yml workflow to bump the wrapper version and open a release PR independently from the server
- Extended publish.yml with wrapper check and publish jobs; CI automatically skips publishing until the bootstrap is complete (see below)
- Isolated server workflows so the wrapper is excluded from server version bumps and publishes
- New setup-node-pnpm and wait-for-npm composite actions to reduce duplication
- Removed "private": true, decoupled wrapper version to 0.0.0, added scripts/updatePackageVersion.ts to keep src/packageInfo.ts in sync on release

**First publish ([MCP-493](https://jira.mongodb.org/browse/MCP-493))**
The first publish needs to be done manually. Once that's done, all subsequent releases will go through CI automatically. 

Prerequisites:
Publish the initial version to npm manually
Configure OIDC trusted publishing for mongodb-atlas-mcp-remote on npmjs.com
From then on, run prepare-atlas-mcp-remote-release → merge the PR → CI publishes

## Checklist


.github/workflows/publish.yml
```diff
@@ -206,3 +183,111 @@ jobs:
...
+
+  publish-atlas-mcp-remote:
+    runs-on: ubuntu-latest
+    environment: Production
+    permissions:
+      contents: write
+      id-token: write
+    needs:
+      - check-atlas-mcp-remote
+    if: needs.check-atlas-mcp-remote.outputs.VERSION_EXISTS == 'false' && needs.check-atlas-mcp-remote.outputs.PACKAGE_EXISTS_ON_NPM == 'true'
```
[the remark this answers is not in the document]
maastha: this is intentional. Added a comment in the file and captured context in this PR's description about this (first release will be manul)

.github/workflows/publish.yml
```diff
@@ -206,3 +183,113 @@ jobs:
...
+      - name: Build
+        run: |
+          pnpm install --frozen-lockfile
+          pnpm --filter 'mongodb-atlas-mcp-remote' run compile
+      - name: Publish to NPM
+        run: pnpm --filter 'mongodb-atlas-mcp-remote' publish --tag ${{ needs.check-atlas-mcp-remote.outputs.RELEASE_CHANNEL }} --no-git-checks
+
+      - name: Publish GitHub release
+        env:
+          GH_TOKEN: ${{ github.token }}
```
EspenAlbert: You can do something similar to the modules if you want to support both a secret or the triggering token as a fallback: https://github.com/terraform-mongodbatlas-modules/terraform-mongodbatlas-atlas-aws/blob/main/.github/workflows/release.yml#L40-L43

```yaml
env:
  # Use APIX_BOT_PAT if available (bypasses branch rulesets), otherwise fall back to GITHUB_TOKEN
  GH_TOKEN: ${{ secrets.APIX_BOT_PAT || github.token }}
```

.github/actions/wait-for-npm/action.yml
```diff
@@ -0,0 +1,35 @@
...
+    description: npm package version to wait for (e.g. 1.2.3)
+    required: true
+
+runs:
+  using: composite
+  steps:
+    - name: Wait for package to be available on npm
+      shell: bash
+      run: |
+        PACKAGE_NAME="${{ inputs.package-name }}"
```
EspenAlbert, line 18: [nit] Why not use a shell script? (Seems a bit much to keep in a GH action)

.github/actions/setup-node-pnpm/action.yml
```diff
@@ -0,0 +1,22 @@
+name: Set up Node.js and pnpm
+description: Installs pnpm, sets up Node.js with npm registry and pnpm cache, and checks out the repository.
+
+inputs:
+  node-version:
```
gagik: I am not sure I see any point in having node-version as an input, let's just have it hardcoded to match all other places with the same node version and ideally use the package.json

.github/workflows/prepare-release.yml
```diff
@@ -40,7 +40,7 @@ jobs:
           set -euo pipefail
           npm version "$INPUTS_VERSION" --no-git-tag-version --allow-same-version
           NEW_VERSION=$(npm pkg get version | tr -d '"')
-          pnpm -r --filter './packages/*' exec npm version "$NEW_VERSION" --no-git-tag-version --allow-same-version
```
gagik, line 43: I don't think this is an issue as the version will be the same

.github/workflows/prepare-release.yml
```diff
@@ -57,7 +57,7 @@ jobs:
         run: |
           set -euo pipefail
 
-          LATEST_TAG=$(gh release list --exclude-drafts --exclude-pre-releases --limit 1 --json tagName --jq '.[0].tagName // empty')
+          LATEST_TAG=$(gh release list --exclude-drafts --exclude-pre-releases --limit 100 --json tagName --jq '[.[] | select(.tagName | test("^v[0-9]"))] | .[0].tagName // empty')
```
gagik, line 62: not sure what the change here is, is it so we can publish many at once?

.github/workflows/publish.yml
```diff
@@ -99,7 +91,7 @@ jobs:
       - name: Publish to NPM
         run: |
           pnpm publish --tag ${{ needs.check.outputs.RELEASE_CHANNEL }} --no-git-checks
-          pnpm -r --filter './packages/*' publish --tag ${{ needs.check.outputs.RELEASE_CHANNEL }} --no-git-checks
+          pnpm -r --filter './packages/*' --filter '!mongodb-atlas-mcp-remote' publish --tag ${{ needs.check.outputs.RELEASE_CHANNEL }} --no-git-checks
```
gagik: this shouldn't be a problem; running publish will be a no-op unless the version has been bumped (which is what we'd expect)

.github/workflows/publish.yml
```diff
@@ -206,3 +183,113 @@ jobs:
             -d "{\"name\": \"vNext\", \"projectId\": ${project_id}, \"startDate\": \"${TODAY}\"}" \
             "${JIRA_URL}/rest/api/2/version" > /dev/null
           echo "Created new vNext version with start date ${TODAY}"
+
+  check-atlas-mcp-remote:
```
gagik: there should be no need for this, let's just let the regular check handle both.

gagik (changes requested):
I believe we don't need any separate logic for publish.yml; we just need to have a way to bump only the remote package as part of prepare-release.

One way is we add an input which lets us choose between "mongodb-mcp-server" and `"mongodb-atlas-mcp-remote"` in `prepare-release`. The former will make it work as usual; the latter will only bump `mongodb-atlas-mcp-remote`. 

`publish.yml` should always just publish the bumped packages and have any special handling that differentiates the 2.

.github/workflows/publish.yml
```diff
@@ -78,17 +78,9 @@ jobs:
...
-        with:
-          persist-credentials: false
-          fetch-depth: 0
-      - uses: pnpm/action-setup@v5
-      - uses: actions/setup-node@v6
+      - uses: ./.github/actions/setup-node-pnpm
         with:
           node-version: "24"
-          registry-url: "https://registry.npmjs.org"
-          cache: "pnpm"
```
gagik, line 91: cache: "pnpm" is important here

packages/mongodb-atlas-mcp-remote/package.json
```diff
@@ -33,6 +33,7 @@
   },
   "devDependencies": {
     "@types/node": "^25.6.0",
+    "tsx": "^4.20.5",
```
gagik: no need for `tsx`. node should run ts files with Node ;latest 24+

maastha:
> I believe we don't need any separate logic for publish.yml; we just need to have a way to bump only the remote package as part of prepare-release.
> 
> One way is we add an input which lets us choose between "mongodb-[mcp-server](http://jira.mongodb.org/browse/MCP-server)" and "mongodb-atlas-mcp-remote" in prepare-release. The former will make it work as usual; the latter will only bump mongodb-atlas-mcp-remote.
> 
> publish.yml should always just publish the bumped packages and have any special handling that differentiates the 2.
> 


@gagik  Thanks for the suggestion! I considered merging them but the workflows have enough concrete differences to justify keeping them separate:

The server release runs generate:arguments (regenerates tool definitions in server.json) — not applicable to the remote package --
Version regression uses different sources: GitHub releases for the server, npm registry for the remote package
The server release has Jira vNext integration; the remote package doesn't track releases there
They version independently (server is 1.x, remote starts at 0.0.0) — a single workflow with a package selector risks triggering the wrong release

Happy to extract the shared boilerplate (checkout, pnpm, bot token) into a composite action if you think that's worth it, but I'd prefer to keep the prepare-release workflows separate. I'll address your other feedback on publish.yml.

.github/actions/setup-node-pnpm/action.yml
maastha: Done.

.github/workflows/publish.yml
maastha: Merged into the `check` job — remote package checks now run as additional steps there with `REMOTE_*` prefixed outputs. `publish-atlas-mcp-remote` now depends on `check`.

.github/workflows/publish.yml
maastha: Kept intentionally — removing it would cause the main `publish` job to attempt publishing `mongodb-atlas-mcp-remote@0.0.0` on every server release, since that version hasn't been published to npm yet and would be accepted.

packages/mongodb-atlas-mcp-remote/package.json
maastha: Removed. Updated the script to use `node --experimental-strip-types`, consistent with how the root project runs other `.ts` scripts.

.github/workflows/publish.yml
maastha: Done.

.github/actions/wait-for-npm/action.yml
maastha: Kept as-is — it's called from two separate jobs, so the composite form avoids duplication.

.github/workflows/prepare-release.yml
maastha: The remote package versions independently (currently `0.0.0`), not in sync with the server. Without the filter, this step would overwrite it with the server version.

.github/workflows/prepare-release.yml
maastha: Now that both server (`v1.x.x`) and remote package (`atlas-mcp-remote-v*`) releases live in the same repo, `--limit 1` could return a remote tag as the latest release. The regex scopes the regression check to server tags only.

.github/workflows/publish.yml
maastha: Already included in the action itself.

.github/workflows/publish.yml
gagik: we can set "private: true" in package.json to prevent release publishing

.github/actions/setup-node-pnpm/action.yml
gagik: actually thinking about it more, could we have this set to v24 instead of package.json? Otherwise Node version will change a lot more often since we have a rather open ended change.

gagik:
SGTM for keeping `prepare-release` separate but wondering if we can normalize the `publish.yaml` further to avoid separate logic

> Version regression uses different sources: GitHub releases for the server, npm registry for the remote package

We seem to be creating GitHub releases for the remote package too, why not use the GitHub releases as a reference for both?

packages/mongodb-atlas-mcp-remote/package.json
gagik: sg, there's actually no need for `--experimental-strip-types` (seems like we use it elsewhere too but it's unnecessary)

.github/workflows/prepare-release.yml
gagik: ah got it.
could we do
```
pnpm -r --filter "@mongodb-js/mcp-*"
```
Instead?

packages/mongodb-atlas-mcp-remote/package.json
```diff
@@ -23,7 +22,8 @@
   },
   "scripts": {
     "compile": "tsc --project tsconfig.json",
-    "test": "vitest --run"
+    "test": "vitest --run",
+    "build:update-package-version": "node --experimental-strip-types scripts/updatePackageVersion.ts"
```
gagik: can we run `../scripts/updatePackageVersion.ts` and avoid duplication? Might require minor tweaking to make it more general

packages/mongodb-atlas-mcp-remote/package.json
maastha: alright, removed.

.github/actions/setup-node-pnpm/action.yml
maastha: set to v24

packages/mongodb-atlas-mcp-remote/package.json
maastha: I'd keep these separate. The root script does a fair bit that's server-specific — it emits mcpServerName/engines into src/common/packageInfo.ts and also syncs packaging/mcpb/manifest.json and the Azure Bicep template — whereas the wrapper only needs { version } in src/packageInfo.ts. 
Generalizing one script to cover both output shapes and conditionally skip the server-only writes adds more branching than the ~15 lines of duplication removes. Happy to revisit if the wrapper's version file grows to need more fields.

.github/workflows/prepare-atlas-mcp-remote-release.yml
```diff
@@ -0,0 +1,85 @@
...
+
+      - name: Check version regression
+        id: version-check
+        continue-on-error: true
+        run: |
+          set -euo pipefail
+          NEW="${{ steps.bump-version.outputs.NEW_VERSION }}"
+          NEW="${NEW#v}"
+
+          PUBLISHED=$(npm view mongodb-atlas-mcp-remote version 2>/dev/null || true)
```
maastha: > We seem to be creating GitHub releases for the remote package too, why not use the GitHub releases as a reference for both?



@gagik  Keeping npm (npm view) as the regression source for the wrapper on purpose. By design CI doesn't own the wrapper's first release — the first publish is manual (MCP-493), and the publish-atlas-mcp-remote job (which is what runs gh release create `mongodbatlas-mcp-remote/v*`) is gated off until the package exists on npm. So right after the manual bootstrap the package is on npm but may have no `mongodbatlas-mcp-remote/v*` GitHub release yet, which means a gh release list check could silently skip the regression warning. 
npm view is guaranteed to reflect what's actually published. The server can rely on GitHub releases since it's always released through CI; I scoped its check to ^v[0-9] tags so the two release lines don't interfere.

gagik: not sure I follow. All of what you describe applied to the `mongodb-mcp-server` as well so the difference is unclear to me.

For the first ever CI-based publish, we can add way to manually type the new version (which we have for regular prepare-release) and handle it this way, then this one-time issue with figuring out the version won't happen. 

And after that first CI publish, we can pretty much rely on GitHub releases going forward.

If there seems to be a problem with relying on GitHub releases, then we can switch both of the packages to rely on NPM but having divergence here seems odd as they _should_ work in similar way. That said, we haven't had issues with the GitHub publishing just yet.

packages/mongodb-atlas-mcp-remote/package.json
gagik: fair enough, sgtm!

.github/workflows/prepare-atlas-mcp-remote-release.yml
maastha: Alright, let's keep both consistent and use GitHub releases for the wrapper's check too. I've updated it to read gh release list scoped to `mongodb-atlas-mcp-remote/v*` tags (so it won't pick up the server's v* releases).

First-release process, so the one gap is explicit:

1. Manual first npm publish (MCP-493) — a person publishes the initial version to npm by hand.
2. Manually create the matching GitHub release (`mongodb-atlas-mcp-remote/v1.0.0`). The publish job that normally creates the release is gated until the package exists on npm, so nothing creates it for the bootstrap version. I'll document this step in the MCP-493 release steps.
3. From the next release on, it's fully automated — the check reads GitHub releases, and each release publishes to npm + creates its release.

Worth flagging that for this package the GitHub release is purely a bookkeeping marker — users install from npm, and nothing in CI triggers off the release (publishing runs on merge to main, not on release/tag events). So creating it by hand is safe; its only consumer is this regression check.

.github/workflows/prepare-release.yml
maastha: yes updated to `pnpm -r --filter '@mongodb-js/mcp-*'`

.github/workflows/publish.yml
maastha: private: true would block all publishing of the package, not just the 0.0.0 placeholder. The REMOTE_NPM_VERSION != '0.0.0' guard is narrower: it only skips the placeholder version while letting genuine releases through. Once the first version is published manually (MCP-493), real versions will publish normally.

maastha:
Manually tested the release/publish workflow changes in a fork ([maastha/mongodb-mcp-server](https://github.com/maastha/mongodb-mcp-server)):

New mongodb-atlas-mcp-remote release path:

- prepare-atlas-mcp-remote-release.yml correctly bumps only the wrapper's version: [PR #1](https://github.com/maastha/mongodb-mcp-server/pull/1)
- Merge → publish.yml builds, dry-run publishes, and creates a real release/tag: [run](https://github.com/maastha/mongodb-mcp-server/actions/runs/28679091664) · [release](https://github.com/maastha/mongodb-mcp-server/releases/tag/mongodb-atlas-mcp-remote/v0.0.1)


Existing main-server release path (regression):

- prepare-release.yml correctly excludes packages/mongodb-atlas-mcp-remote/package.json: [PR #3](https://github.com/maastha/mongodb-mcp-server/pull/3)
- Merge → publish.yml leaves the wrapper's version/tag/release outputs untouched: [run](https://github.com/maastha/mongodb-mcp-server/actions/runs/28680133469)

.github/workflows/publish.yml
```diff
@@ -67,6 +72,59 @@ jobs:
...
+      - name: Check if atlas-mcp-remote version already exists
+        id: check-remote-version
+        shell: bash
+        run: |
+          set +e
+
+          git rev-parse "${{ steps.get-remote-version.outputs.REMOTE_GIT_TAG_VERSION }}" >/dev/null 2>&1
+          if [[ $? -eq 0 ]]; then
+            echo "REMOTE_VERSION_EXISTS=true" >> "$GITHUB_OUTPUT"
+          else
```
gagik, line 93: why is this needed? running `publish` on an already published package should be a no-op as far as I'm aware so there's no issue simply running publish on all packages; whatever needs to be published will be published.

.github/workflows/publish.yml
```diff
@@ -99,7 +152,9 @@ jobs:
       - name: Publish to NPM
         run: |
           pnpm publish --tag ${{ needs.check.outputs.RELEASE_CHANNEL }} --no-git-checks
-          pnpm -r --filter './packages/*' publish --tag ${{ needs.check.outputs.RELEASE_CHANNEL }} --no-git-checks
+          # Only publish the scoped @mongodb-js/mcp-* packages alongside the server.
+          # mongodb-atlas-mcp-remote is intentionally unscoped so it's excluded here and released via publish-atlas-mcp-remote.
+          pnpm -r --filter '@mongodb-js/mcp-*' publish --tag ${{ needs.check.outputs.RELEASE_CHANNEL }} --no-git-checks
```
gagik: unlike bumping, it's fine to run publishing on all tasks so there's no need for special logic here.

.github/workflows/publish.yml
```diff
@@ -206,3 +246,48 @@ jobs:
             -d "{\"name\": \"vNext\", \"projectId\": ${project_id}, \"startDate\": \"${TODAY}\"}" \
             "${JIRA_URL}/rest/api/2/version" > /dev/null
           echo "Created new vNext version with start date ${TODAY}"
+
+  publish-atlas-mcp-remote:
```
gagik, line 236: with the changes above, this should become redundant and we should be able to use the generic publish task for everything

packages/mongodb-atlas-mcp-remote/package.json
```diff
@@ -1,8 +1,7 @@
 {
   "name": "mongodb-atlas-mcp-remote",
   "description": "MCP stdio proxy that forwards requests to the MongoDB Remote MCP server with OAuth2 client credentials authentication",
-  "version": "1.14.0-prerelease.2",
-  "private": true,
```
gagik, line 5: I think we'd still want `private: true` here to exclude it from publishing

.github/workflows/publish.yml
maastha: Done — collapsed the two publish lines into a single generic pnpm -r publish --tag <channel> --no-git-checks. The wrapper is now excluded automatically because it's private (which will be updated after the first manual release)

.github/workflows/publish.yml
maastha: npm publishing is generic now (see above) and removed the REMOTE_PACKAGE_EXISTS_ON_NPM. 

The `publish-atlas-mcp-remote` job stays for the two reasons we discussed in Slack: 
(1) the wrapper needs its own mongodb-atlas-mcp-remote/v* GitHub release/tag — different namespace and notes from the server's v*; and 
(2) a wrapper-only release doesn't trigger the server publish job (it's gated on the server's VERSION_EXISTS), so the wrapper is published from here under its own release channel.

packages/mongodb-atlas-mcp-remote/package.json
maastha: Done — added "private": true.  As discussed, it will be updated at the first manual release (MCP-493), after which CI publishes it like any other package.

.github/workflows/publish.yml
maastha: made it generic (pnpm -r publish), so there's no per-package publish logic anymore and already-published versions are skipped.

I've kept this REMOTE_VERSION_EXISTS check, but note it no longer gates publishing — it only gates the GitHub release creation in the publish-atlas-mcp-remote job. gh release create fails on a tag/release that already exists, so we still need to know whether to create it on a given run.

.github/workflows/publish.yml
```diff
@@ -206,3 +232,48 @@ jobs:
...
+    runs-on: ubuntu-latest
+    environment: Production
+    permissions:
+      contents: write
+      id-token: write
+    needs:
+      - check
+    # The `!= '0.0.0'` guard skips creating github release forthe placeholder version. `private: true` already makes the npm
```
gagik, line 244: nit: seems redundant? `private: false` is already blocking releases and we can always change both version and remove the private: true in one go

.github/workflows/publish.yml
```diff
@@ -14,9 +14,13 @@ jobs:
       NPM_VERSION: ${{ steps.get-version.outputs.NPM_VERSION }}
       GIT_TAG_VERSION: ${{ steps.get-version.outputs.GIT_TAG_VERSION }}
       RELEASE_CHANNEL: ${{ steps.npm-tag.outputs.RELEASE_CHANNEL }}
+      REMOTE_VERSION_EXISTS: ${{ steps.check-remote-version.outputs.REMOTE_VERSION_EXISTS }}
```
gagik: nit: can we call this `ATLAS_MCP_REMOTE_....` otherwise it's unclear these are 2 different packages.

gagik (approved):
Thanks for all the effort and iterations here, LGTM. Some last small nits for consideration.

.github/workflows/publish.yml
maastha: Its not redundant, private: true and this guard block different things:

private: true → only makes the npm publish step a no-op (pnpm skips private packages).
It does not stop this job from running — and the job also does gh release create + wait-for-npm, which run regardless of npm's private flag.

.github/workflows/publish.yml
maastha: updated

PR #1295, opened 2026-06-30
chore: mongodb-atlas-mcp-remote integration tests MCP-539

manupedrozo:
## Description

Integration tests for the `mongodb-atlas-mcp-remote` package.

Ticket: MCP-539

## Checklist


packages/mongodb-atlas-mcp-remote/src/integration.test.ts
```diff
@@ -0,0 +1,108 @@
...
+    it("fails fast when the token endpoint rejects at startup", async () => {
+        remote.failTokenRequests();
+        let stderrOutput = "";
+        await expect(
+            createTestClient(remote, (t) => t.stderr?.on("data", (s) => (stderrOutput += s)))
+        ).rejects.toThrow();
+        expect(stderrOutput).toContain("Failed to acquire access token");
+    });
+
+    it("refreshes the token and retries on 401 from tool calls", async () => {
```
maastha, line 102: qq — this assumes the SDK automatically refetches the token and replays the same failed call after a 401. Is that a guaranteed behavior of the transport, or are the current SDK behavior? Just considering if this is something that could quietly change when we bump the SDK

packages/mongodb-atlas-mcp-remote/src/integration.test.ts
```diff
@@ -0,0 +1,108 @@
...
+        client = await createTestClient(remote);
+        remote.failNextMCPCallWith401();
+        const result = await callTool("mock-project-tool", { projectId: "proj-1" });
+        expect(result.isError).toBeFalsy();
+        expect(remote.getTokenRequestCount()).toBe(2);
```
maastha, line 107: IIUC the mock hands back the same token value both times, so this really only proves the token endpoint got hit twice, not that the new token actually got used on the retry. If you vary the token (mock-token-1, mock-token-2) you could assert the second MCP call carried the fresh one. WDYT?

packages/mongodb-atlas-mcp-remote/src/testHelpers/mockRemote.ts
```diff
@@ -0,0 +1,199 @@
...
+
+    setResponseMode(mode: "sse" | "json"): void {
+        this.responseMode = mode;
+    }
+
+    failNextMCPCallWith401(): void {
+        this.shouldFailNextMCPCall = true;
+    }
+
+    failTokenRequests(): void {
```
maastha: should this be renamed to failNextTokenRequest() to match what it does?

packages/mongodb-atlas-mcp-remote/src/integration.test.ts
```diff
@@ -0,0 +1,108 @@
...
+
+    beforeAll(async () => {
+        remote = await MockRemote.start();
+    });
+
+    afterAll(async () => {
+        await remote.close();
+    });
+
+    afterEach(async () => {
```
maastha, line 49: These tests share one mock with mutable counters and lean on afterEach reset + sequential runs. Might be worth adding a comment something like "// keep sequential — shared counter"  or something so nobody turns these into it.concurrent later and gets confused by flaky counts.

packages/mongodb-atlas-mcp-remote/src/integration.test.ts
```diff
@@ -0,0 +1,108 @@
...
+        expect(stderrOutput).toContain("Failed to acquire access token");
+    });
+
+    it("refreshes the token and retries on 401 from tool calls", async () => {
+        client = await createTestClient(remote);
+        remote.failNextMCPCallWith401();
+        const result = await callTool("mock-project-tool", { projectId: "proj-1" });
+        expect(result.isError).toBeFalsy();
+        expect(remote.getTokenRequestCount()).toBe(2);
+    });
```
maastha, line 108: Also capturing here the issue discussed that was hit during initial manual testing, for a follow-up (not blocking this PR): the wrapper doesn't appear to recover when the remote MCP session expires.


when the remote session is gone, the wrapper doesn't recover. Repro'd with a local mock (establish session → force 404 next call): it never re-initializes (initialize count stays 1) and the call fails. Added logs that show it's not token-related:


```
[DEBUG] tokenManager: Reusing cached access token (expiresIn=3599s)
[ERROR] cli: Failed to forward message to remote MCP server (method=tools/call, code=CLIENT_HTTP_NOT_IMPLEMENTED, status=404)
```

Why: in StreamableHTTPClientTransport._send, a 401 triggers refresh+retry but a 404 just throws — no re-init.

I think we can detect the session-expired 404 and transparently re-initialize + retry, mirroring the token path.

Test: add expireSessionOnNextCall() (404 once) + a second initialize handler + getInitializeCount(), assert retry succeeds and count === 2. Fits alongside the 401 test.

maastha: added logging here https://github.com/mongodb-js/mongodb-mcp-server/pull/1297 to help with this

packages/mongodb-atlas-mcp-remote/src/testHelpers/mockRemote.ts
manupedrozo: Yes! renaming

packages/mongodb-atlas-mcp-remote/src/integration.test.ts
manupedrozo: Added a comment (should be pretty obvious from the shared remote & client variables).

packages/mongodb-atlas-mcp-remote/src/integration.test.ts
manupedrozo: Its the documented SDK behavior for token auth, which is our desired behavior. This test would catch any regression (so not quiet changes).

packages/mongodb-atlas-mcp-remote/src/integration.test.ts
manupedrozo: Thanks! I'll look into reproducing this and cover it in another PR.

packages/mongodb-atlas-mcp-remote/src/integration.test.ts
manupedrozo: Makes sense, changing the mock server to generate sequential tokens 👍

gagik (approved):

PR #1307, opened 2026-07-02
feat: enhance validation when creating new db user - MCP-577

dudaschar:
Adding schema validation for Atlas user roles, enhancing validation for custom roles and escaping values when rendering markdown to request user confirmation.

Also escaping values inside markdown confirmation for delete mongodb tools.
## Checklist

### DIFF
--- a/src/helpers/escapeMarkdown.ts
+++ b/src/helpers/escapeMarkdown.ts
@@ -1,0 +1,6 @@
+/**
+ * Escapes markdown and HTML metacharacters in a string.
+ */
+export function escapeMarkdown(value: string): string {
+    return value.replace(/[\\`*_{}[\]()<>#+\-.!|]/g, "\\$&");
+}
--- a/src/tools/args.ts
+++ b/src/tools/args.ts
@@ -8,10 +8,13 @@ const ALLOWED_USERNAME_CHARACTERS_REGEX = /^[a-zA-Z0-9._-]+$/;
 export const ALLOWED_USERNAME_CHARACTERS_ERROR =
     "Username can only contain letters, numbers, dots, hyphens, and underscores";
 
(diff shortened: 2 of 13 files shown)


src/tools/args.ts
```diff
@@ -8,10 +8,13 @@ const ALLOWED_USERNAME_CHARACTERS_REGEX = /^[a-zA-Z0-9._-]+$/;
 export const ALLOWED_USERNAME_CHARACTERS_ERROR =
     "Username can only contain letters, numbers, dots, hyphens, and underscores";
 
-const ALLOWED_REGION_CHARACTERS_REGEX = /^[a-zA-Z0-9_-]+$/;
+// Generic pattern for identifiers that permit only ASCII letters, numbers, hyphens, and underscores.
+export const ALPHANUMERIC_DASH_UNDERSCORE_REGEX = /^[a-zA-Z0-9_-]+$/;
+
+const ALLOWED_REGION_CHARACTERS_REGEX = ALPHANUMERIC_DASH_UNDERSCORE_REGEX;
 export const ALLOWED_REGION_CHARACTERS_ERROR = "Region can only contain letters, numbers, hyphens, and underscores";
```
dudaschar, line 15: Both region and cluster uses the same regex value, but they could be different. So I moved the regex to a generic ALPHANUMERIC_DASH_UNDERSCORE_REGEX const and assign to both of them. I'm also using this regex pattern to the custom role validation.

src/helpers/escapeMarkdown.ts
```diff
@@ -0,0 +1,6 @@
+/**
+ * Escapes markdown and HTML metacharacters in a string.
+ */
+export function escapeMarkdown(value: string): string {
+    return value.replace(/[\\`*_{}[\]()<>#+\-.!|]/g, "\\$&");
```
lerouxb, line 5: Feels a bit extreme, but I suppose it should do the job.

lerouxb (approved):

PR #1310, opened 2026-07-02
chore: mongodb-atlas-mcp-remote e2e tests MCP-539

manupedrozo:
## Description

Adding sanity e2e tests for the mongodb-atlas-mcp-remote package

## Checklist

### DIFF
--- a/packages/mongodb-atlas-mcp-remote/src/e2e.test.ts
+++ b/packages/mongodb-atlas-mcp-remote/src/e2e.test.ts
@@ -1,0 +1,46 @@
+import { describe, it, expect, beforeAll, afterAll } from "vitest";
+import path from "node:path";
+import { fileURLToPath } from "node:url";
+import { Client, StdioClientTransport } from "@modelcontextprotocol/client";
+
+/**
+ * E2E tests are skipped unless Service Account credentials are set.
+ */
+const shouldSkip = !process.env.REMOTE_MCP_E2E_CLIENT_ID || !process.env.REMOTE_MCP_E2E_CLIENT_SECRET;
+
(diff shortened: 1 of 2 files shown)


packages/mongodb-atlas-mcp-remote/src/e2e.test.ts
```diff
@@ -0,0 +1,46 @@
+import { describe, it, expect, beforeAll, afterAll } from "vitest";
+import path from "node:path";
+import { fileURLToPath } from "node:url";
+import { Client, StdioClientTransport } from "@modelcontextprotocol/client";
+
+/**
+ * E2E tests are skipped unless Service Account credentials are set.
+ */
+const shouldSkip = !process.env.REMOTE_MCP_E2E_CLIENT_ID || !process.env.REMOTE_MCP_E2E_CLIENT_SECRET;
```
maastha, line 9: should this also check for the base url? otherwise this would run against prod no?

gagik: sure, we could be strict about it; it's also fine if they fail because of client ID / secret

gagik (approved):

packages/mongodb-atlas-mcp-remote/src/e2e.test.ts
manupedrozo: Yes, it fails if the client id/secret is not valid for the environment (similar to other tools / tests).

manupedrozo:
Failures on "Run Atlas Local tests" unrelated https://github.com/mongodb-js/mongodb-mcp-server/actions/runs/28792166253/job/85377704584?pr=1310

PR #1313, opened 2026-07-03
chore: rename organizationId arg to orgId for consistency with other tools CLOUDP-419289

gribnoysup:
## Proposed changes

Aligns the arg name with other places that have org id. This is definitely a short-term quick fix, I think for the future we should figure out a way to keep these consistent between tools in a way that is enforced and doesn't rely on us just sort of keeping this in mind all the time

## Checklist


gribnoysup:
I'm not seeing the same CI failure locally when running `pnpm run build:esm`, am I missing some undocumented step I need to do to make CI pass?

gagik (approved):
need to run `pnpm run update:api` but LGTM

blva (approved):

PR #1315, opened 2026-07-03
chore: add session as an argument to addSession CLOUDP-418929

nirinchev:
## Proposed changes

This is required for consumers who need to access session-level fields (such as connection manager) on adding. Additionally, pass headers to getSession so that consumers are able to validate user identity before session is returned to the client.

### DIFF
--- a/packages/types/src/sessionStore.ts
+++ b/packages/types/src/sessionStore.ts
@@ -1,5 +1,6 @@
 import type { ILoggerBase } from "./logging.js";
 import type { IMetrics, MetricDefinitions } from "./metrics.js";
+import type { ISession } from "./session.js";
 
 export type CloseableTransport = {
     close(): Promise<void>;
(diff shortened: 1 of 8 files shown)


packages/types/src/sessionStore.ts
```diff
@@ -8,11 +9,34 @@ export type CloseableTransport = {
...
+     * session does not exist.
+     *
+     * @param headers The headers of the incoming request. Implementations can
+     * use them to validate the caller's identity before returning the session.
+     * To reject a request for an existing session, throw a
+     * `SessionRejectedError` rather than returning `undefined` — the latter is
+     * treated as "session not found" and may trigger implicit session
+     * initialization.
+     */
+    getSession(sessionId: string, headers?: Record<string, unknown>): Promise<T | undefined>;
```
gagik, line 23:
```suggestion
    getSession(sessionId: string, options: { headers?: Record<string, unknown>} = {}): Promise<T | undefined>;
```
Just to future proof against the world where there's more things we might need here?

gagik: we can also turn the whole param into an object param

gagik (approved):
looks good, one suggestion for the function shape to make it more flexible

lerouxb (approved):

packages/types/src/sessionStore.ts
nirinchev: I was thinking of having it as object param, but didn't want to break consumers. Inclined to leave it as a second param and then refactor as part of v2 - wdyt?

lerouxb: I think you can have both? You're adding a second param anyway and you can make options default to {} and then consumers should still be able to use just getSession(sessionId).

nirinchev: Right, what I meant was that eventually, I'd like this to be `getSession(params: { sessionId, headers })` similarly to what we have for addSession. If we're moving in that direction (hopefully soon), then it doesn't matter too much whether the intermediate state is `options: { header }` or just `headers`.

lerouxb: Fine by me.

gagik: v2 sgtm

PR #1316, opened 2026-07-03
fix: update output messages to use "requested collection" and "requested database" - MCP-581

dudaschar:
Avoid interpolating collection and database names into content output.

## Checklist


nirinchev (approved):
Changes look good to me, though the exploit sounds too far-fetched - since the collection name is injectable as an argument, the LLM would have already seen it and passed it to us verbatim. By this point, I'm not sure I see how us returning the name would poison the context anymore than it already is.

PR #1335, opened 2026-07-07
fix: confirmation gate fails closed without elicitation MCP-584

jwongmongodb:
MCP-584

## Proposed changes

`Elicitation.requestConfirmation()` previously returned `true` when the connecting client did not support MCP elicitation, so every confirmation-required tool (`drop-database`, `drop-collection`, `delete-many`, `drop-index`, `atlas-create-db-user`, `atlas-create-access-list`, `atlas-streams-manage`, `atlas-streams-teardown`) was silently auto-approved on non-elicitation clients, bypassing the only safeguard on destructive operations.

Fail closed instead: with no way to obtain human confirmation, the tool refuses to run and returns a clear error telling the user to use an elicitation-capable client. This affects all confirmation-required tools across the server, not just streams, and brings `requestConfirmation()` in line with the existing fail-closed behavior of `requestInput()`.

## Checklist


jwongmongodb:
won't do

PR #1337, opened 2026-07-08
feat: implement redaction of sensitive information in config resource output - MCP-589

dudaschar:
## Proposed changes

Applying redaction for connectOptions and redacting any secret still present in the output

## Checklist

### DIFF
--- a/src/resources/common/config.ts
+++ b/src/resources/common/config.ts
@@ -3,8 +3,7 @@ import type { UserConfig } from "../../common/config/userConfig.js";
 import type { Telemetry } from "../../telemetry/telemetry.js";
 import type { Session } from "../../lib.js";
 import { generateConnectionInfoFromCliArgs } from "@mongosh/arg-parser";
-import { redact } from "mongodb-redact";
-import { Keychain } from "../../common/keychain.js";
+import { Keychain, redactValues } from "../../common/keychain.js";
 
 /**
  * Removes secret material from the driver options before exposing them via the config resource.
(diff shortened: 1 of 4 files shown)


src/resources/common/config.ts
```diff
@@ -3,6 +3,21 @@ import type { UserConfig } from "../../common/config/userConfig.js";
...
+import { Keychain, redactValues } from "../../common/keychain.js";
+
+/**
+ * Removes secret material from the driver options before exposing them via the config resource.
+ * The mongosh arg-mapper places KMS credentials under `autoEncryption.kmsProviders` (e.g. the AWS
+ * access key / secret / session token), so the whole `autoEncryption` block is replaced with a
+ * non-sensitive summary rather than emitted verbatim.
+ */
+function redactDriverOptions(driverOptions: Record<string, unknown>): Record<string, unknown> {
+    const { autoEncryption, ...rest } = driverOptions;
```
nirinchev, line 14: Is there anything else in driver options that could be sensitive?

src/resources/common/config.ts
```diff
@@ -39,13 +54,16 @@ export class ConfigResource extends ReactiveResource<UserConfig, readonly []> {
...
             atlas:
                 this.current.apiClientId && this.current.apiClientSecret
                     ? "set; MongoDB Atlas tools are currently available to use"
                     : "not set; MongoDB Atlas tools are currently unavailable, to have access to MongoDB Atlas tools like creating clusters or connecting to clusters make sure to setup credentials, more info at 'https://github.com/mongodb-js/mongodb-mcp-server'.",
         };
 
-        return JSON.stringify(result);
+        // Backstop: redact any remaining registered secrets (keychain) before egress, matching
+        // the redaction applied on every logging path. Redact per-value so JSON stays valid.
+        const secrets = [...this.session.keychain.allSecrets, ...Keychain.root.allSecrets];
```
nirinchev, line 65: I don't think we need to pull the root keychain secrets - the session ones should be sufficient.

nirinchev (approved):

src/resources/common/config.ts
dudaschar: That's a comment in the keychain file explaining the secrets should be registered in the root or session, I would keep both to make sure we are removing all secrets indeed.

src/resources/common/config.ts
dudaschar: Yes, I updated the comment to be more generic, the initial report was regarding the KMS credentials, but we are redacting other fields as well.

PR #1343, opened 2026-07-10
feat: allow consumers to disable automatic IP access list creation CLOUDP-422265

nirinchev:
## Proposed changes

When hosting the server remotely, the functionality to add IP Access lists doesn't make sense. This change allows consumers to disable it through the API client and adjusts corresponding tools to handle the change.

### DIFF
--- a/src/tools/atlas/create/createAccessList.ts
+++ b/src/tools/atlas/create/createAccessList.ts
@@ -23,9 +23,19 @@ export class CreateAccessListTool extends AtlasToolBase {
     static toolName = "atlas-create-access-list";
     public description = "Allow Ip/CIDR ranges to access your MongoDB Atlas clusters.";
     static operationType: OperationType = "create";
-    public argsShape = {
-        ...CreateAccessListArgs,
-    };
+    // The currentIpAddress arg is omitted on deployments that can't determine the
+    // caller's public IP (e.g. the Atlas-hosted MCP server), so models are never
+    // offered an option that cannot work there. Typed as the full shape because
+    // execute() still receives currentIpAddress as optional either way.
+    public get argsShape(): typeof CreateAccessListArgs {
+        if (this.session.apiClient?.supportsCurrentIpLookup === false) {
+            // eslint-disable-next-line @typescript-eslint/no-unused-vars
+            const { currentIpAddress, ...rest } = CreateAccessListArgs;
+            return rest as typeof CreateAccessListArgs;
+        }
+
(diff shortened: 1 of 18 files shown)


src/tools/atlas/create/createAccessList.ts
```diff
@@ -23,17 +23,31 @@ export class CreateAccessListTool extends AtlasToolBase {
...
     public override outputSchema = CreateAccessListOutputSchema;
 
     protected async execute(
         { projectId, ipAddresses, cidrBlocks, comment, currentIpAddress }: ToolArgs<typeof this.argsShape>,
         context: ToolExecutionContext
     ): Promise<ToolResult<typeof this.outputSchema>> {
         if (!ipAddresses?.length && !cidrBlocks?.length && !currentIpAddress) {
-            throw new Error("One of  ipAddresses, cidrBlocks, currentIpAddress must be provided.");
+            if (!this.apiClient.supportsCurrentIpLookup) {
+                throw new Error("Either ipAddresses or cidrBlocks must be provided.");
```
blva, line 47: do we expect remote MCP to provide the IP upon tool call? I thought we were going to skip it, so we should not throw an error if ip lookup is not supported?

blva: Update: I see your proposal is to request for the IP address from the client. I think that should work. Optionally, we could actually skip it and encourage the user to run the tool to add access list.

nirinchev: With remote, we don't have a reliable way to get the user's IP address, so we don't attempt to infer it. Instead the input args only allow cidr blocks or IP addresses and it's the client's responsibility to supply those.

blva: yes, my point is that it should not be required unless it is a required field in the API call - I didn’t double-check that, but if the field is optional I would go with still creating the cluster and adding a note to call the ip access list tool

nirinchev: This is the createAccessList tool though - for the cluster tool, we don't require it. But here we can't create an access list if we don't have either the IP address or the CIDR blocks

nirinchev:
The test failures are legit - I have forgotten to update those tests. Will fix them later today.

src/tools/atlas/create/createAccessList.ts
blva: oh ok! woosh. missed that, makes sense!

blva (approved):

PR #1351, opened 2026-07-12
feat: multi-connection support (named connections + per-call selection)

ahmednfwela:
## Summary

Adds **multi-connection support** to the MongoDB MCP server so a single server/session can address more than one MongoDB target:

- **Named connections** configured via a new `MDB_MCP_CONNECTIONS` config value — a JSON object mapping a connection name to a target, e.g. `{"analytics":"mongodb://…","reporting":{"connectionString":"mongodb://…"}}` (both the string shorthand and the `{ connectionString }` object form are accepted).
- **Optional per-call `connection` argument** on data tools (via `DBOperationArgs`) that selects which named connection an individual tool call runs against. When omitted, the active/session-default connection is used exactly as before. The per-call name is threaded through `AsyncLocalStorage`, so concurrent/pipelined calls on the shared per-session tool instances never cross state.
- A **`list-connections`** tool to enumerate configured connections.
- **Backward compatible:** the existing single `MDB_MCP_CONNECTION_STRING` continues to work with no `connection` argument, and is folded in as the reserved `"default"` entry.

Each named connection is a lazy, per-session keyed pool of service providers (`connectionRegistry.ts`) that coexists with — and never mutates — the single-slot session-default connection, which keeps a single session safe to share behind a gateway.

### Note: config arg-parser fix included

The `connections` field is expressed as a **catchall object** rather than a `z.record`. The mongosh arg-parser introspects the config schema to build CLI/env options and throws `Unknown field type: ZodRecord`, which (silently) breaks **all** CLI/env config parsing when a `record` field is present. A catchall object is handled the same way as the existing `httpHeaders` field (`coerceObject`), so the whole config-parsing path stays healthy.

This is **Phase 1**: named connections for direct (`connectionString`) targets, with per-call selection on the data tools.

## Deferred (Phase 2 / follow-ups)

- [ ] Metadata tools (`list-databases`, `mongodb-logs`) do not yet accept the per-call `connection` argument.
- [ ] Telemetry still reports the session-default connection (per-call named-connection attribution is not wired).
- [ ] Atlas-typed named connections are deferred to Phase 2 (`connectCluster.ts` is untouched — named connections are direct connection strings only).
- [ ] OIDC device-flow UX for named connections is unwired.
- [ ] `pnpm run update:api` (the api-extractor report) has not been regenerated.

## Testing

- `pnpm run check:types` — clean.
- `pnpm build` — clean.
- Full unit suite green (`vitest run --project unit-and-integration tests/unit`): **960 passed**.
- Adds a permanent unit test (`tests/unit/tools/mongodb/multiConnectionBehavior.test.ts`) proving: (a) backward compat with a single `connectionString`, (b) per-call `connection` selects the correct named connection without touching the session default, and (c) two interleaved `connection` values do not cross state.
- Integration tests were not run here (they require a live Atlas/mongod).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/common/config/configUtils.ts
+++ b/src/common/config/configUtils.ts
@@ -77,6 +77,47 @@ export function commaSeparatedToArray<T extends string[]>(str: string | string[]
 }
 
 /**
+ * Preprocessor for the `connections` config field (env var
+ * `MDB_MCP_CONNECTIONS`). The value arrives as a JSON string mapping connection
+ * names to targets, e.g. `{"analytics":"mongodb://...","reporting":{"connectionString":"mongodb://..."}}`.
+ *
+ * Normalises each target into the `{ connectionString }` object shape and,
+ * mirroring {@link commaSeparatedToArray}, never throws: a malformed JSON blob
+ * is returned unchanged so the downstream Zod `record` schema rejects it and the
+ * error surfaces through the normal config-parsing channel.
+ */
+export function parseNamedConnections(val: unknown): unknown {
+    if (val === undefined || val === null) {
+        return undefined;
+    }
+
+    let raw: unknown = val;
+    if (typeof val === "string") {
+        const trimmed = val.trim();
+        if (trimmed.length === 0) {
+            return undefined;
+        }
+        try {
+            raw = JSON.parse(trimmed);
+        } catch {
+            // Leave the raw string in place; the record schema will reject it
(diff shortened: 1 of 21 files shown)


ahmednfwela:
Closing — opened against upstream in error; this work belongs on our own fork. No upstream contribution intended at this time.

PR #1361, opened 2026-07-16
feat: remove session state and add explicit connection management MCP-601

nirinchev:
## Checklist


src/common/connectionRegistry.ts
```diff
@@ -0,0 +1,522 @@
...
+    readonly connectionId: string;
+    /**
+     * Human/model-readable label: the supplied or generated name, slugified
+     * and suffixed for disambiguation (e.g. `my-project-cluster0-4f2a`).
+     */
+    readonly name: string;
+    readonly source: ConnectionSource;
+    readonly createdAt: Date = new Date();
+    lastUsedAt: Date = new Date();
+    lastError?: string;
```
gagik, line 133: nit, should these be under "metadata" or some other object to separate them from the other args?

src/common/connectionRegistry.ts
```diff
@@ -0,0 +1,522 @@
...
+            );
+        }
+
+        throw new MongoDBError(ErrorCodes.NotConnectedToMongoDB, `Connection "${this.connectionId}" is not connected.`);
+    }
+
+    async isSearchSupported(logger: LoggerBase): Promise<boolean> {
+        const state = this.state;
+        if (state.tag === "connected") {
+            return state.isSearchSupported(logger);
```
gagik, line 198: no need to address it here, this is a pre-existing issue
`state` should have had logger injected into this at initialization, I was kind of confused by logger being passed here.

src/common/connectionRegistry.ts
```diff
@@ -0,0 +1,522 @@
...
+};
+
+/**
+ * Owns the app-level storage and lifecycle of MongoDB connection handles: the
+ * entry map, the preconfigured entry seeded from a configured connection
+ * string, per-scope connection limits, and shutdown. Consumers never hold the
+ * store directly — they access entries through {@link ConnectionRegistry}
+ * views minted with {@link MCPConnectionStore.view}.
+ */
+export class MCPConnectionStore {
```
gagik: move to a different file?

.github/workflows/publish.yml
```diff
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - main
+      - ni/stateless-sessions
```
nirinchev, line 7: This is temp - we should revert before we merge.

src/common/connectionRegistry.ts
```diff
@@ -0,0 +1,533 @@
...
+};
+
+type StoredConnection = {
+    entry: ConnectionEntry;
+    /**
+     * Visibility scope the entry was created under; `undefined` means shared
+     * (the preconfigured entry, or entries created through an unbound view).
+     * Store bookkeeping — deliberately not exposed on the entry itself.
+     */
+    scope?: string;
```
gagik: would it make sense for there to be 
```ts
entries[<connection_id>][<scope>] = entry
```
instead? or can there be multiple entries of the same scope?

src/common/connectionRegistry.ts
```diff
@@ -0,0 +1,533 @@
...
+     * and suffixed for disambiguation (e.g. `my-project-cluster0-4f2a`).
+     */
+    readonly name: string;
+    readonly source: ConnectionSource;
+    readonly createdAt: Date = new Date();
+    lastUsedAt: Date = new Date();
+    lastError?: string;
+
+    /** Revocation cleanup armed at creation; see {@link CreateConnectionEntryOptions.onRevoke}. */
+    private onRevoke?: () => Promise<void>;
```
gagik, line 136: instead of having this hook here, could have an `AtlasConnectionRegistry` that extends this and does `deleteTemporaryUser` internally on revokation? 

It just seems weird for ConnectClusterTool to own this

gagik (commented):
some first look comments, will need to look at it more

src/common/connectionRegistry.ts
```diff
@@ -0,0 +1,533 @@
...
+        }
+    }
+
+    /**
+     * Enforces `maxActiveConnections` per scope, counting explicit entries only
+     * (the preconfigured entry is pinned). Per-scope counting means one scope
+     * (e.g. session) cannot evict another's handles.
+     */
+    private async enforceLimit(scope: string | undefined): Promise<void> {
+        for (;;) {
```
gagik: is this like `while(true)`?

src/common/connectionRegistry.ts
```diff
@@ -0,0 +1,533 @@
...
+                message: `Revoking least-recently-used connection "${lru.entry.connectionId}" because its scope exceeded ${this.userConfig.maxActiveConnections} connections.`,
+            });
+            this.entries.delete(lru.entry.connectionId);
+            await this.revoke(lru.entry);
+        }
+    }
+
+    private async revoke(entry: ConnectionEntry): Promise<void> {
+        this.entries.delete(entry.connectionId);
+        await entry.close().catch(() => undefined);
```
gagik:
```suggestion
		try {
        await entry.close();
        } catch { /** This is best effort, no-op */ }
```

src/transports/base.ts
```diff
@@ -315,14 +312,36 @@ export abstract class TransportRunnerBase<
...
+        // scoping); otherwise the session gets a view of the shared store —
+        // bound to a fresh scope key under the default "session" scope, or
+        // unbound (seeing every entry) under "global". Registries whose
+        // connections become unreachable when the session closes (session-scoped
+        // views, per-session stores) are minted as owned, so `Session.close`
+        // reaps them via `registry.close()`.
+        const sessionScope = userConfig.connectionScope === "session" ? getRandomUUID() : undefined;
+        let connectionRegistry = sessionOptions?.connectionRegistry;
+        if (!connectionRegistry) {
+            if (userConfig.connectionString !== this.userConfig.connectionString) {
```
gagik, line 325: I don't fully understand what scenario this is covering

when would this happen in practice?

tests/integration/tools/atlas/clusters.test.ts
```diff
@@ -422,42 +431,6 @@ describeWithAtlas("clusters", (integration) => {
                     });
                 });
             });
-
-            // Simulates being "connected" to the outer cluster (created by atlas-create-free-cluster above).
-            // The session state is patched so the tool picks up projectId/clusterName without explicit args,
-            // then makes a real upgrade API call against that outer cluster.
-            describe("when connected to the cluster being upgraded", () => {
```
gagik, line 429: should this test not be preserved?

tests/unit/tools/atlas/update/upgradeCluster.test.ts
```diff
@@ -69,7 +69,7 @@ describe("UpgradeClusterTool", () => {
     let mockSession: Partial<Session>;
     let tool: UpgradeClusterTool;
 
-    function buildTool(connectedCluster?: AtlasClusterConnectionInfo): UpgradeClusterTool {
+    function buildTool(): UpgradeClusterTool {
```
gagik, line 72: many things being removed here; is the usecase completely gone with statelessness?

src/common/connectionRegistry.ts
nirinchev: Hm... can you elaborate a bit on this? Not super sure what the proposal is.

src/common/connectionRegistry.ts
nirinchev: Not sure I see an easy way to do that - currently, the connectionRegistry is session-bound, not tool-bound, so `connect` and `atlas-connect-cluster` operate on the same registry.

src/transports/base.ts
nirinchev: Technically, `createServer` could be called with a config that is different from the transport-level config. E.g. a downstream consumer may want to inject a different config values based on the user that's connecting to the server. In that case, we want to make sure to give the server a registry that is fully owned by it and reflects the custom connection string the user should use.

tests/unit/tools/atlas/update/upgradeCluster.test.ts
nirinchev: Previously, there was a mode of operation where you could call upgrade-cluster with no args and then it would work on the currently connected cluster (since we only had at most 1 active connection). With multiple connections, that no longer makes sense, so we force the LLM to call it with explicit project id and cluster name. All these removals were related to that behavior.

tests/integration/tools/atlas/clusters.test.ts
nirinchev: We no longer allow calling `upgrade-cluster` without arguments since we no longer can say unambiguously which cluster you want to upgrade.

src/common/connectionRegistry.ts
gagik: I was wondering whether instead of storing `entries[<connection_id>]` as an array of objects with scopes we should have a 2D array where scope is the second key.

src/transports/base.ts
gagik: hm yeah okay

nirinchev:
Atlas local test failure is unrelated - seems like sample data load issue that's also present on main.

src/common/connectionRegistry.ts
```diff
@@ -0,0 +1,538 @@
...
+    }
+
+    private addEntry({
+        name,
+        clientName,
+        scope,
+        onRevoke,
+    }: CreateConnectionEntryOptions & { scope?: string }): ConnectionEntry {
+        const manager = this.createConnectionManager();
+        if (clientName) {
```
gagik: nit: is there a reason clientName isn't required?

src/common/connectionRegistry.ts
```diff
@@ -0,0 +1,538 @@
...
+        }
+    }
+
+    /**
+     * Enforces `maxActiveConnections` per scope, counting explicit entries only
+     * (the preconfigured entry is pinned). Per-scope counting means one scope
+     * (e.g. session) cannot evict another's handles.
+     */
+    private async enforceLimit(scope: string | undefined): Promise<void> {
+        while (true) {
```
gagik: should we have some kind of infinite loop stopping mechanism here? admittably I'm not sure if it's hypothetically possible but just in case?

gagik (approved):
Seems all reasonable. Don't have any major concerns beyond potential heavy loads caused by resolving telemetry (is most likely fully mitigated by an in-memory store) and again some small nits

src/common/connectionRegistry.ts
nirinchev: `connectionId`s are unique though - they are random uuids, so why would we have more than 1 scope associated with a single id?

src/common/connectionRegistry.ts
nirinchev: `Session.mcpClient` is optional and inside it, `name` is optional as well, so we can't require it without forcing clients to fallback to "unknown" or something like that.

src/common/connectionRegistry.ts
gagik: ah, okay I thought each `entries[<connection_id]` currently pointed to an array of entries... nevermind

gagik (approved):

src/common/connectionRegistry.ts
gagik, line 131: let's make this a constructor arg to prevent https://github.com/10gen/mms/pull/176557/changes#diff-2b60e6310e02abdbdef3eceec17781ef00b8736471a4d35f1bbbb4445e517305R53

packaging/mcpb/manifest.json
```diff
@@ -2,7 +2,7 @@
   "manifest_version": "0.1",
   "name": "mongodb-mcp-server",
   "display_name": "MongoDB",
-  "version": "1.14.0",
+  "version": "1.15.0-alpha.2",
```
blva: [q] why is this needed?

nirinchev: We've been cutting releases from this branch - will revert before we merge

gagik (approved):

PR #1366, opened 2026-07-18
Route request id through elicitation

nirinchev:
## Proposed changes

Without requestId, we'd try to send the elicitation request over the standalone stream, which would always fail on RMCP where we disable it. With requestId, we're eliciting via the tool call POST request, so should be able to get a response from the user. Additionally, we're adding a progress notification loop to get the client to keep the request open past the 60 second tool call limit.


src/tools/tool.ts
```diff
@@ -613,15 +623,37 @@ export abstract class ToolBase<
...
      * @param args - The tool arguments
+     * @param context - The tool execution context, used to relate the
+     * confirmation request to the in-flight tool call
      * @returns A promise resolving to `true` if confirmed or confirmation not
      * required, `false` otherwise
      */
-    public async verifyConfirmed(args: ToolArgs<typeof this.argsShape>): Promise<boolean> {
+    public async verifyConfirmed(
+        args: ToolArgs<typeof this.argsShape>,
+        context?: ToolExecutionContext
```
himanshusinghs: This is likely always available?

src/elicitation.ts
```diff
@@ -1,16 +1,53 @@
...
+    ServerNotification,
+} from "@modelcontextprotocol/sdk/types.js";
 import type { McpServer } from "@modelcontextprotocol/sdk/server/mcp.js";
 
 export type ElicitedInputResult =
     | { accepted: true; fields: Record<string, string> }
     | { accepted: false; fields?: undefined };
 
-const ELICITATION_TIMEOUT_MS = 300_000; // 5 minutes for user interaction
+export type ElicitationOptions = {
```
ciprian-tibulca, line 13: should this contain also `signal?: AbortSignal;` so that cancelled tool calls also cancel the elicitation

nirinchev: Yes, just added in the last commit

himanshusinghs (approved):

src/tools/tool.ts
himanshusinghs: addressed

PR #1380, opened 2026-07-22
feat: Add CMK support in atlas-create-cluster

manupedrozo:
## Description

Add encryption at rest via customer-managed key support to the `atlas-create-cluster` tool.
* The selected CMK provider must already have a valid encryption at rest configuration in the project.
* Defaults to the cluster's provider if a valid configuration exists.

Ticket: [CLOUDP-407393](https://jira.mongodb.org/browse/CLOUDP-407393)

## Checklist


jeroenvervaeke (approved):

tests/accuracy/createCluster.test.ts
```diff
@@ -441,4 +441,82 @@ describeAccuracyTests([
...
+                projectId: PROJECT_ID,
+                clusterName: CLUSTER_NAME,
+                provider: "AWS",
+                region: "US_EAST_1",
+            }),
+        },
+        expectedToolCalls: [
+            ...optionalListProjects,
+            {
+                toolName: "atlas-create-cluster",
```
laylaazouz, line 458: Do these test cases cover the 3 cases specified in the TD? (specifically the no preference/unset case)

- An explicitly requested CMK provider is passed to the tool.
- An explicit ask not to use a CMK results in NONE.
**- Prompts that do not express a CMK preference leave the input unset.**

manupedrozo: Yes, this is covered by existing tests, which do not mention encryption at rest.
If any value is passed for `encryptionAtRestProvider` in those, the expected call no longer matches and accuracy drops.

PR #1386, opened 2026-07-23
fix: report per-index write errors and inserted count when insert-many partially fails

innolope-dev:
## Proposed changes

When `insert-many` hits a `MongoBulkWriteError`, the generic error handler flattens it to a single message. That loses the two things an agent needs to recover: **how many documents actually made it in**, and **which indices failed**. With the driver's default ordered inserts a failure is usually partial — everything before the first failing index was inserted, everything after it was silently skipped — so the LLM can neither tell the user what state the collection is in nor retry with only the missing documents (retrying the whole batch would duplicate the already-inserted ones).

This PR adds a `handleError` override to `InsertManyTool` that, for `MongoBulkWriteError`, reports:

- how many of the requested documents failed and how many were inserted,
- each failing index with its error code and (secret-redacted) server message, wrapped via `formatUntrustedData` since `errmsg` can contain document values,
- the ordered-insert skip semantics, and guidance to retry with only the documents that were not inserted.

The per-index list is capped at 10 entries so a batch with many failures can't blow up the response. All other errors fall through to `super.handleError` unchanged.

Example response for a 3-document batch where index 1 collides with an existing `_id`:

```
Error running insert-many: 1 of 3 document(s) failed to insert into mydb.coll1. 1 document(s) were inserted. Because inserts are ordered, documents after the first failing index were not attempted. Fix or remove the failing document(s) and retry with only the documents that were not inserted.
<untrusted-user-data-...>
- index 1 (code 11000): E11000 duplicate key error collection: mydb.coll1 index: _id_ dup key: ...
</untrusted-user-data-...>
```

Notes:

- Added an integration test covering the partial-failure case (asserts the reported counts, the per-index line, and that the documents before/after the failing index were/weren't inserted). The pre-existing duplicate-key test now flows through the new branch and still passes.
- `api-extractor/reports/tools.public.api.md` regenerated via `pnpm run update:api` (the protected override shows up in the tools report).
- Ran the insert-many suite locally against a real mongod: 14 passed / 3 skipped (the search-enabled suites need a container runtime not available locally).

## Checklist

### DIFF
--- a/api-extractor/reports/tools.public.api.md
+++ b/api-extractor/reports/tools.public.api.md
@@ -1132,6 +1132,8 @@ export class InsertManyTool extends MongoDBToolBase {
     // (undocumented)
     protected execute(input: ToolArgs<typeof InsertManyTool.argsShape>): Promise<ToolResult<typeof InsertManyTool.outputSchema>>;
     // (undocumented)
+    protected handleError(error: unknown, args: ToolArgs<typeof InsertManyTool.argsShape>): Promise<CallToolResult>;
+    // (undocumented)
     static operationType: OperationType;
     // (undocumented)
     outputSchema: {
--- a/api-extractor/reports/tools.public.api.md
+++ b/api-extractor/reports/tools.public.api.md
@@ -1996,6 +1996,8 @@ export class InsertManyTool extends MongoDBToolBase {
     // (undocumented)
     protected execute(input: ToolArgs<typeof InsertManyTool.argsShape>): Promise<ToolResult<typeof InsertManyTool.outputSchema>>;
     // (undocumented)
+    protected handleError(error: unknown, args: ToolArgs<typeof InsertManyTool.argsShape>): Promise<CallToolResult>;
+    // (undocumented)
     static operationType: OperationType;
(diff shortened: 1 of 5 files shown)


himanshusinghs (approved):

himanshusinghs:
Hi @innolope-dev! Thanks for the PR - its definitely a good improvement. Can you please confirm if you have signed the LCA?

addaleax (approved):
Can confirm that the CLA has been signed

himanshusinghs:
@innolope-dev Let me know if you'd be available to resolve the merge conflicts on the PR or if you want me to take over.

innolope-dev:
> @innolope-dev Let me know if you'd be available to resolve the merge conflicts on the PR or if you want me to take over.

Sure, will resolve

innolope-dev:
Note that this wasn't a pure conflict resolution: Keychain.allSecrets no longer exists, so the redaction call had to change to this.session.keychain.redact(...). Since the approvals carried over the rebase, that specific line hasn't been reviewed — worth a second look before merge.

PR #1399, opened 2026-07-27
feat: Add multi-region support in atlas-create-cluster

manupedrozo:
## Description

Adding multi-region cluster support to the `atlas-create-cluster` tool.
* The tool now takes a `regions` array of 1 to 3 regions in priority order.
* Electable node distribution is abstracted following Atlas's recommended resiliency pattern.

## Checklist

- [x] I have signed the [MongoDB CLA](https://www.mongodb.com/legal/contributor-agreement)

PR #1400, opened 2026-07-28
fix: handle createCluster region in a backwards-compatible way CLOUDP-407395

nirinchev:
## Proposed changes

https://github.com/mongodb-js/mongodb-mcp-server/pull/1399 replaced region with regions for the create-cluster tool which may be a breaking change for clients that cache/lock down schemas. This PR proposes a mechanism to handle these types of changes in a backwards compatible way - it introduces a tool-specific normalizeArgs hook that can mutate the model-supplied args to match the desired shape. In this specific case, we look for `region` in the payload and use it to populate `regions` if not supplied.

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -173,7 +173,10 @@ export class Server<
         // TODO: Eventually we might want to make tools reactive too instead of relying on custom logic.
         this.registerTools();
 
-        // This is a workaround for an issue we've seen with some models, where they'll see that everything in the `arguments`
+        // Wrapping the tool call handler to normalize the arguments before the SDK validates them against the
+        // tool's schema. This gives tools a chance to map deprecated arguments through `normalizeRawArgs`.
+        //
+        // It also works around an issue we've seen with some models, where they'll see that everything in the `arguments`
         // object is optional, and then not pass it at all. However, the MCP server expects the `arguments` object to be if
         // the tool accepts any arguments, even if they're all optional.
         //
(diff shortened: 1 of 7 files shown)


src/tools/atlas/create/createCluster.ts
```diff
@@ -233,6 +233,16 @@ export class CreateClusterTool extends AtlasToolBase {
     public override outputSchema = CreateClusterOutputSchema;
     public argsShape = CreateClusterArgsShape;
 
+    /** Accepts the `region` argument that `regions` replaced, mapping it to a single-region cluster. */
+    public override normalizeRawArgs(args: Record<string, unknown>): Record<string, unknown> {
```
manupedrozo, line 237: Nice, really like this approach which keeps the schema clean.
We should probably have a plan for removing this at some point in the future.

manupedrozo (approved):

src/tools/atlas/create/createCluster.ts
nirinchev: Yes, we should be able to remove these types of shims as soon as we publish new versions of the remote mcp connector to all marketplaces.

ciprian-tibulca (approved):

src/server.ts
```diff
@@ -173,7 +173,10 @@ export class Server<
         // TODO: Eventually we might want to make tools reactive too instead of relying on custom logic.
         this.registerTools();
 
-        // This is a workaround for an issue we've seen with some models, where they'll see that everything in the `arguments`
+        // Wrapping the tool call handler to normalize the arguments before the SDK validates them against the
+        // tool's schema. This gives tools a chance to map deprecated arguments through `normalizeRawArgs`.
+        //
+        // It also works around an issue we've seen with some models, where they'll see that everything in the `arguments`
```
himanshusinghs, line 179: The PR looks good but I have a different concern - it feels like with every targeted fix we are diverging away from the standard SDK handling patterns.

We probably should consider raising issues to MCP SDK GitHub and validating with them if this is something they will eventually fix or if the client is doing something completely wrong.

Like for example in this case the arguments not being provided by the LLM seems like a pretty standard thing which the SDK should guard against.

himanshusinghs (approved):

PR #1401, opened 2026-07-29
feat: Add scaling to upgrade cluster tool CLOUDP-407394

laylaazouz:
## Proposed changes
Add scaling to atlas-upgrade-cluster (CLOUDP-407394)
Extends atlas-upgrade-cluster so it can also scale an already-Dedicated cluster in place (resizing its instance size and/or updating compute autoscaling), not just upgrade Free/Flex clusters.

- Resize a Dedicated cluster's instance size (M10–M80) via targetTier.
- Enable/disable/reconfigure compute autoscaling via computeAutoScaling/minInstanceSize/maxInstanceSize (max up to M200).
- Autoscaling defaults on resize, matching atlas-create-cluster's convention.
- Response message reflects change (size, autoscaling, or both).
## Checklist


src/common/atlas/instanceSizes.ts
manupedrozo, line 1: Consider adding this to the common clusters file instead: https://github.com/mongodb-js/mongodb-mcp-server/blob/main/src/common/atlas/cluster.ts

src/telemetry/types.ts
```diff
@@ -281,8 +281,11 @@ export type StreamsToolMetadata = AtlasMetadata & {
 };
 
 export type UpgradeClusterMetadata = AtlasMetadata & {
-    original_tier?: "free" | "flex";
-    target_tier?: "flex" | "m10";
+    original_tier?: "free" | "flex" | "dedicated";
+    target_tier?: "free" | "flex" | "dedicated";
+    original_instance_size?: string;
+    target_instance_size?: string;
```
manupedrozo: Consider merging all 4 into just 2: 
original_tier: "free" | "flex" | instance-sizes
target_tier: "free" | "flex" | instance-sizes

tests/integration/tools/atlas/clusters.test.ts
```diff
@@ -673,5 +673,48 @@ describeWithAtlas("clusters", (integration) => {
                 });
             });
         });
+
+        describe("atlas-upgrade-cluster (scale in place)", () => {
```
manupedrozo: can we move this to the atlas-upgrade-cluster group above?

tests/unit/common/instanceSizes.test.ts
manupedrozo, line 1: I dont see much value in these tests, we are effectively testing an array. Should we remove them?

tests/unit/tools/atlas/update/upgradeCluster.test.ts
manupedrozo, line 211: Any new error cases for dedicated clusters? Example: trying to scale unsupported clusters.

src/tools/atlas/update/upgradeCluster.ts
```diff
@@ -48,17 +54,47 @@ type FlexToM10Body = {
...
     provider?: string,
     region?: string
 ): FreeToM10Body | FlexToM10Body {
+    const enabled = computeAutoScaling ?? true;
+    const autoScaling = {
+        compute: {
+            enabled,
+            scaleDownEnabled: enabled,
+            minInstanceSize: enabled ? (minInstanceSize ?? "M10") : minInstanceSize,
+            maxInstanceSize: enabled ? (maxInstanceSize ?? getDefaultMaxAutoScalingSize("M10")) : maxInstanceSize,
```
manupedrozo: I think this won't cause issues since any `minInstanceSize` above "M10" for the upgrade case *should* get an API validation error, but I would consider this case: if `minInstanceSize` is "M50" and no `maxInstanceSize` is provided, the calculated the max would be "M30", which is lower than the `minInstanceSize`.
Perhaps the simpler approach is enforcing that either both `min` or `max` are passed, or neither.

src/tools/atlas/update/upgradeCluster.ts
```diff
@@ -104,12 +144,17 @@ async function resolveClusterInfo(
         const raw = await apiClient.getCluster({ params: { path: { groupId: projectId, clusterName } } }, context);
         const cluster = formatCluster(raw);
         const firstRegionConfig = raw.replicationSpecs?.[0]?.regionConfigs?.[0] as
-            | { backingProviderName?: string; regionName?: string }
+            | {
+                  autoScaling?: { compute?: { enabled?: boolean; minInstanceSize?: string; maxInstanceSize?: string } };
+              }
             | undefined;
```
manupedrozo: We only consider the first replicationSpec & region? Existing code didn't care about other replication specs but now we do!

src/tools/atlas/update/upgradeCluster.ts
```diff
@@ -104,12 +144,17 @@ async function resolveClusterInfo(
...
         return {
             instanceType: cluster.instanceType,
-            provider: argOverrides.provider ?? firstRegionConfig?.backingProviderName,
-            region: argOverrides.region ?? firstRegionConfig?.regionName,
+            provider: argOverrides.provider ?? cluster.provider,
+            region: argOverrides.region ?? cluster.region,
+            raw,
+            currentInstanceSize: cluster.instanceSize,
+            currentAutoScaling: firstRegionConfig?.autoScaling?.compute,
```
manupedrozo: I dont think this is enough info to verify that we support scaling this cluster (single-shard or replicaset clusters with supported instance sizes). 
E.g. is this a multi-shard cluster? geo-sharded? different autoscaling configs?

Details on naming:
* I think the `current` prefix does not fit here, should just be `instanceSize` & `autoScaling`
* `currentAutoScaling` is actually pointing to `compute`

src/tools/atlas/update/upgradeCluster.ts
```diff
@@ -129,19 +174,76 @@ async function resolveClusterInfo(
     }
 }
 
+// Default max autoscaling size when none is given: two tiers above `size`, capped at M80.
+function getDefaultMaxAutoScalingSize(size: InstanceSize): InstanceSize {
```
manupedrozo: Take a look at `getMaxAutoScalingSize()` from the createCluster tool

src/tools/atlas/update/upgradeCluster.ts
```diff
@@ -129,19 +174,76 @@ async function resolveClusterInfo(
...
+    if (!isStandardInstanceSize(size)) {
```
manupedrozo: The function takes an `InstanceSize`, I dont think we need this validation.

src/tools/atlas/update/upgradeCluster.ts
```diff
@@ -129,19 +174,76 @@ async function resolveClusterInfo(
...
+function buildScaleClusterBody(
+    raw: ClusterDescription20240805 | undefined,
+    targetSize: string,
+    compute: ComputeAutoScaling
+): ScaleClusterBody {
+    const replicationSpecs = (raw?.replicationSpecs ?? []).map((spec) => {
+        const regionConfigs = ((spec.regionConfigs ?? []) as Array<Record<string, unknown>>).map((rc) => {
+            const electableSpecs = rc.electableSpecs as Record<string, unknown> | undefined;
+            const readOnlySpecs = rc.readOnlySpecs as Record<string, unknown> | undefined;
+            const existingAutoScaling = rc.autoScaling as { diskGB?: unknown } | undefined;
```
manupedrozo: Should we calculate the entire autoScaling field outside of this function? Its a bit noisy that we are taking the compute but calculating the diskGB here.
I see that we are marking diskGB as true below as a default. Is that the correct behavior? What if the customer did not intend to touch autoscaling at all, do we enable disk autoscaling regardless?

src/tools/atlas/update/upgradeCluster.ts
```diff
@@ -177,13 +304,95 @@ export class UpgradeClusterTool extends AtlasToolBase {
...
+                    throw new UpgradeClusterError(
+                        `No changes specified for Dedicated cluster "${clusterName}". Provide instanceSize and/or computeAutoScaling to scale it.`
+                    );
+                }
+                if (
+                    clusterInfo.currentInstanceSize === undefined ||
+                    !isStandardInstanceSize(clusterInfo.currentInstanceSize)
+                ) {
+                    throw new UpgradeClusterError(
+                        `Cluster "${clusterName}" has instance size "${clusterInfo.currentInstanceSize ?? "unknown"}", which this tool does not support scaling (NVMe, Gen2, R-series, and sizes outside M10-M80 are unsupported).`
```
manupedrozo: >(NVMe, Gen2, R-series, and sizes outside M10-M80 are unsupported)

Instead of listing all unsupported, can just clarify what IS supported (to avoid having to keep this up to date).

src/tools/atlas/update/upgradeCluster.ts
```diff
@@ -177,13 +304,95 @@ export class UpgradeClusterTool extends AtlasToolBase {
...
+                    );
+                }
+
+                const newSize: InstanceSize = args.instanceSize ?? clusterInfo.currentInstanceSize;
+                const newEnabled = args.computeAutoScaling ?? clusterInfo.currentAutoScaling?.enabled ?? false;
+                const newMin =
+                    args.minInstanceSize ??
+                    clusterInfo.currentAutoScaling?.minInstanceSize ??
+                    (newEnabled ? newSize : undefined);
```
manupedrozo: This does not reconciliate an existing `minInstanceSize` with the target instance size when scaling.
E.g.:
- Existing cluster with instanceSize = M10 and minInstanceSize = M10
- I call this tool to scale to M30, no autoscaling inputs
- The result is instanceSize = M30, minInstanceSize = M10

src/tools/atlas/update/upgradeCluster.ts
```diff
@@ -177,13 +304,95 @@ export class UpgradeClusterTool extends AtlasToolBase {
...
+                const newMax =
```
manupedrozo: Similar comment to minInstanceSize, lets revisit the logic around min/max calculation.

src/tools/atlas/update/upgradeCluster.ts
```diff
@@ -244,6 +490,7 @@ export class UpgradeClusterTool extends AtlasToolBase {
         target: "FLEX" | "M10",
         backingProviderName: string | undefined,
         regionName: string | undefined,
+        args: { computeAutoScaling?: boolean; minInstanceSize?: string; maxInstanceSize?: string },
```
manupedrozo: `args` is too generic. Can we rename to something that indicates these are autoscaling arguments?

src/tools/atlas/update/upgradeCluster.ts
```diff
@@ -268,7 +515,15 @@ export class UpgradeClusterTool extends AtlasToolBase {
                 return await this.apiClient.upgradeTenantUpgrade(
                     {
                         params: { path: { groupId: projectId } },
-                        body: buildM10UpgradeBody("FREE", clusterName, backingProviderName, regionName),
+                        body: buildM10UpgradeBody(
+                            "FREE",
+                            clusterName,
+                            args.computeAutoScaling,
+                            args.minInstanceSize,
+                            args.maxInstanceSize,
```
manupedrozo: should we pass the whole object with autoscaling arguments here instead?

tests/unit/tools/atlas/update/upgradeCluster.test.ts
```diff
@@ -518,6 +618,297 @@ describe("UpgradeClusterTool", () => {
         });
     });
 
+    describe("DEDICATED cluster (scale in place)", () => {
```
manupedrozo:
```suggestion
    describe("DEDICATED cluster", () => {
```

manupedrozo: Didnt review these yet, will take a look once the other comments are resolved since these will likely change.

src/tools/atlas/update/upgradeCluster.ts
laylaazouz: This would break the case where only one of the two is changed while relying on the current stored value for the other. I think validating min<max after computing instead would be a good approach, what do you think?

manupedrozo: Yes, we'd stop supporting that case at the benefit of clearer defaulting logic.
Perhaps a better approach is to validate that `minInstanceSize` is either not passed or M10 for the upgrade case.
Any other invalid input we can leave for the API to validate, I am mainly worried about our own defaulting logic ending up in an invalid config, which would make it hard for a customer/Agent to understand the resulting error.

src/telemetry/types.ts
laylaazouz: merged into original_tier target_tier

tests/integration/tools/atlas/clusters.test.ts
laylaazouz: Moved this to the existing group above

tests/unit/common/instanceSizes.test.ts
laylaazouz: had added these for consistency but I agree there is not much added value, removed

tests/unit/tools/atlas/update/upgradeCluster.test.ts
laylaazouz: i think this is covered in this file, just further down

src/tools/atlas/update/upgradeCluster.ts
laylaazouz: addressed, `minInstanceSize` must now be omitted or 'M10' for upgrade case

src/tools/atlas/update/upgradeCluster.ts
laylaazouz: now rejecting multi-shard and checking that all regions within a shard have the same instance size

src/tools/atlas/update/upgradeCluster.ts
laylaazouz: Naming fixed. Also added checks rejecting multi-shard clusters and clusters with inconsistent instance sizes.

src/tools/atlas/update/upgradeCluster.ts
laylaazouz: fixed, removed redundant validation

src/tools/atlas/update/upgradeCluster.ts
laylaazouz: extracted `buildRegionAutoScaling`, fixed diskGB bug (left as is, not enabled)

src/tools/atlas/update/upgradeCluster.ts
laylaazouz: fixed

src/tools/atlas/update/upgradeCluster.ts
laylaazouz: Fixed. when instanceSize changes, min/max now recompute, same as `createCluster.ts`

src/tools/atlas/update/upgradeCluster.ts
laylaazouz: renamed to autoScalingArgs

src/tools/atlas/update/upgradeCluster.ts
laylaazouz: added AutoScalingArgs type, now passing this in instead of 3 separate arguments

src/tools/atlas/update/upgradeCluster.ts
```diff
@@ -129,19 +189,86 @@ async function resolveClusterInfo(
     }
 }
 
+// Default max autoscaling size when none is given: two tiers above `size`, capped at M80.
+function getDefaultMaxAutoScalingSize(size: StandardInstanceSize): StandardInstanceSize {
```
manupedrozo: I am not sure why we are not reusing `getMaxAutoScalingSize()` from the createCluster tool. 
We are missing the extra logic for `M60`s & `M80`s as well.

src/tools/atlas/update/upgradeCluster.ts
```diff
@@ -129,19 +189,86 @@ async function resolveClusterInfo(
...
+    enabled: boolean;
+    scaleDownEnabled: boolean;
+    minInstanceSize?: string;
+    maxInstanceSize?: string;
+};
+
+type ScaleRegionConfig = {
+    electableSpecs?: { instanceSize: string } & Record<string, unknown>;
+    readOnlySpecs?: { instanceSize: string } & Record<string, unknown>;
+    autoScaling?: { compute: ComputeAutoScaling; diskGB?: unknown };
```
manupedrozo: Should diskGB be `{ enabled: boolean }`?

src/tools/atlas/update/upgradeCluster.ts
```diff
@@ -129,19 +189,86 @@ async function resolveClusterInfo(
...
+} & Record<string, unknown>;
+
+type ScaleClusterBody = {
+    replicationSpecs: Array<{ regionConfigs: ScaleRegionConfig[] } & Record<string, unknown>>;
+};
+
+function buildRegionAutoScaling(
```
manupedrozo: Can we inline this one? A function seems unnecessary.

src/tools/atlas/update/upgradeCluster.ts
```diff
@@ -129,19 +189,86 @@ async function resolveClusterInfo(
...
+            const electableSpecs = rc.electableSpecs as Record<string, unknown> | undefined;
+            const readOnlySpecs = rc.readOnlySpecs as Record<string, unknown> | undefined;
+            if (!electableSpecs && !readOnlySpecs) {
+                return { ...rc };
+            }
+            return {
+                ...rc,
+                ...(electableSpecs && { electableSpecs: { ...electableSpecs, instanceSize: targetSize } }),
+                ...(readOnlySpecs && { readOnlySpecs: { ...readOnlySpecs, instanceSize: targetSize } }),
+                autoScaling: buildRegionAutoScaling(compute, rc.autoScaling as { diskGB?: unknown } | undefined),
```
manupedrozo: Related to the previous comment, we can just build the autoScaling struct here and avoid the casting + function call.

src/tools/atlas/update/upgradeCluster.ts
```diff
@@ -129,19 +189,86 @@ async function resolveClusterInfo(
...
+
 class UpgradeClusterError extends Error {}
 
 export const UpgradeClusterOutputSchema = {
-    originalTier: z.enum(["FREE", "FLEX"]),
-    targetTier: z.enum(["FLEX", "M10"]),
+    originalTier: z.enum(["FREE", "FLEX", "DEDICATED"]),
+    targetTier: z.enum(["FLEX", "DEDICATED"]),
+    originalInstanceSize: z.string().optional(),
+    targetInstanceSize: z.string().optional(),
```
manupedrozo: I think we should merge these just like the telemetry ones (no need for "DEDICATED")

src/tools/atlas/update/upgradeCluster.ts
```diff
@@ -150,16 +277,41 @@ export class UpgradeClusterTool extends AtlasToolBase {
...
             .optional()
-            .describe("Target tier to upgrade to. Defaults to FLEX for Free clusters and M10 for Flex clusters."),
+            .describe(
+                "Target tier to upgrade a Free or Flex cluster to (FLEX or M10 Dedicated). Defaults to FLEX for Free clusters and M10 for Flex clusters. Use instanceSize instead once a cluster is already Dedicated."
+            ),
+        instanceSize: standardInstanceSizeEnum
+            .optional()
+            .describe(
+                "New instance size (M10-M80) to scale an already-Dedicated cluster in place. Use targetTier instead for a Free/Flex source cluster."
+            ),
```
manupedrozo, line 370: Should we merge these into a single input?

src/tools/atlas/update/upgradeCluster.ts
```diff
@@ -129,19 +189,86 @@ async function resolveClusterInfo(
...
+    maxInstanceSize: z.string().optional(),
     resolvedProvider: z.string().optional(),
     resolvedRegion: z.string().optional(),
     clusterId: z.string().optional(),
 };
 
 export class UpgradeClusterTool extends AtlasToolBase {
     static toolName = "atlas-upgrade-cluster";
-    public description = `Upgrade a MongoDB Atlas cluster tier. Upgrades Free (M0) clusters to Flex or M10 Dedicated, or Flex clusters to M10 Dedicated. The upgrade path is determined automatically from the current tier unless overridden with targetTier. Note to LLM: If provider and region are not already known, ask for both together in a single question before calling this tool. ${REGION_RECOMMENDATIONS}`;
+    public description =
```
manupedrozo, line 353: Much better now!

src/tools/atlas/update/upgradeCluster.ts
```diff
@@ -129,19 +189,86 @@ async function resolveClusterInfo(
...
+        "Upgrade or scale a MongoDB Atlas cluster: upgrades Free/Flex clusters to Flex or M10 Dedicated, scales a dedicated cluster's instanceSize, or updates autoscaling. " +
```
manupedrozo: Probably won't cause accuracy drop but the ", or updates autoscaling." reads like you cannot scale + update autoscaling at once.

src/tools/atlas/update/upgradeCluster.ts
```diff
@@ -150,16 +277,41 @@ export class UpgradeClusterTool extends AtlasToolBase {
...
+        instanceSize: standardInstanceSizeEnum
+            .optional()
+            .describe(
+                "New instance size (M10-M80) to scale an already-Dedicated cluster in place. Use targetTier instead for a Free/Flex source cluster."
+            ),
+        computeAutoScaling: z
+            .boolean()
+            .optional()
+            .describe(
+                "When true, enables compute autoscaling for an already-Dedicated cluster being scaled, or for a Free/Flex cluster upgrading to M10 Dedicated (Flex itself has no autoscaling). Min instance size is set to the selected instance size, max is set two tiers above, unless overridden. Omit unless explicitly specified by the user."
```
manupedrozo: "for an already-Dedicated cluster being scaled" - what if we are not scaling?
"(Flex itself has no autoscaling)" - would remove but if needed for accuracy, maybe "Flex clusters do not support autoscaling"
"Min instance size is set to the selected instance size, max is set two tiers above, unless overridden." - we may get away with not mentioning this here since its in the tool description.

src/tools/atlas/update/upgradeCluster.ts
```diff
@@ -150,16 +277,41 @@ export class UpgradeClusterTool extends AtlasToolBase {
...
+            ),
+        minInstanceSize: standardInstanceSizeEnum
+            .optional()
+            .describe(
+                "Minimum instance size (M10-M80) for compute autoscaling, for Dedicated scaling or a Free/Flex-to-M10 upgrade."
```
manupedrozo: "for Dedicated scaling" - would reword a bit
We should add here and for max the mention to omit unless specified by the user.

src/tools/atlas/update/upgradeCluster.ts
```diff
@@ -150,16 +277,41 @@ export class UpgradeClusterTool extends AtlasToolBase {
...
+            ),
+        maxInstanceSize: standardInstanceSizeEnum
+            .optional()
+            .describe(
+                "Maximum instance size (M10-M80) for compute autoscaling, for Dedicated scaling or a Free/Flex-to-M10 upgrade. Defaults to two tiers above the instance size, capped at M80."
```
manupedrozo: This one mentions the default but min doesn't.

src/tools/atlas/update/upgradeCluster.ts
```diff
@@ -150,16 +277,41 @@ export class UpgradeClusterTool extends AtlasToolBase {
...
+            ),
         provider: z
             .string()
             .regex(ALLOWED_PROVIDER_REGEX, "Provider must be uppercase letters and underscores only")
             .optional()
-            .describe("Cloud provider (e.g. AWS, GCP, AZURE). If omitted, the existing value is preserved."),
+            .describe(
+                "Cloud provider (e.g. AWS, GCP, AZURE) for a Free/Flex source cluster. Preserves the existing value if omitted. Does not apply once a cluster is already Dedicated, since scaling never relocates a cluster."
```
manupedrozo:
```suggestion
                "Cloud provider (e.g. AWS, GCP, AZURE) for a Free/Flex source cluster. Preserves the existing value if omitted. Does not apply once a cluster is already Dedicated."
```

src/tools/atlas/update/upgradeCluster.ts
```diff
@@ -177,13 +329,106 @@ export class UpgradeClusterTool extends AtlasToolBase {
...
-                    `Cluster "${clusterName}" is already at the Dedicated tier and cannot be upgraded further.`
+            case "DEDICATED": {
+                if (args.targetTier !== undefined) {
+                    throw new UpgradeClusterError(
+                        `Cluster "${clusterName}" is already Dedicated. Use instanceSize/computeAutoScaling/minInstanceSize/maxInstanceSize instead of targetTier.`
+                    );
+                }
+                if (args.provider !== undefined || args.region !== undefined) {
+                    throw new UpgradeClusterError(
+                        `Invalid Arguments:provider/region. provider/region are not valid when scaling an already-Dedicated cluster "${clusterName}", scaling does not relocate a cluster.`
```
manupedrozo:
```suggestion
                        `Invalid Arguments:provider/region. provider/region are not valid when scaling an already-Dedicated cluster "${clusterName}".`
```

src/tools/atlas/update/upgradeCluster.ts
```diff
@@ -177,13 +329,106 @@ export class UpgradeClusterTool extends AtlasToolBase {
...
+                    );
+                }
+                if (
+                    args.instanceSize === undefined &&
+                    args.computeAutoScaling === undefined &&
+                    args.minInstanceSize === undefined &&
+                    args.maxInstanceSize === undefined
+                ) {
+                    throw new UpgradeClusterError(
+                        `No changes specified for Dedicated cluster "${clusterName}". Provide instanceSize and/or computeAutoScaling to scale it.`
```
manupedrozo: What about min/max?

src/tools/atlas/update/upgradeCluster.ts
```diff
@@ -177,13 +329,106 @@ export class UpgradeClusterTool extends AtlasToolBase {
...
+                    );
+                }
+                if (clusterInfo.instanceSize === undefined || !isStandardInstanceSize(clusterInfo.instanceSize)) {
+                    throw new UpgradeClusterError(
+                        `Cluster "${clusterName}" has instance size "${clusterInfo.instanceSize ?? "unknown"}", which this tool does not support scaling. Only standard M10-M80 instance sizes are supported.`
+                    );
+                }
+                if ((clusterInfo.raw?.replicationSpecs?.length ?? 0) > 1) {
+                    throw new UpgradeClusterError(
+                        `Cluster "${clusterName}" has multiple shards, which this tool does not support scaling. Only single-shard (replica set) clusters are supported.`
```
manupedrozo: Lets adjust a bit: SHARDED with a single shard and REPLICASET are two different types of clusters, see: https://www.mongodb.com/docs/api/doc/atlas-admin-api-v2/operation/operation-updategroupcluster#operation-updategroupcluster-body-application-vnd-atlas-2024-10-23-json-clustertype

src/tools/atlas/update/upgradeCluster.ts
```diff
@@ -177,13 +329,106 @@ export class UpgradeClusterTool extends AtlasToolBase {
...
+                    throw new UpgradeClusterError(
+                        `No changes specified for Dedicated cluster "${clusterName}". Provide instanceSize and/or computeAutoScaling to scale it.`
+                    );
+                }
+                if (clusterInfo.instanceSize === undefined || !isStandardInstanceSize(clusterInfo.instanceSize)) {
+                    throw new UpgradeClusterError(
+                        `Cluster "${clusterName}" has instance size "${clusterInfo.instanceSize ?? "unknown"}", which this tool does not support scaling. Only standard M10-M80 instance sizes are supported.`
+                    );
+                }
+                if ((clusterInfo.raw?.replicationSpecs?.length ?? 0) > 1) {
```
manupedrozo: Should we also use the `clusterType` to help with validation and possibly emit better errors?

src/tools/atlas/update/upgradeCluster.ts
```diff
@@ -177,13 +329,106 @@ export class UpgradeClusterTool extends AtlasToolBase {
...
+                    throw new UpgradeClusterError(
+                        `Cluster "${clusterName}" has multiple shards, which this tool does not support scaling. Only single-shard (replica set) clusters are supported.`
+                    );
+                }
+                const regionElectableSizes = (clusterInfo.raw?.replicationSpecs?.[0]?.regionConfigs ?? [])
+                    .map((rc) => (rc as { electableSpecs?: { instanceSize?: string } }).electableSpecs?.instanceSize)
+                    .filter((size): size is string => size !== undefined);
+                if (new Set(regionElectableSizes).size > 1) {
+                    throw new UpgradeClusterError(
+                        `Cluster "${clusterName}" has regions with different instance sizes, which this tool does not support scaling consistently.`
```
manupedrozo: Maybe something along the lines of: "which this tool does not support scaling to avoid unintended changes."

src/tools/atlas/update/upgradeCluster.ts
manupedrozo: Can we switch to `if` statements for these cases to make it easier to follow the logic, and feel free to use comments to help clarify.
Didn't look at tests yet, this is likely already done: lets make sure to cover the edge cases with unit tests to avoid regressions in this logic.

src/tools/atlas/update/upgradeCluster.ts
```diff
@@ -46,26 +57,64 @@ type FlexToM10Body = {
...
 ): FreeToM10Body | FlexToM10Body {
+    if (autoScalingArgs.minInstanceSize !== undefined && autoScalingArgs.minInstanceSize !== "M10") {
+        throw new UpgradeClusterError(`minInstanceSize must be omitted or "M10" when upgrading to M10 Dedicated.`);
+    }
+
+    const enabled = autoScalingArgs.computeAutoScaling ?? true;
+    const minInstanceSize = enabled ? (autoScalingArgs.minInstanceSize ?? "M10") : autoScalingArgs.minInstanceSize;
+    const maxInstanceSize = enabled
+        ? (autoScalingArgs.maxInstanceSize ?? getDefaultMaxAutoScalingSize("M10"))
+        : autoScalingArgs.maxInstanceSize;
```
manupedrozo: Would not add validation and defaulting logic within this function, would keep it focused on building the body.

tests/unit/tools/atlas/update/upgradeCluster.test.ts
manupedrozo: Got it (still didn't take a full look at these tests), should we move the error cases to this group?

manupedrozo (commented):
A new commit came in while I was taking a look, in case any comment is outdated!

src/tools/atlas/update/upgradeCluster.ts
laylaazouz: Moved `getMaxAutoScalingSize` into common/atlas/cluster.ts so both tools share the same function

src/tools/atlas/update/upgradeCluster.ts
laylaazouz: inlined into buildScaleClusterBody.

src/tools/atlas/update/upgradeCluster.ts
laylaazouz: Merged!

src/tools/atlas/update/upgradeCluster.ts
laylaazouz: Already done when addressing last round of comments

src/tools/atlas/update/upgradeCluster.ts
laylaazouz: reworded this  bit

src/tools/atlas/update/upgradeCluster.ts
laylaazouz: fixed, mentions all 4 now

src/telemetry/types.ts
```diff
@@ -281,8 +283,9 @@ export type StreamsToolMetadata = AtlasMetadata & {
 };
 
 export type UpgradeClusterMetadata = AtlasMetadata & {
```
manupedrozo, line 283: We are missing min/max autoscaling values.

src/telemetry/types.ts
```diff
@@ -281,8 +283,9 @@ export type StreamsToolMetadata = AtlasMetadata & {
...
-    original_tier?: "free" | "flex";
-    target_tier?: "flex" | "m10";
+    original_tier?: "free" | "flex" | Lowercase<StandardInstanceSize>;
```
manupedrozo: No concern on my side on the usage of `StandardInstanceSize` here but I see we are not usually importing types from the tool side here.
May end up going with a string for these, similar to `instanceSize` in `CreateClusterMetadata`

src/tools/atlas/update/upgradeCluster.ts
```diff
@@ -15,6 +22,16 @@ AWS: "East Coast"/"Virginia"/"US East" → US_EAST_1, "Ohio" → US_EAST_2, "Cal
 GCP: "Central US" → CENTRAL_US, "Western US" → WESTERN_US, "Southeast Asia"/"APAC" → SOUTHEASTERN_ASIA_PACIFIC, "Europe"/"EU" → WESTERN_EUROPE.
 AZURE: "East US" → US_EAST_2, "West US" → US_WEST_2, "Europe"/"EU" → EUROPE_NORTH.`;
 
+// Adds M100/M140/M200 since getMaxAutoScalingSize can return those for M60/M80, and maxInstanceSize accepts them directly.
+const INSTANCE_SIZE_ORDER = [...standardInstanceSizeEnum.options, "M100", "M140", "M200"];
```
manupedrozo: I believe M100 is deprecated.

src/tools/atlas/update/upgradeCluster.ts
```diff
@@ -104,12 +185,17 @@ async function resolveClusterInfo(
...
         return {
             instanceType: cluster.instanceType,
-            provider: argOverrides.provider ?? firstRegionConfig?.backingProviderName,
-            region: argOverrides.region ?? firstRegionConfig?.regionName,
+            provider: argOverrides.provider ?? cluster.provider,
+            region: argOverrides.region ?? cluster.region,
+            raw,
+            instanceSize: cluster.instanceSize,
+            autoScaling: { compute: firstRegionConfig?.autoScaling?.compute },
         };
```
manupedrozo: Better now, possible improvement: 
We are only reading the first region here to return an `autoScaling` field for the scaling case. 
But then we still need to read from `raw` and check other regions before using the `autoScaling` field (lines 404-406). 
I think it'd be better to not return `autoScaling` here and instead resolve it in the core logic after the checks.

src/tools/atlas/update/upgradeCluster.ts
```diff
@@ -129,37 +215,112 @@ async function resolveClusterInfo(
...
+        minInstanceSize: standardInstanceSizeEnum
             .optional()
-            .describe("Target tier to upgrade to. Defaults to FLEX for Free clusters and M10 for Flex clusters."),
+            .describe(
+                "Minimum instance size (M10-M80) for compute autoscaling, for a Dedicated cluster or a Free/Flex-to-M10 upgrade. Omit unless explicitly specified by the user."
+            ),
+        maxInstanceSize: maxAutoScalingSizeEnum
+            .optional()
+            .describe(
+                "Maximum instance size (M10-M200) for compute autoscaling, for a Dedicated cluster or a Free/Flex-to-M10 upgrade. Sizes above M80 (M100/M140/M200) are only reachable via autoscaling, not as a direct target size. Omit unless explicitly specified by the user."
```
manupedrozo: Not sure if the extra context is required.
```suggestion
                "Maximum instance size (M10-M200) for compute autoscaling, for a Dedicated cluster or a Free/Flex-to-M10 upgrade. Omit unless explicitly specified by the user."
```

src/tools/atlas/update/upgradeCluster.ts
```diff
@@ -129,37 +215,112 @@ async function resolveClusterInfo(
...
     resolvedProvider: z.string().optional(),
     resolvedRegion: z.string().optional(),
     clusterId: z.string().optional(),
 };
 
 export class UpgradeClusterTool extends AtlasToolBase {
     static toolName = "atlas-upgrade-cluster";
-    public description = `Upgrade a MongoDB Atlas cluster tier. Upgrades Free (M0) clusters to Flex or M10 Dedicated, or Flex clusters to M10 Dedicated. The upgrade path is determined automatically from the current tier unless overridden with targetTier. Note to LLM: If provider and region are not already known, ask for both together in a single question before calling this tool. ${REGION_RECOMMENDATIONS}`;
+    public description =
+        "Upgrade or scale a MongoDB Atlas cluster: upgrades Free/Flex clusters to Flex or M10 Dedicated, and scales a Dedicated cluster's instance size and supports autoscaling configuration. " +
```
manupedrozo: Maybe:
```suggestion
        "Upgrade or scale a MongoDB Atlas cluster. Free and Flex clusters can be upgraded to Flex or M10 Dedicated. Dedicated clusters can be scaled to a different instance size, and compute autoscaling settings can be updated. " +
```

src/tools/atlas/update/upgradeCluster.ts
```diff
@@ -129,37 +215,112 @@ async function resolveClusterInfo(
...
+            .optional()
+            .describe(
+                "For a Free/Flex source cluster: the target tier to upgrade to, defaults to FLEX for Free clusters, M10 for Flex clusters. " +
+                    "For a Dedicated cluster: the new instance size (M10-M80) to scale it to."
+            ),
+        computeAutoScaling: z
+            .boolean()
+            .optional()
+            .describe(
+                "When true, enables compute autoscaling; when false, disables it. Applies to a Dedicated cluster, or to a Free/Flex cluster upgrading to M10 Dedicated (Flex clusters do not support autoscaling). Omit unless explicitly specified by the user."
```
manupedrozo: "When true, enables compute autoscaling; when false, disables it." - this seems too verbose. Maybe "Enable/Disable compute autoscaling". 
I think we can also get away without the "(Flex clusters do not support autoscaling)". Would try to be consistent with the min/max descriptions.

src/tools/atlas/update/upgradeCluster.ts
```diff
@@ -129,37 +215,112 @@ async function resolveClusterInfo(
...
+            .describe(
+                "Maximum instance size (M10-M200) for compute autoscaling, for a Dedicated cluster or a Free/Flex-to-M10 upgrade. Sizes above M80 (M100/M140/M200) are only reachable via autoscaling, not as a direct target size. Omit unless explicitly specified by the user."
+            ),
         provider: z
             .string()
             .regex(ALLOWED_PROVIDER_REGEX, "Provider must be uppercase letters and underscores only")
             .optional()
-            .describe("Cloud provider (e.g. AWS, GCP, AZURE). If omitted, the existing value is preserved."),
+            .describe(
+                "Cloud provider (e.g. AWS, GCP, AZURE) for a Free/Flex source cluster. Preserves the existing value if omitted. Does not apply once a cluster is already Dedicated."
```
manupedrozo:
```suggestion
                "Cloud provider (e.g. AWS, GCP, AZURE) for a Free/Flex source cluster. Preserves the existing value if omitted. Does not apply if a cluster is already Dedicated."
```

src/tools/atlas/update/upgradeCluster.ts
```diff
@@ -129,37 +215,112 @@ async function resolveClusterInfo(
...
+            ),
         region: AtlasArgs.region()
             .optional()
             .describe(
-                "Cloud provider region in Atlas format using uppercase letters and underscores (e.g. US_EAST_1). If omitted, the existing value is preserved."
+                "Cloud provider region in Atlas format using uppercase letters and underscores (e.g. US_EAST_1) for a Free/Flex source cluster. Preserves the existing value if omitted. Does not apply once a cluster is already Dedicated."
```
manupedrozo:
```suggestion
                "Cloud provider region in Atlas format using uppercase letters and underscores (e.g. US_EAST_1) for a Free/Flex source cluster. Preserves the existing value if omitted. Does not apply if a cluster is already Dedicated."
```

src/tools/atlas/update/upgradeCluster.ts
```diff
@@ -177,44 +338,238 @@ export class UpgradeClusterTool extends AtlasToolBase {
             context
         );
 
-        const target = args.targetTier ?? (clusterInfo.instanceType === "FREE" ? "FLEX" : "M10");
+        if (clusterInfo.instanceType === "DEDICATED") {
+            if (args.targetTier === "FLEX") {
```
manupedrozo: Should we have this check in the `switch (clusterInfo.instanceType)` `DEDICATED` case?

Similar for the other checks below, should we move them to their respective cases?

src/tools/atlas/update/upgradeCluster.ts
```diff
@@ -177,44 +338,238 @@ export class UpgradeClusterTool extends AtlasToolBase {
...
+                    `Cluster "${clusterName}" is already Dedicated. targetTier must be an instance size (M10-M80) to scale it in place, not FLEX.`
+                );
+            }
+        } else if (args.targetTier !== undefined && args.targetTier !== "FLEX" && args.targetTier !== "M10") {
+            throw new UpgradeClusterError(
+                `targetTier "${args.targetTier}" is not valid when upgrading from ${clusterInfo.instanceType}. Choose FLEX or M10 - larger instance sizes are only valid once the cluster is already Dedicated.`
+            );
+        }
+
+        const target =
```
manupedrozo: This `target` that is only used for the FREE/FLEX cases. Was confused when reading this on why we were not accounting for other sizes. Similar to the comment above, I think we need to restructure a bit.

src/tools/atlas/update/upgradeCluster.ts
```diff
@@ -177,44 +338,238 @@ export class UpgradeClusterTool extends AtlasToolBase {
...
+                    );
+                }
+                const regionElectableSizes = (clusterInfo.raw?.replicationSpecs?.[0]?.regionConfigs ?? [])
+                    .map((rc) => (rc as { electableSpecs?: { instanceSize?: string } }).electableSpecs?.instanceSize)
+                    .filter((size): size is string => size !== undefined);
+                if (new Set(regionElectableSizes).size > 1) {
+                    throw new UpgradeClusterError(
+                        `Cluster "${clusterName}" has regions with different instance sizes, which this tool does not support scaling to avoid unintended changes.`
+                    );
+                }
```
manupedrozo: Consider moving all of these validations to a function.

src/tools/atlas/update/upgradeCluster.ts
```diff
@@ -177,44 +338,238 @@ export class UpgradeClusterTool extends AtlasToolBase {
...
+
+                const isResizing = args.targetTier !== undefined;
+
+                // Target size: an explicit resize, or unchanged if only autoscaling is being adjusted.
+                let newSize: StandardInstanceSize;
```
manupedrozo: A better name for these could be using a `resolved` prefix instead of `new` since we keep the existing ones if none is passed. 
E.g. `resolvedInstanceSize`, `resolvedMin`, `resolvedMax`, etc.

src/tools/atlas/update/upgradeCluster.ts
```diff
@@ -177,44 +338,238 @@ export class UpgradeClusterTool extends AtlasToolBase {
...
+                }
+
+                // Autoscaling max bound.
+                let newMax: string | undefined;
+                if (args.maxInstanceSize !== undefined) {
+                    // Explicit override always wins.
+                    newMax = args.maxInstanceSize;
+                } else if (!newEnabled) {
+                    // Autoscaling disabled: no max needed.
+                    newMax = undefined;
```
manupedrozo: Curious that this one is different in the `min` resolution logic. Should we do the same there?

src/tools/atlas/update/upgradeCluster.ts
```diff
@@ -177,44 +338,238 @@ export class UpgradeClusterTool extends AtlasToolBase {
...
                     clusterInfo.provider,
                     clusterInfo.region,
+                    args,
                     context
                 ));
                 break;
         }
 
+        if (clusterInfo.instanceType === "DEDICATED") {
+            const isScaling = targetInstanceSize !== clusterInfo.instanceSize;
```
manupedrozo: Small, just for consistency, we use isResizing and isScaling for the same, can we use just 1 name for consistency?
Also, it seems like we can reuse the variable like we do for `clusterId`, `targetInstanceSize`, etc.

tests/accuracy/upgradeCluster.test.ts
```diff
@@ -149,4 +160,100 @@ describeAccuracyTests([
             },
         ],
     },
+    {
```
manupedrozo, line 206: Can we add cases with scaling + autoscaling? With and without explicit min/max

tests/unit/common/cluster.test.ts
```diff
@@ -25,3 +25,16 @@ describe("inspectCluster", () => {
         );
     });
 });
+
+describe("isStandardInstanceSize", () => {
```
manupedrozo: I would personally remove these, up to you!

src/telemetry/types.ts
laylaazouz: Added min_instance_size/max_instance_size to UpgradeClusterMetadata

src/telemetry/types.ts
laylaazouz: Switched to plain string for both, dropped the StandardInstanceSize import

src/tools/atlas/update/upgradeCluster.ts
laylaazouz: Removed M100, kept M140/M200 like before.

src/tools/atlas/update/upgradeCluster.ts
laylaazouz: Moved this t's now resolved from inside the dedicated case, after the region checks.

src/tools/atlas/update/upgradeCluster.ts
laylaazouz: Shortened this and removed flex note

tests/unit/common/cluster.test.ts
laylaazouz: Left these in for now, happy to remove if you prefer that!

src/tools/atlas/update/upgradeCluster.ts
laylaazouz: Yep sorry, added the same to min.

src/tools/atlas/update/upgradeCluster.ts
laylaazouz: Renamed!

src/tools/atlas/update/upgradeCluster.ts
laylaazouz: Done moved to validateDedicatedScaling().

src/tools/atlas/update/upgradeCluster.ts
laylaazouz: Restructured, target resolved inside the FREE/FLEX cases only

src/tools/atlas/update/upgradeCluster.ts
laylaazouz: Moved into the DEDICATED case

src/tools/atlas/update/upgradeCluster.ts
laylaazouz: Consolidated into `isScaling` variable

src/tools/atlas/update/upgradeCluster.ts
```diff
@@ -129,37 +208,173 @@ async function resolveClusterInfo(
...
+        throw new UpgradeClusterError(
+            `No changes specified for Dedicated cluster "${clusterName}". Provide targetTier (new instance size), computeAutoScaling, minInstanceSize, and/or maxInstanceSize to scale it.`
+        );
+    }
+    if (clusterInfo.instanceSize === undefined || !isStandardInstanceSize(clusterInfo.instanceSize)) {
+        throw new UpgradeClusterError(
+            `Cluster "${clusterName}" has instance size "${clusterInfo.instanceSize ?? "unknown"}", which this tool does not support scaling. Only standard M10-M80 instance sizes are supported.`
+        );
+    }
+    if (clusterInfo.raw?.clusterType === "GEOSHARDED" || (clusterInfo.raw?.replicationSpecs?.length ?? 0) > 1) {
```
manupedrozo: Small: We should check for supported values (REPLICASET/SHARDED) instead of unsupported values. If a new clusterType is added tomorrow, we don't want to have to come back here and add a new check.

src/tools/atlas/update/upgradeCluster.ts
```diff
@@ -177,49 +392,215 @@ export class UpgradeClusterTool extends AtlasToolBase {
...
         switch (clusterInfo.instanceType) {
-            case "DEDICATED":
-                throw new UpgradeClusterError(
-                    `Cluster "${clusterName}" is already at the Dedicated tier and cannot be upgraded further.`
+            case "DEDICATED": {
+                if (args.targetTier === "FLEX") {
+                    throw new UpgradeClusterError(
+                        `Cluster "${clusterName}" is already Dedicated. targetTier must be an instance size (M10-M80) to scale it in place, not FLEX.`
+                    );
+                }
```
manupedrozo: Can move inside `validateDedicatedScaling()`

src/tools/atlas/update/upgradeCluster.ts
```diff
@@ -177,49 +392,215 @@ export class UpgradeClusterTool extends AtlasToolBase {
...
+
+                validateDedicatedScaling(clusterInfo, args, clusterName);
+
+                isScaling = args.targetTier !== undefined;
```
manupedrozo: What happens when the target is the same as the current instance size?

src/tools/atlas/update/upgradeCluster.ts
```diff
@@ -177,49 +392,215 @@ export class UpgradeClusterTool extends AtlasToolBase {
...
+
+                // Current autoscaling settings, read from the cluster's single region config (already
+                // validated to be the only one, and consistent across regions, by validateDedicatedScaling).
```
manupedrozo: > single region config (already validated to be the only one

Would adjust, single **replication spec** with possibly multiple regions.

src/tools/atlas/update/upgradeCluster.ts
```diff
@@ -129,37 +208,173 @@ async function resolveClusterInfo(
...
+            `Cluster "${clusterName}" has instance size "${clusterInfo.instanceSize ?? "unknown"}", which this tool does not support scaling. Only standard M10-M80 instance sizes are supported.`
+        );
+    }
+    if (clusterInfo.raw?.clusterType === "GEOSHARDED" || (clusterInfo.raw?.replicationSpecs?.length ?? 0) > 1) {
+        throw new UpgradeClusterError(
+            `Cluster "${clusterName}" is a ${clusterInfo.raw?.clusterType ?? "multi-shard"} cluster, which this tool does not support scaling. Only single-shard clusters (REPLICASET, or SHARDED with one shard) are supported.`
+        );
+    }
+    const regionElectableSizes = (clusterInfo.raw?.replicationSpecs?.[0]?.regionConfigs ?? [])
+        .map((rc) => (rc as { electableSpecs?: { instanceSize?: string } }).electableSpecs?.instanceSize)
```
manupedrozo: Do we need to account for `readOnly` instance sizes here as well? Which we do scale if present.

tests/unit/tools/atlas/update/upgradeCluster.test.ts
```diff
@@ -54,6 +55,83 @@ const DEDICATED_CLUSTER_RAW = {
...
+        },
+    ],
+};
+
+const DEDICATED_CLUSTER_MULTI_REGION_RAW = {
+    id: "dedicated-cluster-id",
+    clusterType: "REPLICASET",
+    replicationSpecs: [
+        {
+            regionConfigs: [
```
manupedrozo, line 82: Would like regions that have more than 1 of electable / readonly / analytics specs. Want to make sure that those are handled correctly. E.g a region with electable + analytics, a region with all 3, etc.

src/tools/atlas/update/upgradeCluster.ts
laylaazouz: good call, changed this to check supported values

src/tools/atlas/update/upgradeCluster.ts
laylaazouz: added a check for this!

src/tools/atlas/update/upgradeCluster.ts
laylaazouz: added the same consistency check for readonly

src/common/atlas/cluster.ts
```diff
@@ -8,6 +9,23 @@ import { requestIdAttr } from "../../helpers/requestIdAttr.js";
 import { LogId } from "../logging/index.js";
 import { ConnectionString } from "mongodb-connection-string-url";
 
+// enum is capped at M80 because that's as far as createCluster
+// and upgradeCluster currently support creating/scaling a cluster directly.
+export const standardInstanceSizeEnum = z.enum(["M10", "M20", "M30", "M40", "M50", "M60", "M80"]);
+
+export type StandardInstanceSize = z.infer<typeof standardInstanceSizeEnum>;
+
+export function isStandardInstanceSize(size: string): size is StandardInstanceSize {
```
gagik: `standardInstanceSizeEnum.parse` would be better

src/common/atlas/cluster.ts
```diff
@@ -8,6 +9,23 @@ import { requestIdAttr } from "../../helpers/requestIdAttr.js";
...
+    return (standardInstanceSizeEnum.options as readonly string[]).includes(size);
+}
+
+// M60 and M80 extend beyond the selectable range. M140 is not supported on Azure.
+export function getMaxAutoScalingSize(size: StandardInstanceSize, provider: string): string {
```
gagik, line 19: this can be a zod validation rule

src/tools/atlas/update/upgradeCluster.ts
```diff
@@ -46,26 +68,79 @@ type FlexToM10Body = {
...
 
-function buildM10UpgradeBody(baseTier: "FREE", clusterName: string, provider?: string, region?: string): FreeToM10Body;
-function buildM10UpgradeBody(baseTier: "FLEX", clusterName: string, provider?: string, region?: string): FlexToM10Body;
+type AutoScalingArgs = { computeAutoScaling?: boolean; minInstanceSize?: string; maxInstanceSize?: string };
+
+type ResolvedM10AutoScaling = { enabled: boolean; minInstanceSize?: string; maxInstanceSize?: string };
+
+// Validates and resolves the autoscaling args for a Free/Flex to M10 upgrade, where the target size is fixed at M10.
+function resolveM10AutoScaling(autoScalingArgs: AutoScalingArgs, provider: string | undefined): ResolvedM10AutoScaling {
+    if (autoScalingArgs.minInstanceSize !== undefined && autoScalingArgs.minInstanceSize !== "M10") {
```
gagik, line 89: not sure it'll cover all the cases here but some of the validation can be done cleaner through a zod schema

src/tools/atlas/update/upgradeCluster.ts
```diff
@@ -129,37 +208,189 @@ async function resolveClusterInfo(
...
+    region?: string;
+};
+
+// Validates that a DEDICATED cluster can be safely scaled.
+function validateDedicatedScaling(
+    clusterInfo: ResolvedClusterInfo,
+    args: DedicatedScalingArgs,
+    clusterName: string
+): void {
+    if (args.targetTier === "FLEX") {
```
gagik, line 276: same here, might be possible to define this through an enum of different Zed schemas. but might be too complicated case also; just a suggestion

src/tools/atlas/update/upgradeCluster.ts
```diff
@@ -177,49 +408,210 @@ export class UpgradeClusterTool extends AtlasToolBase {
...
         switch (clusterInfo.instanceType) {
-            case "DEDICATED":
-                throw new UpgradeClusterError(
-                    `Cluster "${clusterName}" is already at the Dedicated tier and cannot be upgraded further.`
+            case "DEDICATED": {
+                validateDedicatedScaling(clusterInfo, args, clusterName);
+
+                // Current autoscaling settings, read from the first region of the cluster's single replication
+                // spec (validateDedicatedScaling already confirmed there's only one).
+                const firstRegionConfig = clusterInfo.raw?.replicationSpecs?.[0]?.regionConfigs?.[0] as
```
gagik: can we avoid this casting?

maybe the validates clusterInfo should return this

gagik (approved):
A bit hard to parse the validations and sizing logic but nothing blocking; just some nits and suggestions to use Zod if possible to minimize boilerplate.

src/common/atlas/cluster.ts
laylaazouz: went with safeParse instead. This function only needs to return true or false so that the caller can throw its own specific error (parse would throw in the function on an invalid value).

src/common/atlas/cluster.ts
laylaazouz: Moved to zod schema, thanks!

src/tools/atlas/update/upgradeCluster.ts
laylaazouz: Changed to zod schema here as well instead of an inline !== check

src/tools/atlas/update/upgradeCluster.ts
laylaazouz: Fixed, `validateDedicatedScaling` already parses the region config while validating, so it now returns the current `autoScaling.compute` state instead of void.

src/tools/atlas/update/upgradeCluster.ts
laylaazouz: Slightly different to the last case, I think the check is simpler if kept as-is!

src/common/atlas/cluster.ts
manupedrozo: Nice! We can consider removing this function in favor of the inline `safeParse()`.

src/common/atlas/cluster.ts
manupedrozo: Not sure if this is was what @gagik had in mind here (guessing it wasn't).
I would keep the previous version, no value in wrapping this logic into a zod parse() that does the same transformation.

src/tools/atlas/update/upgradeCluster.ts
```diff
@@ -46,26 +68,82 @@ type FlexToM10Body = {
...
+type ResolvedM10AutoScaling = { enabled: boolean; minInstanceSize?: string; maxInstanceSize?: string };
+
+// The target size for a Free/Flex to M10 upgrade is fixed at M10, so minInstanceSize can only be M10 or omitted.
+const m10MinInstanceSizeSchema = z.literal("M10").optional();
+
+// Validates and resolves the autoscaling args for a Free/Flex to M10 upgrade, where the target size is fixed at M10.
+function resolveM10AutoScaling(autoScalingArgs: AutoScalingArgs, provider: string | undefined): ResolvedM10AutoScaling {
+    if (!m10MinInstanceSizeSchema.safeParse(autoScalingArgs.minInstanceSize).success) {
+        throw new UpgradeClusterError(`minInstanceSize must be omitted or "M10" when upgrading to M10 Dedicated.`);
+    }
```
manupedrozo: Similar to my other comment, I don't see value in adding a zod schema + safeParse() to replace a simple if.

I am guessing that what @gagik had in mind was something like:
- Define complete zod schemas for each case (free -> flex, flex -> m10, m10 -> mXX)
- Then once we know the case we run the validation.

This approach would potentially help clarify the validation rules and reduce verbosity.

I'd say we roll back these changes for now and consider a follow-up where we move per-case validations to zod schemas.

src/common/atlas/cluster.ts
laylaazouz: reverted this back to plain if/else

src/tools/atlas/update/upgradeCluster.ts
laylaazouz: Reverted this to a plain check too, can look at that as a separate follow-up rather than as part of this PR if that's preferred.

src/tools/atlas/update/upgradeCluster.ts
```diff
@@ -297,7 +293,10 @@ function validateDedicatedScaling(
             `No changes specified for Dedicated cluster "${clusterName}". Provide targetTier (new instance size), computeAutoScaling, minInstanceSize, and/or maxInstanceSize to scale it.`
         );
     }
-    if (clusterInfo.instanceSize === undefined || !isStandardInstanceSize(clusterInfo.instanceSize)) {
+    if (
+        clusterInfo.instanceSize === undefined ||
```
manupedrozo: Small: I believe `safeParse()` would also fail for an undefined input, so we could remove this undefined check.

tests/unit/common/cluster.test.ts
```diff
@@ -26,15 +26,15 @@ describe("inspectCluster", () => {
     });
 });
 
-describe("isStandardInstanceSize", () => {
+describe("standardInstanceSizeEnum", () => {
```
manupedrozo: Now I would definitely remove this one, we are essentially testing `safeParse()`

manupedrozo (approved):
Good job iterating on this PR!
Left a couple extra non-blocking comments.

src/common/atlas/cluster.ts
gagik: I did mean it as a Zod schema; Zod schema lets us define an expected shape of data so it's easier to see what we're validating. The value to me is readability in what we're trying to valdiate here.

I generally think Zod schemas are much easier to read and maintain then these branches of if statements.

src/common/atlas/cluster.ts
gagik: safeParse is good yeah!

tests/unit/common/cluster.test.ts
laylaazouz: removed this test!

src/tools/atlas/update/upgradeCluster.ts
laylaazouz: good point, removed the undefined check here. thanks!

src/tools/atlas/update/upgradeCluster.ts
gagik: I think it's fine we just don't need these wrapper functions for safeparse. if it's going to throw, we can throw with Zod and customize the error: https://zod.dev/error-customization

src/common/atlas/cluster.ts
gagik: but in this specific case yeah I guess a schema doesn't fit the usecase

tests/unit/common/cluster.test.ts
gagik: it seems to have persisted

src/common/atlas/cluster.ts
manupedrozo: Agree with the idea. However I would not extract every `if` statement into a different schema, it adds even more verbosity. I would instead capture per-case validations in separate schemas.
E.g. `scaleClusterValidationSchema` would validate that provider & regions are unset, one of `targetTier`/`computeAutoScaling`/`minInstanceSize`/`maxInstanceSize` to be set, `targetTier` one of the supported ones, etc.
Then we'd run a single `parse()`/`safeParse()` for the entire schema.

tests/unit/common/cluster.test.ts
laylaazouz: just double checked and can't seem to find it, mind checking again or refreshing and letting me know if it's still there?

src/tools/atlas/update/upgradeCluster.ts
laylaazouz: If we move to using Zod schemas for all validation then we can do this, but since every other validation here throws `UpgradeClusterError`, I think for consistency this should stay as-is for now?

PR #1428, opened 2026-08-14
docs: add remote mcp, skills and plugins instructions

gagik:
Add a clear quick setup section before all the other text which provides quick access to our plugins, skills, and the remote MCP server.

Removes the deprecated HTTP server.
Moves platform-specific instructions (Copilot, OpenCode) to a separate file and adds all missing ones. Updates Windsurf to Devin Desktop


README.md
```diff
@@ -244,93 +313,6 @@ With Atlas API credentials:
 }
 ```
 
-#### Option 5: Running as an HTTP Server
-
-> **⚠️ Security Notice:** This server now supports Streamable HTTP transport for remote connections. **HTTP transport is NOT recommended for production use without implementing proper authentication and security measures.**
```
gagik, line 249: pretty sure HTTP in general is being removed in the future and it doesn't make sense to advertise this

MCP_CLIENT_CONFIGURATION.md
```diff
@@ -0,0 +1,264 @@
...
+      "args": ["-y", "mongodb-mcp-server@latest", "--readOnly"],
+      "env": {
+        "MDB_MCP_CONNECTION_STRING": "mongodb://localhost:27017/myDatabase"
+      }
+    }
+  }
+}
+```
+
+## OpenCode
```
kevinmdb: @gagik Open Code won't work without service accounts currently.

mongodben (commented):
following on our [slack convo](https://mongodb.slack.com/archives/C08LS3TG2BC/p1786716741826659), some high lvl feedback: 


1. i rec including the 'manual' path to mcp config in MCP_CLIENT_CONFIGURATION.md:

```json
{
  "mcpServers": {
     "mongodb-atlas-mcp-remote": {
       "type": "stdio",
       "command": "npx",
       "args": ["-y", "mongodb-atlas-mcp-remote@latest"],
       "env": {
         "MDB_MCP_API_CLIENT_ID": "$CLIENT_ID",
         "MDB_MCP_API_CLIENT_SECRET": "$SECRET"
       }
     }
   }
}
```

2. i'm not sure if it's strictly necessary, but it might be worth adding a note on why you can't configure direct HTTP connection to the atlas mcp, namely b/c there's no static api key auth. 

  this differed from my expectation having configured other MCPs w/ a static API key in this way.

  I think you could accomplish this with something like:

  > You cannot authenticate to the remote MongoDB MCP server using a static API key and HTTP. You must either:
  > 1. Use a client that suppports the OAuth flow
  > 2. Use the `stdio` server `mongodb-atlas-mcp-remote`, which you can authenticate into using static `MDB_MCP_API_CLIENT_ID` and `MDB_MCP_API_CLIENT_SECRET` environment variables

gagik:
@mongodben 

re: 1. I agree, unfortunately for an arbitrary client right now they need to be part of a whitelisted list of client IDs to actually use OAuth so I'm careful to advertise it too much at the moment.

re: 2. I'll add a mention

MCP_CLIENT_CONFIGURATION.md
```diff
@@ -0,0 +1,235 @@
+# MCP Client Configurations
```
blva: why not linking to our public docs where this is maintained?

gagik: good point, it diverted too much originally but now it's essentially the same. will adjust

gagik (commented):
@mongodben 
Added the note explaining that the remote MongoDB MCP server can't be authenticated with a static API key over HTTP (OAuth or the stdio `mongodb-atlas-mcp-remote` server instead), and added the `mongodb-atlas-mcp-remote` stdio config example. We also moved the client configs to link to the public docs at https://www.mongodb.com/docs/mcp-server/get-started/

lerouxb (approved):

PR #1435, opened 2026-08-20
feat: add opt-in GCF output format for tool result data

blackwell-systems:
### Summary

Adds an opt-in output format, `MDB_MCP_OUTPUT_FORMAT=gcf` (or `--outputFormat gcf`), that
serializes tool result documents as [Graph Compact Format](https://gcformat.com) instead of
JSON. The record arrays these tools return (`find` / `aggregate` results, the Atlas `list-*`
tools, collection/index listings) have their repeated field names factored into a single GCF
header, so the response uses fewer tokens when it crosses the LLM boundary.

It hooks a single shared seam: `formatUntrustedData` in `src/tools/tool.ts`, where every tool
wraps its result data. So it covers every data-returning tool with no per-tool changes, and the
GCF wire sits **inside the same untrusted-data security wrapping** as the JSON would.

### Measured

On representative `find` result shapes (row-object documents), o200k tokenizer, lossless
round-trip verified: **~38–41% fewer tokens** than the compact JSON, scaling with document
count.

### Deliberately conservative: it can only ever help

- **Opt-in and additive.** Default output is JSON, unchanged unless the setting is enabled.
- **Never larger.** If the GCF wire is not smaller than the JSON for a given result, the JSON is
  kept (a never-grow guard).
- **Never lossy.** The wire is decoded and compared back to the source value with an
  order-insensitive, Map-aware check; on any parse, encode, or mismatch the JSON is kept. So a
  result is never grown or altered.
- **`structuredContent` untouched.** Only the model-facing text data is re-encoded; the typed
  `structuredContent` a tool sets stays JSON. The two carry equivalent information in different
  representations, per the project's content/structuredContent guidance.
- **Zero new footprint.** `@blackwell-systems/gcf` is a zero-dependency package, pinned exact.

### Why GCF, beyond size

The harder question than size is whether a model reads the compact form as accurately as JSON.
GCF is designed and evaluated for that:

- **Generic-profile comprehension: 100% on every frontier model** (Claude Opus/Sonnet/Haiku,
  GPT-5.5, Gemini 2.5 Pro / 3.1 Pro / 3.5 Flash), 27 runs, 11 models, 4 providers; equal to or
  better than JSON on every model. https://gcformat.com/guide/benchmarks
- The grammar was reverse-engineered from attention-level research into how BPE tokenization
  shapes what a transformer can represent, then validated by controlled from-scratch training at
  1.3B scale, measured causally rather than by an LLM judge. Four papers (2026, under review):
  - [GCF: A Token-Optimized Wire Format for Structured LLM Interactions](https://doi.org/10.5281/zenodo.20579817)
  - [Tokenizer-Attention Coupling](https://doi.org/10.5281/zenodo.20925910)
  - [Stranded Attention](https://doi.org/10.5281/zenodo.21158886)
  - [Developmental Atlas of Attention Head Specialization](https://doi.org/10.5281/zenodo.21205389)

GCF is already in production across observability, developer tooling, and network automation:
[Chrome DevTools MCP](https://github.com/ChromeDevTools/chrome-devtools-mcp) (Google) merged it
as an experimental format; [Speakeasy](https://speakeasy.com) (customers include Google, Verizon,
Mistral AI) ships it in their `oq` CLI; [OmniRoute](https://omniroute.online) vendored it into its
gateway; [NetClaw](https://github.com/automateyournetwork/netclaw) benchmarked it against TOON on
real network data and replaced TOON; and many others across the MCP and agent ecosystem.

### The change

- `MDB_MCP_OUTPUT_FORMAT` / `--outputFormat` config option (`json` default), documented in the
  config schema and README.
- GCF re-encode (never-grow + lossless guards) in `formatUntrustedData`.
- `@blackwell-systems/gcf` dependency.
- Unit tests: the encode + lossless round-trip, the never-grow and non-JSON fallbacks, and the
  default-JSON behavior. `pnpm run check` green (build, types, lint, format, dependencies, api).

GCF is open source and MIT-licensed. Format, spec, SDKs, and benchmarks: https://gcformat.com
Happy to adjust the setting name or extend the scope; reproducible numbers and harness on request.


gagik:
We may revisit in the future but not looking to adopt alternative formats at this time.

PR #1450, opened 2026-08-24
fix: propagate connection error from db tools MCP-627

### LINKED ISSUE #1448: [Bug]: list-databases reports "connection string is not valid" for Network Access failures, discarding the driver's accurate error
### Version

v2.1.0

### App

- [ ] Cursor
- [ ] Windsurf
- [ ] VSCode
- [ ] VSCode Insiders
- [ ] Claude Desktop
- [x] Other

- [ ] Claude 3.5 Sonnet
- [ ] Claude 3.7 Sonnet
- [ ] GPT-4a
- [ ] o4-mini
- [x] Other

## Summary

When an Atlas connection fails because the client IP is not in the cluster's Network Access list, `list-databases` returns:

> The configured connection string is not valid. Please check the connection string and confirm it points to a valid MongoDB instance.

The connection string is valid. The server already knows the real cause — it is stored in `lastError` on the connection record and returned correctly by `list-connections`:

```json
{"connectionId":"preconfigured","state":"errored",
"lastError":"Client network socket disconnected before secure TLS connection was established. It looks like this is a MongoDB Atlas cluster. Please ensure that your Network Access List allows connections from your IP."}
```

That message is accurate and actionable. The one `list-databases` returns is neither, and it is the one the agent acts on. This is most likely to hit during first-time setup, when the connection string is correct but Network Access has not been configured yet.

<details>
<summary>Full <code>list-databases</code> output</summary>

```
The configured connection string is not valid. Please check the connection string and confirm
it points to a valid MongoDB instance.
Alternatively, you can use one of the following tools: "atlas-local-connect-deployment",
"connect" to connect to a MongoDB instance. Note to LLM: For MongoDB Atlas Local deployments,
ask the user to either provide a connection string, specify a deployment name, or use
"atlas-local-list-deployments" to show available local deployments. If a deployment name is
provided, prefer using the "atlas-local-connect-deployment" tool. If a connection string is
provided, prefer using the "connect" tool. Do not invent deployment names or connection
strings unless the user has explicitly specified them. [...]
```

</details>

## Reproduction

1. Configure the server with a valid `MDB_MCP_CONNECTION_STRING` for an Atlas cluster.
2. Ensure your current IP is not in the cluster's **Network Access** list (either never added during setup, or changed since).
3. Start a **new** client session (an existing session holds an open pool and keeps working).
4. Call `list-databases` — generic message. Call `list-connections` — correct `lastError`.
Restoring the IP makes `list-databases` succeed immediately. Nothing else changed.

## Expected

`list-databases` should surface `lastError` rather than asserting a cause it has not established:

> Could not connect to MongoDB. Last error: Client network socket disconnected before secure TLS connection was established. It looks like this is a MongoDB Atlas cluster. Please ensure that your Network Access List allows connections from your IP.

Reserve "the connection string is not valid" for strings that actually fail to parse.

## Why this matters in agent contexts

I ran the same failure twice in two clean directories, differing only by whether MongoDB's `mongodb-mcp-setup` skill was present.

**Without the skill**, the agent accepted the message, told the user their connection string was invalid, followed the tool's suggestion into `atlas-local-list-deployments` (raw Docker error, on a machine not using Atlas Local), asked the user to paste a connection string — which for Atlas carries credentials — advised editing the working MCP config, and never mentioned Network Access.

**With the skill**, it called `list-connections`, found `lastError`, and gave the correct fix in about four steps.

Related: #1402 (tools confusing agents) and #1269 (connect tool reporting misleading connection status).

## Environment

- mongodb-mcp-server: 2.1.0
- Client: Claude Code
- Cluster: MongoDB Atlas, `mongodb+srv://`, SCRAM auth
- OS: Windows 11
- Both runs used `--readOnly`; I have not tested whether the behaviour differs without it.

nirinchev:
## Proposed changes

Fixes https://github.com/mongodb-js/mongodb-mcp-server/issues/1448

### DIFF
--- a/packages/tools-mongodb/src/connectionErrorHandler.ts
+++ b/packages/tools-mongodb/src/connectionErrorHandler.ts
@@ -118,7 +118,7 @@ export const connectionErrorHandler: ConnectionErrorHandler = (error, { availabl
                     content: [
                         {
                             type: "text",
-                            text: "The configured connection string is not valid. Please check the connection string and confirm it points to a valid MongoDB instance.",
+                            text: `Could not connect to MongoDB. Last error: ${error.message}`,
                         },
                         {
                             type: "text",
--- a/packages/integration-tests/src/tools/mongodb/connect/connect.test.ts
+++ b/packages/integration-tests/src/tools/mongodb/connect/connect.test.ts
@@ -180,7 +180,7 @@ describeWithMongoDB("Connect tool", (integration) => {
                 arguments: { connectionString: "mangodb://localhost:12345" },
             });
             const content = getResponseContent(response.content);
-            expect(content).toContain("The configured connection string is not valid.");
+            expect(content).toContain("Could not connect to MongoDB.");
             expect(response.structuredContent).toBeUndefined();
         });
     });
--- a/packages/integration-tests/src/tools/mongodb/mongodbTool.test.ts
+++ b/packages/integration-tests/src/tools/mongodb/mongodbTool.test.ts
@@ -255,9 +255,12 @@ describe("MongoDBTool implementations", () => {
                 });
                 expect(toolResponse?.isError).to.equal(true);
                 const content = toolResponse?.content as CallToolResult["content"];
-                expect(
-                    content.some((c) => c.type === "text" && c.text.startsWith("Could not connect to MongoDB."))
-                ).toBe(true);
+                const message = content.find(
+                    (c) => c.type === "text" && c.text.startsWith("Could not connect to MongoDB.")
+                );
+                expect(message?.type).toBe("text");
+                const messageText = message?.type === "text" ? message.text : "";
+                expect(messageText).toMatch(/^Could not connect to MongoDB\. Last error: .+/);
             });
         });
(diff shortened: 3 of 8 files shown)


packages/tools-mongodb/src/connectionErrorHandler.ts
```diff
@@ -118,7 +118,7 @@ export const connectionErrorHandler: ConnectionErrorHandler = (error, { availabl
                     content: [
                         {
                             type: "text",
-                            text: "The configured connection string is not valid. Please check the connection string and confirm it points to a valid MongoDB instance.",
+                            text: `Could not connect to MongoDB. Last error: ${error.message}`,
```
gagik, line 121: maybe we should redact here instead of at a level above?

nirinchev: We could, though that would mean we'd be passing an unredacted error to the error handler, which would mean each implementation would need to add redacting logic.

gagik (approved):

PR #1454, opened 2026-08-24
chore: verify tools and core api-extractor reports MCP-629

nirinchev:
## Proposed changes

Fixes a few issues with api-extractor reports:
1. `api-extractor/reports/tools.public.api.md` was generated but not verified by `check:api-report`.
2. `packages/core/api-extractor/reports/core.api.md` didn't have an api-extractor.json so nothing regenerated or verified it.

## Checklist


gagik (approved):
LGTM, later we should probably start switching to library-level reports, I largely kept it preserved for comparison with v1/v2 purposes

gagik:
I remember now why this was removed, it's because `tools.public.api.md` is terribly flaky across Node versions and OSes.... I'm going to disable it again but will follow-up with a more strict script which enforces same ordering...

PR #1459, opened 2026-08-25
chore: rework keychain API MCP-634

nirinchev:
## Proposed changes

Refactors the keychain API to encapsulate the redacting capabilities and avoid handing out secrets. This required some restructuring of the Logger base classes to more explicitly convey whether a logger redacts or not.


packages/core/src/logging/redactingLoggerBase.ts
```diff
@@ -0,0 +1,83 @@
...
+} from "@mongodb-js/mcp-types";
+import { LoggerBase } from "./loggerBase.js";
+
+/**
+ * Base class for loggers that emit a message somewhere a secret could escape to. The keychain is
+ * required rather than optional: every such logger redacts, so there is no valid way to construct
+ * one without the means to do it.
+ */
+export abstract class RedactingLoggerBase<T extends EventMap<T> = DefaultEventMap> extends LoggerBase<T> {
+    private readonly keychain: IKeychain;
```
addaleax: It's probably just fine in this context here, but conceptually it seems pretty clear that we'd only ever be interested in the `redact()` part of the keychain interface, and that using the full `IKeychain` interface is tying the logging interface to the keychain concept without the two actually being semantically related beyond this one piece.

Feel free to ignore, but I'd either use a separate interface or just do e.g. `Pick<IKeychain, 'redact'>` here

packages/core/src/logging/redactingLoggerBase.ts
```diff
@@ -0,0 +1,83 @@
...
+        noRedaction: LogPayload["noRedaction"]
+    ): Record<string, string> | undefined {
+        if (!attributes) {
+            return undefined;
+        }
+        const redacted: Record<string, string> = {};
+        for (const [key, value] of Object.entries(attributes)) {
+            redacted[key] = this.redactIfNecessary(value, noRedaction);
+        }
+        return redacted;
```
addaleax: Either:

```suggestion
        const redacted: Record<string, string> = Object.create(null);
        for (const [key, value] of Object.entries(attributes)) {
            redacted[key] = this.redactIfNecessary(value, noRedaction);
        }
        return redacted;
```

or:

```suggestion
        return Object.fromEntries(
            Object.entries(attributes).map(([key, value]) => [key, this.redactIfNecessary(value, noRedaction)])
        );
```

since `key` is arbitrary and could be e.g. `__proto__`.

addaleax: Also, I can see this matches https://github.com/mongodb-js/mongodb-mcp-server/blob/e884faaf595f3c60503c3c56265fc9e803004fa8/packages/mongodb-atlas-mcp-remote/src/logging/loggerBase.ts#L28 pretty much 1:1 – do we want the duplication? (And in any case, the comment above ^ applies there too)

packages/core/src/keychain.ts
```diff
@@ -33,29 +36,78 @@ export class Keychain implements IKeychain {
...
 /**
- * Recursively redacts registered secrets from every string value of a value, leaving the structure
- * intact. Redaction is applied per-value (not on serialized JSON) so it can never corrupt the
- * resulting JSON, regardless of what the redactor substitutes.
+ * Recursively redacts `secrets` from the strings in `value`.
+ *
+ * mongodb-redact's own `redact` only descends into plain objects and arrays, so a secret held in
+ * a class instance field would survive it. This walks non-plain objects too, but rebuilds them on
+ * their original prototype and only when a nested value actually changed - so a `Date`, `Buffer`
+ * or `Map` is returned as-is rather than being flattened into a bare object, and the caller's
```
addaleax: `Map` and `Set` are good examples of collections that still wouldn't be inspected during redaction, it might be worth mentioning that

packages/core/src/keychain.ts
```diff
@@ -33,29 +36,78 @@ export class Keychain implements IKeychain {
...
+        const entries = Object.entries(value);
+        const redacted = entries.map(([key, entry]) => [key, redactDeep(entry, secrets, ancestors)] as const);
+        if (redacted.every(([, entry], index) => entry === entries[index]?.[1])) {
+            return value;
+        }
+
+        return Object.setPrototypeOf(Object.fromEntries(redacted), Object.getPrototypeOf(value) as object | null);
+    } finally {
+        // Only ancestors block recursion; a value referenced twice from different branches is
+        // not a cycle and must still be redacted on the second visit.
```
addaleax: This is true, but is there a good reason not to use a `WeakMap` that maps original values to redacted values instead of a `WeakSet` to only track originals? Redacting the same object multiple times isn't great, it duplicates work and breaks referential integrity at very little benefit (paying for a `WeakMap` vs a `WeakSet` isn't all that different)

addaleax (approved):

PR #1461, opened 2026-08-25
feat: serve protocol revision 2026-07-28 (both eras) on HTTP and stdio

gagik:
Support protocol revision 2026-07-28 per the [SDK migration guide](https://github.com/modelcontextprotocol/typescript-sdk/blob/main/docs/migration/support-2026-07-28.md), stacked on the v2 SDK upgrade (#1460).

## Serving 2026-07-28 (both eras)

- **HTTP**: `MCPHttpServer` gains a modern `createMcpHandler(..., { legacy: 'reject' })` leg; POSTs route via `isLegacyRequest` — the guide's prescribed composition for keeping the sessionful Streamable-HTTP wiring serving 2025 clients.
- **stdio**: `StdioRunner` now serves through `serveStdio` with a per-connection `createServer` factory. `CliServer` splits registration from transport connect (`register()`), idempotent `close()`, and wires session/telemetry cleanup into the protocol server's `onclose`.

## Elicitation → multi-round-trip (`inputRequired`)

2026-07-28 removes the server→client request channel. Confirmation/input flows are now written once in the `inputRequired` style and serve both eras — the SDK's legacy shim turns the same return into real server→client requests on 2025-era connections:

- `Elicitation`/`IElicitation` become `inputRequired` builders + `inputResponses` readers.
- Converted `ToolBase` confirmation gate, MongoDB `confirmWriteStages` (aggregate / aggregate-db), and Atlas streams build connection-field elicitation.
- Capability-less clients proceed without prompting (prior behavior preserved).

## Notable

- New integration tests exercise the modern era end-to-end over HTTP and stdio (client negotiates `modern`, lists tools; confirmation flow via client auto-fulfilment).
- Elicitation/unit/aggregate/drop-index/streams tests updated to drive client-side `elicitation/create` handlers or the new builder/reader API.
- api-extractor reports regenerated; `MCP_SERVER_LIBRARY.md` and the v3-migration skill updated.
- **Follow-up (out of scope)**: `mongodb-atlas-mcp-remote` proxy stays legacy bridging (session-recovery is initialize-centric); modern passthrough needs SEP-2243 standard-header synthesis.


.prettierignore
```diff
@@ -1,6 +1,7 @@
 dist
 node_modules
 *.lock
+.pi/todos/
```
gagik: need to remove

packages/cli/src/cliServer.ts
```diff
@@ -123,6 +123,12 @@ export class CliServer<TMetrics extends DefaultMetricDefinitions = DefaultMetric
     private readonly startTime: number;
     private readonly subscriptions = new Set<string>();
 
+    /** Guard against double-close of session-scoped resources. */
+    private closed = false;
+
+    /** Whether {@link register} has run (guards against repeated registration). */
+    private registered = false;
```
gagik, line 130: can we avoid this guard?

packages/cli/src/cliServer.ts
```diff
@@ -123,6 +123,12 @@ export class CliServer<TMetrics extends DefaultMetricDefinitions = DefaultMetric
     private readonly startTime: number;
     private readonly subscriptions = new Set<string>();
 
+    /** Guard against double-close of session-scoped resources. */
+    private closed = false;
```
gagik, line 127: can we avoid this?

packages/types/src/elicitation.ts
```diff
@@ -1,34 +1,84 @@
...
+     * Builds the `inputRequired` result asking the user for structured input
+     * via a form. The handler returns it on the first entry; on re-entry the
+     * answers are read back with {@link IElicitation.readInput}.
+     *
+     * @param key - The identifier the answers arrive under in `inputResponses`.
+     * @param message - The message/title to display in the form.
+     * @param schema - A JSON Schema describing the fields to collect.
+     * @returns The `InputRequiredResult` to return from the request handler.
+     */
+    inputRequired(key: string, message: string, schema: ElicitRequestSchema): InputRequiredResult;
```
gagik: use object params

packages/tools-atlas/src/tools/connect/connectCluster.test.ts
```diff
@@ -100,8 +100,16 @@ describe("ConnectClusterTool", () => {
         } as unknown as ITelemetry;
 
         const mockElicitation = {
-            requestConfirmation: vi.fn(),
-        } as unknown as IElicitation;
+            supportsElicitation: (): boolean => true,
```
gagik: convert this (and other similar uses of this) into a shared helper (if it doesn't exist already)

packages/types/src/server.ts
```diff
@@ -10,6 +10,13 @@ import type { ICompositeLogger } from "./logging.js";
  */
 export type SessionServer<TTransport extends Transport = Transport> = {
     connect(transport: TTransport): Promise<void>;
+    /**
+     * Registers resources, capabilities, tools, request handlers and lifecycle
+     * hooks on the underlying MCP server without connecting it to a transport.
+     * Required by the 2026-07-28 serving entries (`serveStdio`,
+     * `createMcpHandler`), which build/register instances through a factory.
+     */
+    register?(): Promise<void>;
```
gagik: make this always required

gagik (commented):
Self-review complete: addressed the note — removed `.pi/todos/` from `.prettierignore`.

.prettierignore
gagik: Fixed — removed `.pi/todos/` from `.prettierignore`. It's a personal agent-tool path (ignored via local git config), not something this repo should reference.

packages/integration-tests/src/transports/modernProtocol.test.ts
```diff
@@ -0,0 +1,41 @@
...
+describe("2026-07-28 protocol (modern era) over HTTP", () => {
+    let runner: StreamableHttpTestRunnerComponents;
+    let client: Client;
+    let transport: StreamableHTTPClientTransport;
+
+    beforeAll(async () => {
+        // Random port: other suites bind the default httpPort (3000), and
+        // vitest runs test files in parallel workers.
+        runner = createStreamableHttpTestRunner({ ...defaultTestConfig, httpPort: 0 });
+        await runner.runner.start();
```
gagik: this is awkard. const { runner } = createStreamableHttpTestRunner is better

packages/tools-atlas/src/tools/connect/connectCluster.test.ts
gagik: Fixed — added `createMockElicitation()` to `@mongodb-js/mcp-test-utils` and migrated connectCluster.test.ts plus the ~22 other unit tests with the same inline mock to use it.

packages/types/src/elicitation.ts
gagik: Fixed — `IElicitation.inputRequired` now takes object params (`ElicitInputRequiredParams`), with the implementation, the `StreamsBuildTool` caller and tests updated.

packages/cli/src/cliServer.ts
gagik: Kept — `register()` is now a required `SessionServer` method (per the related comment on server.ts) that serving entries call directly (`CliStdioRunner`, the modern HTTP factory) as well as through `connect()` on the legacy HTTP / dry-run paths. Keeping the documented idempotency contract guards against double-registration of tools/resources/handlers if an instance ever receives more than one of those entry points.

packages/cli/src/cliServer.ts
gagik: Kept — this guard is required. `close()` runs `closeResources()` directly and then `mcpServer.close()`, whose `onclose` handler invokes `closeResources()` again (the stdio path relies solely on that `onclose` path since the runner only closes the McpServer), and `onerror` can fire alongside. Without the flag, telemetry/session would be closed multiple times.

packages/types/src/server.ts
gagik: Fixed — `SessionServer.register` is now always required: the optional call in the modern HTTP handler became a direct `await server.register()`, and the fake servers in the http-runners/integration tests implement it.

packages/integration-tests/src/transports/modernProtocol.test.ts
gagik: Fixed — destructured `const ({ runner } = createStreamableHttpTestRunner(...))` so the local `runner` is the actual `StreamableHttpRunner`, matching the pattern used in negotiatedClientState.test.ts; `getServerAddress(runner)` and `runner.close()` updated accordingly.

packages/cli/src/cliServer.ts
```diff
@@ -152,6 +158,24 @@ export class CliServer<TMetrics extends DefaultMetricDefinitions = DefaultMetric
...
+     *
+     * Idempotent: the 2026-07-28 serving entries (`serveStdio`,
+     * `createMcpHandler`) build one instance per connection/request through a
+     * factory and call this before handing the `McpServer` to the entry, while
+     * the legacy sessionful HTTP path calls it through {@link connect}.
+     */
+    async register(): Promise<void> {
+        if (this.registered) {
+            return;
+        }
```
addaleax, line 194: This doesn't account for calling `.register()` while `.register()` is already running but not yet finished. Is that a concern here?

packages/cli/src/cliServer.ts
```diff
@@ -237,20 +261,37 @@ export class CliServer<TMetrics extends DefaultMetricDefinitions = DefaultMetric
...
     async close(): Promise<void> {
+        await this.closeResources();
+        await this.mcpServer.close();
+    }
+
+    private async closeResources(): Promise<void> {
+        if (this.closed) {
+            return;
+        }
+        this.closed = true;
```
addaleax, line 315: Should `.register()` handle the case where this is `true` (e.g. throw)?

addaleax (commented):
I guess I'm a bit unclear on the degree to which we'd support the pre-multi-roundtrip model here – do we do that at all? The code changes don't look like that would be the case

gagik:
@addaleax I think the SDK is meant to have a [built-in shim to replicate old functionality](https://github.com/modelcontextprotocol/typescript-sdk/blob/main/docs/migration/support-2026-07-28.md#legacy-shim-for-input_required) but it's true there's no confirmation of that here.

The larger issue is that we don't have proper end-to-end tests which use an actual harness... I will look into setting those up and see if we can replicate some of these things.

packages/core/src/elicitation.ts
```diff
@@ -68,131 +46,97 @@ export class Elicitation {
...
+    public readConfirmation(inputResponses: ElicitationInputResponses): boolean | undefined {
+        if (inputResponses === undefined) {
+            return undefined;
         }
+        const content = acceptedContent<{ confirmation?: string }>(inputResponses, CONFIRMATION_INPUT_KEY);
+        if (content === undefined) {
+            return undefined;
+        }
+        return content.confirmation === "Yes";
+    }
```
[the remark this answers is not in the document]
gagik: Fixed — `readConfirmation()` now reads the response via the SDK's `inputResponse()` view: a missing key resolves to `undefined` (re-prompt), while `decline`/`cancel` (or a non-elicitation response) resolves to `false`, so callers no longer re-prompt after an explicit answer. Added decline/cancel test coverage.

packages/test-utils/src/elicitationMocks.ts
```diff
@@ -64,3 +96,26 @@ export function createMockElicitInput(): {
...
+        confirmationRequired: (): never => {
+            throw new Error("not implemented");
+        },
+        readInput: (): undefined => undefined,
+        inputRequired: (): never => {
+            throw new Error("not implemented");
+        },
+        ...overrides,
+    };
+}
```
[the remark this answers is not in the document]
gagik: Fixed — the mock's `readConfirmation()` now defaults to `undefined` ("no answer yet", matching first entry), so unexpected confirmation gating routes into the throwing `confirmationRequired()` and fails loudly instead of silently proceeding.

packages/types/src/tool.ts
```diff
@@ -63,6 +63,15 @@ export type ToolExecutionContext = {
...
     sendNotification?: (notification: unknown) => Promise<void>;
+    /**
+     * Responses to a previous `input_required` round (protocol revision
+     * 2026-07-28 multi-round-trip requests). Present only when this request
+     * is a client retry carrying the answers to elicitation/sampling/roots
+     * requests the handler returned as `inputRequired(...)`. Keyed by the
+     * server-assigned identifiers of the embedded requests. Values are
+     * untrusted client input.
+     */
+    inputResponses?: Record<string, unknown>;
```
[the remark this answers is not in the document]
gagik: Fixed — `ToolExecutionContext.inputResponses` is now typed as the shared `ElicitationInputResponses` alias from `@mongodb-js/mcp-types` (core's local duplicate was removed in favor of the same shared alias).

gagik:
@addaleax addressed your comments here; I will defer testing to https://github.com/mongodb-js/mongodb-mcp-server/pull/1464 (and an elicitation-specific follow up).

We'd want to run these tests against our last Node-published v2.x version and compare accordingly

packages/http-runners/src/mcpHttpServer.ts
```diff
@@ -95,6 +109,30 @@ export abstract class MCPHttpServer<
...
+            async (ctx: McpRequestContext) => {
+                const request: TransportRequestContext = {
+                    headers: Object.fromEntries(ctx.requestInfo?.headers ?? []),
+                    query: ctx.requestInfo?.url
+                        ? Object.fromEntries(new URL(ctx.requestInfo.url).searchParams)
+                        : undefined,
+                };
+                const server = await this.createServerForRequest(request);
+                await server.register();
+                return server.mcpServer as unknown as McpServer;
```
addaleax: This typecast is necessary?

packages/tools-mongodb/src/tools/read/aggregate.ts
```diff
@@ -231,7 +232,10 @@ export class AggregateTool extends MongoDBToolBase {
 
             const writeStageTargets = getWriteStageTargets(pipeline, database);
             if (writeStageTargets.length > 0) {
-                await this.confirmWriteStages(writeStageTargets, context);
+                const writeConfirmation = this.confirmWriteStages(writeStageTargets, context);
+                if (writeConfirmation) {
+                    return writeConfirmation;
```
addaleax: This naming doesn't feel particularly intuitive – takes a few loops to get that `writeConfirmation` is not a confirmation at all, and rather the request for a confirmation

addaleax (approved):
Looks great overall – no blocking comments!

packages/http-runners/src/mcpHttpServer.ts
gagik: Done in 3c8b96965 — SessionServer.mcpServer is now typed as the SDK's McpServer, so the cast is gone.

packages/tools-mongodb/src/tools/read/aggregate.ts
gagik: Done in 3c8b96965 — renamed to getInputRequiredResult / inputRequiredResult with an explanatory comment.

gagik:
failing streams test is unrelated to our changes but should be fixed

PR #1463, opened 2026-08-26
fix(atlas): classify invalid Streams arguments MCP-628

jwongmongodb:
## Proposed changes

Streams tools threw plain `Error` objects for caller-correctable validation failures such as a missing `workspaceName`. Remote MCP could not distinguish those failures from programming or infrastructure errors, so they were classified as unexpected and contributed to alerting totals.

Add a stable `InvalidArgument` error code and use `MongoDBError` for caller-addressable validation failures across the Streams build, discover, manage, and teardown tools. This lets the existing Remote MCP error classifier mark them as expected without matching user-facing message text. Infrastructure and programming errors retain their existing unexpected classification.

Includes regression coverage for the Streams error helper and missing teardown arguments.

Validation:
- Built `@mongodb-js/mcp-tools-atlas` and its workspace dependencies
- Passed `@mongodb-js/mcp-tools-atlas` typecheck
- Passed ESLint for changed TypeScript files
- Passed Streams error helper test (1 test)
- Passed Streams teardown tests (26 tests)

Jira: https://jira.mongodb.org/browse/MCP-628

## Checklist

### DIFF
--- a/packages/tools-atlas/src/streams/errors.ts
+++ b/packages/tools-atlas/src/streams/errors.ts
@@ -1,0 +1,11 @@
+import { ErrorCodes, MongoDBError } from "@mongodb-js/mcp-tools-mongodb";
+
+/**
+ * Creates a caller-addressable error for invalid or incomplete Streams tool arguments.
+ *
+ * Using MongoDBError rather than a plain Error allows Remote MCP to classify the
+ * failure as expected without relying on the caller-facing message.
+ */
+export function streamsInvalidArgument(message: string): MongoDBError<typeof ErrorCodes.InvalidArgument> {
+    return new MongoDBError(ErrorCodes.InvalidArgument, message);
+}
--- a/packages/tools-atlas/src/streams/errors.ts
+++ b/packages/tools-atlas/src/streams/errors.ts
@@ -1,11 +1,9 @@
 import { ErrorCodes, MongoDBError } from "@mongodb-js/mcp-tools-mongodb";
 
-/**
- * Creates a caller-addressable error for invalid or incomplete Streams tool arguments.
- *
- * Using MongoDBError rather than a plain Error allows Remote MCP to classify the
- * failure as expected without relying on the caller-facing message.
- */
-export function streamsInvalidArgument(message: string): MongoDBError<typeof ErrorCodes.InvalidArgument> {
-    return new MongoDBError(ErrorCodes.InvalidArgument, message);
+/** A caller-addressable error for invalid or incomplete Streams tool arguments. */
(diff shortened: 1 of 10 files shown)


packages/tools-atlas/src/streams/errors.ts
```diff
@@ -0,0 +1,11 @@
+import { ErrorCodes, MongoDBError } from "@mongodb-js/mcp-tools-mongodb";
+
+/**
+ * Creates a caller-addressable error for invalid or incomplete Streams tool arguments.
+ *
+ * Using MongoDBError rather than a plain Error allows Remote MCP to classify the
+ * failure as expected without relying on the caller-facing message.
+ */
+export function streamsInvalidArgument(message: string): MongoDBError<typeof ErrorCodes.InvalidArgument> {
+    return new MongoDBError(ErrorCodes.InvalidArgument, message);
```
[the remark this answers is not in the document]
jwongmongodb: Remote MCP wraps each tool’s `handleError` with `withToolErrorClassification` before invocation. `ToolBase.invoke` catches the raw error and passes it directly to `this.handleError(error, args)`, so the wrapper receives the original `MongoDBError` and can inspect `ErrorCodes.InvalidArgument` before conversion to a `CallToolResult`.

I added an `invoke`-level regression test in 0974b68d that installs a wrapped `handleError` and verifies it receives the original `MongoDBError` with `code: ErrorCodes.InvalidArgument`. The log and duration metric do not carry the code, but they are not the Remote MCP classification boundary.

packages/tools-atlas/src/streams/errors.ts
```diff
@@ -0,0 +1,11 @@
+import { ErrorCodes, MongoDBError } from "@mongodb-js/mcp-tools-mongodb";
+
+/**
+ * Creates a caller-addressable error for invalid or incomplete Streams tool arguments.
+ *
+ * Using MongoDBError rather than a plain Error allows Remote MCP to classify the
+ * failure as expected without relying on the caller-facing message.
+ */
+export function streamsInvalidArgument(message: string): MongoDBError<typeof ErrorCodes.InvalidArgument> {
```
gagik: maybe we can use a new class and extend `MongoDBError` instead of having this function?

addaleax (approved):

packages/tools-atlas/src/streams/errors.ts
jwongmongodb: Good suggestion. I replaced the factory with a dedicated `StreamsInvalidArgumentError extends MongoDBError<typeof ErrorCodes.InvalidArgument>` in 6c726187 and updated all Streams call sites and the helper test. This keeps the stable classification code while making the error intent and generic error name explicit.

lerouxb (approved):

PR #1472, opened 2026-09-01
feat(streams): add processor autoscaling STREAMS-2618

jwongmongodb:
## Jira and API contract

- Jira: STREAMS-2618
- API contract: https://docs.google.com/document/d/1q5mmlV63bEk006CXpDEWM3nXDa41USzgvE_nvHkZGGs/edit?tab=t.9igxibskrbi1#heading=h.gz89gh34tjpx
- Atlas API: GA Atlas OpenAPI v2 (`openapi-2025-03-12.json`, selected by `packages/scripts/generate.sh`)

## Proposed changes

- add stream processor autoscaling configuration to processor create, modify, and `startWith` workflows
- expose the configured baseline `tier`, read-only `effectiveTier`, and autoscaling configuration when listing, inspecting, and diagnosing processors
- preserve the Atlas API contract's omission semantics, explicit `null` resets, and explicit disable behavior
- keep `atlas-streams-manage` classified as an update and confirmation-required tool; credential handling is unchanged
- add unit, Streams integration, and accuracy coverage for autoscaling workflows

## Drive-by fix

- update the live Schema Registry integration fixture from `https://cloud-dev.mongodb.com/` to `https://httpbin.org/`
- stricter Atlas Schema Registry URL validation rejects the former as `Invalid url`; the latter is a resolvable public HTTPS endpoint
- no production Schema Registry behavior changes

## Generated surfaces

- update `api-extractor/reports/tools.public.api.md` for the intentional public tool-schema changes
- API client generation produced no Atlas client diff because the autoscaling contract is already present in the dated GA specification
- the unrelated `packages/scripts/generate.sh` tooling change was removed before review

## Safety and behavior

- create remains a mutating build operation; processor start and modify remain confirmation-required manage operations
- omitted autoscaling fields preserve persisted values during modify/start
- `autoscaling: null`, `enabled: false`, and `enabled: null` explicitly disable and clear autoscaling
- `minTier: null` and `maxTier: null` reset those bounds to workspace defaults
- Atlas API errors are propagated; the tools do not claim success before Atlas confirms the operation

### Unit

- exact create request with baseline tier and autoscaling bounds
- top-level `startWith` autoscaling request shape
- processor modify request shape
- explicit `null` disable behavior
- baseline/effective tier and autoscaling structured output from build, discover, and manage

### Streams integration

- create a processor with autoscaling and inspect the persisted configuration
- start and stop the processor
- modify baseline tier and autoscaling bounds, including a `null` bound reset
- disable autoscaling with `autoscaling: null`
- use a resolvable Schema Registry fixture URL

### Accuracy

Representative prompts cover:

- creating a processor with a baseline tier and min/max autoscaling bounds
- enabling autoscaling while modifying a stopped processor
- starting a processor with autoscaling options

Runtime accuracy execution requires a configured model-provider key and the `accuracy-tests` label. Only the accuracy package type check was run locally.

## Validation

- `pnpm check:types`
- `pnpm run check:api`
- focused Streams unit tests — 150 passed
- `pnpm --filter @mongodb-js/mcp-accuracy-tests check`
- formatting check for all changed source and test files
- prior full branch validation: `pnpm run build`, `pnpm run generate`, `pnpm check`, and `pnpm test:unit` — 1,086 passed, 2 skipped
- local live Streams execution cannot provision the shared project because the local credential lacks the organization group-creator role (`NOT_ORG_GROUP_CREATOR`); the `Run Atlas Streams tests` CI job is authoritative
- Inspector not run; tool registration and schema metadata are covered by the Streams integration suite

## Rollout and compatibility

- requires the processor autoscaling fields available in the dated GA Atlas v2 specification
- no new dependencies or lockfile changes
- no Agent Skill update is required because this extends existing create, inspect, manage, and diagnose workflows without changing their sequencing or safety model

## Checklist

- [x] Branch rebased onto current `main`
- [x] Generated surface and API report reviewed
- [x] Unit coverage added
- [x] Streams integration coverage added
- [x] Accuracy coverage added
- [x] Unrelated API-generation tooling change removed


packages/accuracy-tests/src/streamsBuild.test.ts
```diff
@@ -331,7 +331,7 @@ describeAccuracyTests(
             mockedTools,
         },
         {
-            prompt: `Add a Confluent Schema Registry connection named 'registry' with URL https://schema-registry.example.com to workspace '${workspaceName}'`,
+            prompt: `Add a Confluent Schema Registry connection named 'registry' with URL https://198.51.100.10 to workspace '${workspaceName}'`,
```
jwongmongodb: this is a driveby change that we had to fix internally due to a SECBUG

packages/tools-atlas/src/tools/streams/build.ts
```diff
@@ -3,21 +3,32 @@ import { StreamsToolBase } from "../../streams/streamsToolBase.js";
...
-    "AWSLambda",
-    "SchemaRegistry",
-    "Sample",
-]);
+const BuildResource: z.ZodEnum<{
+    processor: "processor";
+    workspace: "workspace";
+    connection: "connection";
+    privatelink: "privatelink";
+}> = z.enum(["workspace", "connection", "processor", "privatelink"]);
```
nirinchev: Why do we need the explicit type declarations here? Does the inferred type not work?

packages/tools-atlas/src/streams/streamsArgs.ts
```diff
@@ -162,6 +174,22 @@ export const PrivateLinkConfig = z
...
+            .optional()
+            .describe("Enable autoscaling. Explicit false or null disables autoscaling and clears its configuration."),
+        minTier: StreamsTier.nullable()
+            .optional()
+            .describe("Autoscaling floor. Null resets the floor to the workspace default tier."),
+        maxTier: StreamsTier.nullable()
+            .optional()
+            .describe("Autoscaling ceiling. Null resets the ceiling to the workspace maximum tier."),
+    })
+    .nullable();
```
nirinchev: Do we want to declare this as nullable?

packages/integration-tests/src/tools/atlas/streams/build.test.ts
```diff
@@ -142,7 +142,7 @@ describeWithStreams("atlas-streams-build", (integration) => {
                         connectionName: srConnName,
                         connectionType: "SchemaRegistry",
                         connectionConfig: {
-                            schemaRegistryUrls: ["https://cloud-dev.mongodb.com/"],
+                            schemaRegistryUrls: ["https://httpbin.org/"],
```
addaleax, line 151: What's this related to? If the exact value doesn't matter, let's just use `example.org`, that's an unambiguous way of saying that it doesn't matter

packages/tools-atlas/src/tools/streams/manage.ts
```diff
@@ -214,7 +217,10 @@ export class StreamsManageTool extends StreamsToolBase {
     private static mapUpdateWorkspaceStructuredContent(
         updated: {
             dataProcessRegion?: { cloudProvider?: string; region?: string };
-            streamConfig?: { tier?: string; maxTierSize?: string } | null;
+            streamConfig?: {
+                tier?: "SP2" | "SP5" | "SP10" | "SP30" | "SP50";
+                maxTierSize?: "SP2" | "SP5" | "SP10" | "SP30" | "SP50";
```
nirinchev:
```suggestion
                tier?: z.infer<typeof StreamsTier>;
                maxTierSize?: z.infer<typeof StreamsTier>;
```

packages/tools-atlas/src/streams/streamsArgs.ts
```diff
@@ -4,6 +4,18 @@ import { AtlasArgs } from "../args.js";
...
+    SP50: "SP50";
+    SP30: "SP30";
+    SP10: "SP10";
+    SP5: "SP5";
+    SP2: "SP2";
+};
+
+export const StreamsTier: z.ZodEnum<StreamsTierValues> = z.enum(["SP2", "SP5", "SP10", "SP30", "SP50"]);
+
+const DBRoleType: z.ZodEnum<{ CUSTOM: "CUSTOM"; BUILT_IN: "BUILT_IN" }> = z.enum(["BUILT_IN", "CUSTOM"]);
```
addaleax: We can probably just omit the explicit typing, and if we need to explicitly need to refer to the types, infer them from the zod value rather than the other way around, e.g. like

https://github.com/mongodb-js/mongodb-mcp-server/blob/7d32cc426a7d14f732c0add68f4ffc7bc29c193f/packages/tools-atlas/src/args.ts#L24-L25

(not blocking of course)

packages/tools-atlas/src/tools/streams/manage.ts
```diff
@@ -413,28 +421,40 @@ export class StreamsManageTool extends StreamsToolBase {
...
             },
             context
         );
 
+        const structuredContent: ManageOutput = { processorState: "STOPPED" };
+        if (args.tier !== undefined && updated.tier !== undefined) structuredContent.tier = updated.tier;
+        if ((args.tier !== undefined || args.autoscaling !== undefined) && updated.effectiveTier !== undefined) {
+            structuredContent.effectiveTier = updated.effectiveTier;
+        }
+        if (args.autoscaling !== undefined) structuredContent.autoscaling = updated.options?.autoscaling ?? null;
```
nirinchev: Why are we only populating these if the args were supplied? Any drawback to just populating them indiscriminately?

packages/tools-atlas/src/tools/streams/build.ts
```diff
@@ -3,21 +3,32 @@ import { StreamsToolBase } from "../../streams/streamsToolBase.js";
...
+const ConnectionType: z.ZodEnum<{
+    Cluster: "Cluster";
+    Kafka: "Kafka";
+    S3: "S3";
+    Https: "Https";
+    AWSKinesisDataStreams: "AWSKinesisDataStreams";
+    AWSLambda: "AWSLambda";
+    SchemaRegistry: "SchemaRegistry";
+    Sample: "Sample";
+}> = z.enum(["Kafka", "Cluster", "S3", "Https", "AWSKinesisDataStreams", "AWSLambda", "SchemaRegistry", "Sample"]);
```
addaleax: ditto

packages/tools-atlas/src/streams/streamsArgs.ts
jwongmongodb: technically the API allows null to be set for autoscaling but it is not documented, I can remove this for now since disabling the autoscaling field can be done with `autoscaling.enabled`= false

packages/tools-atlas/src/streams/streamsArgs.ts
himanshusinghs: nit: You can also just define the StreamsTier schema and derive the type out of it with:

```
type StreamsTierValues = z.infer<typeof StreamsTier>
```

That way there is only single source of truth - the schema itself.

packages/tools-atlas/src/tools/streams/discover.ts
```diff
@@ -178,6 +200,12 @@ function buildProcessorStructuredContent(
     if (proc.tier !== undefined) {
         structuredContent.tier = proc.tier;
     }
+    if (proc.effectiveTier !== undefined) {
+        structuredContent.effectiveTier = proc.effectiveTier;
+    }
+    if (proc.options?.autoscaling !== undefined) {
+        structuredContent.autoscaling = proc.options.autoscaling;
+    }
```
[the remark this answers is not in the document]
addaleax: The Copilot comment doesn't seem too important (unless TS compilation fails), but generally speaking it seems like a valid point – should we be validating the contents of these fields here (e.g. by parsing them through the zod schema)?

packages/tools-atlas/src/streams/streamsArgs.ts
```diff
@@ -47,7 +59,7 @@ export const ConnectionConfig = z
         dbRoleToExecute: z
             .object({
                 role: z.string().optional(),
-                type: z.enum(["BUILT_IN", "CUSTOM"]).optional(),
+                type: DBRoleType.optional(),
```
himanshusinghs, line 56: Since we have a default in the description, might as well add the `.default()` chain to make it visible to the agent via the schema itself.

packages/tools-atlas/src/tools/streams/build.ts
```diff
@@ -203,6 +214,14 @@ export class StreamsBuildTool extends StreamsToolBase {
...
                     "The DLQ connection must already exist in the workspace."
             ),
+        processorTier: StreamsTier.optional().describe(
+            "Baseline processing tier. Only for resource='processor'. Defaults to the workspace tier when omitted."
+        ),
+        autoscaling: StreamsAutoscaling.optional().describe(
+            "Autoscaling configuration. Only for resource='processor'. " +
+                "Set enabled=true to enable it; omitted means disabled at creation. " +
+                "minTier and maxTier default to workspace bounds when enabled."
+        ),
```
himanshusinghs: Please correct me if I am wrong but I think all the `optional()` chain is coming out of the idea that some args are applicable to only specific resource type like in this case processorTier and autoScaling applies only to resource type "processor".

We can provide better guidance to the model by using [discriminated unions](https://zod.dev/api#discriminated-unions). We do something similar for aggregate, explain and export tool. Its worth looking into and it can simplify some of the optional case handling as well.

packages/tools-atlas/src/streams/streamsArgs.ts
```diff
@@ -162,6 +174,22 @@ export const PrivateLinkConfig = z
     })
     .passthrough();
 
+export const StreamsAutoscaling = z
+    .object({
+        enabled: z
+            .boolean()
+            .nullable()
+            .optional()
+            .describe("Enable autoscaling. Explicit false or null disables autoscaling and clears its configuration."),
```
himanshusinghs: Can this just be a boolean?

packages/tools-atlas/src/tools/streams/build.ts
himanshusinghs: Will be taken up in a separate refactor of the schema

packages/tools-atlas/src/streams/streamsArgs.ts
himanshusinghs: This is a smaller level change that might just get missed in the refactor. If it does not change the behaviour then its worth doing it now.

packages/tools-atlas/src/streams/streamsArgs.ts
himanshusinghs: The same for this. If it can just be boolean and does not change the behaviour (actual handling of it) then its worth doing it now since its a smaller change that might get missed in the larger schema refactor.

packages/integration-tests/src/tools/atlas/streams/build.test.ts
jwongmongodb: this is a driveby change to a secbug ticket that was fixed with our api, it needs a valid url to be resolvable which was failing the mcp tests. Here is the pr: https://github.com/10gen/mms/pull/180436. Note: this was also failing for the terraform provider repo from the integration test as well which was fixed. We can use https://example.org, but the text in the website was why I decided not to use it: `This domain is for use in documentation examples without needing permission. Avoid use in operations.`

packages/tools-atlas/src/tools/streams/manage.ts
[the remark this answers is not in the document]
jwongmongodb: fixed, good catch. `links` is a read-only field the api echoes back, so i now project through `toStreamsAutoscaling` (enabled/minTier/maxTier only) and drop the null/cleared response. didn't re-parse with zod on purpose, the payload is already typed, we just don't want to leak the extra keys.

packages/tools-atlas/src/tools/streams/discover.ts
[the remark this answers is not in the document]
jwongmongodb: didn't re-parse with zod on purpose, the api response is already typed by the generated client so a parse would just add failure risk for nothing. the projection only keeps enabled/minTier/maxTier out of structured output.

packages/tools-atlas/src/tools/streams/build.ts
jwongmongodb: done, dropped the explicit `z.ZodEnum<{...}>` annotations and derive the types from the schema now (`StreamsTierValue = z.infer<typeof StreamsTier>`), same as `args.ts`.

packages/tools-atlas/src/tools/streams/manage.ts
jwongmongodb: done, dropped the explicit `z.ZodEnum<{...}>` annotations and derive the types from the schema now (`StreamsTierValue = z.infer<typeof StreamsTier>`), same as `args.ts`.

packages/tools-atlas/src/streams/streamsArgs.ts
jwongmongodb: done, dropped the explicit `z.ZodEnum<{...}>` annotations and derive the types from the schema now (`StreamsTierValue = z.infer<typeof StreamsTier>`), same as `args.ts`.

jwongmongodb: done, dropped the explicit `z.ZodEnum<{...}>` annotations and derive the types from the schema now (`StreamsTierValue = z.infer<typeof StreamsTier>`), same as `args.ts`.

packages/tools-atlas/src/streams/streamsArgs.ts
jwongmongodb: as i said, done in `a278e019`. top-level `autoscaling: null` shorthand removed, disable is `enabled: false`. `minTier`/`maxTier` stayed nullable though, since a null bound is the only way to reset to workspace default.

packages/tools-atlas/src/streams/streamsArgs.ts
jwongmongodb: done, `enabled` is just a boolean now, disable via `enabled: false`. dropped the `null` variants too since they weren't documented.

packages/tools-atlas/src/streams/streamsArgs.ts
jwongmongodb: added a declarative `default("BUILT_IN")` in `1b8f3aab` so it shows up in the schema. no wire change, build.ts already applies it at runtime.

packages/tools-atlas/src/tools/streams/manage.ts
jwongmongodb: these fields are confirmations of the change the caller requested, so we only populate them when that field was actually passed. emitting them on unrelated modifies would imply tier/autoscaling were just changed, and the api response won't reliably echo them anyway. same as update-workspace's includeRegion/includeTier; full state is discover inspect-processor.

packages/tools-atlas/src/tools/streams/build.ts
jwongmongodb: agreed this is the right shape long-term, just out of scope for this pr, tracking it in a separate schema refactor.

packages/tools-atlas/src/tools/streams/build.ts
jwongmongodb: done, dropped the explicit `z.ZodEnum<{...}>` annotations and derive the types from the schema now (`StreamsTierValue = z.infer<typeof StreamsTier>`), same as `args.ts`.

addaleax (approved):

PR #1474, opened 2026-09-02
refactor: remove session state

gagik:
## What / why

Removes the per-client `session` object as a first-class concept. The server is now request-scoped and stateless: every tool call is driven by its request (args, headers, `_meta`), the static `UserConfig`, and app-level shared services — no server-side per-client `Session` object held in memory. Per-client state (MongoDB connections, exports, identity) is either keyed by `connectionId`/client identity or carried on the request.

The session object is replaced by:
- `ToolServer` / `ToolServices` (instead of `IToolSession` / `ISession`) — services injected individually into each request-scoped server, not accessed via a `session`.
- A request-centric `ToolExecutionContext.request` — per-request server, signal, id, headers, client info live on the request.
- `BaseServer` (register + `mcpServer`) for the HTTP/stdio runners, replacing `SessionServer`.
- App-level `SharedServices` (connections, exports, keychain, API client, telemetry) built once per process, referenced by each request's server.

The public session surface (`ISession`, `IToolSession`, `ISessionStore`, `SessionServer`) is removed from the types. The modern (2026-07-28) protocol is served statelessly — one fresh request-scoped server per request, with client identity carried per request in the `_meta` envelope.

## 2025-era (legacy) HTTP

Legacy (2025-era) requests are served **sessionfully** through a small in-memory session store, split out of `MCPHttpServer` into `LegacyMcpHttpHandler`. This is deliberate: the 2025 protocol negotiates client capabilities (incl. `elicitation`) in the `initialize` handshake and delivers elicitation as real server→client requests, both of which need a live, correlated transport. The SDK's `legacy: 'stateless'` path cannot do that (each request gets a fresh instance: no capabilities, no return channel), so confirmation/elicitation was broken for 2025-era clients like Claude Code and Codex. Serving legacy sessionfully restores the SDK's elicitation shim without touching the modern stateless path.

- `initialize` creates a session (a fresh request-scoped server connected to a sessionful streamable HTTP transport); later POSTs, the `GET /mcp` SSE idle stream, and `DELETE /mcp` session close are routed to the session's transport.
- Session count is capped by the new **`maxSessions`** config (`MDB_MCP_MAX_SESSIONS` / `--maxSessions`, default 1000); sessions beyond the cap are rejected with `503` / `-32008`.
- `createMcpHandler` uses `legacy: 'reject'` for the modern path; legacy traffic is routed by `isLegacyRequest`. `LegacyMcpHttpHandler` is created via an overridable `createLegacyHandler()` hook on `MCPHttpServer`.

## Breaking changes / notes

- **Tool constructor**: `ToolBase` subclasses now receive the server as a single argument (`new (server: ToolServer)`). Use `this.server.*` (e.g. `this.server.config`, `this.server.keychain`, `this.server.elicitation`) — `this.session` and the old injected `telemetry`/`elicitation`/`metrics` fields are gone.
- **Execution context**: per-request fields are nested under `context.request` (`request.server`, `request.signal`, `request.id`, `request.headers`, `request.clientInfo`) instead of flat `context.signal`/`context.requestId`/`context.requestInfo`.
- **Keychain API** (from main MCP-634) is now `IKeychain extends IRedactor` with `redact(value)`; `allSecrets` and the standalone `redact(value, secrets)` helper are removed.
- **Custom tools / consumers** must update to the `ToolServer` shape; `IToolSession`/`ISession`/`SessionServer` are removed from the public types. For legacy HTTP the server now holds a session per 2025-era client (capped by `maxSessions`).

## Caveats

- **Legacy (2025-era) HTTP is sessionful, not stateless.** It uses a minimal in-memory session store (`mcp-session-id` routing) so the SDK's elicitation shim works. Modern-only deployments are unaffected; legacy clients that do not need elicitation still work, but they now occupy a session (bounded by `maxSessions`) instead of a per-request stateless transport. GET/DELETE (`/mcp`) serve the 2025 session stream/close rather than returning `405`.
- **Logging correlation**: `sessionId` is gone from the request-scoped server logs; correlation uses `request.id` plus the forwarded `x-request-id` header, and (once client identity mapping lands) the client identity.
- **Per-request cost**: the modern path rebuilds and re-registers a server per request (fresh tools/resources on a new `McpServer`). Static schema caches and shared `AppServices` amortize this; verifying this is acceptable under load is a follow-up.
- **Exports cleanup** no longer rides on a `Session.close()`; TTL/expiry or client-keyed ownership is the server's responsibility.


docs/plans/remove-session-concept.md
```diff
@@ -0,0 +1,342 @@
...
+- Per-client data (MongoDB connections created by the client, its exports directory,
+  its identity for `appName`/telemetry) is either **explicitly keyed** (by
+  `connectionId` / client identity) or **request-carried** (from negotiated state).
+
+## 2. Current state (why this is feasible)
+
+Exploration of the codebase shows the tool layer is already ~90% de-sessioned:
+
+- **`Session` (`packages/cli/src/cliSession.ts`)** is a thin, mostly-immutable context
+  bag: `logger`, `config`, `keychain` (already a process-global singleton),
```
addaleax, line 31: gonna talk more about `keychain` being problematic on Slack, but I wanted to read this PR first for context

gagik:
This WIP branch has been split into an ordered stack of 4 standalone PRs (same content, sliced by dependency; each builds, lints, and passes unit tests on its own):

1. **#1477** — remove session concept from core/types/http/cli + request-centric tool context
2. **#1478** — Phase 2: shared AppServices (api-client ctor, resources({server}), startup config validation)
3. **#1479** — Phase 3A: per-client connection scoping + authenticated multi-tenant HTTP
4. **#1480** — config on request.server + lint/api sweep

All four are currently drafts. #1477 targets chore/protocol-revision-2026-07-28 (this PR's base); each subsequent PR targets the previous one. Leaving this PR open as the tracking issue for the stack.

gagik:
Updated split: this branch is now a **2-PR stack** (the earlier 4-part split collapsed — its phases overrode each other: request.config → request.server landed after the envelope, AppServices refactored commit-1 glue, auth iterated across 4 commits. Splitting those was artificial):

1. **#1481** — the complete refactor: session removal + request-centric tool context + AppServices + Phase 3A scoping/auth + request.server (build, unit tests, integration typecheck, API reports green; lint/format swept in part B)
2. **#1482** — the branch's final lint/format/api-report sweep on top of #1481

Both drafts. #1481 targets chore/protocol-revision-2026-07-28 (this PR's base); #1482 targets #1481's branch. The earlier 4-part PRs (#1477-1480) are closed as superseded. Leaving this PR open as the tracking issue.

packages/cli/src/createServerServices.ts
```diff
@@ -38,23 +46,109 @@ export type SharedServerServices = {
...
+        }
+    }
+
+    // Validate API client credentials
+    if (config.apiClientId && config.apiClientSecret) {
+        try {
+            try {
+                const apiBaseUrl = new URL(config.apiBaseUrl);
+                if (apiBaseUrl.protocol !== "https:") {
+                    // Log a warning, but don't error out. This is to allow for testing against local or non-HTTPS endpoints.
```
addaleax, line 93: I've definitely seen the approach "HTTPS or known localhost/loopback, otherwise error or require explicit opt-in" a number of times. Not saying we need to do that, just saying that it's something other apps do for security reasons

packages/cli/src/createServerServices.ts
```diff
@@ -119,28 +250,44 @@ export function createServerFromConfig({
...
         telemetry,
-        connectionErrorHandler,
         elicitation,
         metrics,
         tools,
         resources,
         serverMetadata,
+        // Validated once at startup by `validateAppConfig`; the per-request
+        // server must not re-run the (network) credential validation.
+        configValidated: true,
```
gagik: there should be a way to avoid this

docs/plans/remove-session-concept.md
```diff
@@ -0,0 +1,342 @@
...
+App-level (built once at startup, shared, immutable config):
+  UserConfig, Metrics, CompositeLogger (root), Keychain (root),
+  MCPConnectionStore, ApiClient, ExportsManager (app-rooted),
+  AtlasLocalClient, DeviceId, MonitoringServer, AtlasTelemetry.
+  (No McpServer — the SDK's HTTP entry requires a fresh instance per request,
+  so the app level holds only heavy services; see Phase 2.)
+
+Request-scoped (derived per request, never stored):
+  the minimal per-request server: fresh McpServer + Elicitation + CliServer,
+  tool instantiation/registration, and a client-scoped ConnectionRegistry view
```
nirinchev: Is this true, even for v2 of the SDK? Seems quite wasteful to be instantiating a brand new MCP server per request.

packages/atlas-api-client/src/apiClient.test.ts
```diff
@@ -9,6 +9,18 @@ const testHttpClient = {
     Request: globalThis.Request,
 };
 
+function credentialedProvider(baseUrl = "https://api.test.com", httpClient = testHttpClient): AuthProvider | undefined {
```
gagik: `createMockAuthProvider`, move to a separate file, avoid using api.test.com and use either example.com or localhost

docs/plans/remove-session-concept.md
gagik: we can follow up with a PR on top of this

packages/cli/src/createServerServices.ts
gagik: I'd be open to do this separately as that's another behavior change

packages/atlas-api-client/src/apiClient.test.ts
gagik: Fixed — extracted `createMockAuthProvider` into `src/test-helpers/createMockAuthProvider.ts` (excluded from the package build so it isn't shipped) and replaced `https://api.test.com` with `https://example.com` throughout the atlas-api-client tests.

packages/cli/src/createServerServices.ts
gagik: Went further and removed the flag entirely: `CliServer` no longer validates config on register, and `CliServer.validateConfig` (which duplicated `validateAppConfig`) is gone. Validation is owned solely by `validateAppConfig`, called once at startup in `createSharedServicesFromConfig` — the per-request server never needs to revalidate because the connection string / Atlas credentials are `overrideBehavior: "not-allowed"` and can't differ per request. No boolean anywhere.

packages/cli/src/cliServer.ts
```diff
@@ -401,56 +377,4 @@ export class CliServer<TMetrics extends DefaultMetricDefinitions = DefaultMetric
     public get clientInfo(): Implementation | undefined {
         return this.mcpServer.server.getClientVersion();
     }
-
-    private async validateConfig(): Promise<void> {
```
gagik, line 396: this was duplicated from https://github.com/mongodb-js/mongodb-mcp-server/blob/e7a534e3584b9e60d2d91498548b4dfe359767a9/packages/cli/src/createServerServices.ts#L65

gagik: so now it's CLI parsing that validates but not the server itself

gagik:
need to move `ToolBase` into `types` as a follow-up, some minor duplication here to minimize changes

docs/plans/remove-session-concept.md
```diff
@@ -0,0 +1,345 @@
+# Plan: Remove "Session" as a Concept from the MCP Server
```
gagik, line 1: keeping this for reference, will remove before merging

gagik:
Will need to do another round of self-review but reaching a stable state!

packages/cli/src/resources/common/exportedData.ts
```diff
@@ -35,19 +35,19 @@ export class ExportedData {
...
-        this.session.exportsManager.on("export-available", (uri: string): void => {
-            server.sendResourceListChanged();
-            server.sendResourceUpdated(uri);
+        this.server.exportsManager.on("export-available", (uri: string): void => {
+            this.server.sendResourceListChanged();
+            this.server.sendResourceUpdated(uri);
         });
-        this.session.exportsManager.on("export-expired", (): void => {
-            server.sendResourceListChanged();
+        this.server.exportsManager.on("export-expired", (): void => {
```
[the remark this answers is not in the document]
gagik: Done in b0e4ce8ec — gated the export-available/export-expired listeners to long-lived transports (stdio/dry-run), so HTTP request-scoped servers no longer accumulate them on the shared ExportsManager.

packages/integration-tests/src/tools/atlas/projects.test.ts
```diff
@@ -42,9 +42,9 @@ describeWithAtlas("projects", (integration) => {
     const projectsToCleanup: string[] = [];
 
     afterAll(async () => {
-        const session = integration.mcpServer().session;
+        const session = integration.mcpServer();
         assertApiClientIsAvailable(session);
-        const apiClient = session.apiClient;
+        const apiClient = session.getApiClient();
```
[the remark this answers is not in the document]
gagik: This compiles: `integration.mcpServer()` returns a `CliServer` augmented with `getApiClient()` via `Object.assign` in the test helper, so `session.getApiClient()` is a valid member — confirmed with `tsc --noEmit` on the integration-tests package.

packages/integration-tests/src/tools/atlas/dbUsers.test.ts
```diff
@@ -36,9 +36,9 @@ describeWithAtlas("db users", (integration) => {
             }
 
             try {
-                const session = integration.mcpServer().session;
+                const session = integration.mcpServer();
                 assertApiClientIsAvailable(session);
-                const apiClient = session.apiClient;
+                const apiClient = session.getApiClient();
```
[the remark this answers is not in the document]
gagik: This compiles: `integration.mcpServer()` returns a `CliServer` augmented with `getApiClient()` via `Object.assign` in the test helper, so `session.getApiClient()` is a valid member — confirmed with `tsc --noEmit` on the integration-tests package.

packages/integration-tests/src/tools/atlas/accessLists.test.ts
```diff
@@ -23,17 +23,17 @@ describeWithAtlas("ip access lists", (integration) => {
         const values = [...ips, ...cidrBlocks];
 
         beforeAll(async () => {
-            const session = integration.mcpServer().session;
+            const session = integration.mcpServer();
             assertApiClientIsAvailable(session);
-            const apiClient = session.apiClient;
+            const apiClient = session.getApiClient();
```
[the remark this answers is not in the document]
gagik: This compiles: `integration.mcpServer()` returns a `CliServer` augmented with `getApiClient()` via `Object.assign` in the test helper, so `session.getApiClient()` is a valid member — confirmed with `tsc --noEmit` on the integration-tests package.

packages/integration-tests/src/tools/atlas/projects.test.ts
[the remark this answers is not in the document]
gagik: Follow-up (250ec15c0): switched these tests to use the direct `server.apiClient` property and removed the test helper's `getApiClient()` shim, so they no longer rely on a method `CliServer` doesn't expose.

packages/integration-tests/src/tools/atlas/dbUsers.test.ts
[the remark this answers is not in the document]
gagik: Follow-up (250ec15c0): switched these tests to use the direct `server.apiClient` property and removed the test helper's `getApiClient()` shim, so they no longer rely on a method `CliServer` doesn't expose.

packages/integration-tests/src/tools/atlas/accessLists.test.ts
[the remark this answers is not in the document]
gagik: Follow-up (250ec15c0): switched these tests to use the direct `server.apiClient` property and removed the test helper's `getApiClient()` shim, so they no longer rely on a method `CliServer` doesn't expose.

packages/cli/src/cliServer.ts
```diff
@@ -192,16 +208,13 @@ export class CliServer<TMetrics extends DefaultMetricDefinitions = DefaultMetric
         if (this.registered) {
             return;
         }
-        this.registerPromise = this.doRegister().finally(() => {
+        this.registerPromise = Promise.resolve(this.doRegister()).finally(() => {
```
addaleax: Doesn't really make sense, does it?

packages/cli/src/createServerServices.ts
```diff
@@ -68,47 +165,96 @@ export async function createSharedServicesFromConfig(
...
+ *  2. the `x-mcp-client-name` header — opt-in label for unauthenticated
+ *     deployments (see {@link CLIENT_SCOPE_HEADER}).
+ *  3. none — the caller falls back to an ephemeral scope.
+ */
+function clientScopeFromRequest(request?: TransportRequestContext): string | undefined {
+    if (request?.authInfo?.mode === "authenticated") {
+        return request.authInfo.state.clientId;
+    }
+    const header = request?.headers?.[CLIENT_SCOPE_HEADER];
+    const name = typeof header === "string" && header.length > 0 ? header.trim() : undefined;
```
addaleax: This logic probably isn't necessary anyway, but it's probably also not intentional that a whitespace-only string would be converted to an empty one while an empty would result in `undefined`?

```suggestion
    const name = (typeof header === "string" && header.trim()) || undefined;
```

packages/integration-tests/src/integrationHelpers.ts
```diff
@@ -96,14 +105,18 @@ export const DEFAULT_LONG_RUNNING_TEST_WAIT_TIMEOUT_MS = 1_200_000;
...
+ * The request-centric design delivers config to tools through the
+ * request-scoped server (`mcpServer().config`, read by tools off `this.server`), which
+ * is constructed from the server's effective config for every call. Tests
+ * that mutate `mcpServer().config` (e.g. readOnly / disabledTools) see those
+ * mutations on the next tool call automatically, so no per-tool config
+ * syncing is needed.
+ */
+export function syncMongoToolsConfigFromUserConfig(_mcpServer: CliServer): void {
+    void _mcpServer;
 }
```
addaleax: Where is this actually being used? I may not understand what "call-site compatibility" means in this context

packages/mongodb-mcp-server/server.json
```diff
@@ -734,13 +658,6 @@
           "format": "string",
           "isSecret": false
         },
-        {
-          "name": "MDB_MCP_EXTERNALLY_MANAGED_SESSIONS",
-          "description": "When true, the HTTP transport allows requests with a session ID supplied externally through the 'mcp-session-id' header. When an external ID is supplied, the initialization request is optional.",
-          "isRequired": false,
-          "format": "boolean",
-          "isSecret": false
-        },
```
addaleax, line 743: Does e.g. `packages/mongodb-mcp-server/deploy/aws/` automatically sync somehow? This option is still mentioned there

packages/mongodb-mcp-server/server.json
```diff
@@ -1174,13 +1043,6 @@
           "isRequired": false,
           "format": "number"
         },
-        {
-          "type": "named",
-          "name": "--notificationTimeoutMs",
-          "description": "Notification timeout for a client to be aware of disconnect (only applies to http transport).",
-          "isRequired": false,
-          "format": "number"
-        },
```
addaleax, line 1183: Also still mentioned in `MCP_SERVER_LIBRARY.md` (which might need a bit of a general overhaul to account for the changes in this PR – may be best done as a separate PR)

packages/mongodb-mcp-server/server.json
gagik: hmm nope; we should probably update our generate script to match but we also do some weird handpicking there

packages/cli/src/cliServer.ts
gagik: yeah this was a weird result of register being async and then back to sync again, will remove

packages/integration-tests/src/integrationHelpers.ts
gagik: this used to be a hacky patch that is no longer needed 🙈

packages/mongodb-mcp-server/server.json
gagik: will defer to a separate PR

addaleax (approved):

lerouxb (approved):

PR #1477, opened 2026-09-02
refactor: remove the session concept (part 1/4) — core/types/http/cli + tools

gagik:
First slice of the session-removal work previously carried on refactor/remove-session-concept (#1474): deletes the session object from server core, types, http-runners, and CLI, introduces the request-centric `ToolExecutionContext.request` envelope, and threads the request object through the mongodb/atlas/atlas-local tool packages.

Stack: part 2 (Phase 2 services) → part 3 (Phase 3A scoping + auth) → part 4 (request.server + lint/api sweep). Each part builds, lints, and passes unit tests independently.

PR #1478, opened 2026-09-02
refactor: Phase 2 — shared AppServices over minimal per-request servers (part 2/4)

gagik:
Second slice: single-arg `ApiClient` constructor, resources constructed with (`{ server }`) reading services off the server, and startup-only app-config validation — the app-level shared services that per-request servers reference but never own.

Part 2 of the session-removal stack (#1474); builds on part 1 (#1477), continues in part 3.


gagik:

PR #1479, opened 2026-09-02
feat: Phase 3A — per-client connection scoping + authenticated multi-tenant HTTP (part 3/4)

gagik:
Third slice: clients' MongoDB connections survive across their HTTP requests via stable identity-scoped registry views (anonymous requests stay ephemeral), plus the authenticated multi-tenant mode: `RequestAuthState` on every request and a required `HttpServerOptions.authMode` field.

Part 3 of the session-removal stack (#1474); builds on part 2 (#1478), continues in part 4.


gagik:

PR #1480, opened 2026-09-02
refactor: read effective config off request.server + lint/api sweep (part 4/4)

gagik:
Final slice: the effective (request-overridden) config moves from the request envelope to the request-scoped server (`request.server.config`), and the branch-wide lint/format/api-report sweep lands here so every part of the stack is standalone-green.

Final part of the session-removal stack (#1474); builds on part 3 (#1479). The net of parts 1-4 equals the original #1474 branch.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,5 +1,6 @@
 dist
 node_modules
+.pi/
 *.lock
 resources/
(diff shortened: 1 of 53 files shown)


gagik:

PR #1481, opened 2026-09-02
refactor: remove the session concept — complete refactor (part A)

gagik:
Removes the server-side session object end-to-end: deletes the session concept from core/types/http-runners/cli and all tool packages, introduces the request-centric `ToolExecutionContext.request` envelope, adds app-level shared services (`AppServices`), per-client connection scoping + authenticated multi-tenant HTTP (`RequestAuthState`/required `authMode`), and reads effective config off `request.server`.

This is the entire refactor previously on `refactor/remove-session-concept` (#1474) in one PR — the earlier phases (request context, AppServices, scoping/auth) intentionally build on and supersede each other, so they are not split. The branch's final lint/format/api-report sweep lands as part B on top of this. Verification: build, unit tests (1057), integration-tests typecheck, API reports all green (lint/format sweep is part B).

PR #1482, opened 2026-09-02
chore: resolve lint issues and regenerate reports (part B)

gagik:
The branch's final cleanup commit on top of part A: applies `eslint --fix` and prettier across the tree, removes unused imports/dead generics, adds explicit return types, drops the unused `@modelcontextprotocol/node` deps, and regenerates the api-extractor reports.

Part B of the session-removal stack (part A: #1481). Net of A + B equals the original `refactor/remove-session-concept` branch (#1474).

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,5 +1,6 @@
 dist
 node_modules
+.pi/
 *.lock
 resources/
 packages/ui/src/lib/tools/*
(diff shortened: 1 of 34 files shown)


gagik:
ok #1474 is just as easy to follow...

PR #1483, opened 2026-09-02
chore: Add classifiable ToolArgumentValidationError for tool argument failures MCP-644

himanshusinghs:
## Proposed changes

Tool calls that fail because the caller passed bad arguments were throwing bare `Error`s, so downstream (the remote-mcp error classifier) couldn't tell a caller-fixable input rejection apart from an infrastructure failure - the false-positives that set off the `atlas-create-access-list` page. This adds a `ToolArgumentValidationError` (in core, re-exported from `mongodb-mcp-server`) that tools throw when they reject arguments, giving the classifier a single type to key on and mark those failures as expected.

`createAccessList` is the immediate case, and its recurring `atlas_api_400` from over-long comments is now caught at the schema layer (comments are capped at Atlas's 80-character limit) instead of round-tripping to the API to fail. The same error type is adopted by the other tools that already did argument validation by hand — `explain`, `export`, `createIndex`, and the existing `createCluster` / `upgradeCluster` local error classes now extend it. Messages are unchanged, so tool output is identical.

## Checklist

### DIFF
--- a/packages/tools-atlas/src/tools/create/createAccessList.ts
+++ b/packages/tools-atlas/src/tools/create/createAccessList.ts
@@ -5,12 +5,17 @@ import { AtlasToolBase } from "../../atlasTool.js";
 import { makeCurrentIpAccessListEntry, DEFAULT_ACCESS_LIST_COMMENT } from "../../helpers/accessListUtils.js";
 import { AtlasArgs, CommonArgs } from "../../args.js";
 
+// Atlas rejects access list entry comments longer than this (INVALID_NETWORK_PERMISSION_COMMENT).
+const ACCESS_LIST_COMMENT_MAX_LENGTH = 80;
+export const ACCESS_LIST_COMMENT_TOO_LONG_ERROR = "Comment must be 80 characters or less.";
+
 export const CreateAccessListArgs = {
     projectId: AtlasArgs.projectId().describe("Atlas project ID"),
     ipAddresses: z.array(AtlasArgs.ipAddress()).describe("IP addresses to allow access from").optional(),
     cidrBlocks: z.array(AtlasArgs.cidrBlock()).describe("CIDR blocks to allow access from").optional(),
     currentIpAddress: z.boolean().describe("Add the current IP address").default(false),
     comment: CommonArgs.asciiOnlyString()
+        .max(ACCESS_LIST_COMMENT_MAX_LENGTH, ACCESS_LIST_COMMENT_TOO_LONG_ERROR)
         .describe("Comment for the access list entries")
         .default(DEFAULT_ACCESS_LIST_COMMENT)
         .optional(),
--- a/packages/tools-atlas/src/tools/create/createAccessList.ts
+++ b/packages/tools-atlas/src/tools/create/createAccessList.ts
@@ -7,7 +7,7 @@ import { AtlasArgs, CommonArgs } from "../../args.js";
 
 // Atlas rejects access list entry comments longer than this (INVALID_NETWORK_PERMISSION_COMMENT).
 const ACCESS_LIST_COMMENT_MAX_LENGTH = 80;
-export const ACCESS_LIST_COMMENT_TOO_LONG_ERROR = "Comment must be 80 characters or less.";
+export const ACCESS_LIST_COMMENT_TOO_LONG_ERROR = `Comment must be ${ACCESS_LIST_COMMENT_MAX_LENGTH} characters or less.`;
 
 export const CreateAccessListArgs = {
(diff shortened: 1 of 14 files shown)


packages/tools-atlas/src/tools/create/createAccessList.ts
```diff
@@ -1,16 +1,21 @@
...
 export const CreateAccessListArgs = {
     projectId: AtlasArgs.projectId().describe("Atlas project ID"),
     ipAddresses: z.array(AtlasArgs.ipAddress()).describe("IP addresses to allow access from").optional(),
     cidrBlocks: z.array(AtlasArgs.cidrBlock()).describe("CIDR blocks to allow access from").optional(),
     currentIpAddress: z.boolean().describe("Add the current IP address").default(false),
     comment: CommonArgs.asciiOnlyString()
+        .max(ACCESS_LIST_COMMENT_MAX_LENGTH, ACCESS_LIST_COMMENT_TOO_LONG_ERROR)
         .describe("Comment for the access list entries")
         .default(DEFAULT_ACCESS_LIST_COMMENT)
         .optional(),
```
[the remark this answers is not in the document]
himanshusinghs: Addressed in https://github.com/mongodb-js/mongodb-mcp-server/pull/1483/commits/179ca9dd71461daa42f533aa120b9e673b7de273

packages/tools-atlas/src/tools/create/createAccessList.ts
```diff
@@ -1,16 +1,21 @@
...
+import { type ToolArgs, type ToolResult, ToolArgumentValidationError } from "@mongodb-js/mcp-core";
 import type { OperationType, ToolExecutionContext } from "@mongodb-js/mcp-types";
 import { AtlasToolBase } from "../../atlasTool.js";
 import { makeCurrentIpAccessListEntry, DEFAULT_ACCESS_LIST_COMMENT } from "../../helpers/accessListUtils.js";
 import { AtlasArgs, CommonArgs } from "../../args.js";
 
+// Atlas rejects access list entry comments longer than this (INVALID_NETWORK_PERMISSION_COMMENT).
+const ACCESS_LIST_COMMENT_MAX_LENGTH = 80;
+export const ACCESS_LIST_COMMENT_TOO_LONG_ERROR = "Comment must be 80 characters or less.";
+
```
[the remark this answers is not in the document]
himanshusinghs: Addressed in https://github.com/mongodb-js/mongodb-mcp-server/pull/1483/commits/179ca9dd71461daa42f533aa120b9e673b7de273

packages/tools-atlas/src/tools/create/createAccessList.ts
```diff
@@ -1,19 +1,23 @@
...
+import { type ToolArgs, type ToolResult, ToolArgumentValidationError } from "@mongodb-js/mcp-core";
 import type { OperationType, ToolExecutionContext } from "@mongodb-js/mcp-types";
 import { AtlasToolBase } from "../../atlasTool.js";
 import { makeCurrentIpAccessListEntry, DEFAULT_ACCESS_LIST_COMMENT } from "../../helpers/accessListUtils.js";
 import { AtlasArgs, CommonArgs } from "../../args.js";
 
+// Atlas rejects access list entry comments longer than this (INVALID_NETWORK_PERMISSION_COMMENT).
+const ACCESS_LIST_COMMENT_MAX_LENGTH = 80;
+export const ACCESS_LIST_COMMENT_TOO_LONG_ERROR = `Comment must be ${ACCESS_LIST_COMMENT_MAX_LENGTH} characters or less.`;
+
```
addaleax: Why are we doing this validation ourselves if Atlas rejects it anyway?

himanshusinghs: The schema also serves as guide for agent to construct their parameters correctly. Having the schema specify the max allowed length would actively guide agents to not call tool with comments longer than 80 characters.

gagik: makes sense but probably useless to have a custom error message for this, zod will handle it

himanshusinghs: that's fair - yea we don't use custom error message pattern elsewhere. Will remove it.

himanshusinghs: pushed a [commit](https://github.com/mongodb-js/mongodb-mcp-server/pull/1483/commits/ff9f5fdfd86fa71378018c19932c2ca595197915) removing the extra validation error message

addaleax (approved):

gagik (approved):

PR #1489, opened 2026-09-04
chore: flush logs before server exits in error conditions MCP-486

himanshusinghs:
## Proposed changes

Adds `logger.flush` before exit path.

As a drive-by, removed the process.exit() in StdioRunner's startup failure path and instead we now let the error bubble up to the CLI entry point which logs additional logs and exit cleanly.

## Checklist

### DIFF
--- a/packages/cli/src/runMcpCli.ts
+++ b/packages/cli/src/runMcpCli.ts
@@ -85,14 +85,22 @@ export async function runMcpCli({
 
     // Create logger, then the transport runner (stdio or HTTP based on config)
     const logger = await createLoggerFromConfig({ config, keychain: Keychain.root });
-    const transportRunner = await createRunnerFromConfig({
-        config,
-        serverMetadata,
-        tools,
-        resources,
-        logger,
-    });
 
-    // Start the transport runner
-    await startRunner({ transportRunner, logger, onExit });
+    try {
+        const transportRunner = await createRunnerFromConfig({
+            config,
+            serverMetadata,
+            tools,
+            resources,
+            logger,
+        });
+
+        // Start the transport runner
+        await startRunner({ transportRunner, logger, onExit });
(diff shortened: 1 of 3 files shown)


packages/integration-tests/src/transports/startupFailureLogFlush.test.ts
```diff
@@ -0,0 +1,179 @@
...
+                    "--disabledTools",
+                    "atlas-local",
+                ],
+                serverMetadata: { mcpServerName: "MongoDB MCP Server", version: "1.2.3-test" },
+                consoleLogger: console,
+                onExit: vi.fn(),
+                tools: [],
+                resources: [],
+            })
+        ).rejects.toThrow("stdio-crash");
```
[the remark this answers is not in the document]
himanshusinghs: addressed in https://github.com/mongodb-js/mongodb-mcp-server/pull/1489/commits/dcfa07f45202d5af20e5e870d2d896abf2df5de1

packages/cli/src/runMcpCli.ts
```diff
@@ -85,14 +85,22 @@ export async function runMcpCli({
...
+        });
+
+        // Start the transport runner
+        await startRunner({ transportRunner, logger, onExit });
+    } catch (error) {
+        // Flush buffered logs (e.g. async disk writes) before the error reaches the
+        // caller, which exits the process without waiting for pending log writes.
+        await logger.flush();
+        throw error;
+    }
```
addaleax, line 105: Should this just be try/finally?

addaleax (approved):

packages/cli/src/runMcpCli.ts
himanshusinghs: I was hoping to keep flush only on the failure path.

PR #1501, opened 2026-09-08
fix: redact connection-string secrets from ConnectionEntry.lastError

SyedAnas01:
`ConnectionEntry.lastError` is set from the raw driver/arg-parser error message when a connect attempt fails. That message can embed the connection string verbatim (for example, `mongodb+srv://user:pass@` with no host, which throws `Protocol and host list are required in "..."` with the credentials still in it). `lastError` gets read back unredacted by `summarizeConnection()` and returned by `list-connections`, which takes no arguments and needs no confirmation.

The redaction in `MongoDBToolBase.handleError()` only covers the response for the call that's currently failing — it never touches `ConnectionEntry.lastError`, which is written one layer earlier in `ConnectionEntry.connect()`.

This mostly matters for the preconfigured connection: it dials lazily, sits outside per-session scoping, and isn't removed from the store when a dial fails, so if `MDB_MCP_CONNECTION_STRING` is malformed this way, the raw string stays in `lastError` until the config is fixed — visible to any client calling `list-connections`, including other sessions in a shared HTTP deployment.

**Fix:** redact at the point of capture in `ConnectionEntry.connect()`, using `Keychain.root.redact()` — same pattern already used elsewhere in this codebase for driver errors that might carry secrets.

**Test:** added a regression test that fails a connect with a real (unmocked) `@mongosh/arg-parser` call against a truncated `mongodb+srv://` string and asserts `lastError` doesn't contain the credential. Confirmed it fails without the fix, passes with it.

Related to GHSA-grqv-8942-f78m.

### DIFF
--- a/packages/tools-mongodb/src/common/connectionEntry.test.ts
+++ b/packages/tools-mongodb/src/common/connectionEntry.test.ts
@@ -134,9 +134,6 @@ describe("ConnectionEntry with MCPConnectionManager", () => {
 
     describe("lastError", () => {
         it("should not contain the raw connection string when the connect attempt fails", async () => {
-            // No mock needed here: a connection string with credentials but no host
-            // fails inside @mongosh/arg-parser before the driver is ever called, and
-            // the real arg-parser embeds the string verbatim in the thrown error.
             const entry = new ConnectionEntry({
                 connectionId: "preconfigured",
                 name: "preconfigured",
--- a/packages/tools-mongodb/src/common/connectionEntry.test.ts
+++ b/packages/tools-mongodb/src/common/connectionEntry.test.ts
@@ -5,7 +5,7 @@ import { NodeDriverServiceProvider } from "@mongosh/service-provider-node-driver
 import { CompositeLogger } from "@mongodb-js/mcp-core";
 import { MCPConnectionManager, type ConnectionManager } from "./connectionManager.js";
 import { MCPConnectionStore, type ConnectionStoreConfig } from "./connectionStore.js";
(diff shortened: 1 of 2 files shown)


packages/tools-mongodb/src/common/connectionEntry.test.ts
```diff
@@ -132,6 +132,33 @@ describe("ConnectionEntry with MCPConnectionManager", () => {
         });
     });
 
+    describe("lastError", () => {
+        it("should not contain the raw connection string when the connect attempt fails", async () => {
+            // No mock needed here: a connection string with credentials but no host
+            // fails inside @mongosh/arg-parser before the driver is ever called, and
+            // the real arg-parser embeds the string verbatim in the thrown error.
```
gagik: this comment can be removed

gagik:
It's not a critical leakage but it is always good to harden these things so changes seem good to me 🙂 

Please confirm that you have signed the [MongoDB CLA](https://www.mongodb.com/legal/contributor-agreement). One small comment beyond that.

packages/tools-mongodb/src/common/connectionEntry.test.ts
SyedAnas01: Removed, thanks.

SyedAnas01:
Hi @gagik, confirming I've signed the MongoDB CLA. Let me know if anything else is needed.

gagik (approved):

gagik:
@SyedAnas01 LGTM, thank you for your contribution!

PR #1504, opened 2026-09-08
fix: disallow write stages in export aggregations MCP-651

nirinchev:
## Proposed changes

Rejects $out and $merge in export aggs.


gagik (approved):
Should we consider adding elicitation for write stage aggregations? and/or track usage of this

nirinchev:
I don't think the export tool should support write stages at all - the intention is to store the result of a large aggregation on disk for future referencing. If you want to run a write agg, you should just the aggregate tool.

gagik:
ah right, I was referring to the actual aggregation tool

PR #1508, opened 2026-09-09
chore: move atlas info from state to connection entry CLOUDP-443824

nirinchev:
## Proposed changes

Telemetry reads the cluster details off the live connection state, so failed or torn-down calls lost project_id / cluster_id. Since the cluster is property of the connection handle, not the connection itself, we move it to ConnectionEntry.atlasCluster, set at creation. The connection manager no longer knows about Atlas; ConnectionSettings.atlas becomes a hostType override.


lerouxb (approved):

himanshusinghs (approved):

gagik:
@copilot please deal with merge origin/main into this and deal with conflicts and re-run `update:api` so that passes

PR #1513, opened 2026-09-10
chore: Publish @mongodb-js/harness-tester MCP-652

lerouxb:
MCP-652

Make @mongodb-js/harness-tester a publishable npm package so it can be consumed from outside this monorepo (e.g. by the MongoDB MCP Server standalone build).

Currently the package is private: true and its exports point directly at the TypeScript source (./src/index.ts), so it can't be installed from npm. This PR publishes a built dist instead, matching the existing @mongodb-js/mcp-* package convention.

**Changes:**

* package.json: drop private: true; point exports/main/types at ./dist; add files: ["dist"], a build script, a repository field, and publishConfig.access: "public".
* tsconfig.json: enable declaration/composite emit, set outDir: ./dist, enable erasableSyntaxOnly.
* tsconfig.build.json (new): build from src/, excluding test files, so output lands flat at dist/index.js (not dist/src/...).
* src/harness/logger.ts: replace constructor parameter-property shorthand with an explicit field (not erasable syntax).
* knip.ts: register src/index.ts! as an entry point, since knip can no longer trace usage through the source exports.

Verified: pnpm run build, check:types, check:format, and check:dependencies (knip) all pass. check:lint has 6 pre-existing errors in cli/parseUserConfig.ts / integration-tests unrelated to this change (confirmed on the base branch).

*Note:* .github/workflows/publish.yml runs pnpm -r publish, so harness-tester will be picked up on the next release once this merges.


gagik (approved):
should we get someone to do the first manual publish?

PR #1517, opened 2026-09-10
fix: shut down stdio server on EOF

### LINKED ISSUE #1515: [Bug]: stdio server does not exit after stdin EOF, leaving orphaned processes
### Version

1.13.0

Environment:

- Node.js 22.23.2
- npm 10.9.8
- Codex CLI 0.153.4
- ChatGPT desktop 26.903.71938
- macOS 26.6.2 (25G83)
- stdio transport
- `--readOnly`

### App

- [x] Other: Codex / ChatGPT desktop

Not applicable. The behaviour occurs at the stdio process-lifecycle level.

### Bug Description

When `mongodb-mcp-server` runs over stdio, it does not terminate after stdin reaches EOF.

This leaves MongoDB MCP processes running after the MCP client exits. Over multiple Codex/ChatGPT sessions, these accumulated into orphaned process families consuming substantial memory.

The issue was reproduced by running the MongoDB MCP executable directly, without Codex or an `npx` parent process. This isolates the failure to exit to the server's stdio lifecycle rather than the client or npm wrapper.

#### Steps to reproduce

1. Install version 1.13.0:

   ```bash
   npm install mongodb-mcp-server@1.13.0
   ```

2. Configure a valid `MDB_MCP_CONNECTION_STRING`.

3. Send valid MCP initialisation messages through a finite stdin pipe:

   ```bash
   printf '%s\n' \
     '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2025-03-26","capabilities":{},"clientInfo":{"name":"stdio-eof-repro","version":"1.0"}}}' \
     '{"jsonrpc":"2.0","method":"notifications/initialized","params":{}}' \
     | node ./node_modules/mongodb-mcp-server/dist/esm/index.js --readOnly
   ```

4. After `printf` completes, stdin reaches EOF.

5. Observe from another terminal that `mongodb-mcp-server` continues running instead of exiting:

   ```bash
   pgrep -fl mongodb-mcp-server
   ```

#### Controlled reproduction result

I ran the installed 1.13.0 entry point directly with Node.js, sent `initialize` and `notifications/initialized`, and then closed stdin.

- The server returned a successful initialisation response.
- No stderr output was produced.
- The Node process remained alive for at least 10 seconds after EOF.
- It exited only after receiving `SIGTERM`.
- The test did not involve an `npx` wrapper or an MCP client process.

#### Expected behaviour

When the stdio input stream closes, the server should close its MCP transport and MongoDB connections, dispose of timers or other active handles, and exit normally without requiring an external signal.

#### Actual behaviour

The server remains alive after stdin closes.

When launched through an MCP client using `npx`, the retained process tree typically looks like this:

```text
npm exec mongodb-mcp-server@1.13.0 --readOnly
  └─ node .../node_modules/.bin/mongodb-mcp-server --readOnly
```

After the client exits, the npm wrapper can be adopted by `launchd` (`PPID 1`) while both processes continue running.

#### Observed impact

During one investigation:

- 112 MongoDB MCP worker processes were running.
- 94 process families were stale and no longer had their original MCP client.
- The MongoDB MCP workers and npm wrappers had approximately 10.1 GiB of summed RSS.
- The oldest stale instances had remained alive for more than five days.
- Separate one-shot shell pipelines remained alive for more than five hours after their stdin producers finished.

The RSS value is summed across processes and may include shared memory pages, but the retained process count and resource impact are substantial.

The server may have active handles, connection-monitoring tasks, or timers that keep the Node.js event loop alive after the stdio transport closes. The stdio runner may need to handle EOF or transport closure explicitly and shut down all associated resources.

lambegraham:
## Proposed changes

Fixes #1515.

When an MCP client closes stdin, the stdio server remains alive and retains memory. Repeated sessions can therefore accumulate orphaned Node processes.

Handle stdin EOF by invoking the existing graceful shutdown path, which closes the transport and shared services, flushes logs, and exits. HTTP behaviour is unchanged.

Adds a regression test that initialises a real server process, closes stdin, and verifies exit code `0`.

### Live verification

Compared the fixed build with the same build loading the exact upstream `startRunner` source from `0738bcb1`.

| Version | Run | RSS before EOF | Result after EOF |
| --- | --- | ---: | --- |
| Upstream | 1 | 177.5 MiB | Still running after 10 s; 177.5 MiB RSS |
| Upstream | 2 | 177.3 MiB | Still running after 10 s; 177.3 MiB RSS |
| Upstream | 3 | 176.7 MiB | Still running after 10 s; 142.9 MiB RSS |
| Fixed | 1 | 175.1 MiB | Exit `0` within 16 ms; PID gone |
| Fixed | 2 | 176.4 MiB | Exit `0` within 16 ms; PID gone |
| Fixed | 3 | 177.0 MiB | Exit `0` within 14 ms; PID gone |

All upstream test processes required `SIGTERM` for cleanup. Fixed processes exited without an external signal, releasing their process-owned memory.

These checks used stdio initialisation without an active MongoDB connection; live database connection cleanup was not exercised.

### Tests

- Repository build passed.
- Repository tests: 1,142 passed, 2 skipped.
- EOF integration regression passed.
- Affected TypeScript, ESLint, Prettier and diff checks passed.

## Checklist

### DIFF
--- a/packages/cli/src/startRunner.ts
+++ b/packages/cli/src/startRunner.ts
@@ -1,5 +1,4 @@
-import { LogId } from "@mongodb-js/mcp-core";
-import type { StdioRunner } from "@mongodb-js/mcp-core";
+import { LogId, StdioRunner } from "@mongodb-js/mcp-core";
 import type { StreamableHttpRunner } from "@mongodb-js/mcp-http-runners";
 import type { CompositeLogger } from "@mongodb-js/mcp-core";
 import type { OnExit } from "./types.js";
@@ -47,6 +46,9 @@ export async function startRunner({ transportRunner, logger, onExit }: StartRunn
     process.on("SIGABRT", () => void shutdown());
     process.on("SIGTERM", () => void shutdown());
     process.on("SIGQUIT", () => void shutdown());
+    if (transportRunner instanceof StdioRunner) {
+        process.stdin.once("end", () => void shutdown());
+    }
 
     try {
         await transportRunner.start();
--- a/packages/integration-tests/src/transports/stdioEof.test.ts
+++ b/packages/integration-tests/src/transports/stdioEof.test.ts
@@ -1,0 +1,69 @@
+import { spawn, type ChildProcessWithoutNullStreams } from "child_process";
+import { once } from "events";
+import path from "path";
+import { afterEach, describe, expect, it } from "vitest";
+
+const currentDir = import.meta.dirname;
+const projectRoot = path.resolve(currentDir, "../../../..");
+const serverPath = path.resolve(projectRoot, "packages/mongodb-mcp-server/dist/esm/index.js");
+const PROCESS_TIMEOUT_MS = 10_000;
+
+function cleanEnv(): NodeJS.ProcessEnv {
+    const env: NodeJS.ProcessEnv = { ...process.env };
+    for (const key of Object.keys(env)) {
+        if (key.startsWith("MDB_MCP_")) {
+            delete env[key];
+        }
+    }
+    return env;
+}
+
+describe("stdio EOF lifecycle", () => {
+    let child: ChildProcessWithoutNullStreams | undefined;
+
+    afterEach(() => {
+        if (child?.exitCode === null && child.signalCode === null) {
+            child.kill("SIGKILL");
+        }
+    });
+
+    it("exits cleanly after an initialized client closes stdin", async () => {
+        child = spawn(process.execPath, [serverPath, "--telemetry", "disabled", "--disabledTools", "atlas-local"], {
+            env: {
+                ...cleanEnv(),
+                MDB_MCP_TRANSPORT: "stdio",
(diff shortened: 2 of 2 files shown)


nirinchev:
This looks reasonable, though do you think it'd be better to handle that upstream - e.g. somewhere here: https://github.com/modelcontextprotocol/typescript-sdk/blob/b65426158ed9f29aea8ef3dc09ca22d7d9d6f970/packages/server/src/server/stdio.ts#L75? That way all MCP servers based on the typescript SDK would get it out of the box instead of handling it individually?

lambegraham:
Yes, we could handle EOF upstream, and that would benefit other TypeScript MCP servers too. But we’d still need to handle shutdown here in this project. MongoDB MCP’s current onclose handler only records telemetry, so closing the transport alone wouldn’t close the shared services and exit the process. I think in this case it is best to handle it here.

lerouxb (approved):

PR #1527, opened 2026-09-15
feat(connections): reap idle MongoDB connections to bound resource lifetime

gagik:
## What

Add an idle/TTL reaper to `MCPConnectionStore` so connections that sit unused past a configurable timeout get closed and removed, restoring the lifetime bound the pre-sessionless model provided via per-session idle timeouts.

- New `connectionIdleTimeoutMs` config (default 600000, `<= 0` disables), set via `MDB_MCP_CONNECTION_IDLE_TIMEOUT_MS` / `--connectionIdleTimeoutMs`.
- An unref'd 60s interval in `MCPConnectionStore` revokes any entry whose `lastUsedAt` is older than the timeout. It runs through the existing `revoke()` path, so temp-user cleanup, list-connections removal, and the revocation log behave exactly like an LRU overflow.
- The preconfigured connection is pinned (it auto-redials), and entries still dialing (`state === "connecting"`) are left alone.
- Background exports stream after their triggering request returns, so `ConnectionEntry` gains `acquire()`/`release()` lease counting and the export tool holds a lease for the stream's lifetime, released in the export's `finally`.

## Why

The sessionless refactor (removal of `Session`/per-session state) means connections in named/global scopes are unowned and have no natural teardown except explicit disconnect, per-scope LRU overflow (`maxActiveConnections`), or process shutdown. On a long-lived shared server that's an unbounded accumulation of MongoDB connection pools/sockets. This restores the v2 behavior where a session's connections were closed (and its resources freed) once the session went idle.

## Notes

- No behavior change under the default single-client stdio/local use case beyond eventually closing genuinely idle connections.
- Reaping is skipped for a connection holding an active export lease, so it can't be cut mid-stream.
- Requires an explicit opt-out (`0`) only if a deployment wants connections to persist forever.


gagik:
Superseded by #1528 (simplified to idle/TTL reaping only; dropped the export-lease plumbing per review). Branch now lives on `gagik/connection-entry`.

PR #1528, opened 2026-09-15
chore(connections): reap idle MongoDB connections to bound resource lifetime

gagik:
## What

Add an idle/TTL reaper to `MCPConnectionStore` so MongoDB connections that go unused past a configurable timeout get closed and removed, restoring the resource-lifetime bound the pre-sessionless model provided via per-session idle timeouts.

- New `connectionIdleTimeoutMs` config (default `600000`), set via `MDB_MCP_CONNECTION_IDLE_TIMEOUT_MS` / `--connectionIdleTimeoutMs`. `0` disables reaping; negatives are rejected (schema `.min(0)`).
- The reaper is an unref'd interval in `MCPConnectionStore` that runs on the timeout cadence itself and revokes any entry whose `lastUsedAt` is older than the timeout. It routes through the existing `revoke()` path, so temp-user cleanup, list-connections removal, and the revocation log behave exactly like an LRU overflow.
- The preconfigured connection is pinned (it auto-redials), and entries still dialing (`state === "connecting"`) are left alone.
- `connectionIdleTimeoutMs` is a required member of `ConnectionStoreConfig` (mirroring `maxActiveConnections`) and a lower-only override in the config schema.

## Why

The sessionless refactor (removal of per-session state) means connections in named/global scopes are unowned and have no natural teardown except explicit disconnect, per-scope LRU overflow (`maxActiveConnections`), or process shutdown. On a long-lived shared server that's an unbounded accumulation of MongoDB connection pools/sockets. This restores the v2 behavior where a session's connections were closed once the session went idle.

## Notes

- Browser-safe: the `setInterval` id is typed as `ReturnType<typeof setInterval>` and `unref()` is guarded so a web build works.
- No upper bound on the cadence (a long timeout just reaps less often); a short timeout runs closer to on-time.
- No behavior change under the default single-client stdio/local use case beyond eventually closing genuinely idle connections.


packages/tools-mongodb/src/common/connectionStore.ts
```diff
@@ -251,11 +262,71 @@ export class MCPConnectionStore {
...
+        }
+        const now = Date.now();
+        const idle: ConnectionEntry[] = [];
+        for (const { entry } of this.entries.values()) {
+            if (entry.source === "preconfigured") {
+                continue;
+            }
+            if (entry.state.tag === "connecting") {
+                continue;
+            }
```
himanshusinghs, line 294: This made me realise that we don't take care of connecting state in Remote MCP. Need to cross check now if that was intentional.

himanshusinghs (approved):

PR #1536, opened 2026-09-17
chore: sync root README.md into mongodb-mcp-server package

gagik:
The `generate` script now copies the root `README.md` into `packages/mongodb-mcp-server/README.md` so the published npm package ships a README that stays in sync with the repo.

Note: the monorepo package's `files` list only includes `dist`, so this ensures the package directory has the README on disk before publish.


gagik:
we probably want to make the root readme more basic and redirect to the nested one... but this should work for now

PR #1542, opened 2026-09-18
fix: return correct error code for modern-era methods called without _meta MCP-660

nirinchev:
## Proposed changes

Looks like when routing legacy requests to `server/discover`, we would respond with 400/32004, which is a non-standard response type. While those shouldn't happen since that's a modern-only endpoint, we should still return the correct error code.


gagik:
Interesting, this feels like something that can/should be upstreamed to the SDK

gagik (approved):

PR #1543, opened 2026-09-18
fix(cli): restore greedy array parsing and reject swallowed arguments…

### LINKED ISSUE #1540: [Bug]: Space-separated `--disabledTools` values are silently ignored since 1.3.0, and the second value becomes the connection target
### Version

Reproduced on 3.0.2, 2.1.1, 2.0.0, 1.8.0, 1.6.0, 1.5.0, 1.3.1 and 1.3.0. Works correctly on 1.2.0 and earlier, so this looks like a regression in 1.3.0.

### App

- [ ] Cursor
- [ ] Windsurf
- [ ] VSCode
- [ ] VSCode Insiders
- [ ] Claude Desktop
- [x] Other

- [ ] Claude 3.5 Sonnet
- [ ] Claude 3.7 Sonnet
- [ ] GPT-4a
- [ ] o4-mini
- [x] Other

### Bug Description

The README and the [Enable or Disable MCP Server Features documentation](https://www.mongodb.com/docs/mcp-server/local-mcp/configuration/enable-or-disable-features/#disabling-mcp-server-tools) both show a space-separated list for `--disabledTools`. Although some part of the README seem to have been updated to reflect the change, the Configuration Options reference for `--disabledTools` shows

```
Comma separated values of tool names, operation types, and/or categories of tools that will be disabled.
```
Running the documented example:

```bash
npx -y mongodb-mcp-server@2.1.1 --disabledTools create update delete atlas --dryRun
```

**What I see**

- `"disabledTools": ["create"]`
- `"connectionString": "mongodb://127.0.0.1:27017/update?directConnection=true&serverSelectionTimeoutMS=2000"`
- `update-many`, `delete-many`, `drop-database` and `atlas-list-clusters` are still enabled.
- No warning is printed about the values that were dropped.

**What I'd expect**

All four values disabled, and the connection string left as configured. That's what 1.2.0 and earlier do.

Worth noting the practical effect: someone using `disabledTools` rather than `--readOnly` can end up with write tools exposed to an agent while believing they're disabled, and the server also connects somewhere other than the intended target. `--readOnly` itself isn't affected.

**Forms that do work on current versions**

- Comma-separated: `--disabledTools create,update,delete,atlas`
- Repeating the flag: `--disabledTools create --disabledTools update`
- `MDB_MCP_DISABLED_TOOLS="create,update,delete,atlas"`

Mixing the two, such as `--disabledTools create,update delete`, doesn't work either: `commaSeparatedToArray` splits a plain string or a single-element array, but returns a multi-element array untouched, so `create,update` stays a single entry and matches no tool. That only surfaces once space-separated values parse into several entries, so it's worth fixing alongside.

### When it regressed

I installed published versions and ran `--disabledTools find aggregate`, then listed tools over MCP to see whether both disappeared, with controls (no flag, and a single value):

| Version | Result |
| --- | --- |
| 0.0.4, 0.1.0, 0.3.0, 1.0.0, 1.1.0, 1.2.0 | both disabled |
| 1.3.0 and every release since, including 3.0.2 | only `find` disabled |

Up to 1.2.0, `common/config.js` carried its own parser options, copied from mongosh's arg-parser but with `"greedy-arrays": true`. In 1.3.0 those were replaced by `@mongosh/arg-parser`'s shared `defaultParserOptions`, which sets `'greedy-arrays': false`, and the override seems to have been dropped in the process. `yargs-parser` then takes one value per occurrence of an array option, `processPositionalCliOptions` treats the first leftover positional as the connection specifier, and the remaining positionals land in `fileNames`, which `parseUserConfigSources` deletes without a warning.

This should apply to every array option, not just `disabledTools`, though that's the only one I tested end to end. The existing config tests appear to cover only the comma-separated form.

Note: 1.3.0 and 1.4.0 wouldn't start for me until I pinned `@mongosh/arg-parser` to 3.23.0, since `^3.23.0` now resolves to a 3.x release with incompatible module exports. The 1.3.0 result above comes from that pinned install.

### Suggested fix

In `packages/cli/src/config/parseUserConfig.ts`:

1. Set `"greedy-arrays": true` in `defaultParserOptions.configuration`, restoring the pre-1.3.0 behaviour.
2. Return an error for leftover positional arguments instead of deleting `fileNames`.
3. Return an error when an array option receives a value that looks like a connection string.

And in `packages/cli/src/config/configUtils.ts`:

4. Split every entry in `commaSeparatedToArray`, rather than only a string or a single-element array, so the two forms can be mixed.

Points 2 and 3 are there because neither change alone covers everything. Greedy arrays would silently consume `--disabledTools find mongodb://…`, and the leftover check alone misses `--disabledTools find aggregate`, where the second value still becomes the connection target with no leftovers.

I have a patch with tests ready that I'll attach to the MCP ticket.

### Related

- The README example uses `collectionSchema`, but the tool is named `collection-schema`, so that entry matches nothing even with the right separator.
- A bare word after an array option, such as `--disabledTools find mydb`, is still accepted as a tool name, since nothing validates the names. A startup warning for entries matching no tool, operation type or category would cover that as well as the typo above.

comtev:
… MCP-659

## Proposed changes

Fixes #1540

# Pull request description

Repository: https://github.com/mongodb-js/mongodb-mcp-server
Patch: `mcp-disabledtools-greedy-arrays.patch` (applies to `main` at `ca4dcc7`, the 3.0.2 release commit)

**Title:** fix(cli): restore greedy array parsing and reject swallowed arguments

### What this changes

Restores the pre-1.3.0 behaviour of space-separated values for array options, and turns two silent failures into errors.

In `packages/cli/src/config/parseUserConfig.ts`:

1. `"greedy-arrays": true` in `defaultParserOptions.configuration`, so an array option takes the values that follow it.
2. Leftover positional arguments return an error instead of being deleted along with `fileNames`.
3. An array option that receives a `mongodb://` or `mongodb+srv://` value returns an error pointing at the positional argument or `MDB_MCP_CONNECTION_STRING`.

In `packages/cli/src/config/configUtils.ts`:

4. `commaSeparatedToArray` now splits every entry, instead of only a plain string or a single-element array. Without this, `--disabledTools create,update delete` leaves `create,update` as one entry that matches no tool. The case only arises once greedy arrays produce several entries, so the two changes belong together.

### Why

Until 1.2.0, `common/config.js` set `"greedy-arrays": true` in its own parser options. In 1.3.0 those were replaced by `@mongosh/arg-parser`'s `defaultParserOptions`, which sets it to `false`, and the override was lost. Since then the documented form disables only the first value, takes the second as the connection target, and drops the rest without a warning. Every release from 1.3.0 to 3.0.2 is affected.

Points 2 and 3 are included because neither change on its own is sufficient. Greedy arrays would silently consume a connection string placed after an array option, and the leftover check alone misses the two-value case, where the second value still becomes the connection target without leaving leftovers.

### Behaviour

| Arguments | Before | After |
| --- | --- | --- |
| `--disabledTools create update delete atlas` | only `create`; target becomes `update` | all four disabled |
| `--disabledTools find aggregate` | only `find`; target becomes `aggregate` | both disabled |
| `--disabledTools find --disabledTools aggregate` | both disabled | unchanged |
| `--disabledTools find aggregate --readOnly` | only `find`; target becomes `aggregate` | both disabled, read-only |
| `mongodb://localhost --disabledTools find aggregate` | only `find`; `aggregate` dropped | both disabled, URI kept |
| `--disabledTools find mongodb://localhost` | URI treated as a tool name | clear error |
| `--disabledTools create,update delete atlas` | `create,update` kept as one entry, matching nothing | all four disabled |
| `--disabledTools create,update, delete` | `create,update,` kept as one entry | `create`, `update` and `delete` disabled |

This affects every array option (`loggers`, `previewFeatures`, `confirmationRequiredTools`, `monitoringServerFeatures`), which matches both the documentation and the pre-1.3.0 behaviour. The `commaSeparatedToArray` change likewise applies to every option using that preprocessor, which only makes the input they accept more forgiving. Please flag it if any of them is meant to differ.

### Tests

I added eight regression tests to `packages/integration-tests/src/config/userConfig.test.ts`, covering space-separated lists, repeated flags, a list followed by a boolean flag, a connection string before and after a list option, and mixed comma and space forms. Seven of them fail on `main`; the repeated-flag case already passed.

All 285 tests in `packages/cli` and `packages/integration-tests/src/config` pass with the change, and ESLint and Prettier are clean. I built and tested this against `main` at `ca4dcc7`; I didn't run the suites that need MongoDB or Atlas infrastructure, so those are worth watching in CI.

### Docs

The docs page and README examples become correct again with this change, so they need no update. The README's `collectionSchema` should still become `collection-schema`, since that's a separate typo; happy to include it here if you'd like.

### Not covered

A bare word after an array option, such as `--disabledTools find mydb`, is still accepted as a tool name. Validating names against the registered tools felt like it belonged where tools are registered rather than in the parser, so I left it out.

## Checklist


gagik:
Going to close this in favor of improving our documentation and avoiding making potentially breaking syntax changes.
`--disabledTools=tool1,tool2` should be used instead

PR #1557, opened 2026-09-21
fix: handle space-separated array args MCP-659

nirinchev:
## Proposed changes

Addresses a regression with yargs not parsing space-separated arrays correctly.


gagik:
I'm not sure I'd frame it as a fix or a regression.
We were misaligned with mongosh before and we became aligned later and at this point space separation has been unsupported for longer that it has been supported.

This will probably work OK for most cases but it's a bit of a bandaid with trying to guess when the argument is positional and when it isn't (which is why mongosh is avoiding greedy-arrays cc: @addaleax). I think people can have all kinds of weird existing connection string or mongosh connection configurations which would not match this anymore. 

We _could_ also support only `npx mongodb-mcp-server --disabledTools="connect find"` perhaps?

I'd rather we throw on e.g. unknown disabled tools or similar cases, that'd improve the UX.

PR #1563, opened 2026-09-23
fix: return correct error code for modern-era mcp requests MCP-660

nirinchev:
## Proposed changes

If a modern client makes a request to a modern endpoint (e.g. `server/discover`), return in-band JsonRPC error to trigger the fallback to the older protocol version.


src/transports/mcpHttpServer.ts
```diff
@@ -26,6 +31,31 @@ import {
...
+    if (typeof body !== "object" || body === null || Array.isArray(body)) {
+        return false;
+    }
+    const { method } = body as { method?: unknown };
+    return typeof method === "string" && MODERN_ONLY_METHODS.has(method);
+}
+
+/** The JSON-RPC id to echo in a response, when the body carries a usable one. */
+function requestIdOf(body: unknown): string | number | null {
+    const { id } = (body ?? {}) as { id?: unknown };
```
himanshusinghs: nit: Should this apply the same body checks as above?

src/transports/mcpHttpServer.ts
```diff
@@ -89,6 +119,29 @@ export class MCPHttpServer<
...
+     */
+    private reportMethodNotFound(req: express.Request, res: express.Response): void {
+        this.logger.debug({
+            id: LogId.streamableHttpTransportRequestFailure,
+            context: "streamableHttpTransport",
+            message: `Received a request for unimplemented method ${String((req.body as { method?: unknown }).method)}`,
+            attributes: requestIdAttr(req.headers),
+        });
+
+        res.status(200).json({
```
himanshusinghs, line 142: just confirming - method not found is still a 200 or should it be 404?

himanshusinghs (approved):

src/transports/mcpHttpServer.ts
nirinchev: Seems like SDK is kind of inconsistent on this front - I asked claude to dig through it and seems like sometimes it returns 200 and errors in body, sometimes it returns status codes. If a session exists, that would be a 200, so decided to keep it consistent.

src/transports/mcpHttpServer.ts
nirinchev: Yeah, should have just combined both methods instead of implementing them separately.

